PR #11, opened 2017-06-29
docs: add previously undocumented zone_id attribute

nicolai86:
this PR adds a previously undocumented, computed property to the documentation.


nicolai86:
:tada: thank you!

PR #15, opened 2017-07-06
Document default value of `proxied` on record resources

### LINKED ISSUE #4: cloudflare_record: Docs unclear on default value of "proxied"
_This issue was originally opened by @jleclanche as hashicorp/terraform#10082. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

https://www.terraform.io/docs/providers/cloudflare/r/record.html

Since `proxied` is a boolean (which the docs doesn't specify) and is Optional, it would be good to specify whether it defaults to false or true.

OJFord:
Resolves #4.


stack72:
LGTM thanks for this

PR #23, opened 2017-10-22
[IMP] Docs

### LINKED ISSUE #22: Suport for 'auto' ttl, or documentation thereof
Without further knowledge and looking at the 'ttl' type ('int') I assume that auto 'ttl' (as in the UI) is not supported, unless it's '0'. It would be nice, would it be either implemented or documented, as in the latter case, it's not obvious.

blaggacao:
Un-bias user doubts
Closes #22


elithrar:
LGTM.

PR #30, opened 2018-02-26
Rate limit for zone resource

benjvi:
Adds the new resource and updates the cloudflare-go library to a recent version

Partially due to the way structured maps are (not) handled in terraform the expanding/flattening logic is pretty long for the nested attributes. I've split things out so hopefully its a bit clearer. The only other change from the go client should be that bypass_url_patterns was specified as a key-value struct, here its simplified as a string list. 

Since we need both zone + resource id to do lookups the ID is a amalgamation of the two, and we generate two computed fields for the separate zone + resource ids which is what I refer to in any operations after read. Hopefully this makes sense - this approach not common since normally resource id is sufficient for lookups but there is some prior art in the AWS provider for this


catsby:
Hey there! Due to the very large file count included here, this pull request is very difficult to review. I see that perhaps much of the file count is due to dependency changes? If at all possible, please break those out into a separate pull request. That of course assumes that those changes are possible to merge without this feature (which is usually the case, but I admit I don't know CloudFlare's SDK well 😄 )

The same applies to https://github.com/terraform-providers/terraform-provider-cloudflare/pull/31

benjvi:
@catsby Is the separate PR in #32 with this rebased on top enough to make this easier to review?

catsby:
@benjvi well, it _should have_, but I merged #32 and this is still very large. Perhaps I did the merge wrong, by making a merge commit.

Would it be possible to rebase again off of `master` now?

cloudflare/resource_cloudflare_rate_limit.go
```diff
@@ -0,0 +1,490 @@
...
+			return errors.Wrap(err,
+				fmt.Sprintf("Error reading rate limit resource from API for resource %s in zone %s", zoneId, rateLimitId))
+		}
+	}
+	log.Printf("[DEBUG] Read CloudFlare Rate Limit from API as struct: %+v", rateLimit)
+
+	d.Set("threshold", rateLimit.Threshold)
+	d.Set("period", rateLimit.Period)
+	d.Set("match", flattenRateLimitTrafficMatcher(rateLimit.Match))
+	d.Set("action", flattenRateLimitAction(rateLimit.Action))
```
catsby: Errors need to be checked when setting non-scalar attributes such as `action` and `match`:

```go
if err := d.Set("action", ...); err != nil {
  log.Printf("[WARN] Error setting action: %s", err)
}
```

benjvi:
@catsby I updated with your comments. Also I changed the ID a bit so it could be consistent with the existing DNS record resource + a bit more intuitive ( it also seems that underscore is relatively common in subdomains so I changed the separator to "/" on import )

PR #31, opened 2018-02-26
Load balancer + load balancer pool resources

benjvi:
These load balancers do geographically aware load-balancing based on defined pools.

As such, I realized that load balancer pools are necessary as a prereq to doing load balancers so I have both in this PR. Can split this up if it proves too difficult to digest

Note - there are also associated 'monitors' for load balancing pools, I intend to add that as a separate resource in a subsequent PR. Then it would be useful to have it in at least one of the load balancer pool tests.

Also worth noting that the load balancer uses the same kind of compound id as rate_limit

### DIFF
--- a/cloudflare/import_cloudflare_load_balancer_pool_test.go
+++ b/cloudflare/import_cloudflare_load_balancer_pool_test.go
@@ -1,0 +1,37 @@
+package cloudflare
+
+import (
+	"testing"
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccCloudFlareLoadBalancerPool_Import(t *testing.T) {
+	t.Parallel()
+	var loadBalancerPool cloudflare.LoadBalancerPool
+	rnd := acctest.RandString(10)
+	name := "cloudflare_load_balancer_pool." + rnd
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckCloudFlareLoadBalancerPoolConfigBasic(rnd),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCloudFlareLoadBalancerPoolExists(name, &loadBalancerPool),
+				),
+			},
+			{
+				ResourceName:      name,
+				ImportState:       true,
+				ImportStateVerify: true,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCloudFlareLoadBalancerPoolExists(name, &loadBalancerPool),
+				),
+			},
+		},
+	})
+}
--- a/cloudflare/import_cloudflare_load_balancer_test.go
+++ b/cloudflare/import_cloudflare_load_balancer_test.go
@@ -1,0 +1,41 @@
+package cloudflare
+
+import (
+	"os"
+	"testing"
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccCloudFlareLoadBalancer_Import(t *testing.T) {
+	t.Parallel()
+	var loadBalancer cloudflare.LoadBalancer
+	zone := os.Getenv("CLOUDFLARE_DOMAIN")
+	rnd := acctest.RandString(10)
+	name := "cloudflare_load_balancer." + rnd
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckCloudFlareLoadBalancerConfigBasic(zone, rnd),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCloudFlareLoadBalancerExists(name, &loadBalancer),
+					testAccCheckCloudFlareLoadBalancerIDIsValid(name, zone),
+				),
+			},
+			{
+				ResourceName:      name,
+				ImportState:       true,
+				ImportStateVerify: true,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCloudFlareLoadBalancerExists(name, &loadBalancer),
+					testAccCheckCloudFlareLoadBalancerIDIsValid(name, zone),
+				),
+			},
+		},
+	})
+}
--- a/cloudflare/import_cloudflare_load_balancer_test.go
+++ b/cloudflare/import_cloudflare_load_balancer_test.go
@@ -4,6 +4,8 @@ import (
 	"os"
 	"testing"
 
+	"fmt"
+
 	"github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
@@ -28,9 +30,10 @@ func TestAccCloudFlareLoadBalancer_Import(t *testing.T) {
 				),
 			},
 			{
-				ResourceName:      name,
-				ImportState:       true,
-				ImportStateVerify: true,
+				ResourceName:        name,
+				ImportStateIdPrefix: fmt.Sprintf("%s/", zone),
+				ImportState:         true,
+				ImportStateVerify:   true,
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckCloudFlareLoadBalancerExists(name, &loadBalancer),
 					testAccCheckCloudFlareLoadBalancerIDIsValid(name, zone),
--- a/cloudflare/resource_cloudflare_load_balancer.go
+++ b/cloudflare/resource_cloudflare_load_balancer.go
@@ -1,0 +1,361 @@
+package cloudflare
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"time"
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/pkg/errors"
+)
+
+func resourceCloudFlareLoadBalancer() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceCloudFlareLoadBalancerCreate,
+		Read:   resourceCloudFlareLoadBalancerRead,
+		Update: resourceCloudFlareLoadBalancerUpdate,
+		Delete: resourceCloudFlareLoadBalancerDelete,
+		Exists: resourceCloudFlareLoadBalancerExists,
+		Importer: &schema.ResourceImporter{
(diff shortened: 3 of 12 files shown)


website/docs/r/load_balancer.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+    name = "example-1"
+    address = "192.0.2.1"
+    enabled = false
+  }
+}
+
+# Define a load balancer which always points to the pool we defined
+# In normal usage, would have different pools set for different pops/regions
+# And some failover defined within that
+resource "cloudflare_load_balancer" "bar" {
```
catsby: Nitpick: let's move the `cloudflare_load_balancer` config to the top of the `hcl` example, so it's the first thing users see. Otherwise, the user is learning about `cloudflare_load_balancer` and the first thing they see is `cloudflare_load_balancer_pool` configuration.

catsby: >different pools set for different pops/regions

Also, what's a `pop`? It's best to not assume users are familiar with acronyms specific to a service. I'm not familiar with load balancing in detail, so I can't tell if `pop` and `region` are synonymous, or two distinct acceptable options.

cloudflare/import_cloudflare_load_balancer_pool_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"testing"
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccCloudFlareLoadBalancerPool_Import(t *testing.T) {
```
catsby: I ran this test locally but it failed: 

```console
$ make testacc TESTARGS=" run=TestAccCloudFlareLoadBalancerPool_Import" 
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccCloudFlareLoadBalancerPool_Import -timeout 120m
?       github.com/terraform-providers/terraform-provider-cloudflare    [no test files]
=== RUN   TestAccCloudFlareLoadBalancerPool_Import
=== PAUSE TestAccCloudFlareLoadBalancerPool_Import
=== CONT  TestAccCloudFlareLoadBalancerPool_Import
--- FAIL: TestAccCloudFlareLoadBalancerPool_Import (0.76s)
        testing.go:434: Step 0 error: Error applying: 1 error(s) occurred:

                * cloudflare_load_balancer_pool.c2p8b4gigl: 1 error(s) occurred:

                * cloudflare_load_balancer_pool.c2p8b4gigl: error creating load balancer pool: error from makeRequest: HTTP status 400: content "{\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 1002,\n      \"message\": \"the origin list length must be in range [1, 0]: validation failed\"\n    }\n  ],\n  \"messages\": []\n}\n"
FAIL
FAIL    github.com/terraform-providers/terraform-provider-cloudflare/cloudflare 0.787s
make: *** [testacc] Error 1
```

cloudflare/import_cloudflare_load_balancer_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"testing"
+
+	"fmt"
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccCloudFlareLoadBalancer_Import(t *testing.T) {
```
catsby: Similar error as the other import test:

```console
$ make testacc TESTARGS="-run=TestAccCloudFlareLoadBalancer_Import" 

TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccCloudFlareLoadBalancer_Import -timeout 120m
?       github.com/terraform-providers/terraform-provider-cloudflare    [no test files]
=== RUN   TestAccCloudFlareLoadBalancer_Import
=== PAUSE TestAccCloudFlareLoadBalancer_Import
=== CONT  TestAccCloudFlareLoadBalancer_Import
--- FAIL: TestAccCloudFlareLoadBalancer_Import (0.58s)
        testing.go:434: Step 0 error: Error applying: 1 error(s) occurred:

                * cloudflare_load_balancer_pool.tdu3ryy6ap: 1 error(s) occurred:

                * cloudflare_load_balancer_pool.tdu3ryy6ap: error creating load balancer pool: error from makeRequest: HTTP status 400: content "{\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 1002,\n      \"message\": \"the origin list length must be in range [1, 0]: validation failed\"\n    }\n  ],\n  \"messages\": []\n}\n"
FAIL
FAIL    github.com/terraform-providers/terraform-provider-cloudflare/cloudflare 0.601s
make: *** [testacc] Error 1
```

cloudflare/resource_cloudflare_load_balancer.go
```diff
@@ -0,0 +1,335 @@
...
+
+			"zone_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: false, // allow reusing same load balancer for different dns names
```
catsby: `ForceNew` is `false` by default, so this can be omitted. I am curious about what the comment means. It reads as if Load Balancer names are not unique? I'm not sure what `ForceNew` would influence there.

cloudflare/resource_cloudflare_load_balancer.go
```diff
@@ -0,0 +1,335 @@
...
+			},
+
+			"fallback_pool_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringLenBetween(1, 32),
+			},
+
+			"default_pool_ids": {
+				Type:     schema.TypeList,
```
catsby: Is the ordering of pool ids significant such that it affects the behavior of the load balancer? Will these pool id's always be returned from the API in the same order (assuming no external modifications)?

cloudflare/resource_cloudflare_load_balancer.go
```diff
@@ -0,0 +1,335 @@
...
+							Type:     schema.TypeList,
+							Required: true,
+							Elem: &schema.Schema{
+								Type:         schema.TypeString,
+								ValidateFunc: validation.StringLenBetween(1, 32),
+							},
+						},
+					},
+				},
+				Set: HashByMapKey("pop"),
```
catsby: Is there a specific need for a custom hash function? `Set` should not be set unless there are very specific requirements to the hashing, otherwise please omit `Set` on any `TypeList`,`TypeMap`, or `TypeSet`, and let Terraform apply the default hashing for each type.

cloudflare/resource_cloudflare_load_balancer.go
```diff
@@ -0,0 +1,335 @@
...
+
+	log.Printf("[INFO] Creating CloudFlare Load Balancer from struct: %+v", newLoadBalancer)
+
+	r, err := client.CreateLoadBalancer(zoneId, newLoadBalancer)
+	if err != nil {
+		return errors.Wrap(err, "error creating load balancer for zone")
+	}
+
+	if r.ID == "" {
+		return fmt.Errorf("cailed to find id in Create response; resource was empty")
```
catsby: Typo `cailed` -> `failed`

cloudflare/resource_cloudflare_load_balancer.go
```diff
@@ -0,0 +1,335 @@
...
+
+func resourceCloudFlareLoadBalancerRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	zoneId := d.Get("zone_id").(string)
+	loadBalancerId := d.Id()
+
+	loadBalancer, err := client.LoadBalancerDetails(zoneId, loadBalancerId)
+	if err != nil {
+		if strings.Contains(err.Error(), "HTTP status 404") {
+			log.Printf("[INFO] Load balancer %s in zone %s no longer exists", loadBalancerId, zoneId)
```
catsby: Nitpick: `no longer exists` -> `not found`. We don't know with certainty that it doesn't exist, we only know for certain that in this call the LB was not found in that Zone

cloudflare/resource_cloudflare_load_balancer.go
```diff
@@ -0,0 +1,335 @@
...
+			d.SetId("")
+			return nil
+		} else {
```
catsby: Nitpick/Go idiom/**not a blocker:** the `} else { ` is not needed because the statement above returns, the `else` returns, and the containing block ends.

The following are logical equivalent:

```go
{
  if true {
    return true
  } else {
    return false
  }
}
```

and

```go
{
  if true {
    return true
  }
  return false
}
```

cloudflare/resource_cloudflare_load_balancer.go
```diff
@@ -0,0 +1,335 @@
...
+			return errors.Wrap(err,
+				fmt.Sprintf("Error reading load balancer resource from API for resource %s in zone %s", zoneId, loadBalancerId))
+		}
+	}
+	log.Printf("[INFO] Read CloudFlare Load Balancer from API as struct: %+v", loadBalancer)
```
catsby: Nitpick/non-blocker: this debug is probably not needed or is excessive for logs

website/docs/r/load_balancer.html.markdown
catsby: I tried to apply this example config and got similar error to the test failures

benjvi:
Closing in favour of #40

cloudflare/resource_cloudflare_load_balancer.go
benjvi: I want to enforce uniqueness on the pop field (point-of-presence name) , I thought this was the best way to do it?

cloudflare/resource_cloudflare_load_balancer.go
benjvi: yes they should be, this is intended to define the failover order for these pools

cloudflare/resource_cloudflare_load_balancer.go
benjvi: I think initially I thought that this name was immutable for some reason, which it isn't. The actual behaviour is as you would normally expect, so I'm OK to delete this comment

cloudflare/resource_cloudflare_load_balancer.go
catsby: 👍

cloudflare/import_cloudflare_load_balancer_pool_test.go
benjvi: @catsby I can't reproduce this in my account. I am going to do some more testing relating to plan levels - but I thought you were also using an enterprise account?

benjvi: @catsby having done some reading up it seems this error message is the one that occurs when you didn't enable load balancing in the account (https://community.cloudflare.com/t/cloudflare-warp-beta/5656/252). You will need to enable load balancing in the cloudflare dashboard

cloudflare/import_cloudflare_load_balancer_test.go
benjvi: As above, this is because load balancing is not enabled

PR #33, opened 2018-03-02
Add Load balancer monitor resource 

benjvi:
A simple, fairly standard-looking resource which enables load balancers created via the load balancer resource to work correctly. Relates to and follows the same patterns used in #31 , maybe helpful to consider them together

### DIFF
--- a/cloudflare/resource_cloudflare_load_balancer_monitor.go
+++ b/cloudflare/resource_cloudflare_load_balancer_monitor.go
@@ -1,0 +1,258 @@
+package cloudflare
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"time"
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/pkg/errors"
+)
+
+func resourceCloudFlareLoadBalancerMonitor() *schema.Resource {
+	return &schema.Resource{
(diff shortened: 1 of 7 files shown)


cloudflare/resource_cloudflare_load_balancer_monitor.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func resourceCloudFlareLoadBalancerPoolMonitorDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	log.Printf("[INFO] Deleting CloudFlare Load Balancer Monitor: %s ", d.Id())
+
+	err := client.DeleteLoadBalancerMonitor(d.Id())
+	if err != nil {
+		return errors.Wrap(err, "error deleting cloudflare load balancer monitor")
```
catsby: Check the error here for the equivalent of a 404 like it's done on line 208 above. If it's delete externally to Terraform, we don't want users to be stuck here.

benjvi: Is this in case deletions occur concurrently with terraform run? I would assume that otherwise terraform will always run read first. In any case I've addressed this now

benjvi:
Closed in favour of #42

cloudflare/resource_cloudflare_load_balancer_monitor.go
catsby: It likely `READ` will be ran first, but it's plausible that we're executing a saved plan in which case I believe it won't

PR #12, opened 2017-07-05
Implement page rules

OJFord:
This PR is a rebasing of hashicorp/terraform#11249 due to the new (-ish, which is testament to how stale I let this get!) providers structure.

I'm working on tests right now, but (assuming no side-effect of the new structure)  it is currently functional by my manual tests.

Messrs. @thasmo @jamestyrrell @jonleighton @mwarkentin - you've expressed interest (either in hashicorp/terraform#9040 or in #3) of support for CloudFlare page rules; if you're able to verify this is working too that would be much appreciated :slightly_smiling_face:

N.B.
I retained the history from my branch in the core repo; an unfortunate side-effect of which is that Github seems to have a new (?) element for authors/committers which is listing my name twice. :laughing:

Presumably if this is merged I'll no longer be the committer, and so I would be listed only as the author, (OJFord + Maintainer)  so I'm tempted just to leave it as it is. If you'd prefer I clean it up though; I *think* it would just be a matter of rebasing with `--reset-author`.

**Note**: this isn't _quite_ ready to be merged, waiting for cloudflare/cloudflare-go#138 because the alternative is a nasty hack.

PR #34, opened 2018-03-02
Adding Page rule resource (updated)

benjvi:
This is based on the work in #12 and so includes those commits. The structure is mostly the same as defined there by @OJFord , with just some minor changes.

I've changed the booleans for the unitary fields on the api so that 'false' is the same as not setting the field (which should have the same semantics anyway). I had hopes of changing on/off to boolean as well but due to some corner cases in the api, this seemed to cause more problems than it solved.

Other than that theres a few minor changes to get tests passing and I also added import. This is rebased on top of the vendoring change #32


OJFord:
Thanks for picking this up! Is this still dependant on the upstream API changes in cloudflare/cloudflare-go/pull/154?

OJFord:
@catsby Yes, @benjvi's finished up my abandonware in #12 😆 - I've closed that PR in favour of this.

Thanks @benjvi!

catsby:
Closing in favor of https://github.com/terraform-providers/terraform-provider-cloudflare/pull/38 where I resolve the merge conflicts. Thanks!

PR #35, opened 2018-03-02
Zone Settings Override resource

benjvi:
This is a terraform resource that allows users to manage their zone settings. For context, the lifecycle neither of the zone nor the zone settings is fully manageable through the API (ie zone settings are tied to the zone lifecycle; the zone cannot be deleted through the API). So the correspondence with API objects is looser here than it normally is.

By taking the initial / default settings specified on the zone, we can track the changes against that group of settings and manage it as a resource in terraform, which is the idea of this resource. 

The API settings a specified as a list of objects  with more-or-less arbitrary structure. Since that didn't seem to fit too well into terraforms model or how people perceive the API (e.g. in the documentation) the flattened approach with an entry for each setting type seemed to make the most sense. This also drops some metadata on each field , such as the modified date or whether the field is editable. Beyond that, settings are modelled much as they are in the API. As the surface area here is quite large it doesn't seem very practical to test all the different combinations of values and how they work from the API side


cloudflare/validators.go
```diff
@@ -79,3 +81,24 @@ func validateRecordName(t string, value string) error {
 
 	return nil
 }
+
+// validateIntInSlice returns a SchemaValidateFunc which tests if the provided value
+// is of type int
+func validateIntInSlice(valid []int) schema.SchemaValidateFunc {
+	return func(i interface{}, k string) (s []string, es []error) {
```
catsby: Named result parameters are discouraged, recommend changing to:

```
return func(i interface{}, k string) ([]string, []error) {
  var es []error 
  //...
  // es if invalid
  return nil, es
// nil, nil if valid
  return nil, nil
}
```

This way, `s` is never initialized at all, and both are returned as `nil` if there are no errors.

cloudflare/resource_cloudflare_zone_settings_override.go
```diff
@@ -0,0 +1,665 @@
...
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: resourceCloudFlareZoneSettingsSchema,
+				},
+			},
+
+			"initial_settings": {
+				Type:     schema.TypeList,
```
catsby: Is the ordering of settings significant, or guaranteed to always be returned in the same order?

catsby: Maybe `TypeMap` is more appropriate here? I'll defer to you

cloudflare/resource_cloudflare_zone_settings_override.go
```diff
@@ -0,0 +1,665 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+var resourceCloudFlareZoneSettingsSchema = map[string]*schema.Schema{
+	"advanced_ddos": {
+		Type: schema.TypeString, ValidateFunc: validation.StringInSlice([]string{"on", "off"}, false),
```
catsby: Nitpick: `ValidateFunc` should be on it's online

benjvi:
Closing in favour of #41

cloudflare/resource_cloudflare_zone_settings_override.go
benjvi: I think this runs into the problem with structured map values in terraform's schema language

benjvi: The order doesn't matter per se

catsby: If ordering doesn't matter, or more importantly if ordering will not be guaranteed, then we should use a `TypeSet` here.

benjvi: This is always restricted to a single item in the terraform config so ordering will not matter (one to one relationship with a sub-resource that holds all the settings). It looks more understandable to use a TypeList in this case and I thought this was also the convention from looking at AWS/Google providers ?

cloudflare/resource_cloudflare_zone_settings_override_test.go
```diff
@@ -0,0 +1,131 @@
...
+						name, "settings.0.challenge_ttl"),
+					resource.TestCheckResourceAttrSet(
+						name, "settings.0.security_level"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccCloudFlareZone_Overrides(t *testing.T) {
```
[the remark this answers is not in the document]
benjvi: I can't verify this in my zones right now, everything is passing for me. I did notice in the TestAccCloudFlareZone_Basic there were some assertions that could fail based on the state of the underlying account which I have fixed ( the idea in this test case is to prove that the settings are unchanged from the initial settings). It doesn't look like any assertions in TestAccCloudFlareZone_Overrides should depend on the initial state of the zone. What error messages are you getting?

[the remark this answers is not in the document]
benjvi: @catsby Found some errors running tests when using a free instead of enterprise plan as I was before. Hopefully this will now be working for you in #41

PR #41, opened 2018-03-18
Zone Settings Override resource (updated)

benjvi:
Continued from #35. Reopening this so others can address issues identified there in the next few days


nickithewatt:
@catsby, @benjvi answered some of your original review comments in #35, however there are still a few outstanding review items to address in this PR - we will get onto those this week.

benjvi:
Hopefully, all the comments from the previous review are now addressed. I had to make quite a few changes to get this to work in free plans as well, which included a fair amount of cleanup - so it should be a bit more understandable now as well. Let me know what you think @catsby

PR #37, opened 2018-03-06
[WIP] Api rate limiting

benjvi:
This introduces options to the provider to allow for rate limiting, backoff and logging. Options are just passed through to the API client. 

This can't be merged until the upstream change https://github.com/cloudflare/cloudflare-go/pull/163 gets included. In the meantime, the updated code is added manually, but it should be replaced by a proper vendored version of the API client

PR #43, opened 2018-03-18
Api rate limiting (updated)

benjvi:
Continued from #37

### DIFF
--- a/cloudflare/config.go
+++ b/cloudflare/config.go
@@ -8,13 +8,14 @@ import (
 )
 
 type Config struct {
-	Email string
-	Token string
+	Email   string
+	Token   string
+	Options []cloudflare.Option
 }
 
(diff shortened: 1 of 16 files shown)


nickithewatt:
Just to re-iterate @benjvi's note from previous PR #37, This PR is still failing because it is dependent on the upstream change cloudflare/cloudflare-go#163 getting accepted and merged. 

To test this PRin the meanwhile, you will need to updated the code manually to refer/include the  cloudflare/cloudflare-go#163 changes. Once that change is finally merged, this code will need to be updated again to refer to the proper vendored version of the API client

benjvi:
Now that the upstream cloudflare/cloudflare-go#163  is merged, I've updated the vendoring so this should be ready to go now (smaller number of vendor changes so I just left the changes in this PR). @catsby can you let me know if this is OK now?

PR #47, opened 2018-04-15
Fixed "Example Usage" problem in "page_rule.html.markdown"

223n:
Fixed an issue where "zone" was incorrectly entered as "domain" in the content of "Example Usage".


patryk:
Thanks for sharp eye!

PR #48, opened 2018-04-16
Add link to geographic regions, detail on address field. 

prdonahue:
The load balancing documentation indicates that omitting check regions is optional, but this is true only for Enterprise accounts. Make this clear.

The origin definition doesn't include detail on the address field, so this is added.

Fixed a couple typos.


patryk:
Thanks!

PR #51, opened 2018-04-18
IPv6 CIDR blocks should return IPv6 addresses

williamlord:
Currently returning IPv4 addresses.


patryk:
Yep, thanks for noticing!

PR #52, opened 2018-04-22
Remove incorrect import support docs

tdmalone:
Import is not supported for this resource - running terraform import with the documented format returns 'resource cloudflare_record doesn't support import'


garrettgalow:
@patryk we should be able to close this now that we have added import support (in #36 ) for records correct?

patryk:
Correct. Closing.

PR #62, opened 2018-06-02
Update page_rule.html.markdown

jsonmaur:
`cloudflare_page_rule.foobar` used the `domain` argument when it's supposed to be `zone` (according to the reference.)


jsonmaur:
Looks like this is was already PR'ed: https://github.com/terraform-providers/terraform-provider-cloudflare/pull/47/files. This should really get merged....

PR #61, opened 2018-05-16
Page Rules bugfix / enhancements

SteveGoldthorpe-Work:
Fixes issues:
* BugFix: rules status uses disabled not paused #60 
* Enhancements: Page Rule - Host Header Override #46 (and Bypass Cache on Cookie and Resolve Override).

PR #68, opened 2018-06-20
Page rule additions

simpsora:
This PR adds a new page rule action and adds a new value to an existing action.

It is based on unmerged upstream PR terraform-providers#61, so currently includes the changes from that PR as well.

- Adds action `explicit_cache_control action`
- Adds missing `off` value to `security_level` action

I've been using this code for the past couple of days and it's working as expected -- I can successfully enable the `explicit_cache_control` action and set `security_level` to `off` via Terraform, and see the results reflected in the Page Rules UI.

Note: I haven't written any tests because the existing test suite doesn't test each action and value. Happy to add tests that exercise the new functionality if you'd prefer.

Thanks for your consideration!
Ross

### DIFF
--- a/website/docs/r/page_rule.html.markdown
+++ b/website/docs/r/page_rule.html.markdown
@@ -57,11 +57,13 @@ Action blocks support the following:
 * `cache_level` - (Optional) Whether to set the cache level to `"byypass"`, `"basic"`, `"simplified"`, `"aggressive"`, or `"cache_everything"`.
 * `forwarding_url` - (Optional) The URL to forward to, and with what status. See below.
 * `host_header_override` - (Optional) The Host Header to override on the origin servers.
+* `origin_error_page_pass_thru` - (Optional) Whether this action is `"on"` or `"off"`.
 * `resolve_override` - (Optional) Override the origin server with this host.
 * `rocket_loader` - (Optional) Whether to set the rocket loader to `"off"`, `"manual"`, or `"automatic"`.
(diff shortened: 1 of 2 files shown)


jacobbednarz:
@simpsora For the purpose of reviewing, would it be possible to exclude the changes from #61?

website/docs/r/page_rule.html.markdown
```diff
@@ -59,7 +59,7 @@ Action blocks support the following:
 * `host_header_override` - (Optional) The Host Header to override on the origin servers.
 * `resolve_override` - (Optional) Override the origin server with this host.
 * `rocket_loader` - (Optional) Whether to set the rocket loader to `"off"`, `"manual"`, or `"automatic"`.
-* `security_level` - (Optional) Whether to set the security level to `"essentially_off"`, `"low"`, `"medium"`, `"high"`, or `"under_attack"`.
+* `security_level` - (Optional) Whether to set the security level to `"off"`, `"essentially_off"`, `"low"`, `"medium"`, `"high"`, or `"under_attack"`.
```
jacobbednarz: "off" is an enterprise only feature - do we need to make note of this somewhere? Or just rely on the API to return sane error messages if it isn't available?

simpsora: I don't _think_ it's the responsibility of the provider to know or care about a customer's contract type with Cloudflare.  The API should just To The Right Thing ™️  if a user without appropriate permissions tries to use it.

simpsora:
Closing in favor of https://github.com/terraform-providers/terraform-provider-cloudflare/pull/81, which is based on current `master` and has far fewer changes.

PR #72, opened 2018-06-25
Add support for `min_tls_version`

jacobbednarz:
Updates the crypto settings to accept `min_tls_version` and it's
associated values.

See https://api.cloudflare.com/#zone-settings-get-minimum-tls-version-setting


patryk:
Thanks Jakob!

PR #77, opened 2018-07-01
Fix incorrect code example

Vinnl:
As mentioned in the argument reference below the code example (and as I experienced when using it), `domain` should be `zone`.


patryk:
Thanks for noticing!

PR #78, opened 2018-07-01
Add `cloudflare_account_member` resource

jacobbednarz:
This adds a new resource `cloudflare_account_member` which is used to
manage account members.

There is a caveat here, which is noted in the README but I'll reiterate,
that once an account member is created, the user object (which is where
the email address is defined) becomes [read only](https://api.cloudflare.com/#account-members-update-member) and cannot be
updated using the account member endpoints. 

This resource does rely on cloudflare/cloudflare-go#196 which in turn
relies on a few PRs in the `cloudflare-go` library. Would be good to get
those merged and then update the vendored version before merging this to
confirm everything works as expected.

Tests are still outstanding while I get a test account fully setup.

### Screenshots (website)

**Sidebar navigation**

 

**Main page**


cloudflare/resource_cloudflare_account_member.go
```diff
@@ -0,0 +1,114 @@
...
+	"log"
+	"os"
+	"strings"
+
+	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+// This needs to be hooked up to the provider schema instead.
+var cloudflareOrgID = os.Getenv("CLOUDFLARE_ORG_ID")
```
jacobbednarz: Need to work out why `d.GetOk("org_id")` isn't working in the methods below.

Using `org_id` also changes a bunch of functionality to be managed since it overrides `user_owner_from_zone` which isn't used in the account member resource but changing any behaviour to include it may break other things.

https://github.com/terraform-providers/terraform-provider-cloudflare/blob/b9b9853d9b539375104f0a4eb22a1cd41b415fb9/cloudflare/provider.go#L73-L78

cloudflare/resource_cloudflare_account_member.go
```diff
@@ -0,0 +1,114 @@
...
+		Delete: resourceCloudflareAccountMemberDelete,
+
+		SchemaVersion: 0,
+		Schema: map[string]*schema.Schema{
+			"email_address": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"role_ids": {
```
jacobbednarz: Decided to use IDs over Names here as the name _could_ change but the IDs are pretty unique and won't trigger a bunch of extra updates.

README.md
```diff
@@ -42,7 +42,7 @@ If you wish to work on the provider, you'll first need [Go](http://www.golang.or
 To compile the provider, run `make build`. This will build the provider and put the provider binary in the `$GOPATH/bin` directory.
 
 ```sh
-$ make bin
```
jacobbednarz: I'll throw up a PR soon that will expand on the "how to develop the provider" section since I've had a few people ask me and the documentation is lacking.

jacobbednarz:
#101 brings in the required functionality for this PR.

patryk:
Ok, I've broke the PR 🤦🏻‍♂️. Sorry! Need to revert it back.

patryk:
@jacobbednarz do you want to make any additional changes to this PR or are we good to go?

jacobbednarz:
We do need to look at https://github.com/terraform-providers/terraform-provider-cloudflare/pull/78/files/2da08fc70772bb4073470055e5ec3886309c10bd#diff-2e16a4318cb61e160f7ac9dc64bca334 otherwise the account number doesn't get applied correctly. I haven't quite found the link for this yet but can have another peek.

jacobbednarz:
@patryk This should be good to rock and roll now. You've already updated the vendored version of `cloudflare-go` to support this in the last couple of days. 🚢

patryk:
Awesome stuff, guys! Thanks so much!

jacobbednarz:
Thanks @patryk 🌮

PR #79, opened 2018-07-03
forwarding_urls in page rules are lists

MrTravisB:
Currently when importing a page rule with a forwarding_url, the following error is given.

        2018-07-03T14:43:23.366-0700 [DEBUG] plugin.terraform-provider-cloudflare_v1.0.0_x4: 2018/07/03 14:43:23 [WARN] Error setting actions in page rule "XXXXXXXXXXXXXX": actions.0.forwarding_url: '': source data must be an array or slice, got map

This fixes that. forwarding_url is defined as a list and there needs to be set to a list.

### DIFF
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -422,7 +422,7 @@ func transformFromCloudFlarePageRuleAction(pageRuleAction *cloudflare.PageRuleAc
 		break
 
 	case pageRuleAction.ID == "forwarding_url":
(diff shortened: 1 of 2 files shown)


jacobbednarz:
@MrTravisB Nice find. Is it possible to add (or expand on) a test case to prevent this from regressing?

MrTravisB:
@jacobbednarz done

patryk:
Indeed, good catch. Thanks Travis and thanks Jacob for taking time in reviewing it.

jacobbednarz:
appreciate it @MrTravisB 🙇

PR #83, opened 2018-07-05
Fix resolve override and host header override

simpsora:
`host_header_override` and `resolve_override` should both take a plain
text string value, but they're currently treated as OnOff fields.

This PR moves them both to plain string fields with no validation
(there's no suitable validation function for a plain text string).

It also adds them to the docs.

This fixes https://github.com/terraform-providers/terraform-provider-cloudflare/issues/82.


patryk:
Ops! Sorry about that!

simpsora:
This fixes https://github.com/terraform-providers/terraform-provider-cloudflare/issues/46 as well.

PR #85, opened 2018-07-05
Add missing waf action

simpsora:
`waf` is an OnOff value.

Also added to the docs.

This PR, along with https://github.com/terraform-providers/terraform-provider-cloudflare/pull/76 and https://github.com/terraform-providers/terraform-provider-cloudflare/pull/83, together supersede everything that's in https://github.com/terraform-providers/terraform-provider-cloudflare/pull/61.


patryk:
Thank you Ross.

PR #86, opened 2018-07-06
Update ip_ranges example

Chupaka:
Google Compute Firewall currently supports only IPV4 addresses, and using current example produces an error:

* google_compute_firewall.allow_cloudflare_ingress: Error creating firewall: googleapi: Error 400: Invalid value for field 'resource.sourceRanges[0]': '2405:b500::/32'. Must be a valid IPV4 CIDR address range.

So, moving from cidr_blocks to ipv4_cidr_blocks. Also, updating google_compute_firewall syntax to the actual one.


patryk:
Thanks Pavel!

PR #88, opened 2018-07-09
bypass_cache_on_cookie is not limited to on/off

MrTravisB:
https://github.com/terraform-providers/terraform-provider-cloudflare/commit/1d787b2c49f8b7c1e5910f09450a41a10c77b2b8#diff-f1bba0bb24bfe33e56d7c9f834a25c82R64

Added `bypass_cache_on_cookie` in the above commit but that field is not limited to on/off values. It should be a freeform string. Can't find CF API documentation on this field but in the UI it is a text field and no on/off toggle.


patryk:
Yes, `bypass_cache_on_cookie` and `cache_on_cookie` are both strings.

PR #98, opened 2018-07-24
Add support for WAF Rule configuration

gerardc:
Note that since WAF rules already exist for an account we're not
really "creating" them, only pointing to them and storing their
state.

Since the ID is known at creation time and we need to "point"
to the actual WAF rule, I've added "rule_id" as an attribute that
the user can set based on the ID they can see in the UI.

An example configuration looks like this:

```
resource "cloudflare_waf_rule" "100000" {
  rule_id = "100000"
  zone = "domain.com"
  mode = "simulate"
}
```

The create function will look up the WAF rule 100000, update
it's state depending on the mode passed in by the user, then
store the result.

The read function checks for any updates to the mode of the WAF rule.

The update function attempts to modify the mode of a WAF rule.

The delete function attempts to set the WAF rule to its default
mode.

Import is also implemented and tested in the convention of other
import test files in the repo.

### DIFF
--- a/cloudflare/import_cloudflare_waf_rule_test.go
+++ b/cloudflare/import_cloudflare_waf_rule_test.go
@@ -1,0 +1,32 @@
+package cloudflare
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccCloudflareWAFRule_Import(t *testing.T) {
+	t.Parallel()
+	zone := os.Getenv("CLOUDFLARE_DOMAIN")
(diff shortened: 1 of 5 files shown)


gerardc:
cc @patryk

patryk:
Hey Gerard, thanks for sending this is.

My remarks:
- could you change `name` to `rule_id`, for which it actually is?
- could you add some documentation in `/website` directory?

Thanks.

gerardc:
cc @patryk thanks for the feedback, I've implemented both changes.

patryk:
Looks good, thanks Gerard!

Issue #114, opened 2018-09-06
Page rule lacks support for Auto Minify and Polish


bradenex:
Currently there is no support for Auto Minify in `resource_cloudflare_page_rule.go` and, in the same file, Polish settings only allow for `on` and `off` . When `on` is used, this error is returned:

`cloudflare_page_rule.cloudflare_rule_www: Failed to create page rule: error from makeRequest: HTTP status 400: content "{\"success\":false,\"errors\":[{\"code\":1004,\"message\":\"Page Rule validation failed: See messages for details.\"}],\"messages\":[{\"code\":1,\"message\":\".settings[0].value: Polish can only be off, lossless or lossy\",\"type\":null}],\"result\":null}"`

When using `lossless` or `lossy`, this error is returned:
`Error: cloudflare_page_rule.cloudflare_rule_www: expected actions.0.polish to be one of [on off], got lossless`

Both of these settings have support in `resource_cloudflare_zone_settings_override.go` but not in Page Rules.

patryk:
Fixed by #116. Thanks for report Braden!

bradenex:
Thanks! However, only half of this issue has been resolved. I'm still not seeing any support for Auto-minify.

SteveGoldthorpe-Work:
Yip I only fixed the Polish issue - the other is more complex and I wanted to get the Polish fix out of the way whilst I had the time.
Technically the Polish issue was a bug and the Auto Minify issue is an enhancement.
Probably easier to create a new issue for Auto Minify and keep to one thing per issue.

PR #116, opened 2018-09-13
Fix page rule polish (off, lossless or lossy)

SteveGoldthorpe-Work:
#114 (first part - fix polish options)


patryk:
Looks good, thanks Steve!

PR #121, opened 2018-10-08
Add support for importing `account_members`

jacobbednarz:
Creates the ability to use Terraform's import functionality for the
`account_member` resource.


patryk:
Thanks!

jacobbednarz:
@patryk What is the policy on cutting a new release? Now that `account_members` is fully supported, we're looking to use it but not sure if we should worry about building the provider locally ourselves or just wait for a new release to be cut?

jacobbednarz:
🙇 much appreciated!

PR #122, opened 2018-10-11
Allow setting SSL level to Strict (SSL-Only Origin Pull)

bobrik:
This is only available on Enterprise plan.


patryk:
Thanks @bobrik !

PR #124, opened 2018-10-12
Correctly set `multi_script` on Enterprise worker imports

### LINKED ISSUE #123: Import multi-script worker route problem
Importing `cloudflare_worker_route` always imports it as if it is single-script worker filter even `CLOUDFLARE_ORG_ID` is set. 
What is the correct way to import multi-script worker route?

### Terraform Version
Terraform v0.11.7

### Affected Resource(s)
cloudflare_worker_route

```hcl
terraform {
  required_version = "0.11.7"
}

resource "cloudflare_worker_route" "route" {
  zone = ""
  pattern = ""
  script_name = "${cloudflare_worker_script.script.name}"
}

resource "cloudflare_worker_script" "script" {
  name = "my_script"
  content = "${file("./scripts/my_script")}"
}
```

### Expected Behavior
Resource imported as multi-script route

### Actual Behavior
Resource imported as single-script filter

### Steps to Reproduce
CLOUDFLARE_ORG_ID=xxx terraform import cloudflare_worker_route.route zone/route_id

### References
#111

jacobbednarz:
This updates the import for `cloudflare_worker_route` resource. Prior to
this it relied on `d.Get("script_name")` which was never accessible so
it would always return an empty string (even when using
`getRouteFromResource`).

With this change, it now queries the Cloudflare API fetching all worker
routes and when the route ID matches one in the API response, it will
set the `isEnterpriseWorker` flag to `true` which will correctly set the
`multi_script` option. This includes the prerequisitie that the
organisation ID must also be set to enable it which maintains
compatibility with the existing behaviour.

Fixes #123

### DIFF
--- a/cloudflare/resource_cloudflare_worker_route.go
+++ b/cloudflare/resource_cloudflare_worker_route.go
@@ -175,6 +175,7 @@ func resourceCloudflareWorkerRouteDelete(d *schema.ResourceData, meta interface{
 
 func resourceCloudflareWorkerRouteImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
 	client := meta.(*cloudflare.API)
+	isEnterpriseWorker := false
 
 	// split the id so we can lookup
 	idAttr := strings.SplitN(d.Id(), "/", 2)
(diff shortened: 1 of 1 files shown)


patryk:
Thanks Jacob!

jacobbednarz:
It looks like this just missed the 1.7.0 release. There have been quite 
a few changes since the last release 
(https://github.com/terraform-providers/terraform-provider-cloudflare/compare/v1.7.0...master 

) so I can't imagine a fresh release is too far away.

jacobbednarz:
Alternatively, you could build the provider locally until a new release 
has been cut. See 
https://github.com/terraform-providers/terraform-provider-cloudflare#building-the-provider 
for full details.

PR #128, opened 2018-10-19
Ignore `account_member` role ID ordering

jacobbednarz:
After we imported our users to Terraform we ordered the roles
alphabetically using a variable map for readability. Once we started
applying changes we noticed that the resource would be marked as needing
replacement despite nothing changing. Turns out the reason was the role
ID ordering.

There are a few ways I could have solved this however I opted for the
simplest of just checking that the old and new values are both present
somewhere in the roles and if they were, suppress the change diff.

This doesn't impact any functionality however it does allow a little
more flexibility if you have a particular way you want the roles ordered
and not want to follow Cloudflare's API response of lexically sorting on
the ID.


patryk:
Yup, nice find, thanks!

PR #129, opened 2018-10-19
Update rate limit validation to allow 1

jacobbednarz:
Despite what the documentation states, 1 is valid here (at least for
Enterprise customers).

Have raised ticket 1589750 to get the documentation updated.


patryk:
Thanks Jacob!

Issue #134, opened 2018-10-24
Error during import of cloudflare_zone_lockdown


jphilaine:
Hi there,

I can't import a Cloudflare zone lockdown, probably due to recently API update.
Terraform use `zones/firewall/lockdowns/xxxxxxxxxxxxxxxxxxxxxxxx`
 instead of `zones/yyyyyyyyyyyyyyyyyyyyyyyyy/firewall/lockdowns/xxxxxxxxxxxxxxxxxxxxxxxx`.

Cloudflare API return this error:
```
{"success":false,"errors":[{"code":7003,"message":"Could not route to \/zones\/firewall\/lockdowns\/xxxxxxxxxxxxxxxxxxxxxxxx, perhaps your object identifier is invalid?"},{"code":7000,"message":"No route for that URI"}],"messages":[],"result":null}
```

### Terraform Version
terraform version is 0.11.10
terraform-provider-cloudflare version is 1.7.0

### Affected Resource(s)
- cloudflare_zone_lockdown

### Terraform Configuration Files
```hcl
provider "cloudflare" {
  email = "${var.cloudflare["email"]}"
  token = "${var.cloudflare["token"]}"
}

# Restrict access to these endpoints to requests from a known IP address.
resource "cloudflare_zone_lockdown" "endpoint_lockdown" {
  zone        = "api.mysite.com"
  paused      = "false"
  description = "Restrict access to these endpoints to requests from a known IP address"
  urls = [
    "api.mysite.com/some/endpoint*",
  ]
  configurations = [
    {
      "target" = "ip"
      "value" = "198.51.100.4"
    },
  ]
}
```

### Debug Output
See an anonymous debug output:
https://gist.github.com/jphilaine/3f7db1eb37971d4650fe8b2da1a9bdbc

### Panic Output
Nothing

### Expected Behavior
Terraform can refreshing state at the end of import and successfuly  import Cloudflare zone lockdown.

### Actual Behavior
An error occured due to malformed Cloudflare API URLs: missing zone ID

### Steps to Reproduce
1. Create a zone lockdown into Cloudflare
2. Create a terraform cloudflare_zone_lockdown ressource
3. Import existing Cloudflare lockdown into state using `terraform import` command
```
$ terraform import cloudflare_zone_lockdown  api.mysite.com/d41d8cd98f00b204e9800998ecf8427e
```

### References
[...]

PR #135, opened 2018-10-24
Fix import of zone lockdowns

pdecat:
As mentioned in #134, zone lockdowns imports are failing, and the regression is caused by https://github.com/terraform-providers/terraform-provider-cloudflare/commit/4bc4aa2cbf18a92adbadea9b3e6f47e7d23df482#diff-0e4cc11a1763d8ef0318546d3027d928

Sadly, I do not have a test account with the appropriate subscription level to run the acceptance tests:

```
                * cloudflare_zone_lockdown.zuwfrow271: error creating zone lockdown for zone "myzone.org": error from makeRequest: HTTP status 400: content "{\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 10012,\n      \"messag
e\": \"zonelockdown.api.not_entitled.max_rules\"\n    }\n  ],\n  \"messages\": []\n}\n"                                                                                                                                                                                  
```

(and I'd rather not run acceptance tests on production accounts)


patryk:
Ops! Sorry about that. Just how one missing line can break things. Thanks for the effort.

PR #136, opened 2018-10-24
Update `cloudflare_load_balancer_monitor` interval validation

### LINKED ISSUE #133: error creating load balancer monitor: interval is not in range [60, 3600]
According to the example usage at https://www.terraform.io/docs/providers/cloudflare/r/load_balancer_monitor.html `55` should be a valid value for 'interval` – but when using this example code, I see the following error:

```
cloudflare_load_balancer_monitor.test: error creating load balancer monitor: error from makeRequest: HTTP status 400: content "{\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 1002,\n      \"message\": \"interval is not in range [60, 3600]: validation failed\"\n    }\n  ],\n  \"messages\": []\n}\n"
* module.swarm-ams.cloudflare_load_balancer_monitor.test: 1 error(s) occurred:
```

### Terraform Version
v0.11.10

### Affected Resource(s)

```hcl
resource "cloudflare_load_balancer_monitor" "test" {
  expected_body = "alive"
  expected_codes = "2xx"
  method = "GET"
  timeout = 7
  path = "/health"
  interval = 55
  retries = 5
  description = "example load balancer"
}
```

### Expected Behavior
What should have happened?

According to the example usage at https://www.terraform.io/docs/providers/cloudflare/r/load_balancer_monitor.html `55` should be a valid value for 'interval`

### Actual Behavior
The error states that the value should be between `60` and `3600`

jacobbednarz:
Updates the `cloudflare_load_balancer_monitor` document and schema
validation for the `interval` argument to only accept values between 60
and 3600 seconds.

Fixes #133

### DIFF
--- a/cloudflare/resource_cloudflare_load_balancer_monitor.go
+++ b/cloudflare/resource_cloudflare_load_balancer_monitor.go
@@ -55,9 +55,10 @@ func resourceCloudflareLoadBalancerMonitor() *schema.Resource {
 			},
 
 			"interval": {
-				Type:     schema.TypeInt,
-				Optional: true,
-				Default:  60,
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      60,
+				ValidateFunc: validation.IntBetween(60, 3600),
 			},
 			// interval has to be larger than (retries+1) * probe_timeout:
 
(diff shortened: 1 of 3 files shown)


patryk:
👌

PR #144, opened 2018-10-28
Remove validation for `cloudflare_load_balancer_monitor` interval

jacobbednarz:
#136 introduced some basic validation based on the publicly documented
API to provide some safety rails for new users (originally addressing
#133). Unfortunately the `interval` attribute minimum can be lowered for
Enterprise users below the documented defaults and the validation [gets
in the way of the Enterprise users](https://github.com/terraform-providers/terraform-provider-cloudflare/issues/133#issuecomment-433359115). We could have just lowered the
minimum but I feel this would have created a game of whack-a-mole
whereby we would be slowly decreasing this until it was no longer
useful. Instead, I've just opted to remove it and keep the default and
documentation pointing to the supported defaults.


patryk:
Good spot. Thanks.

PR #152, opened 2018-11-14
`s/status/statuses` update for documentation

jacobbednarz:
Updates the rate limit resource docs to use the correct status key name
for `match.response`.


patryk:
Nice catch!

PR #57, opened 2018-04-30
Support load balancer origin weights and session affinity

lawrencejones:
Updating the cloudflare-go library brings support for the weight
parameter of load balancer origins. This allows configuration of traffic
priority to each of the backend origins, which is critical in more
complex load balancing setups.

I don't have a full-featured Cloudflare account, so I can't run all the tests. The fully specified load balancer test is within reach of my pauper setup, so I can provide test output for that!

```
λ (lab) ~> make testacc TEST=./cloudflare TESTARGS='-run=TestAccCloudFlareLoadBalancerPool_FullySpecified.*'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./cloudflare -v -run=TestAccCloudFlareLoadBalancerPool_FullySpecified.* -timeout 120m
=== RUN   TestAccCloudFlareLoadBalancerPool_FullySpecified
=== PAUSE TestAccCloudFlareLoadBalancerPool_FullySpecified
=== CONT  TestAccCloudFlareLoadBalancerPool_FullySpecified
--- PASS: TestAccCloudFlareLoadBalancerPool_FullySpecified (3.70s)
PASS
ok      github.com/terraform-providers/terraform-provider-cloudflare/cloudflare 3.728s
```

PR #153, opened 2018-11-14
Support load balancer origin weights and session affinity

patryk:
Continuation of PR #57, thanks @lawrencejones.

### DIFF
--- a/cloudflare/resource_cloudflare_load_balancer.go
+++ b/cloudflare/resource_cloudflare_load_balancer.go
@@ -57,6 +57,11 @@ func resourceCloudflareLoadBalancer() *schema.Resource {
 				},
 			},
 
+			"session_affinity": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
 			"proxied": {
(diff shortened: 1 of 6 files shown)


patryk:
cc @jacobbednarz, @deuill.

cloudflare/resource_cloudflare_load_balancer.go
```diff
@@ -165,6 +172,7 @@ func resourceCloudflareLoadBalancerCreate(d *schema.ResourceData, meta interface
 		Proxied:        d.Get("proxied").(bool),
 		TTL:            d.Get("ttl").(int),
 		SteeringPolicy: d.Get("steering_policy").(string),
+		Persistence:    d.Get("session_affinity").(string),
```
jacobbednarz: Something that we might have a discussion about in `cloudflare/cloudflare-go` is the fact that we're using `Persistence` on `cloudflare.LoadBalancer` despite the API response using `session_affinity`. While not wrong, it does seems like a potential for misunderstanding.

PR #157, opened 2018-11-19
Allow zones to be properly imported

ooesili:
Without this change, when zones are imported they don't have the `zone`
attribute set inside the state. This causes the next plan to attempt to
recreate the resource because it thinks the existing resource has an
empty string for the zone attribute. The `zone` attribute is indeed set
when zones are created through Terraform which explains why this may
have gone unnoticed up until now.


patryk:
Thanks Wesley! Great find.

PR #160, opened 2018-11-20
adding plan attribute to fix #159

### LINKED ISSUE #159: Plan readable/settable when creating/updating zones
This is a feature request to add Plan attribute to the `cloudflare_zone` resource.
Asking as since we are working with multiple plans we would like to manage zones subscribing them to a plan or another.

### Terraform Version
```
Terraform v0.11.10
provider.cloudflare: version = "~> 1.9"
```

### Affected Resource(s)
cloudflare_zone

in the desired config
```hcl
{
  name : "example.org"
  jump_start = false
  paused = false
  plan = "Enterprise Website"
}
```

I think I already did what needed in the code, will raise PR.
Thank you

inge4pres:
Fixes #159 
Could not run the whole AcceptanceTests suite for some un-related errors on other resources, but did run the Acceptance for the interested resource using
```
CLOUDFLARE_EMAIL="my@account.com" CLOUDFLARE_TOKEN="..." CLOUDFLARE_DOMAIN="example.org" TESTARGS="-run TestZone$" make testacc
```
and it's passing

### DIFF
--- a/cloudflare/resource_cloudflare_zone.go
+++ b/cloudflare/resource_cloudflare_zone.go
@@ -202,7 +202,7 @@ func flattenMeta(d *schema.ResourceData, meta cloudflare.ZoneMeta) map[string]in
 }
 
 func setRatePlan(d *schema.ResourceData, client *cloudflare.API, zoneID string) error {
-	if planName, ok := d.GetOk("planName"); ok {
+	if planName, ok := d.GetOk("plan"); ok {
 		ratePlan, err := getZonePlanIDFor(client, zoneID, planName.(string))
 		if err != nil {
 			return fmt.Errorf("Error fetching planName %s for zone %q: %s", planName, zoneID, err)
--- a/cloudflare/resource_cloudflare_zone.go
+++ b/cloudflare/resource_cloudflare_zone.go
@@ -48,11 +48,10 @@ func resourceCloudflareZone() *schema.Resource {
 				},
 			},
 			"plan": {
-				Type:     schema.TypeString,
-				Computed: true,
-				Optional: true,
-				//TODO ask for complete Enum
-				ValidateFunc: validation.StringMatch(regexp.MustCompile("^([a-zA-Z0-9 ]+)$"), "Not a valid plan: use Free Website, Pro Plan, Business, Enterprise"),
+				Type:         schema.TypeString,
+				Computed:     true,
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice([]string{"Free Website", "Business Website", "Pro Website", "Enterprise Website"}, false),
 			},
 			"meta": {
 				Type:     schema.TypeMap,
@@ -86,6 +85,7 @@ func resourceCloudflareZone() *schema.Resource {
 		},
 	}
 }
+
 func resourceCloudflareZoneCreate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*cloudflare.API)
 
@@ -115,8 +115,10 @@ func resourceCloudflareZoneCreate(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
-	if err := setRatePlan(d, client, zone.ID); err != nil {
-		return err
+	if plan, ok := d.GetOk("plan"); ok {
+		if err := setRatePlan(plan.(string), client, zone.ID); err != nil {
+			return err
+		}
 	}
 
 	return resourceCloudflareZoneRead(d, meta)
@@ -168,8 +170,10 @@ func resourceCloudflareZoneUpdate(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
-	if err := setRatePlan(d, client, zoneID); err != nil {
-		return err
+	if plan, ok := d.GetOk("plan"); ok {
+		if err := setRatePlan(plan.(string), client, zoneID); err != nil {
+			return err
+		}
 	}
 
 	return resourceCloudflareZoneRead(d, meta)
@@ -201,15 +205,13 @@ func flattenMeta(d *schema.ResourceData, meta cloudflare.ZoneMeta) map[string]in
 	return cfg
 }
 
-func setRatePlan(d *schema.ResourceData, client *cloudflare.API, zoneID string) error {
-	if planName, ok := d.GetOk("plan"); ok {
-		ratePlan, err := getZonePlanIDFor(client, zoneID, planName.(string))
-		if err != nil {
-			return fmt.Errorf("Error fetching planName %s for zone %q: %s", planName, zoneID, err)
-		}
-		if _, err := client.ZoneSetRatePlan(zoneID, *ratePlan); err != nil {
-			return fmt.Errorf("Error setting planName %s for zone %q: %s", planName, zoneID, err)
-		}
+func setRatePlan(planName string, client *cloudflare.API, zoneID string) error {
+	ratePlan, err := getZonePlanIDFor(client, zoneID, planName)
+	if err != nil {
+		return fmt.Errorf("Error fetching planName %s for zone %q: %s", planName, zoneID, err)
+	}
+	if _, err := client.ZoneSetRatePlan(zoneID, *ratePlan); err != nil {
+		return fmt.Errorf("Error setting planName %s for zone %q: %s", planName, zoneID, err)
 	}
 	return nil
 }
@@ -224,5 +226,5 @@ func getZonePlanIDFor(client *cloudflare.API, zoneID, planName string) (*cloudfl
 			return &p, nil
 		}
 	}
-	return nil, fmt.Errorf("Not found amongst plans")
+	return nil, fmt.Errorf("plan %s not found amongst the available plans", planName)
 }
--- a/cloudflare/resource_cloudflare_zone.go
+++ b/cloudflare/resource_cloudflare_zone.go
@@ -47,6 +47,13 @@ func resourceCloudflareZone() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"plan": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
+				//TODO ask for complete Enum
+				ValidateFunc: validation.StringMatch(regexp.MustCompile("^([a-zA-Z0-9 ]+)$"), "Not a valid plan: use Free Website, Pro Plan, Business, Enterprise"),
+			},
 			"meta": {
 				Type:     schema.TypeMap,
 				Computed: true,
@@ -108,6 +115,10 @@ func resourceCloudflareZoneCreate(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
+	if err := setRatePlan(d, client, zone.ID); err != nil {
+		return err
+	}
+
 	return resourceCloudflareZoneRead(d, meta)
 }
 
@@ -136,6 +147,7 @@ func resourceCloudflareZoneRead(d *schema.ResourceData, meta interface{}) error 
 	d.Set("name_servers", zone.NameServers)
 	d.Set("meta", flattenMeta(d, zone.Meta))
 	d.Set("zone", zone.Name)
+	d.Set("plan", zone.Plan.Name)
 
 	return nil
 }
@@ -154,6 +166,10 @@ func resourceCloudflareZoneUpdate(d *schema.ResourceData, meta interface{}) erro
 		if err != nil {
 			return fmt.Errorf("Error updating zone_id %q: %s", zoneID, err)
 		}
+	}
+
+	if err := setRatePlan(d, client, zoneID); err != nil {
+		return err
 	}
 
 	return resourceCloudflareZoneRead(d, meta)
@@ -183,4 +199,30 @@ func flattenMeta(d *schema.ResourceData, meta cloudflare.ZoneMeta) map[string]in
 	log.Printf("[DEBUG] flattenMeta %#v", cfg)
 
 	return cfg
+}
+
+func setRatePlan(d *schema.ResourceData, client *cloudflare.API, zoneID string) error {
+	if planName, ok := d.GetOk("planName"); ok {
+		ratePlan, err := getZonePlanIDFor(client, zoneID, planName.(string))
+		if err != nil {
+			return fmt.Errorf("Error fetching planName %s for zone %q: %s", planName, zoneID, err)
+		}
+		if _, err := client.ZoneSetRatePlan(zoneID, *ratePlan); err != nil {
+			return fmt.Errorf("Error setting planName %s for zone %q: %s", planName, zoneID, err)
+		}
+	}
+	return nil
+}
+
+func getZonePlanIDFor(client *cloudflare.API, zoneID, planName string) (*cloudflare.ZoneRatePlan, error) {
+	plans, err := client.AvailableZoneRatePlans(zoneID)
+	if err != nil {
+		return nil, err
+	}
+	for _, p := range plans {
+		if strings.EqualFold(p.Name, planName) {
+			return &p, nil
+		}
+	}
+	return nil, fmt.Errorf("Not found amongst plans")
 }
--- a/cloudflare/resource_cloudflare_zone.go
+++ b/cloudflare/resource_cloudflare_zone.go
@@ -49,7 +49,6 @@ func resourceCloudflareZone() *schema.Resource {
 			},
 			"plan": {
 				Type:         schema.TypeString,
-				Computed:     true,
 				Optional:     true,
 				ValidateFunc: validation.StringInSlice([]string{"Free Website", "Business Website", "Pro Website", "Enterprise Website"}, false),
 			},
--- a/cloudflare/resource_cloudflare_zone.go
+++ b/cloudflare/resource_cloudflare_zone.go
@@ -7,10 +7,26 @@ import (
 	"strconv"
 	"strings"
 
-	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
 )
+
+// we enforce the use of the Cloudflare API 'legacy_id' field until the mapping of plan is fixed in cloudflare-go
+const (
+	planIDFree       = "free"
+	planIDPro        = "pro"
+	planIDBusiness   = "business"
+	planIDEnterprise = "enterprise"
+)
+
+// we keep a private map and we will have a function to check and validate the descriptive name from the RatePlan API with the legacy_id
+var idForName = map[string]string{
+	"Free Website":       planIDFree,
+	"Pro Website":        planIDPro,
+	"Business Website":   planIDBusiness,
+	"Enterprise Website": planIDEnterprise,
+}
 
 func resourceCloudflareZone() *schema.Resource {
 	return &schema.Resource{
@@ -50,7 +66,8 @@ func resourceCloudflareZone() *schema.Resource {
 			"plan": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				ValidateFunc: validation.StringInSlice([]string{"Free Website", "Business Website", "Pro Website", "Enterprise Website"}, false),
+				Computed:     true,
+				ValidateFunc: validation.StringInSlice([]string{planIDFree, planIDPro, planIDBusiness, planIDEnterprise}, false),
 			},
 			"meta": {
 				Type:     schema.TypeMap,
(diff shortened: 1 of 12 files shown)


inge4pres:
Pushed a commit more with added docs for the website, Travis failing probably due to updated `cloudflare-go` deps as I run `govendor fetch github.com/cloudflare/cloudflare-go` and pushed the vendored updates... need to be reverted?

jacobbednarz:
> Could not run the whole AcceptanceTests suite for some un-related errors on other resources

Since the acceptance tests create real resources, you'll need to have the products enabled in your account to fully run them. As long as you can confirm the changes you're making are 👌 the others _should_ remain the same but myself or @patryk can confirm the full suite for you if needed.

cloudflare/resource_cloudflare_zone.go
```diff
@@ -47,6 +47,13 @@ func resourceCloudflareZone() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"plan": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
+				//TODO ask for complete Enum
+				ValidateFunc: validation.StringMatch(regexp.MustCompile("^([a-zA-Z0-9 ]+)$"), "Not a valid plan: use Free Website, Pro Plan, Business, Enterprise"),
```
jacobbednarz: We should use the available values from the API for this value validation.

```suggestion
				ValidateFunc: validation.StringInSlice([]string{"free", "business", "pro", "enterprise"}, false),
```

cloudflare/resource_cloudflare_zone.go
```diff
@@ -108,6 +115,10 @@ func resourceCloudflareZoneCreate(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
+	if err := setRatePlan(d, client, zone.ID); err != nil {
```
jacobbednarz: This is no longer needed if we use the `ValidateFunc` in the schema.

cloudflare/resource_cloudflare_zone.go
```diff
@@ -156,6 +168,10 @@ func resourceCloudflareZoneUpdate(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
+	if err := setRatePlan(d, client, zoneID); err != nil {
```
jacobbednarz: This is no longer needed if we use the `ValidateFunc` in the schema.

cloudflare/resource_cloudflare_zone.go
```diff
@@ -184,3 +200,29 @@ func flattenMeta(d *schema.ResourceData, meta cloudflare.ZoneMeta) map[string]in
 
 	return cfg
 }
+
+func setRatePlan(d *schema.ResourceData, client *cloudflare.API, zoneID string) error {
```
jacobbednarz: This is no longer needed if we use the `ValidateFunc` in the schema.

cloudflare/resource_cloudflare_zone.go
```diff
@@ -184,3 +200,29 @@ func flattenMeta(d *schema.ResourceData, meta cloudflare.ZoneMeta) map[string]in
...
+			return fmt.Errorf("Error fetching planName %s for zone %q: %s", planName, zoneID, err)
+		}
+		if _, err := client.ZoneSetRatePlan(zoneID, *ratePlan); err != nil {
+			return fmt.Errorf("Error setting planName %s for zone %q: %s", planName, zoneID, err)
+		}
+	}
+	return nil
+}
+
+func getZonePlanIDFor(client *cloudflare.API, zoneID, planName string) (*cloudflare.ZoneRatePlan, error) {
```
jacobbednarz: This is no longer needed if we use the `ValidateFunc` in the schema.

cloudflare/resource_cloudflare_zone_test.go
```diff
@@ -20,6 +20,7 @@ func TestZone(t *testing.T) {
 					resource.TestCheckResourceAttr(name, "zone", "example.org"),
 					resource.TestCheckResourceAttr(name, "paused", "true"),
 					resource.TestCheckResourceAttr(name, "name_servers.#", "2"),
+					resource.TestCheckResourceAttr(name, "plan", "Free Website"),
```
jacobbednarz:
```suggestion
					resource.TestCheckResourceAttr(name, "plan", "free"),
```

vendor/vendor.json
```diff
@@ -251,10 +251,10 @@
 			"revisionTime": "2017-07-27T06:48:18Z"
 		},
 		{
-			"checksumSHA1": "ANR/7MdWTtjTzxFZSuKE0Ss5j0s=",
```
jacobbednarz: Let's break this change out into it's own Pull Request. We shouldn't bump library versions with new features unless they are dependent on one another and even then, two PRs is preferred.

website/docs/r/zone.html.markdown
```diff
@@ -25,6 +25,7 @@ The following arguments are supported:
 * `zone` - (Required) The DNS zone name which will be added.
 * `paused` - (Optional) Boolean of whether this zone is paused (traffic bypasses Cloudflare). Default: false.
 * `jump_start` - (Optional) Boolean of whether to scan for DNS records on creation. Ignored after zone is created. Default: false.
+* `plan` - (Optional) The name of the commercial plan to apply to the zone, can be updated once the one is created; one of `Free Website`, `TODO`, `TODO`, `TODO`
```
jacobbednarz:
```suggestion
* `plan` - (Optional) The name of the commercial plan to apply to the zone. Valid values: `free`, `business`, `pro`, `enterprise`.
```

inge4pres:
Hi @jacobbednarz thanks for your help :)
The reason why I added thos constants is that the cloudflare-go client returns those names instead of the one you mentioned.
When changing to `Free Website` to `free`, the test fails
```
=== RUN   TestZone
--- FAIL: TestZone (13.86s)
    testing.go:527: Step 0 error: Check failed: Check 4/4 error: cloudflare_zone.test: Attribute 'plan' expected "free", got "Free Website"
FAIL
FAIL    github.com/terraform-providers/terraform-provider-cloudflare/cloudflare 13.897s
?       github.com/terraform-providers/terraform-provider-cloudflare/version    [no test files]
```

I've used `cloudlfare.API.AvailableZoneRatePlans(zoneID)` to fetch all the plans from and cycling through matching the required in config.
Is there any other method of client to get those constants?
Thanks

vendor/vendor.json
patryk: Please don't bump `cloudflare-go` in PRs. If needed, just let us know and we will commit updated upstream version.

cloudflare/resource_cloudflare_zone.go
inge4pres: From what I can see in the `cloudflare-go` impl, these are `legacy_id`s, not available in the `ZoneRatePlan` returned by `client.ZoneDetails(zoneID)`. Should I add code in `resourceCloudflareZoneRead` to fetch the legacy ids from another part of the client?

jacobbednarz:
> The reason why I added thos constants is that the cloudflare-go client returns those names instead of the one you mentioned.

[`AvailableZoneRatePlans`](https://github.com/cloudflare/cloudflare-go/blob/c69a36b9a2daf6c67a332305ce6d1b94cd0f7860/zone.go#L496-L508) returns a [`ZoneRatePlan`](https://github.com/cloudflare/cloudflare-go/blob/c69a36b9a2daf6c67a332305ce6d1b94cd0f7860/zone.go#L62-L70) which has the `Id` ("free" in this case). You've been relying on `Name` which is the "Free Website" value.

> When changing to Free Website to free, the test fails

Yep, you'll need to `assert` on `Zone.Plan.Id` instead of `Zone.Plan.Name` to get this to work.

cloudflare/resource_cloudflare_zone.go
jacobbednarz: IIRC, this is a discussion we've had with our TAM and the "legacy_id" reference is named such as Cloudflare transitioned to UUIDs instead of generic strings for identifiers. 

Having a look at `AvailableRatePlans` and what `ZoneDetails` returns, there isn't a common field between the two which will make association a touch harder.

```go
// AvailableRatePlans
(cloudflare.ZoneRatePlan) {
  ID: (string) (len=10) "enterprise",
  Name: (string) (len=6) "CF_ENT",
  Price: (int) 0,
  Currency: (string) (len=3) "USD",
  Duration: (int) 0,
  Frequency: (string) (len=7) "monthly",
  Components: ([]cloudflare.zoneRatePlanComponents) (len=4 cap=4) {
   (cloudflare.zoneRatePlanComponents) {
    Name: (string) (len=22) "dedicated_certificates",
    Default: (int) 0,
    UnitPrice: (int) 5
   },
   (cloudflare.zoneRatePlanComponents) {
    Name: (string) (len=29) "dedicated_certificates_custom",
    Default: (int) 0,
    UnitPrice: (int) 10
   },
   (cloudflare.zoneRatePlanComponents) {
    Name: (string) (len=10) "page_rules",
    Default: (int) 125,
    UnitPrice: (int) 0
   },
   (cloudflare.zoneRatePlanComponents) {
    Name: (string) (len=5) "zones",
    Default: (int) 1,
    UnitPrice: (int) 0
   }
  }
 }
```

```go
// ZoneDetails
Plan: (cloudflare.ZoneRatePlan) {
  ID: (string) (len=32) "94f3b7b768b0458b56d2cac4fe5ec0f9",
  Name: (string) (len=18) "Enterprise Website",
  Price: (int) 0,
  Currency: (string) (len=3) "USD",
  Duration: (int) 0,
  Frequency: (string) (len=7) "monthly",
  Components: ([]cloudflare.zoneRatePlanComponents) <nil>
 }
```

Perhaps instead of trying to associate the two through existing methods, we just statically define the plan levels (free, enterprise, etc) and then have a fuzzy match function that ties it to an existing plan and errors if it doesn't match? Similar to your `setRatePlan` method but just using a static definition of available plans. Here is a diff demonstrating the idea

```diff
diff --git a/cloudflare/resource_cloudflare_zone.go b/cloudflare/resource_cloudflare_zone.go
index eeea482..51b308e 100644
--- a/cloudflare/resource_cloudflare_zone.go
+++ b/cloudflare/resource_cloudflare_zone.go
@@ -10,8 +10,11 @@ import (
 	cloudflare "github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/pkg/errors"
 )
 
+var validPlanTypes = []string{"free", "business", "pro", "enterprise"}
+
 func resourceCloudflareZone() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceCloudflareZoneCreate,
@@ -51,8 +54,7 @@ func resourceCloudflareZone() *schema.Resource {
 				Type:         schema.TypeString,
 				Computed:     true,
 				Optional:     true,
-				//TODO ask for complete Enum
-				ValidateFunc: validation.StringMatch(regexp.MustCompile("^([a-zA-Z0-9 ]+)$"), "Not a valid plan: use Free Website, Pro Plan, Business, Enterprise"),
+				ValidateFunc: validation.StringInSlice(validPlanTypes, false),
 			},
 			"meta": {
 				Type:     schema.TypeMap,
@@ -115,10 +117,6 @@ func resourceCloudflareZoneCreate(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
-	if err := setRatePlan(d, client, zone.ID); err != nil {
-		return err
-	}
-
 	return resourceCloudflareZoneRead(d, meta)
 }
 
@@ -147,7 +145,12 @@ func resourceCloudflareZoneRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("name_servers", zone.NameServers)
 	d.Set("meta", flattenMeta(d, zone.Meta))
 	d.Set("zone", zone.Name)
-	d.Set("plan", zone.Plan.Name)
+
+	planName, err := normalisePlanName(zone.Plan.Name)
+	if err != nil {
+		return err
+	}
+	d.Set("plan", planName)
 
 	return nil
 }
@@ -168,10 +171,6 @@ func resourceCloudflareZoneUpdate(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
-	if err := setRatePlan(d, client, zoneID); err != nil {
-		return err
-	}
-
 	return resourceCloudflareZoneRead(d, meta)
 }
 
@@ -201,28 +200,22 @@ func flattenMeta(d *schema.ResourceData, meta cloudflare.ZoneMeta) map[string]in
 	return cfg
 }
 
-func setRatePlan(d *schema.ResourceData, client *cloudflare.API, zoneID string) error {
-	if planName, ok := d.GetOk("planName"); ok {
-		ratePlan, err := getZonePlanIDFor(client, zoneID, planName.(string))
-		if err != nil {
-			return fmt.Errorf("Error fetching planName %s for zone %q: %s", planName, zoneID, err)
-		}
-		if _, err := client.ZoneSetRatePlan(zoneID, *ratePlan); err != nil {
-			return fmt.Errorf("Error setting planName %s for zone %q: %s", planName, zoneID, err)
-		}
-	}
-	return nil
+func normalisePlanName(planName string) (string, error) {
+	planName = strings.ToLower(planName)
+	planName = strings.Replace(planName, " website", "", 1)
+
+	if !sliceContains(planName, validPlanTypes) {
+		return "", errors.New("invalid plan name detected")
 	}
 
-func getZonePlanIDFor(client *cloudflare.API, zoneID, planName string) (*cloudflare.ZoneRatePlan, error) {
-	plans, err := client.AvailableZoneRatePlans(zoneID)
-	if err != nil {
-		return nil, err
+	return planName, nil
 }
-	for _, p := range plans {
-		if strings.EqualFold(p.Name, planName) {
-			return &p, nil
+
+func sliceContains(a string, list []string) bool {
+	for _, b := range list {
+		if b == a {
+			return true
 		}
 	}
-	return nil, fmt.Errorf("Not found amongst plans")
+	return false
 }
diff --git a/cloudflare/resource_cloudflare_zone_test.go b/cloudflare/resource_cloudflare_zone_test.go
index 45043e6..ff01392 100644
--- a/cloudflare/resource_cloudflare_zone_test.go
+++ b/cloudflare/resource_cloudflare_zone_test.go
@@ -20,7 +20,7 @@ func TestZone(t *testing.T) {
 					resource.TestCheckResourceAttr(name, "zone", "example.org"),
 					resource.TestCheckResourceAttr(name, "paused", "true"),
 					resource.TestCheckResourceAttr(name, "name_servers.#", "2"),
-					resource.TestCheckResourceAttr(name, "plan", "Free Website"),
+					resource.TestCheckResourceAttr(name, "plan", "free"),
 				),
 			},
 		},

```

While looking into this, I've found that `resourceCloudflareZoneUpdate` doesn't actually do much of an update right now (just pauses the domain). I'll open a separate issue for this.

inge4pres: Looks good, just when we are creating the zone in `resourceCloudflareZoneCreate` how do we instruct the zone to be assigned the desired plan?

vendor/vendor.json
inge4pres: Restored from master

cloudflare/resource_cloudflare_zone.go
patryk: No need for fuzzy logic. You may use `AvailableZoneRatePlans` to get the list of Plans that can be applied to a zone, including their IDs, which can be used to `PATCH` the zone resource.

patryk: You cannot do it in single step, you'll need to edit the zone after creation with `ZoneSetRatePlan` or `EditZone`.

inge4pres: Agreed :) I will leave the `setRatePlan` function in the create and update function

inge4pres:
@jacobbednarz thanks for the hint, however the ID returned is not what we expect. The [documentation](https://api.cloudflare.com/#zone-zone-details) of Cloudflare API ZoneDetails as implemented in cloudflare-go client is also wrong as instead of returning a struct like the one in the link, it returns this struct

```type ZoneRatePlan struct {
	ID         string                   `json:"id"`
	Name       string                   `json:"name,omitempty"`
	Price      int                      `json:"price,omitempty"`
	Currency   string                   `json:"currency,omitempty"`
	Duration   int                      `json:"duration,omitempty"`
	Frequency  string                   `json:"frequency,omitempty"`
	Components []zoneRatePlanComponents `json:"components,omitempty"`
}```

where the ID field is "0feeeeeeeeeeeeeeeeeeeeeeeeeeeeee". When changing the test using `free` and the implementation with `d.Set("plan", zone.Plan.ID)` 

```
=== RUN   TestZone
--- FAIL: TestZone (3.57s)
    testing.go:527: Step 0 error: Check failed: Check 4/4 error: cloudflare_zone.test: Attribute 'plan' expected "free", got "0feeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
```

So why not just use the descriptive names?

**Comment by jacobbednarz** [comment]
@patryk mentioned we don't need the fuzzy logic (which is the simplest way forward); I was just concerned about using the descriptive values in configuration as Cloudflare is more than just websites now and I felt including that would make some of these configurations would be a misnomer. It's not a big deal though, just personal preference which is weakly held.

In summary, I think now we just need to update the `plan` `ValidateFunc` to restrict those values and this should be close to being wrapped up. All the other feedback has been addressed from my side.

File: website/docs/r/zone.html.markdown (hunk 8 above)
Comment (inge4pres) [reply website/docs/r/zone.html.markdown]: As discussed, used literal names from Cloudflare-go

File: cloudflare/resource_cloudflare_zone_test.go (hunk 6 above)
Comment (inge4pres) [reply cloudflare/resource_cloudflare_zone_test.go]: Not changed as discussed

**Comment by inge4pres** [comment]
Last commit passed the single acceptance test on added `plan`; documentation has been updated. Waiting for your LGTM :)

File: cloudflare/resource_cloudflare_zone.go [hunk 9]
```diff
@@ -47,6 +47,12 @@ func resourceCloudflareZone() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"plan": {
+				Type:         schema.TypeString,
+				Computed:     true,
```
jacobbednarz: I don't think we want the computed option here. We have the ability to set this in the resource which negates the need for computing it later (such as a UUID or interpolated values)

```suggestion
```

jacobbednarz:
I'm not sure if I missed it before or it was swapped in the tidy up but we don't need the `Computed` option on the plan schema. 

`Computed` is used for schema values that the user cannot define and are generally generated one offs or at runtime. The `plan` value here isn't that. 

I'm 👍🏻 once that is fixed.

inge4pres:
Thanks @jacobbednarz for clarifying 😄applied your suggestion

inge4pres:
I guess this is just waiting for @patryk to approve?

cloudflare/resource_cloudflare_zone.go
```diff
@@ -47,6 +47,11 @@ func resourceCloudflareZone() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"plan": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice([]string{"Free Website", "Business Website", "Pro Website", "Enterprise Website"}, false),
```
patryk: Would be better to use `free`, `pro`, `business`, `enterprise` here and convert internally between them and API view. e.g.
`free` <=> `Free Website`

jacobbednarz: Apologies @patryk but I think there is a little bit of confusion here as an earlier comment mentioned we don't want to apply any fuzzy matches to the values and instead you wanted the full values. I might have misunderstood your intentions so would you mind clarifying again to avoid extra back and forth? Perhaps a patch demonstrating what changes you intend here?

inge4pres: @patryk please let us know... this is just a small change compared to my next PR 😄

inge4pres:
Been 2 weeks since no activity, and for us this is quite important.
We are Enterprise customers of Cloudflare and are trying to fill the gap in the Terraform provider for such features.
Please let us know if this can be carried on, currently we're using a mixture of custom built provider and curls to Cloudflare API and we'd really like to streamline into Terraform the whole thing.
Understand you (Hashicorp) are busy with the 0.12 release but we would like to help if we can.

cloudflare/resource_cloudflare_zone.go
inge4pres: Let me summarize to see if I got this: you'd still be ok to use the ID as they are now, but instead of maping them with their descriptive names in the HCL and docs you'd prefer to show the legacy IDs to the user?
This means keeping a 1:1 match between the descriptive name of the plan and the legacy ID, like a `map[string]string` in a private const?

inge4pres:
No worries, thanks for getting back, just trying to see this one go through 😄 
If I got the matter right in the last unresolved discussion just let me know and I should be able to push the required code not long after your confirmation

inge4pres:
@patryk @jacobbednarz [last commit](https://github.com/terraform-providers/terraform-provider-cloudflare/pull/160/commits/1dfae34c238ab8741e72eb7dee709c1bf534cbea) bundles everything discussed and should be good to go, let me know in case it isn't

cloudflare/resource_cloudflare_zone.go
patryk: Yes, this is what I meant. In the future we will clean this mess up, but IDs will stay the same. Thanks!

PR #162, opened 2018-11-23
Allow `zone_id` to set `zone` and vice versa

### LINKED ISSUE #161: cloudflare_filter import issue - zone not set
- terraform import for cloudflare_filter resource forces a new resource on next apply as zone is not added up on import.
`    zone:        "" => "domain.name" (forces new resource)`

- terraform import for cloudflare_filter resource requires a zone_id rather than zone unlike the other resources.  We should be consistent.
`$ terraform import cloudflare_filter.default d41d8cd98f00b204e9800998ecf8427e/9e107d9d372bb6826bd81d3542a419d6`

provider.cloudflare 1.9

jacobbednarz:
During an import of the `clouflare_filter` resource, the `zone_id` is
within `resourceCloudflareFilterRead` to match the expected schema
resource. However, we don't currently set `zone` (the zone name)
anywhere. This becomes problematic on subsequent terraform operations as
the `Read` attempts to sync the schema with the filter and ends up
attempting to recreate the resource due to detecting a change.

Initially I thought we could just iterate over all zones and pull out
the one that matched the zone ID but this only works _if_ the zone ID
has been provided (which isn't guaranteed since both fields are optional
but one is required). Instead, I've opted to ensure we update both the
`zone` and `zone_id` properties when we have one of them. I.e when we
only have the `zone_id`, use that value to find and set the `zone` value
(and vice versa). This will ensure that both values are defined and a
change in either will retrigger the update.

Fixes #161.

### DIFF
--- a/cloudflare/resource_cloudflare_filter.go
+++ b/cloudflare/resource_cloudflare_filter.go
@@ -25,6 +25,7 @@ func resourceCloudflareFilter() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
+				Computed: true,
 			},
 			"zone_id": {
 				Type:     schema.TypeString,
--- a/cloudflare/resource_cloudflare_filter.go
+++ b/cloudflare/resource_cloudflare_filter.go
@@ -111,6 +111,26 @@ func resourceCloudflareFilterCreate(d *schema.ResourceData, meta interface{}) er
 func resourceCloudflareFilterRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*cloudflare.API)
 	zoneID := d.Get("zone_id").(string)
+	zoneName := d.Get("zone").(string)
+
+	if zoneID == "" {
+		zoneID, _ = client.ZoneIDByName(zoneName)
+	} else {
+		zones, err := client.ListZones()
+		if err != nil {
+			return fmt.Errorf("failed to lookup all zones: %s", err)
+		}
+
+		for _, zone := range zones {
+			if zone.ID == zoneID {
+				zoneName = zone.Name
+			}
+		}
+
+		if zoneName == "" {
+			return fmt.Errorf("failed to find zone name")
+		}
+	}
 
 	log.Printf("[DEBUG] Getting a Filter record for zone %q, id %s", zoneID, d.Id())
 	filter, err := client.Filter(zoneID, d.Id())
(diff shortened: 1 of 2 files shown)


cloudflare/resource_cloudflare_filter.go
```diff
@@ -111,6 +111,26 @@ func resourceCloudflareFilterCreate(d *schema.ResourceData, meta interface{}) er
...
+			return fmt.Errorf("failed to lookup all zones: %s", err)
+		}
+
+		for _, zone := range zones {
+			if zone.ID == zoneID {
+				zoneName = zone.Name
+			}
+		}
+
+		if zoneName == "" {
```
jacobbednarz: this is a safe guard to ensure that if we do get to this point, we really do have a `zoneName` set

patryk: We should not need to check for `zoneID` being empty here, will be always populated by `Create` or `Import`.
We should also mark `zone` field as `Computed: true`.

patryk: `Computed` here means that the code may set it itself, so it should not be taken into consideration on diff unless explicitly set (yes, `Computed` fields can be set in HCL).

Because now in `Create` we set `zone_id` if we provide `zone`, but we will be setting `zone` when we import with `zone_id` only, we need both to be `Computed`.

I made a mistake coding in the `zone` field, I was young and stupid at that time ;)
We really need to clean `zone` vs `zone_id` mess in v2.0 😰

jacobbednarz: > `Computed` here means that the code may set it itself, so it should not be taken into consideration on diff unless explicitly set (yes, `Computed` fields can be set in HCL).

TIL, other uses of this that I have seen have all been for things like UUIDs or generating values at runtime.

> We really need to clean `zone` vs `zone_id` mess in v2.0 😰

Do you want to start a v2.0 issue where we can keep track of these things so that we have a record of it?

jacobbednarz:
I've updated this to mark `zone` as `Computed` and removed the additional guard clause via b9ab72e and b14b7f8  respectively.

cloudflare/resource_cloudflare_filter.go
patryk: Yup, good idea. Issue and/or milestone (will need to read GitHub guides on this).

PR #174, opened 2018-12-04
Allow `zone_id` to set `zone` and vice versa

### LINKED ISSUE #165: cloudflare_firewall_rule import issue - zone not set
terraform import for cloudflare_firewall_rule resource forces a new resource on next apply as zone is not added up on import.
`zone: "" => "domain.name" (forces new resource)`

Similar to #161

provider.cloudflare 1.9

jacobbednarz:
During an import of the `cloudflare_firewall_rule` resource, the `zone_id` is
within `resourceCloudflareFirewallRuleRead` to match the expected schema
resource. However, we don't currently set zone (the zone name)
anywhere. This becomes problematic on subsequent terraform operations as
the Read attempts to sync the schema with the firewall rule and ends up
attempting to recreate the resource due to detecting a change.

This is a similar issue to #161 and is addressed in a similar manner.

Fixes #165

### DIFF
--- a/cloudflare/resource_cloudflare_firewall_rule.go
+++ b/cloudflare/resource_cloudflare_firewall_rule.go
@@ -25,6 +25,7 @@ func resourceCloudflareFirewallRule() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
+				Computed: true,
 			},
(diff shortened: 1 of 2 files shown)


patryk:
Yup, you're right. Thanks!

PR #175, opened 2018-12-04
Ensure `zone` and `zone_id` are always set

### LINKED ISSUE #166: cloudflare_access_rule import issue - zone not set
importing a zone based access rule doesn't set zone forcing a new resource next plan/apply.

```
terraform import cloudflare_access_rule.whitelist_mywhitelist zone/85e6c246febfda393702006b823326ab/e9536f2f59b330327d3292c23b508ea9
Import successful!
```

on next plan
`      zone:                 "" => "domain.name" (forces new resource)`

Similar to #161 and #165 

provider.cloudflare 1.9

jacobbednarz:
During an import of the `cloudflare_access_rule` resource, the `zone_id` is
within `resourceCloudflareAccessRuleRead` to match the expected schema
resource. However, we don't currently set zone (the zone name) anywhere.
This becomes problematic on subsequent terraform operations as the Read
attempts to sync the schema with the access rule and ends up attempting to
recreate the resource due to detecting a change.

This is a similar issue to #161 and is addressed in a similar manner.

Fixes #166

### DIFF
--- a/cloudflare/resource_cloudflare_access_rule.go
+++ b/cloudflare/resource_cloudflare_access_rule.go
@@ -94,11 +94,7 @@ func resourceCloudflareAccessRuleCreate(d *schema.ResourceData, meta interface{}
 			r, err = client.CreateUserAccessRule(newRule)
 		}
 	} else {
-		var zoneID string
-
-		if zoneID != "" {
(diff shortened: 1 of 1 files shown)


patryk:
Thanks!

PR #176, opened 2018-12-05
'cache_key' now ignored for Page Rule updates

grogancolin:
The Page Rule resource will now ignore the `cache_key` field when
performing an Update.

As the `cache_key` can _only_ be created via Cloudflare Support, and
this would be done outside of terraform, it means no Page Rules with a
custom `cache_key` can be managed by terraform.
By ignoring (at least, until the Cloudflare API is updated as mentioned in https://github.com/terraform-providers/terraform-provider-cloudflare/issues/117)
people can go forward with using the provider until that work is
completed.

### DIFF
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -441,6 +441,9 @@ func resourceCloudflarePageRuleUpdate(d *schema.ResourceData, meta interface{}) 
 
 		for _, action := range actions {
 			for id, value := range action.(map[string]interface{}) {
+				if contains(pageRuleKeysToIgnoreOnUpdate, id) {
+					continue
+				}
 				newPageRuleAction, err := transformToCloudflarePageRuleAction(id, value)
 				if err != nil {
 					return err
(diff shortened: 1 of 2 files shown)


garrettgalow:
There is an open ticket that we are investigating internally for the inability to PATCH a page rule that Jacob mentioned.

jacobbednarz:
I've just got a reply back on this issue that it is currently being looked at and is intending to be addressed in an upcoming release. No word on release date but this PR isn't required anymore a the source of the issue will be addressed.

We have a couple of niggles in this area so I'm happy to follow this ticket and update anything we need in the provider based on our testing.

jacobbednarz:
I've got word back that this functionality should be fixed within the API now. I'm going to test it manually and then confirm the functionality in the provider is setup correctly.

grogancolin:
Great to hear, thanks for the update Jacob.

jacobbednarz:
Great news! I've confirmed this working manually and with the Terraform provider. @grogancolin would you mind confirming this caters for your `cache_key` use case too? (We don't rely on it anymore so can't test but other fields look good!)

patryk:
Awesome, thanks everyone for contribution!

PR #183, opened 2018-12-20
Page rule `zone` attribute change to trigger new resource

### LINKED ISSUE #180: Changing a Page Rule's zone doesn't force recreation
### Terraform Version
```
Terraform v0.11.8
+ provider.cloudflare v1.10.0
```

### Affected Resource(s)
- cloudflare_page_rule

```hcl
variable "cloudflare_api_key" {
  type = "string"
}

variable "cloudflare_email" {
  type = "string"
}

provider "cloudflare" {
  version = "~> 1.10.0"
  email   = "${var.cloudflare_email}"
  token   = "${var.cloudflare_api_key}"
}

variable "domain" {
  type = "string"
}

resource "cloudflare_page_rule" "strict_ssl" {
  zone   = "${var.domain}"
  target = "https://foobar.${var.domain}/"

  actions = {
    ssl = "strict"
  }
}
```

### Expected Behavior
When the Page Rule zone has changed, it should destroy the Page Rule in the old zone and create a new one in the new zone.

### Actual Behavior
It tries to modify the Page Rule in place, which leads to an error message from the Cloudflare API similar to the following:

```
* cloudflare_page_rule.strict_ssl: Failed to update Cloudflare Page Rule: error from makeRequest: HTTP status 400: content "{\"success\":false,\"errors\":[{\"code\":1004,\"message\":\"Page Rule validation failed: See messages for details.\"}],\"messages\":[{\"code\":1,\"message\":\".targets[0]: Your URL should reference the domain '<thedomain>' in some way.\",\"type\":null}],\"result\":null}"
```

### Steps to Reproduce
Prerequisites:
- The above configuration
- `cloudflare_email` and `cloudflare_api_key` in `terraform.tfvars`
- 2 domains in a Cloudflare account (in this example, `domain1.com` and `domain2.com`)

Steps:
1. `terraform init`
2. `terraform apply -var 'domain=domain1.com'`
2. `terraform apply -var 'domain=domain2.com'`

jacobbednarz:
#180 raised an issue whereby using a `cloudflare_page_rule` resource
with a single resource identifier but multiple variables to change the
`zone` attribute would attempt to update the page rule in place instead
of re-creating it (see the attached issue for full reproduction
steps). This isn't good as the `target` attribute needs to contain
references to the `zone` in order to be successfully updated.

To fix this, the schema details for `zone` has been updated to force a
new resource on change. On the surface this change is fine however I can
see a couple of scenarios where a `DiffSuppressFunc` that checks the
host against the stored state file host might be needed in order to not
create and delete unnecessarily and instead only create a new resource
if the hosts are different.

Fixes #180

### DIFF
--- a/cloudflare/provider_test.go
+++ b/cloudflare/provider_test.go
@@ -44,6 +44,12 @@ func testAccPreCheck(t *testing.T) {
 	}
 }
 
+func testAccPreCheckAltDomain(t *testing.T) {
+	if v := os.Getenv("CLOUDFLARE_ALT_DOMAIN"); v == "" {
+		t.Fatal("CLOUDFLARE_ALT_DOMAIN must be set for this acceptance test")
+	}
+}
+
 func testAccPreCheckOrg(t *testing.T) {
 	if v := os.Getenv("CLOUDFLARE_ORG_ID"); v == "" {
 		t.Fatal("CLOUDFLARE_ORG_ID must be set for this acceptance test")
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -26,6 +26,7 @@ func resourceCloudflarePageRule() *schema.Resource {
 			"zone": {
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 			},
 
 			"zone_id": {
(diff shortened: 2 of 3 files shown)


patryk:
Thanks!

PR #185, opened 2018-12-24
Update page_rule.html.markdown

m1kola:
It was missing mention of the `explicit_cache_control` action.


patryk:
Thanks @m1kola !

PR #186, opened 2018-12-31
Document provider configuration via ENV Variables

rtizzy:
As noted in title, I documented some information that wasn't already covered.


patryk:
Thanks @rtizzy, as @jacobbednarz mentioned we have it documented already. Closing.

PR #197, opened 2019-01-23
Add log action for firewall rules

Johnee:
CloudFlare has an `Enterprise plan` which allows firewall rules with an action of type `log`.

While this works exclusively for Enterprise users, it is something that should be available in the Terraform provider itself.

### DIFF
--- a/website/docs/r/firewall_rule.markdown
+++ b/website/docs/r/firewall_rule.markdown
@@ -36,7 +36,7 @@ The following arguments are supported:
 
 * `zone` - (Optional) The DNS zone to which the Firewall Rule should be added. Will be resolved to `zone_id` upon creation.
 * `zone_id` - (Optional) The DNS zone to which the Filter should be added.
-* `action` - (Required) The action to apply to a matched request. Allowed values: "block", "challenge", "allow", "js_challenge".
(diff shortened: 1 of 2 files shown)


website/docs/r/firewall_rule.markdown
```diff
@@ -36,7 +36,7 @@ The following arguments are supported:
 
 * `zone` - (Optional) The DNS zone to which the Firewall Rule should be added. Will be resolved to `zone_id` upon creation.
 * `zone_id` - (Optional) The DNS zone to which the Filter should be added.
-* `action` - (Required) The action to apply to a matched request. Allowed values: "block", "challenge", "allow", "js_challenge".
+* `action` - (Required) The action to apply to a matched request. Allowed values: "block", "challenge", "allow", "js_challenge", "log".
```
jacobbednarz: Let’s call out that log is Enterprise specific to avoid any confusion with other plan types trying to use this value.

Johnee: I've changed the wording to be `Enterprise plan also allows "log".`

patryk:
Thanks for contribution, Jan!

PR #218, opened 2019-02-26
test/cloudflare_page_rule: Remove not working unitary field updates

jacobbednarz:
A little while ago, Cloudflare [fixed the page rule PATCH endpoint](https://github.com/terraform-providers/terraform-provider-cloudflare/pull/176) to
handle individual attribute updates however this unearthed a few
problems. Most notably, you're unable to delete an individual page rule
attribute without calling the PUT endpoint and [replacing the rule
itself](https://github.com/terraform-providers/terraform-provider-cloudflare/issues/198) (which has it's own issues in some cases).

To get the acceptance test suite green, I'm ripping the unitary test
assertion out as it's currently broken and without a fix from
Cloudflare's API, there isn't much we can do about it so there is little
point testing it as-is.

Fixes CI failure

```
------- Stdout: -------
=== RUN   TestAccCloudflarePageRule_Updated
--- FAIL: TestAccCloudflarePageRule_Updated (2.48s)
    testing.go:527: Step 1 error: Check failed: Check 2/4 error: 'disable_apps' found at api, but we should have removed it
FAIL
```

FYI @radeksimko

### DIFF
--- a/cloudflare/resource_cloudflare_page_rule_test.go
+++ b/cloudflare/resource_cloudflare_page_rule_test.go
@@ -302,14 +302,6 @@ func testAccCheckCloudflarePageRuleAttributesBasic(pageRule *cloudflare.PageRule
 			return fmt.Errorf("'always_online' not specified at api")
 		}
 
-		if val, ok := actionMap["disable_apps"]; ok {
-			if val != nil {
-				return fmt.Errorf("'disable_apps' is a unitary value, expect nil value at api, but found: '%v'", val)
-			}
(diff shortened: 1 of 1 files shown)


jacobbednarz:
I covered this question a bit in the linked threads but the short version is either way, it's broken for someone. If we go back to using PUT for all requests, it will potentially cause some breakages for those using things like custom cache keys and others with page rule priority issues. 

Without a fix from cloudflare on the API itself, it's not really a good experience.

PR #226, opened 2019-03-01
Organization api is deprecated

pollosp:
Hello 👋 guys.

From cloudflare doc
```
This API is deprecated, please use equivalent /accounts API endpoints where possible. Account APIs provide a broader range of features, and are backwards-compatible to organization APIs.
```

https://api.cloudflare.com/#organizations-properties 

In the  cloudflare-go client they are using accounts endpoint https://github.com/cloudflare/cloudflare-go/blob/master/workers.go#L197 but still using OrganizationID instead accountID, I think is another PR there and here to be consistent.
Because in all doc they refer to this as `account_identifier` or `account_id`
```
PUT accounts/:account_identifier/workers/scripts/:script_name
```

To give you context, I'm doing this PR because I spent some time today looking for Organization ID and it isn't available anymore and is not obious that account id is the samething at least it wasn’t for me today 🤔

Problem of this PR is that it is a breaking change with the current interface and it is going to  affect users. I know some resource notify before they are going to be deprecated , don’t know if there is something similar to attributes, I will appreciate help here in order to keep the rosource working or backward compatible PR if you see this change worthy.

### DIFF
--- a/cloudflare/provider.go
+++ b/cloudflare/provider.go
@@ -75,10 +75,10 @@ func Provider() terraform.ResourceProvider {
 				Description: "If specified zone is owned by an organization, configure API client to always use that organization",
 			},
 
-			"org_id": &schema.Schema{
+			"account_id": &schema.Schema{
 				Type:        schema.TypeString,
 				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("CLOUDFLARE_ORG_ID", nil),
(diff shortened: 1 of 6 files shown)


pollosp:
@jacobbednarz Totally agree, I'm going to change thus PR only updating the doc as you tool is a soft change.

holms:
This would be nice to have :)

jacobbednarz:
This has been noting for the next breaking change release so I’m going to close this unless someone else wants to take a pass at the documentation to clarify the variable intents.

PR #230, opened 2019-03-04
Move to Go Modules

### LINKED ISSUE #206: [PROPOSAL] Switch to Go Modules
As part of the preparation for Terraform v0.12, we would like to migrate all providers to use [Go Modules](https://github.com/golang/go/wiki/Modules). We plan to continue checking dependencies into vendor/ to remain compatible with existing tooling/CI for a period of time, however go modules will be used for management. Go Modules is the official solution for the go programming language, we understand some providers might not want this change yet, however we encourage providers to begin looking towards the switch as this is how we will be managing all Go projects in the future. Would maintainers please react with :+1: for support, or :-1: if you wish to have this provider omitted from the first wave of pull requests. If your provider is in support, we would ask that you avoid merging any pull requests that mutate the dependencies while the Go Modules PR is open (in fact a total codefreeze would be even more helpful), otherwise we will need to close that PR and re-run `go mod init`. Once merged, dependencies can be added or updated as follows:

```
$ GO111MODULE=on go get github.com/some/module@master
$ GO111MODULE=on go mod tidy
$ GO111MODULE=on go mod vendor
```

GO111MODULE=on might be unnecessary depending on your environment, this example will fetch a module @ master and record it in your project's go.mod and go.sum files. It's a good idea to tidy up afterward and then copy the dependencies into vendor/. To remove dependencies from your project, simply remove all usage from your codebase and run:

```
$ GO111MODULE=on go mody tidy
$ GO111MODULE=on go mod vendor
```

Thank you sincerely for all your time, contributions, and cooperation!

jacobbednarz:
Updates the provider to move to `go mod` for dependency management along
with some helpful documentation on how we expect dependency management
to work.

Closes #206

cc @appilon @radeksimko


appilon:
Looking good so far thank you @jacobbednarz . We would just ask for a few extra tweaks in our effort to keep all the providers consistent.

1) Could you please add the following to the travis config to ensure compatibility
```yaml
env:
  - GO111MODULE=on GOFLAGS=-mod=vendor
```
I also noticed this travis config contains several versions of Go. I would recommend just sticking to one, while we continue to vendor for the time being, eventually we plan to fully adopt go modules and older versions (pre 1.11) would not be supported.

2) This also means that for development go1.11 is required, could you update the README to reflect that.

3) Unfortunately go modules has had some growing pains through go1.11's releases, as such we have begun encoding the specific version of go used into a `goenv` file: `.go-version`. In this case I can see you have used go1.12 (it was a new change from go1.11 to go1.12 that they would add a line of the version of go used for modules in the go.mod file). I think right now we prefer that file to have `1.11.5`, but `1.12.0` is likely fine (basically anything >= go1.11.4 is fine, there was a bug prior to that release).

4) Could you delete all mention/use of `govendor` and the `vendor-status` make target from the travis config and makefile. Go modules does not have an equivalent to `govendor status`

5) (Optional) We have a reference to a version of Terraform v0.11 that uses go modules, this will significantly clean up this project's `go.mod` file. To update, simply run
```
GO111MODULE=on go get github.com/hashicorp/terraform@sdk-v0.11-with-go-modules
GO111MODULE=on go mod tidy
GO111MODULE=on go mod vendor
```

My apologies I know that is a lot of little requests. To better illustrate the changes I am linking a merged PR from another provider. Thank you for your time and contribution 😄 !
https://github.com/terraform-providers/terraform-provider-gitlab/pull/95/files

jacobbednarz:
> Could you please add the following to the travis config to ensure compatibility
>
> ```yaml
> env:
>   - GO111MODULE=on GOFLAGS=-mod=vendor
> ```
> 
> I also noticed this travis config contains several versions of Go. I would recommend just sticking to one, while we continue to vendor for the time being, eventually we plan to fully adopt go modules and older versions (pre 1.11) would not be supported.

Sounds good, addressed in 79dbf68. I agree we should only test on the single version considering we intend to package it up anyway. I was only perusing the Travis CI UI when you mentioned this and realised we weren't actually using go modules so great find!

> This also means that for development go1.11 is required, could you update the README to reflect that.

Fixed in 3db01ea 

> Unfortunately go modules has had some growing pains through go1.11's releases, as such we have begun encoding the specific version of go used into a goenv file: .go-version. In this case I can see you have used go1.12 (it was a new change from go1.11 to go1.12 that they would add a line of the version of go used for modules in the go.mod file). I think right now we prefer that file to have 1.11.5, but 1.12.0 is likely fine (basically anything >= go1.11.4 is fine, there was a bug prior to that release).

I added the `.go-version` file in 2f80b43 however for anyone following along with this, I needed to manually fetch `1.11.5` to get it to work due to the version not being in the latest 1.x release of `goenv` (via Homebrew, probably won't be an issue if you clone it).

```
$ brew install goenv
$ cd $(brew --prefix)/Cellar/goenv/1.23.3/plugins/go-build/share/go-build
$ wget https://raw.githubusercontent.com/syndbg/goenv/master/plugins/go-build/share/go-build/1.11.5
$ goenv install 1.11.5
```

>Could you delete all mention/use of govendor and the vendor-status make target from the travis config and makefile. Go modules does not have an equivalent to govendor status

Thanks for raising this. I did remove it in the README but missed the Makefile references. Cleaned up in b6c22cc.


>  (Optional) We have a reference to a version of Terraform v0.11 that uses go modules, this will significantly clean up this project's `go.mod` file.

Awesome, done in 24495ca.

Have confirmed this PR builds and applies changes to infrastructure as expected.

appilon:
@jacobbednarz I ran the acceptance tests, there are no new failures except 1 test that is reported as being flaky, feel free to merge this whenever

jacobbednarz:
Thanks for taking the time to review this one! I've also just had @radeksimko mention the same thing about the flakey test so I'll address that before getting a new release cut later today (potentially tomorrow).

PR #232, opened 2019-03-05
test/data_source_zones: Fix flakey tests due to domain count

jacobbednarz:
Updates two zone data source tests to account for situations where the
tests aren't being run against a clean or empty Cloudflare account which
results in flakey CI runs.

In both cases, the count was off by at least one due to the account
already have domains setup in them.


jacobbednarz:
As the reporter, @radeksimko do you mind casting your eyes over this one and checking this solves your issue?

PR #270, opened 2019-03-26
custom_page: Define `SetID` to save state

jacobbednarz:
Prior to this change we weren't setting the `Id` value for these
resources when creating them from new. If you imported, it worked as the
composite ID (`customPageLevel`/`identifier`/`pageType`) was defined.

To fix the issue, we're calling `d.SetId()` and setting the ID
explicitly as a md5 checksum of the `identifier` and `pageType`. This is
in order to maintain consistency with the other resources which use a 32
character ID.

Fixes #260

PR #273, opened 2019-03-28
Added missing SetId for custom_pages

mphoratiu:
For delete, the URL value cannot be nil, must match an URL.

Relates to https://github.com/terraform-providers/terraform-provider-cloudflare/issues/260

cc @jacobbednarz


jacobbednarz:
@mphoratiu My apologies, in my previous comment I forgot to link to the changes which has already implemented this functionality. Check out #270 which addresses this slightly better as it allows management of account and zone level configurations.

PR #275, opened 2019-03-28
data_source_zone: Return ID and Name

### LINKED ISSUE #211: Cloudflare zones datasource is not fetching all zones & doesn't return zone id
### Affected Resource(s)
- cloudflare_zones data source

### Expected Behavior
Should return all zones or at least give more configuration (as API permits: https://api.cloudflare.com/#zone-list-zones), like passing `page`, `per_page`....
Also returned data should provide more information about each zone (e.g.: zone ID) instead just returning the zone name.

### Actual Behavior
If there're more than 20 zones, only the first 20 are returned (I guess because of the default value). 
Each returned zone is just the zone name (no ID).

### Steps to Reproduce
1. Create more than 20 zones in Cloudflare.
1. 
```
data "cloudflare_zones" "test" {
  filter {}
}

output "zones" {
  value = "${data.cloudflare_zones.test.zones}"
}
```

### Important Factoids
This prevents using this data source to dynamically get zone IDs & forces zone ID hard coding.

### References
- [Cloudflare Zones API documentation](https://api.cloudflare.com/#zone-list-zones)

jacobbednarz:
Updates the `cloudflare_zones` data source to return a map of the ID and·
name for the filtered results. This is to better help facilitate dynamic·
configurations where end users would like to pass in a zone filter and·
automate creation of resources.

There are a couple of potential downsides though:

- The first is that this introduces a somewhat breaking change as it
 doesn't perform as it once did. An entry will be added to the CHANGELOG
 but it could be easy to miss.
- The second is that the syntax to access a single result is pretty
 ugly. This will be nicer in 0.12 but for now it's 🤮 .

Included in this PR is fixing all the spectrum integration tests. They
have been failing since they were created due to relying on a data
source that never made it into the provider. It now uses the working
`cloudflare_zones` data source and has been confirmed working.

Closes #211

### DIFF
--- a/cloudflare/resource_cloudflare_spectrum_application_test.go
+++ b/cloudflare/resource_cloudflare_spectrum_application_test.go
@@ -31,7 +31,7 @@ func TestAccCloudflareSpectrumApplication_Basic(t *testing.T) {
 					testAccCheckCloudflareSpectrumApplicationIDIsValid(name),
 					resource.TestCheckResourceAttr(name, "protocol", "tcp/22"),
 					resource.TestCheckResourceAttr(name, "origin_direct.#", "1"),
-					resource.TestCheckResourceAttr(name, "origin_direct.0", "tcp://192.0.2.1:23"),
+					resource.TestCheckResourceAttr(name, "origin_direct.0", "tcp://1.2.3.4:23"),
 					resource.TestCheckResourceAttr(name, "origin_port", "22"),
 				),
 			},
@@ -82,7 +82,7 @@ func TestAccCloudflareSpectrumApplication_Update(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckCloudflareSpectrumApplicationExists(name, &spectrumApp),
 					testAccCheckCloudflareSpectrumApplicationIDIsValid(name),
-					resource.TestCheckResourceAttr(name, "origin_direct.0", "tcp://192.0.2.1:23"),
+					resource.TestCheckResourceAttr(name, "origin_direct.0", "tcp://1.2.3.4:23"),
 				),
 			},
 			{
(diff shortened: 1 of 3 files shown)


cloudflare/resource_cloudflare_spectrum_application_test.go
```diff
@@ -31,7 +31,7 @@ func TestAccCloudflareSpectrumApplication_Basic(t *testing.T) {
 					testAccCheckCloudflareSpectrumApplicationIDIsValid(name),
 					resource.TestCheckResourceAttr(name, "protocol", "tcp/22"),
 					resource.TestCheckResourceAttr(name, "origin_direct.#", "1"),
-					resource.TestCheckResourceAttr(name, "origin_direct.0", "tcp://192.0.2.1:23"),
```
jacobbednarz: 192.0.2.1 was throwing validation errors so swapped it out to pass validation.

cloudflare/resource_cloudflare_spectrum_application_test.go
```diff
@@ -225,63 +225,69 @@ func testAccManuallyDeleteSpectrumApplication(name string, spectrumApp *cloudfla
 func testAccCheckCloudflareSpectrumApplicationConfigBasic(zoneName, ID string) string {
 	return fmt.Sprintf(`
 resource "cloudflare_spectrum_application" "%[2]s" {
-  zone_id  = "${data.cloudflare_zone.test.id}"
+  zone_id  = "${lookup(data.cloudflare_zones.test.zones[0], "id")}"
```
jacobbednarz: this isn't the prettiest thing out there but it works. open to suggestions on improvements

shiftie: `data.cloudflare_zones.test.zones.*.id[0]` ?

shiftie:
will this be available here: http://terraform-0.12.0-dev-snapshots.s3-website-us-west-2.amazonaws.com/terraform-provider-cloudflare/ for 0.12 beta usage?

jacobbednarz:
Not to my knowledge. I don’t know who maintains that or who owns it. Instead, you will have to build the provider locally until a release is cut.

shiftie:
ok thank you.

PR #276, opened 2019-03-29
Upgrade to Terraform 0.12.0-beta1

bltavares:
This commit bumps Terraform to work with the new version of Terraform.

It also cleans up the `go.mod` file. Now that our dependencies are also
using `go mod`, Go won't populate the `// indirect` on the file anymore.

There is a small caveat to compile the transitive dependencies without
needing to install `bazaar`, or having it failing to work on Windows. We
need to replace them with alternative sources that are not on Bazaar.

Related to: 
- https://github.com/terraform-providers/terraform-provider-cloudflare/issues/259
- https://github.com/hashicorp/terraform/pull/20870


jacobbednarz:
I’m still not sold on the stability of 0.12 just yet. Considering that many consumers of this provider run critical production workloads using it, I think we wait until it leaves beta or the milestone is empty of known issues. 

@patryk any thoughts here?

bltavares:
Sure thing @jacobbednarz. It could wait for an official release of the new version.

I was sharing more so that others could give this patch a try if they are inclined to test the new HCL2 features and use Cloudflare provider, like I'm doing :)

bltavares:
I've opened an issue with Terraform to avoid one of the suggested workarounds on this PR.

I'll close this PR until a stable version has been released, to avoid the workarounds as well.

patryk:
@jacobbednarz agreed. Although we're close to 0.12 release, it's not stable yet, and I'd want to wait for HashiCorp for giving green light first. Compiling against beta API is not good idea, provided there are production users of the plugin.

Issue #208, opened 2019-02-11
resource "cloudflare_load_balancer" has no `enabled` field


SteveGoldthorpe-Work:
### Terraform Version
Terraform v0.11.11
+ provider.cloudflare v1.11.0

### Affected Resource(s)
Please list the resources as a list, for example:
- cloudflare_load_balancer

Although not documented in the API docs, there is a parameter "enabled" which can have the value true or false.  This has the effect of enabling or disabling the load balancer.  The lack of this parameter means we turn the load balancer on or off or ensure it's in a given state.

SteveGoldthorpe-Work:
I think this needs an update in cloudflare-go first.

SteveGoldthorpe-Work:
Raised issue:
https://github.com/cloudflare/cloudflare-go/issues/272

jacobbednarz:
Awesome! Once this lands in `cloudflare-go` we can add it to Terraform. Until then, let's park this.

jacobbednarz:
This had landed upstream so if you want to add it, I'd happily review a PR for it. If not, I can take a look in the coming days.

SteveGoldthorpe-Work:
I've already got the code stashed, will update and create PR tomorrow.

jallentech:
Thanks, this is a really useful piece of work. I was just wondering whether it will be available in the next release and when the next release is likely to happen?

jacobbednarz:
It will be in the next release but no ETA yet on when it will be cut. Shouldn't be too far away though.

jallentech:
That's great, thanks for the update

jacobbednarz:
This was released in [1.13.0](https://github.com/terraform-providers/terraform-provider-cloudflare/blob/master/CHANGELOG.md#1130-april-12-2019)

PR #281, opened 2019-04-02
Add enabled field to load_balancer (#208)

SteveGoldthorpe-Work:
Adds support for enabled field in load-balancers (#208).
Load balancers that have been disabled via the dashboard (or other tooling) will want to change to enabled, you should add the field 'enabled: false' to them to match their current state.
Load balancers purely managed via terraform will be fine as there was no way to disable them before.


patryk:
Looks good, thank you @SteveGoldthorpe-Work !

Issue #107, opened 2018-08-20
browser_cache_ttl setting doesn't work


sixcorners:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v0.11.8
+ provider.cloudflare v1.2.0

### Affected Resource(s)
- cloudflare_zone_settings_override.settings

### Terraform Configuration Files
```hcl
resource "cloudflare_zone_settings_override" "settings" {
  name = "${var.cloudflare_zone}"

  settings {
    browser_cache_ttl = 0
  }
}

```

### Debug Output
https://gist.github.com/sixcorners/31d365e40d7aa20dec7f7d812c09d6b6

### Expected Behavior
What should have happened?
browser_cache_ttl should have been set to 0.

### Actual Behavior
What actually happened?
browser_cache_ttl didn't change.

### Steps to Reproduce
1. `terraform state rm cloudflare_zone_settings_override.settings && terraform apply`

### References

patryk:
Hey @sixcorners, the issue will be fixed in `v1.3.0`.

garrettgalow:
This doesn't appear to be fixed in the latest version:
```
cloudflare_page_rule.cf_test_Browser_cache: Modifying... (ID: a1c9307ff4dd125f8b50ecd4a23f6c04)
  actions.0.browser_cache_ttl: "0" => "10"
cloudflare_page_rule.cf_test_Browser_cache: Modifications complete after 0s (ID: a1c9307ff4dd125f8b50ecd4a23f6c04)

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
some_tf_example$ terraform apply
cloudflare_page_rule.cf_test_Browser_cache: Refreshing state... (ID: a1c9307ff4dd125f8b50ecd4a23f6c04)

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
some_tf_example$ vim cloudflare.tf
some_tf_example$ terraform apply
cloudflare_page_rule.cf_test_Browser_cache: Refreshing state... (ID: a1c9307ff4dd125f8b50ecd4a23f6c04)

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  ~ cloudflare_page_rule.cf_test_Browser_cache
[...]

PR #291, opened 2019-04-16
cloudflare_page_rule: Allow `browser_cache_ttl` of 0

jacobbednarz:
In c1799f4, @patryk made some improvements on how the `Update` handles
working out which rules need to be changed. With the introduction of
that change, the `transformToCloudflarePageRuleAction` now accepts a
third argument of whether the action has changed. For the majority of
rules, this isn't problematic however it has caused issues with
`browser_cache_ttl` as a value of 0 tells Cloudflare to respect the
origin cache control headers however, here it would end up setting the
page rule action value to `nil` (as it hadn't changed and was equal to
0).

To start with, I was going to remove the `intValue == 0` portion of the
check and just rely on whether the rule had been changed but this didn't
work due to the way disabling `edge_cache_ttl` works (it is set to 0
for removal). Instead, I moved the check into the conditional for the
`pageRuleAPIFloatFields` `elseif` which will allow the functionality for
both to work actions to work as desired.

Fixes https://github.com/terraform-providers/terraform-provider-cloudflare/issues/107#issuecomment-482353024

### DIFF
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -326,7 +326,7 @@ func resourceCloudflarePageRuleCreate(d *schema.ResourceData, meta interface{}) 
 			newPageRuleAction, err := transformToCloudflarePageRuleAction(id, value, false)
 			if err != nil {
 				return err
-			} else if newPageRuleAction.Value == nil {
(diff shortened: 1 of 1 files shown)


jacobbednarz:
@garrettgalow this addresses your issue from #107 (we had an internal team hit this too)

PR #316, opened 2019-04-30
page_rule: Add ability to manage minify action

### LINKED ISSUE #120: Add Auto-minify capability to Page Rules
Currently there is no support for Auto Minify in Page Rules `resource_cloudflare_page_rule.go`. Looks like it would be a quick add, given feature support at the zone level.

jacobbednarz:
Adds the ability to manage the `minify` action from within Page Rules.

Closes #120

### DIFF
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -245,6 +245,33 @@ func resourceCloudflarePageRule() *schema.Resource {
 							},
 						},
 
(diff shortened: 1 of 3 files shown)


patryk:
Because it's not a simple on/off toggle, we would benefit from test here.

jacobbednarz:
Sounds good @patryk. Added some more coverage in 92a39d4.

patryk:
That's much better, thanks.

PR #320, opened 2019-04-30
Fix page rule and spectrum application import acceptance tests

jacobbednarz:
This fixes a handful of our remaining broken acceptance tests. Two main
candidates:

- `page_rule`: Removes a bunch of the test setup that hasn't worked
  since it was introduced.
- `spectrum_application_import`: Adds the correct zone ID argument to
  perform the import with.


zackproser:
LGTM! :+1:

PR #321, opened 2019-05-01
Add sweeper for `cloudflare_zones`

### LINKED ISSUE #296: Implement acceptance test sweepers
While running acceptance tests in Hashicorp's CI, we often hit issues with left over resources which cause CI to fail. To work around this, we can use [sweepers](https://www.terraform.io/docs/extend/testing/acceptance-tests/sweepers.html). In essence, they are post test run tasks that remove any dangling resources and will assist with any failed resources that persist.

jacobbednarz:
Creates a [Terraform sweeper](https://www.terraform.io/docs/extend/testing/acceptance-tests/sweepers.html) to clean up any dangling resources. Here,
for Cloudflare zones that are being used a test for management via the
API. I've restricted the lookup to the ephemeral zones we want to
remove instead of all of them as I suspect removing all would make the
other tests fail that depend on it always being there.

Closes #296

### DIFF
--- a/cloudflare/data_source_zones_test.go
+++ b/cloudflare/data_source_zones_test.go
@@ -2,13 +2,44 @@ package cloudflare
 
 import (
 	"fmt"
+	"log"
+	"os"
 	"strconv"
 	"testing"
 
+	"github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform/terraform"
 
 	"github.com/hashicorp/terraform/helper/resource"
 )
+
+func init() {
+	resource.AddTestSweepers("cloudflare_zones", &resource.Sweeper{
+		Name: "cloudflare_zones",
+		F:    testSweepCloudflareZones,
+	})
+}
+
+func testSweepCloudflareZones(r string) error {
+	client, _ := cloudflare.New(os.Getenv("CLOUDFLARE_TOKEN"), os.Getenv("CLOUDFLARE_EMAIL"))
+	zones, _ := client.ListZones("baa.com", "baa.net", "baa.org", "foo.net")
+
+	if len(zones) == 0 {
+		log.Print("[DEBUG] No Cloudflare Zones to sweep")
+		return nil
+	}
+
+	for _, zone := range zones {
+		log.Printf("[INFO] Deleting Cloudflare Zone ID: %s", zone.ID)
+		_, err := client.DeleteZone(zone.ID)
+
+		if err != nil {
+			log.Printf("[ERROR] Failed to delete Cloudflare Zone (%s): %s", zone.Name, err)
+		}
+	}
+
+	return nil
+}
 
 func TestAccCloudflareZonesMatchName(t *testing.T) {
 	resource.Test(t, resource.TestCase{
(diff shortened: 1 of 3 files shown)


jacobbednarz:
If I understand how this works, we will need Hashicorp to update the SWEEPER_REGION CI environment variable to be "cloudflare_zones" for this to actually do anything.

@radeksimko Could you provide any insight here and perhaps a review that I'm not holding it wrong?

cloudflare/data_source_zones_test.go
```diff
@@ -2,14 +2,45 @@ package cloudflare
...
 
+func init() {
+	resource.AddTestSweepers("cloudflare_zones", &resource.Sweeper{
+		Name: "cloudflare_zones",
+		F:    testSweepCloudflareZones,
+	})
+}
+
+func testSweepCloudflareZones(r string) error {
+	client, _ := cloudflare.New(os.Getenv("CLOUDFLARE_TOKEN"), os.Getenv("CLOUDFLARE_EMAIL"))
```
radeksimko: Nitpick: Do you mind moving this setup line into a function, like `sharedClient() error`, or even better - create a more reusable function in `cloudflare/provider.go` that can be used by both provider code and tests?

cloudflare/data_source_zones_test.go
```diff
@@ -2,14 +2,45 @@ package cloudflare
...
+	zones, _ := client.ListZones("baa.com", "baa.net", "baa.org", "foo.net")
```
radeksimko: Could you please add error check here?

cloudflare/data_source_zones_test.go
jacobbednarz: Sure, I'll update it for the tests only as the provider auth setup needs some love and rejigging.

cloudflare/data_source_zones_test.go
jacobbednarz: Good pick up! I've included this as a part of 7878e60.

jacobbednarz:
Good idea on the resource naming. We've only been running this in a dedicated testing account so we haven't thought about that. I'll create an issue to go through and update all of these. Thanks!

cloudflare/data_source_zones_test.go
```diff
@@ -21,8 +19,15 @@ func init() {
...
-	client, _ := cloudflare.New(os.Getenv("CLOUDFLARE_TOKEN"), os.Getenv("CLOUDFLARE_EMAIL"))
-	zones, _ := client.ListZones("baa.com", "baa.net", "baa.org", "foo.net")
+	client, clientErr := sharedClient()
+	if clientErr != nil {
+		log.Fatalf("[ERROR] Failed to create Cloudflare client: %s", clientErr)
+	}
+
+	zones, zoneErr := client.ListZones("baa.com", "baa.net", "baa.org", "foo.net")
+	if zoneErr != nil {
+		log.Fatalf("[ERROR] Failed to fetch Cloudflare zones: %s", zoneErr)
```
radeksimko: Logging via `Fatal*` also has a "side effect" of `os.Exit(1)` which I don't think is necessary here? We can achieve pretty much the same by just returning the error, which also gives us the opportunity to perform any cleanup higher up the call stack.

jacobbednarz: Fair enough 👍

jacobbednarz:
@radeksimko This one should be ready for some more review again.

PR #333, opened 2019-05-06
page_rules: Fix logic error for `transformToCloudflarePageRuleAction`

jacobbednarz:
In order to be able to use `browser_cache_ttl` of 0 in page rules (to
respect the origin cache control value) I updated the required logic in
#291. Unfortunately, this only partially fixed the issue. Updating an
existing resource worked fine however creating a new resource fell into
the old behaviour due to using `||` where it should have been a `&&` to
ensure both conditions were met.


jacobbednarz:
This has uncovered another edge case that I'll hopefully address in another PR.

PR #334, opened 2019-05-06
page_rules: Make cache TTLs behave sensibly

jacobbednarz:
This changes the schema definition for `browser_cache_ttl` and
`edge_cache_ttl` from integers to strings (to be cast as integers
instead). This has come about because I've been battling getting
`browser_cache_ttl` to work consistently for the following scenarios:

- Creating a new action for a page rule where `browser_cache_ttl` is 0.
- Updating an action in a page rule where `browser_cache_ttl` is set to
  1 and I want to update it to 0.

The current logic where `browser_cache_ttl` and `edge_cache_ttl` are
grouped and treated similar made this quite difficult as only
`browser_cache_ttl` accepts 0 as a valid API value whereas 1 is the
minimum for `edge_cache_ttl`. To get around this, I've ditched the
checks that assumed 0 was a default value and instead explicitly set
the default for these keys as empty strings. This means they will end up
as `nil` and not get triggered for API updates but can still be set to
integers (cast as strings) for when you actually want those values.

One of the trade offs I've made here is removing the integer validation.
While the validation is great, I feel that this work around is better to
get the `{browser,edge}_cache_ttl` functionality working as expected is
more important. The validation will now just be enforced by the API
instead.

### DIFF
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -3,6 +3,7 @@ package cloudflare
 import (
 	"fmt"
 	"log"
+	"strconv"
 
 	"strings"
 
@@ -206,15 +207,15 @@ func resourceCloudflarePageRule() *schema.Resource {
 						},
 
 						"browser_cache_ttl": {
-							Type:         schema.TypeInt,
-							Optional:     true,
-							ValidateFunc: validation.IntAtMost(31536000),
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "",
 						},
 
 						"edge_cache_ttl": {
-							Type:         schema.TypeInt,
-							Optional:     true,
-							ValidateFunc: validation.IntAtMost(31536000),
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "",
 						},
 
 						"cache_level": {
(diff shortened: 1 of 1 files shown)


cloudflare/resource_cloudflare_page_rule.go
```diff
@@ -206,15 +207,15 @@ func resourceCloudflarePageRule() *schema.Resource {
 						},
 
 						"browser_cache_ttl": {
-							Type:         schema.TypeInt,
-							Optional:     true,
-							ValidateFunc: validation.IntAtMost(31536000),
+							Type:     schema.TypeString,
```
jacobbednarz: HCL still allows the following even though it's listed as a string.

```hcl
resource "..." "..." {
  actions {
    browser_cache_ttl = 3
  }
}
```

patryk:
While in general I agree with your reasoning, I'd like to point out that you could try using `GetOkExists` here, which allows you to distinguish between 0 and value not set.

jacobbednarz:
I had issues with `GetOkExists` as it was defaulting to 0. Could have been my implementation but happy to also leave these as integers if we can make that work instead.

jacobbednarz:
It's worth noting, this will only be an issue for Enterprise customers who have the ability to set the value as 0. Other than that, this won't be an issue. If we don't want to change this behaviour a workaround is to set it to `1` and then back to `0` to have Terraform state and Cloudflare in sync. #338 will help with this as it will also allow deletion as we'll be swapping to the PUT endpoint (full replacement) for each rule update.

jacobbednarz:
Turns out I was holding it wrong and there is a simpler, better approach to both of these options 😄

jacobbednarz:
Added the fix in 44ac57c as it is a dependency there and I was getting sick of jumping back and forth between branches.

PR #335, opened 2019-05-06
cloudflare_logpush: Remove acceptance test

jacobbednarz:
Logpush unfortunately doesn't lend itself to acceptance tests as it
wants manual, human intervention by design so automation is off the
table. At least for now.

This cleans up the logpush acceptance test as we can't run it in CI.

cc @zackproser


zackproser:
LGTM! Thanks for handling this.

PR #338, opened 2019-05-06
page_rules: Swap to completely replacing rules

### LINKED ISSUE #198: Deleting page rule action doesn't work with inline updates
This morning I was looking to apply a routine change to some page rules and noticed that actions which are removed cannot be performed using the current implementation of `resourceCloudflarePageRuleUpdate` due to it using [`ChangePageRule`](https://github.com/cloudflare/cloudflare-go/blob/1e1d4e0b8ae7aad64146f9f8f6a366209ce39e8b/page_rules.go#L186) which performs a HTTP PATCH request (and was recently [fixed](https://github.com/terraform-providers/terraform-provider-cloudflare/pull/176#issuecomment-452603705) to match the documented behaviour). The reason page rule actions cannot be removed using this API call is that the endpoint accepts partial updates to replace the existing action and to remove an action, it cannot be present or pass in an empty value (which the API doesn't allow). The Cloudflare Dashboard works around this issue by only using HTTP PUT requests and replace the rules each time.

This leaves us in a bit of a conundrum. If we swap to using the PUT endpoint, we can restore the previous functionality where deleting actions works but we can't perform partial updates to work around issues such as using actions like `cache_key` (due to requiring a Cloudflare Solutions Engineer to implement). If we leave the functionality as is, individual actions cannot be deleted but we can perform in place updates without recreating page rules each time.

I have considered a third option where we use a hybrid of the two but I'm unsure if the complexity is worth it as we might end up in a situation with lots of path branching to account for each scenario.

Thoughts?

cc @patryk as you might have some insight and thoughts here.

### LINKED ISSUE #265: Page_Rule change to and from a forwarding_url fails due to faulty validation error.
### Terraform Version
v0.11.11

### Affected Resource(s)
cloudflare_page_rule

```hcl

Changing a page_rule from this:

resource "cloudflare_page_rule" "subdomain_settings" {
  zone     = "${var.domain}"
  target   = "https://subdomain.website.com/*"
  actions = {
      ssl         = "flexible"
      cache_level = "aggressive"
  }
  priority = 43
  status   = "active"
}

to this:

resource "cloudflare_page_rule" "subdomain_settings" {
  zone     = "${var.domain}"
  target   = "https://subdomain.website.com/*"
  actions = {
      forwarding_url {
          url         = "https://offline.website.com/maintenance.html"
          status_code = 302
      }
  }
  priority = 43
  status   = "active"
}


```

### Debug Output
```

* cloudflare_page_rule.subdomain_settings: 1 error(s) occurred:

* cloudflare_page_rule.subdomain_settings: Failed to update Cloudflare Page Rule: error from makeRequest: HTTP status 400: content "{\"success\":false,\"errors\":[{\"code\":1004,\"message\":\"Page Rule validation failed: See messages for details.\"}],\"messages\":[{\"code\":1,\"message\":\".settings: \\\"forwarding_url\\\" may not be used with \\\"any setting\\\"\",\"type\":null}],\"result\":null}"

```

### Expected Behavior
The ssl and cache_level settings should be removed from the pagerule and the forwarding_url setting added to the page_rule

### Actual Behavior
The change is aborted due to a validation error

### Steps to Reproduce
1. `terraform plan`
2. `terraform apply`

jacobbednarz:
For a long time, this provider relied on sending page rules using
cloudflare-go's `ChangePageRule` which was documentated to allow inline
updates using a PATCH request. Until recently, this wasn't the case and
it actually mimicked the PUT alternative (`UpdatePageRule`) where it
attempted to replace the entire rule. This was [fixed](https://github.com/terraform-providers/terraform-provider-cloudflare/pull/176#issuecomment-452906541) and while you
could now send inline updates to actions, there wasn't a way to remove
them once they were there. To address this, I've swapped back to
`UpdatePageRule` (the PUT request) which will resolve the aforementioned
issue with removing rules but it does re-introduce the caveat of not
being able to manage actions that require a Solutions Engineer. This
means, you can't manage them using this endpoint as you won't have
permission to apply them. This extends to page rule changes like
re-ordering or updating other actions which are in the same rule. If you
hit this use case, you should move the cache key into Cloudflare
Workers.

This commit backs out most of the changes from c1799f4 as they are no
longer required for differentiating the actions to send but it does
introduce some changes to the way we determine whether or not we send
`browser_cache_ttl` and `edge_cache_ttl` based on some issues
encountered with their default values. It now performs a `HasChange` on 
their values to help determine whether or not they should be a part of 
the payload.

Closes #198 
Closes #265

### DIFF
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -350,7 +350,7 @@ func resourceCloudflarePageRuleCreate(d *schema.ResourceData, meta interface{}) 
 	log.Printf("[DEBUG] Actions found in config: %#v", actions)
 	for _, action := range actions {
 		for id, value := range action.(map[string]interface{}) {
-			newPageRuleAction, err := transformToCloudflarePageRuleAction(id, value, false)
+			newPageRuleAction, err := transformToCloudflarePageRuleAction(id, value)
 			if err != nil {
 				return err
 			} else if newPageRuleAction.Value == nil || newPageRuleAction.Value == "" {
@@ -463,25 +463,24 @@ func resourceCloudflarePageRuleUpdate(d *schema.ResourceData, meta interface{}) 
 		}
 	}
 
-	old, new := d.GetChange("actions")
+	if v, ok := d.GetOk("actions"); ok {
+		actions := v.([]interface{})
+		newPageRuleActions := make([]cloudflare.PageRuleAction, 0, len(actions))
 
-	oldActions := old.([]interface{})[0].(map[string]interface{})
-	newActions := new.([]interface{})[0].(map[string]interface{})
-
-	newPageRuleActions := make([]cloudflare.PageRuleAction, 0, len(newActions))
-
-	for id, value := range newActions {
-		hasChanged := id != "forwarding_url" && id != "minify" && oldActions[id] != value
-		newPageRuleAction, err := transformToCloudflarePageRuleAction(id, value, hasChanged)
-		if err != nil {
-			return err
-		} else if newPageRuleAction.Value == nil {
-			continue
+		for _, action := range actions {
+			for id, value := range action.(map[string]interface{}) {
+				newPageRuleAction, err := transformToCloudflarePageRuleAction(id, value)
+				if err != nil {
+					return err
+				} else if newPageRuleAction.Value == nil {
+					continue
+				}
+				newPageRuleActions = append(newPageRuleActions, newPageRuleAction)
+			}
 		}
-		newPageRuleActions = append(newPageRuleActions, newPageRuleAction)
-	}
 
-	updatePageRule.Actions = newPageRuleActions
+		updatePageRule.Actions = newPageRuleActions
+	}
 
 	if priority, ok := d.GetOk("priority"); ok {
 		updatePageRule.Priority = priority.(int)
@@ -493,9 +492,7 @@ func resourceCloudflarePageRuleUpdate(d *schema.ResourceData, meta interface{}) 
 
 	log.Printf("[DEBUG] Cloudflare Page Rule update configuration: %#v", updatePageRule)
 
-	// contrary to docs, change page rule actually does a full replace
-	// this part of the api needs some work, so it may change in future
-	if err := client.ChangePageRule(zoneID, d.Id(), updatePageRule); err != nil {
+	if err := client.UpdatePageRule(zoneID, d.Id(), updatePageRule); err != nil {
 		return fmt.Errorf("Failed to update Cloudflare Page Rule: %s", err)
 	}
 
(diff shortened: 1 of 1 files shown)


jacobbednarz:
Note: CI will be red until #334 is merged and the conflicts are resolved in this branch.

jacobbednarz:
Have a failing integration test for `TestAccCloudflarePageRule_Updated` that I'll take a look at tonight.

jacobbednarz:
We're also green on integration tests now.

```
=== RUN   TestAccCloudflarePageRule_Import
=== PAUSE TestAccCloudflarePageRule_Import
=== RUN   TestAccCloudflarePageRule_Basic
--- PASS: TestAccCloudflarePageRule_Basic (4.13s)
=== RUN   TestAccCloudflarePageRule_FullySpecified
--- PASS: TestAccCloudflarePageRule_FullySpecified (4.25s)
=== RUN   TestAccCloudflarePageRule_ForwardingOnly
--- PASS: TestAccCloudflarePageRule_ForwardingOnly (3.66s)
=== RUN   TestAccCloudflarePageRule_ForwardingAndOthers
--- PASS: TestAccCloudflarePageRule_ForwardingAndOthers (0.02s)
=== RUN   TestAccCloudflarePageRule_Updated
--- PASS: TestAccCloudflarePageRule_Updated (6.15s)
=== RUN   TestAccCloudflarePageRule_CreateAfterManualDestroy
--- PASS: TestAccCloudflarePageRule_CreateAfterManualDestroy (6.35s)
=== RUN   TestAccCloudflarePageRule_UpdatingZoneForcesNewResource
=== RUN   TestAccCloudflarePageRuleMinifyAction
--- PASS: TestAccCloudflarePageRuleMinifyAction (4.71s)
=== CONT  TestAccCloudflarePageRule_Import
--- PASS: TestAccCloudflarePageRule_Import (6.14s)
PASS
PASS	github.com/terraform-providers/terraform-provider-cloudflare/cloudflare	35.441s
```

patryk:
Thank you @jacobbednarz, merged!

grogancolin:
I understand your reasoning on this, however I do feel it's a mistake that this provider explicitly backs out of managing specific page rules.

Not all organisations use workers, and assuming we can and should isn't something that the provider should be forcing on end users.

For our Page Rule configuration, this PR breaks completely our ability to manage them via code.

jacobbednarz:
@grogancolin I totally sympathise with you. This wasn't just a decision based on what made my maintainer life easier as internally, we are facing the issues both you and I have outlined in these PRs.

- We have custom cache keys in page rules
- We don't use Workers
- We manage everything in code

One thing that helped make this decision was that up until about three months ago, this behaviour was consistent regardless of whether we used the PATCH or PUT endpoint (due to a regression in the API). We've been lucky enough that the number of page rules we're changing on the zones has really slowed down and importing from what changes are made in the UI is sufficient. This doesn't mean by any means that I'm settling for this; it just means I've opted for the masses and not the minority to allow bandwidth for focusing on the issue.

I will open a new issue with the current state of affairs and see if we can brainstorm some potential work arounds for the Cloudflare API (trying to handle it in the provider isn't really viable) and drag in the Cloudflare folks for their 2c. WDYT?

grogancolin:
@jacobbednarz - apologies I didn't get back to you on this. 

> I will open a new issue with the current state of affairs and see if we can brainstorm some potential work arounds for the Cloudflare API (trying to handle it in the provider isn't really viable) and drag in the Cloudflare folks for their 2c. WDYT?

I think that's a fair position - however it would be good to give it more urgency as this is a regular irritant for us.

Any updates on the current state of affairs?

Anything I can do to help with the task? Obviously I can't modify Cloudflare's APIs but I can and am more than willing to work on the provider to keep this issue moving in the right direction.

PR #351, opened 2019-05-15
Update note on doc

charlesverdad:
Fixes little typo probably caused by copy-paste of the previous block.


patryk:
Sharp eye there! Thanks a lot!

Issue #277, opened 2019-03-29
Add support for worker KV stores 


jbergstroem:
### Terraform Version
Terraform v0.11.13

### Affected Resource(s)
- cloudflare_worker_script
- cloudflare..

I'd like to express my KV stores as well as binding them to a script via terraform. Couldn't find any references to this in the issues nor in documentation.

jbergstroem:
Relevant docs: https://developers.cloudflare.com/workers/kv/api/

jimsorock:
Any update here?

adam-janis:
cloudflare-go bindings support is merged
https://github.com/cloudflare/cloudflare-go/pull/363

garrettgalow:
@patryk I think we can close this now

danielnaab:
Given that Cloudflare is suggesting one store secrets in Workers KV, it would be nice if the Terraform provider supported adding key/value pairs into a namespace, also.

jacobbednarz:
@danielnaab please create a new issue if this is something you’d like to have a go at submitting a PR for or if you’d like to request a new feature. Be sure to follow the template provided.

PR #363, opened 2019-05-26
Add support for Workers KV

Electroid:
Addresses the resource request from #277. I'm _relatively_ new to Go, so feel free to nit pick! Since KV is eventually consistent, acceptance tests could prove to be more of a hassle than it's worth, thoughts?

* Adds `cloudflare_worker_kv_namespace` and `cloudflare_worker_kv_pair` resources 
* Convince methods for casting the client were added to `utils.go`<br>(may need future refactoring with account overhaul)
* Updates website documentation

### DIFF
--- a/cloudflare/provider.go
+++ b/cloudflare/provider.go
@@ -108,6 +108,8 @@ func Provider() terraform.ResourceProvider {
 			"cloudflare_waf_rule":               resourceCloudflareWAFRule(),
 			"cloudflare_worker_route":           resourceCloudflareWorkerRoute(),
 			"cloudflare_worker_script":          resourceCloudflareWorkerScript(),
+			"cloudflare_worker_kv_namespace":    resourceCloudflareWorkerKVNamespace(),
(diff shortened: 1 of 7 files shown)


jacobbednarz:
heya @Electroid! Thanks for the PR!

> Since KV is eventually consistent, acceptance tests could prove to be more of a hassle than it's worth, thoughts?

While the service itself might be eventually consistent, Terraform won't be and the state needs to be in sync between the Terraform state and the remote API.

> Adds cloudflare_worker_kv_namespace and cloudflare_worker_kv_pair resources

I can see this is still pending on some changes in `cloudflare-go` so we probably want to ensure that lands before we do too much in here just to limit the back and forth of the changes you'll need to make. I'm happy to have you wait on this one until upstream lands and then come back to this one?

Electroid:
Yep, I'll promote this out of draft when those changes come in. Thanks for taking a look!

jbergknoff-rival:
Any status update on this?

cooper2989:
Is this going to be covered anytime in the near future? currently looking to pipeline KV worker assignments via terraform but hitting a dead end

PR #379, opened 2019-06-13
Fix browser_cache_ttl page rule action bugs

stevehodgkiss:
Addresses 2 issues with `browser_cache_ttl` due to `0` being the default value and also a valid & desirable value to define it with.

- Deleting: After a `browser_cache_ttl` page rule action has been applied, it can't be deleted because its value is the default of `0` when it's removed, so it ends up in the update PUT request as being set to `0`.
- Setting to 0 initially: Defining `browser_cache_ttl = 0` initially doesn't work because it's not seen as a change (current workaround is to set to something else first, then 0).

Using an integer for this value in the schema makes that quite challenging. Another option might be to make `-1` a special and default value used to mean "not set". The approach in this PR is to use a string instead.

I'd like to take a closer look at some of the test cases before this merged (hence WIP in the title).

### DIFF
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -3,7 +3,7 @@ package cloudflare
 import (
 	"fmt"
 	"log"
-
+	"strconv"
 	"strings"
 
 	cloudflare "github.com/cloudflare/cloudflare-go"
@@ -206,9 +206,9 @@ func resourceCloudflarePageRule() *schema.Resource {
 						},
 
 						"browser_cache_ttl": {
-							Type:         schema.TypeInt,
-							Optional:     true,
-							ValidateFunc: validation.IntAtMost(31536000),
+							Type:     schema.TypeString,
+							Optional: true,
+							//ValidateFunc: validation.IntAtMost(31536000),
 						},
 
 						"edge_cache_ttl": {
@@ -596,7 +596,12 @@ func transformToCloudflarePageRuleAction(id string, value interface{}, d *schema
 	changed := d.HasChange(fmt.Sprintf("actions.0.%s", id))
 
 	if strValue, ok := value.(string); ok {
-		if strValue == "" && !changed {
+		if id == "browser_cache_ttl" && changed {
+			intValue, err := strconv.Atoi(strValue)
+			if err == nil {
+				pageRuleAction.Value = intValue
+			}
+		} else if strValue == "" && !changed {
 			pageRuleAction.Value = nil
 		} else {
 			pageRuleAction.Value = strValue
@@ -616,9 +621,7 @@ func transformToCloudflarePageRuleAction(id string, value interface{}, d *schema
 			}
 		}
 	} else if intValue, ok := value.(int); ok {
-		if id == "browser_cache_ttl" && changed {
-			pageRuleAction.Value = intValue
-		} else if id == "edge_cache_ttl" && intValue > 0 && changed {
+		if id == "edge_cache_ttl" && intValue > 0 && changed {
 			pageRuleAction.Value = intValue
 		} else {
 			pageRuleAction.Value = nil
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -208,7 +208,6 @@ func resourceCloudflarePageRule() *schema.Resource {
 						"browser_cache_ttl": {
 							Type:     schema.TypeString,
 							Optional: true,
-							//ValidateFunc: validation.IntAtMost(31536000),
 						},
 
 						"edge_cache_ttl": {
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -540,12 +540,6 @@ var pageRuleAPINilFields = []string{
 	"disable_railgun",
 	"disable_security",
 }
-var pageRuleAPIFloatFields = []string{
-	"edge_cache_ttl",
-}
-var pageRuleAPIFloatAsStringFields = []string{
-	"browser_cache_ttl",
-}
 var pageRuleAPIStringFields = []string{
 	"bypass_cache_on_cookie",
 	"cache_key",
@@ -571,15 +565,15 @@ func transformFromCloudflarePageRuleAction(pageRuleAction *cloudflare.PageRuleAc
 		value = true
 		break
 
-	case contains(pageRuleAPIFloatFields, pageRuleAction.ID):
-		value = pageRuleAction.Value.(float64) // we use TypeInt but terraform seems to do the right thing converting from float
-		break
-
 	case contains(pageRuleAPIStringFields, pageRuleAction.ID):
 		value = pageRuleAction.Value.(string)
 		break
 
-	case contains(pageRuleAPIFloatAsStringFields, pageRuleAction.ID):
+	case pageRuleAction.ID == "edge_cache_ttl":
+		value = pageRuleAction.Value.(float64) // we use TypeInt but terraform seems to do the right thing converting from float
+		break
+
+	case pageRuleAction.ID == "browser_cache_ttl":
 		value = fmt.Sprintf("%.0f", pageRuleAction.Value.(float64))
 		break
 
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -541,8 +541,10 @@ var pageRuleAPINilFields = []string{
 	"disable_security",
 }
 var pageRuleAPIFloatFields = []string{
-	"browser_cache_ttl",
 	"edge_cache_ttl",
+}
+var pageRuleAPIFloatAsStringFields = []string{
+	"browser_cache_ttl",
 }
 var pageRuleAPIStringFields = []string{
 	"bypass_cache_on_cookie",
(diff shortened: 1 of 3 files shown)


cloudflare/resource_cloudflare_page_rule_test.go
```diff
@@ -2,17 +2,124 @@ package cloudflare
...
+		for _, pageRuleAction := range pageRule.Actions {
+			if pageRuleAction.ID == key && pageRuleAction.Value == value {
+				return nil
+			}
+		}
+		return fmt.Errorf("cloudflare page rule action not found %#v:%#v\nAction State\n%#v", key, value, pageRule.Actions)
+	}
+}
+
+//#################################################################
```
jacobbednarz:
```suggestion
```

cloudflare/resource_cloudflare_page_rule_test.go
```diff
@@ -2,17 +2,124 @@ package cloudflare
...
+}
+
+func assertStatePageRuleActionExists(resourceName string, key string, value string) resource.TestCheckFunc {
+	return resource.TestCheckResourceAttr(
+		resourceName,
+		fmt.Sprintf("actions.0.%s", key),
+		value)
+}
+
+func assertCloudflarePageRuleActionExists(resourceName string, key string, value interface{}) resource.TestCheckFunc {
```
jacobbednarz: Is there a reason why we didn't use `testAccCheckCloudflarePageRuleExists` here? The signature and logic looks similar. 

https://github.com/terraform-providers/terraform-provider-cloudflare/blob/b1dbaa88a803d4b069bb77c2b6c702ba2c507253/cloudflare/resource_cloudflare_page_rule_test.go#L498

cloudflare/resource_cloudflare_page_rule_test.go
```diff
@@ -2,17 +2,124 @@ package cloudflare
...
+			Config: buildPageRule(`browser_cache_ttl = 0`),
+		},
+		{
+			Config: buildPageRule(`browser_check = "on"`),
+			Check:  assertActionNotExists("browser_cache_ttl"),
+		},
+	})
+}
+
+func buildPageRule(actions string) string {
```
jacobbednarz: Let's swap this one for a specific configuration setup (like we do for [`testAccCheckCloudflarePageRuleConfigMinify`](https://github.com/terraform-providers/terraform-provider-cloudflare/blob/b1dbaa88a803d4b069bb77c2b6c702ba2c507253/cloudflare/resource_cloudflare_page_rule_test.go#L542-L555)). It's probably a bit more verbose but it will be explicit about what is getting setup.

jacobbednarz:
Thanks for getting this up @stevehodgkiss! :cake: I like the additional tests we've added here however a couple things we should update to align with the other tests.

1. The integration tests you've added here should be named `TestAccCloudflarePageRule_*`. The `TestAcc` is important as that is what the CI runner looks for when it's searching for tests to run. See https://www.terraform.io/docs/extend/testing/acceptance-tests/index.html#test-files for more details. Right now, these new tests aren't being automatically run.
1. We should structure the tests similar to the ones we already have where each uses an explicit resource definition and the test setup just relies on that. An example:
- Rule configuration: https://github.com/terraform-providers/terraform-provider-cloudflare/blob/b1dbaa88a803d4b069bb77c2b6c702ba2c507253/cloudflare/resource_cloudflare_page_rule_test.go#L557-L567
- Usage: https://github.com/terraform-providers/terraform-provider-cloudflare/blob/b1dbaa88a803d4b069bb77c2b6c702ba2c507253/cloudflare/resource_cloudflare_page_rule_test.go#L123-L146

cloudflare/resource_cloudflare_page_rule_test.go
```diff
@@ -2,17 +2,124 @@ package cloudflare
...
+			if pageRuleAction.ID == key && pageRuleAction.Value == value {
+				return nil
+			}
+		}
+		return fmt.Errorf("cloudflare page rule action not found %#v:%#v\nAction State\n%#v", key, value, pageRule.Actions)
+	}
+}
+
+//#################################################################
+
```
jacobbednarz:
```suggestion
```

patryk:
Yes, I think changing the field to string is a good compromise, and in the end should not deal any major damage, provided Terraform uses strings implicitly everywhere in `.tf` files anyway.

jacobbednarz:
I think this is all the scenarios.

| scenario                                                                                                   | state expected outcome                    | remote expected outcome                                          |
|------------------------------------------------------------------------------------------------------------|-------------------------------------------|------------------------------------------------------------------|
| page rule with `browser_cache_ttl = 0` can be imported                                                     | `browser_cache_ttl` = 0                   | page rule exists with `browser_cache_ttl` value of 0             |
| page rule with `browser_cache_ttl = 10` can be imported                                                    | `browser_cache_ttl` = 10                  | page rule exists with `browser_cache_ttl` value of 10            |
| page rule with `browser_cache_ttl` absent can be imported                                                  | `browser_cache_ttl` should not be present | page rule without `browser_cache_ttl` should exist               |
| create a new rule `browser_cache_ttl =  0`                                                                 | `browser_cache_ttl` = 0                   | page rule exists with `browser_cache_ttl` value of 0             |
| create a new rule `browser_cache_ttl = 10`                                                                 | `browser_cache_ttl` = 10                  | page rule exists with `browser_cache_ttl` value of 10            |
| create a new rule `browser_cache_ttl` absent                                                               | `browser_cache_ttl` should not be present | page rule doesn't get created with `browser_cache_ttl` attribute |
| delete an existing page rule where `browser_cache_ttl` was set to 0                                        | `browser_cache_ttl` should not be present | page rule isn't present with `browser_cache_ttl` value           |
| delete an existing page rule where `browser_cache_ttl` was set to 10                                       | `browser_cache_ttl` should not be present | page rule isn't present with `browser_cache_ttl` value           |
| delete an existing page rule where `browser_cache_ttl` is absent                                           | `browser_cache_ttl` should not be present | page rule isn't present with `browser_cache_ttl` value           |
| update an existing page rule from not having `browser_cache_ttl` present to `browser_cache_ttl` value of 0 | `browser_cache_ttl` = 0                   | page rule is updated with `browser_cache_ttl` value of 0         |
| update an existing page rule from `browser_cache_ttl = 0` to be 10                                         | `browser_cache_ttl` = 10                  | page rule is updated with `browser_cache_ttl` value of 10        |
| update an existing page rule from `browser_cache_ttl = 10` to be 0                                         | `browser_cache_ttl` = 0                   | page rule is updated with `browser_cache_ttl` value of 0         |

jacobbednarz:
Looking great @stevehodgkiss! The only two test cases I think we're missing here is the ones for importing resources (one with a value of 0, one with a value of > 0 and one without the attribute entirely). Here is an example of another resource testing the import.

https://github.com/terraform-providers/terraform-provider-cloudflare/blob/5d7a3d1b81a79f6489691f3d74330a841fe2e55e/cloudflare/resource_cloudflare_zone_lockdown_test.go#L36-L57

Looking at the code, this should Just Work™ once the tests are added.

I've run the integration tests on CI and in another couple of accounts and so far, 💚 all the way.

cloudflare/resource_cloudflare_page_rule.go
```diff
@@ -578,6 +579,10 @@ func transformFromCloudflarePageRuleAction(pageRuleAction *cloudflare.PageRuleAc
 		value = pageRuleAction.Value.(string)
 		break
 
+	case contains(pageRuleAPIFloatAsStringFields, pageRuleAction.ID):
```
jacobbednarz: Nit: `browser_cache_ttl` is a pretty unique case (similar to `forwarding_url` below). We could probably just do a `pageRuleAction.ID == "browser_cache_ttl"` check here instead of the `contains`. I wouldn't expect to add to `pageRuleAPIFloatAsStringFields`.

```suggestion
	case pageRuleAction.ID == "browser_cache_ttl":
```

jacobbednarz: Similarly, if we do that, we can also do away with `pageRuleAPIFloatFields` only containing `edge_cache_ttl` and put that in a single `pageRuleAction.ID == "edge_cache_ttl"` check.

jacobbednarz: Here's a `diff` of what I mean.

```diff
diff --git a/cloudflare/resource_cloudflare_page_rule.go b/cloudflare/resource_cloudflare_page_rule.go
index 7fcfc3b..4a870c3 100644
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -540,12 +540,6 @@ var pageRuleAPINilFields = []string{
        "disable_railgun",
        "disable_security",
 }
-var pageRuleAPIFloatFields = []string{
-       "edge_cache_ttl",
-}
-var pageRuleAPIFloatAsStringFields = []string{
-       "browser_cache_ttl",
-}
 var pageRuleAPIStringFields = []string{
        "bypass_cache_on_cookie",
        "cache_key",
@@ -571,15 +565,15 @@ func transformFromCloudflarePageRuleAction(pageRuleAction *cloudflare.PageRuleAc
                value = true
                break

-       case contains(pageRuleAPIFloatFields, pageRuleAction.ID):
-               value = pageRuleAction.Value.(float64) // we use TypeInt but terraform seems to do the right thing converting from float
-               break
-
        case contains(pageRuleAPIStringFields, pageRuleAction.ID):
                value = pageRuleAction.Value.(string)
                break

-       case contains(pageRuleAPIFloatAsStringFields, pageRuleAction.ID):
+       case case pageRuleAction.ID == "edge_cache_ttl"::
+               value = pageRuleAction.Value.(float64) // we use TypeInt but terraform seems to do the right thing converting from float
+               break
+
+       case pageRuleAction.ID == "browser_cache_ttl":
                value = fmt.Sprintf("%.0f", pageRuleAction.Value.(float64))
                break
```

patryk:
Big thanks @stevehodgkiss for helping out with this issue! Appreciated.

PR #390, opened 2019-06-24
Bugfix: browser_cache_ttl is incorrectly sent as a string when unchanged

stevehodgkiss:
Fixes an issue added in https://github.com/terraform-providers/terraform-provider-cloudflare/pull/379 where changing an unrelated value will result in `browser_cache_ttl` being sent as a string rather than an integer. The error message returned from CloudFlare is "TTL values must be whole numbers only". Removing the `changed` from the conditional fixes that.


jacobbednarz:
CI has 3 failures at the moment but look unrelated.

```
TestAccCloudflareWorkerRoute_MultiScriptEnt 
TestAccCloudflareWorkerRoute_SingleScriptEnt 
TestAccCloudflareWorkerRoute_SingleScriptNonEnt
```

I'll investigate these separately.

jacobbednarz:
Thanks a bunch @stevehodgkiss! 🎈

PR #407, opened 2019-07-10
Fix: examples of cloudflare_zone_lockdown in documentation

jakshi:
Examples of cloudflare_zone_lockdown that provided in documentation doesn't work in terraform 0.12


patryk:
Thanks @jakshi !

PR #409, opened 2019-07-11
test/spectrum_application: Add test sweeper

jacobbednarz:
Recently we have been seeing some test flakiness for spectrum
applications due to left over resources or partially deleted resources.
To combat this, we can use a test sweeper (similar to what we've done
with `zones`) which will ensure a clean test bed for all runs.

This fixes the following CI failure.

```
------- Stdout: -------
=== RUN   TestAccCloudflareSpectrumApplication_Basic
--- FAIL: TestAccCloudflareSpectrumApplication_Basic (1.46s)
    testing.go:568: Step 0 error: errors during apply:

        Error: error creating spectrum application for zone: error from makeRequest: HTTP status 400: content "{\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 12004,\n      \"message\": \"Failed to associate application: Port(s) conflict with existing application.\"\n    }\n  ],\n  \"messages\": []\n}\n"
```

### DIFF
--- a/cloudflare/resource_cloudflare_spectrum_application_test.go
+++ b/cloudflare/resource_cloudflare_spectrum_application_test.go
@@ -3,6 +3,7 @@ package cloudflare
 import (
 	"errors"
 	"fmt"
+	"log"
 	"testing"
 
 	"os"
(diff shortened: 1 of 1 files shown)


cloudflare/resource_cloudflare_spectrum_application_test.go
```diff
@@ -12,6 +13,50 @@ import (
...
+	}
+
+	zone := os.Getenv("CLOUDFLARE_DOMAIN")
+	zones, zoneErr := client.ListZones(zone)
+
+	if zoneErr != nil {
+		log.Printf("[ERROR] Failed to fetch Cloudflare zones: %s", zoneErr)
+	}
+
+	for _, zone := range zones {
```
jacobbednarz: Realistically, I could just pull the first entry as we are only using a single domain however I think this is fine for the testing accounts.

jacobbednarz:
Confirmed with a handful of CI integration runs and random manual injections that this has addressed the flakiness.

PR #412, opened 2019-07-15
resource/cloudflare_zone: Add unicode support

jacobbednarz:
Introduce unicode support for using zone names and ensure that the
schema comparison is performed after first converting both values to
unicode. This change ensures that the punycode and unicode values are
both stored as the unicode version to match what the Cloudflare API
returns.

### To do

- [x] Setup/purchase unicode domain for tests

Closes #399

### DIFF
--- a/cloudflare/resource_cloudflare_zone_test.go
+++ b/cloudflare/resource_cloudflare_zone_test.go
@@ -58,7 +58,7 @@ func TestAccCloudflareZone(t *testing.T) {
 	})
 }
 
-func TestZoneWithUnicodeIsStoredAsUnicode(t *testing.T) {
+func TestAccZoneWithUnicodeIsStoredAsUnicode(t *testing.T) {
 	name := "cloudflare_zone.tf-acc-unicode-test-1"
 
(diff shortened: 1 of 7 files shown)


jacobbednarz:
I've also updated a bunch of acceptance test names that didn't start with `TestAcc` and weren't being run before.

cloudflare/resource_cloudflare_zone_test.go
```diff
@@ -58,10 +58,84 @@ func TestZone(t *testing.T) {
...
 
+func TestAccZoneWithUnicodeIsStoredAsUnicode(t *testing.T) {
+	name := "cloudflare_zone.tf-acc-unicode-test-1"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testZoneConfig("tf-acc-unicode-test-1", "zajęzyk.pl", "true", "false"),
```
jacobbednarz: This domain needs to be swapped out with our own domain before merging.

PR #413, opened 2019-07-16
Update test names to trigger acceptance tests

jacobbednarz:
The test names didn't start with `TestAcc` so the integration suite
wasn't running them despite being integration tests.


jacobbednarz:
CI is failing on `example.org` which we're dealing with out of band.

PR #416, opened 2019-07-16
resource/cloudflare_zone: Update test domain

jacobbednarz:
CI has recently started failing the `TestAccCloudflareZone` test as the
`example.org` domain is using old Cloudflare nameservers and the work to
address this is on a backlog to be fixed. Instead of using an example
domain, we can swap to using an ephemeral subdomain and perform the
assertions against that.

### DIFF
--- a/cloudflare/resource_cloudflare_zone_test.go
+++ b/cloudflare/resource_cloudflare_zone_test.go
@@ -65,7 +65,6 @@ func TestAccCloudflareZonePartialSetup(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(name, "zone", "foo.net"),
 					resource.TestCheckResourceAttr(name, "paused", "true"),
-					resource.TestCheckResourceAttr(name, "name_servers.#", "2"),
 					resource.TestCheckResourceAttr(name, "plan", planIDFree),
 					resource.TestCheckResourceAttr(name, "type", "partial"),
 				),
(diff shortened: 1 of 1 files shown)


cloudflare/resource_cloudflare_zone_test.go
```diff
@@ -2,52 +2,90 @@ package cloudflare
...
 					resource.TestCheckResourceAttr(name, "name_servers.#", "2"),
 					resource.TestCheckResourceAttr(name, "plan", planIDFree),
 					resource.TestCheckResourceAttr(name, "type", "full"),
 				),
 			},
+		},
+	})
+}
+
+func TestAccCloudflareZonePartialSetup(t *testing.T) {
```
jacobbednarz: This is still failing. 

```
------- Stdout: -------
=== RUN   TestAccCloudflareZonePartialSetup
--- FAIL: TestAccCloudflareZonePartialSetup (21.10s)
    testing.go:568: Step 0 error: errors during apply:
        
        Error: Error creating zone "example.cfapi.net": error from makeRequest: HTTP status 400: content "{\"success\":false,\"errors\":[{\"code\":1104,\"message\":\"Partial zone signup not allowed\"}],\"messages\":[],\"result\":null}"
        
          on /opt/teamcity-agent/temp/buildTmp/tf-test190919551/main.tf line 2:
          (source code not available)
        
        
FAIL
```

Chasing it up with Cloudflare folks.

patryk:
Should work now. Merged.

Issue #227, opened 2019-03-03
Breaking changes for a major release


jacobbednarz:
We've spoken a couple of times regarding cleaning up the Terraform interface or interactions which would result in breaking backwards compatibility. To discuss how we want to do this and how it will be executed, I've raised this issue. 

Here is a table from a quick search on the changes that we'd like to introduce:

Change | Discussion link | Reason
------------ | -------------| -------------
Swap `org_id` for `account_id` | https://github.com/terraform-providers/terraform-provider-cloudflare/pull/226#pullrequestreview-209887811 | Move off deprecated APIs and use newer nomenclature
Deprecate `zone_name` in favour of `zone_id` | https://github.com/terraform-providers/terraform-provider-cloudflare/issues/161#issuecomment-441128985 | More explicitness as domain name isn't unique but ID is.

I'll keep this table updated so if you have other things to add please pop them below and I'll update this accordingly.

Once we decide on what we want to change, I'll create a new milestone and we can start picking off the tasks to complete.

cc @patryk @garrettgalow @prdonahue

garrettgalow:
Based on #109 we should focus on zone_name -> zone_id.

To clarify, from the customer's perspective, Is org_id -> account_id just changing the single var name in the provider configuration?

jacobbednarz:
> To clarify, from the customer's perspective, Is org_id -> account_id just changing the single var name in the provider configuration?

Yep! End users will only have that single thing to change but there will need to be a bit of shuffling in the provider.

garrettgalow:
It would likely be ideal to shift zone name to id in a staged approached - aka allowing any resource with zone name to instead use zone id and then at a later time remove the zone_name fields - not sure how feasible that is.

if we have to make breaking changes for TF .12 then it would be a good time for these things, but if not I'm still in favor of scheduling time for them - I'd like to kill off orgs sometime this year, and this would be one of the checkboxes on the list

jacobbednarz:
> It would likely be ideal to shift zone name to id in a staged approached - aka allowing any resource with zone name to instead use zone id and then at a later time remove the zone_name fields - not sure how feasible that is.

We can definitely make it possible for both values to be present in the schema as is (with some deprecation warnings) and then in a later version just deprecate the one we no longer want.

> if we have to make breaking changes for TF .12 then it would be a good time for these things, but if not I'm still in favor of scheduling time for them

[...]

PR #417, opened 2019-07-17
provider: Rename "token" => "key"

jacobbednarz:
API "token" is currently used for HTTP authentication requests where the
`X-Auth-Key` needs to be set. Until recently, it was fine to use the
wrong term for this authentication method (as there was only the one)
however there is a new feature upcoming from Cloudflare which is
actually called API tokens and it requires a different authentication
procedure.

This means we need to rename the existing variables and references to
allow us to build in the new functionality when it is announced.

I've updated all the references to use the term "key" as that is what
the documentation and values are and I think it's a better reflection of
what method is in use than what we have in place at the moment.

### DIFF
--- a/cloudflare/cloudflare_sweeper_test.go
+++ b/cloudflare/cloudflare_sweeper_test.go
@@ -15,7 +15,7 @@ func TestMain(m *testing.M) {
 // sharedClient returns a common Cloudflare client setup needed for the
 // sweeper functions.
 func sharedClient() (*cloudflare.API, error) {
-	client, err := cloudflare.New(os.Getenv("CLOUDFLARE_TOKEN"), os.Getenv("CLOUDFLARE_EMAIL"))
+	client, err := cloudflare.New(os.Getenv("CLOUDFLARE_KEY"), os.Getenv("CLOUDFLARE_EMAIL"))
 
 	if err != nil {
 		return client, err
--- a/cloudflare/config.go
+++ b/cloudflare/config.go
@@ -9,13 +9,13 @@ import (
 
 type Config struct {
 	Email   string
-	Token   string
+	Key     string
 	Options []cloudflare.Option
 }
 
 // Client() returns a new client for accessing cloudflare.
 func (c *Config) Client() (*cloudflare.API, error) {
-	client, err := cloudflare.New(c.Token, c.Email, c.Options...)
+	client, err := cloudflare.New(c.Key, c.Email, c.Options...)
 	if err != nil {
 		return nil, fmt.Errorf("Error creating new Cloudflare client: %s", err)
 	}
--- a/cloudflare/provider.go
+++ b/cloudflare/provider.go
@@ -26,11 +26,11 @@ func Provider() terraform.ResourceProvider {
 				Description: "A registered Cloudflare email address.",
 			},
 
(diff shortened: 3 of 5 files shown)


jacobbednarz:
cc @garrettgalow @hunts

jacobbednarz:
`CLOUDFLARE_KEY` has been added to CI (without deleting `CLOUDFLARE_TOKEN` while the cut over is happening)

hunts:
Thanks for making this change. This looks a breaking change to me, will it a problem?

jacobbednarz:
No problem at all. We can determine what type of version bump we want when we perform a release so if we want a major bump, then we can do it. If we're going to do a major bump, there are a couple of other things I'll also add to the release (see #227 for the list) to knock them all over at once.

We might be able to release this as a minor bump as it's renaming a field, not making the change backwards incompatible.

garrettgalow:
Feels like will end up as a breaking change - unless for new API Tokens are going to be under a different name and not `CLOUDFLARE_TOKEN`?

I can understand that in this change allowing it to be either wouldn't be breaking

jacobbednarz:
I hadn’t thought too far ahead yet but perhaps we could use `CLOUDFLARE_API_TOKEN` if that would avoid the breaking change. But, as I mentioned we can cut a breaking homage release if we’d like and we might just get the other things listed to be included there so users get a single major bump for now.

garrettgalow:
If we are going to go ahead and do the breaking change two things:
1. Agree on doing the other stuff with it.
2. We then might as well just use the old name for the new tokens since that name is correct.

jacobbednarz:
>     1. Agree on doing the other stuff with it.

Agreed. I've been using #227 as that list. Is there anything missing there you'd like to include? 

>     2. We then might as well just use the old name for the new tokens since that name is correct.

👍

garrettgalow:
Nope that list is excellent

hunts:
Not know what is the best way to make forward changes based on this change, but I made this PR: https://github.com/jacobbednarz/terraform-provider-cloudflare/pull/1

Or can we create a dedicated feature branch for these breaking changes?

jacobbednarz:
As it stands, I was going to do these three independently and let them sit until we were ready for a breaking change to be rolled out and cut a new release for it.

We can make a single intermediate branch  for merging all these changes into however I haven't seen great success with that approach in the past for long lived branches as the waters end up very muddied on what is the the source of truth.

hunts:
Chatted with @garrettgalow offline. We can also consider introduce the `api_token` field before the release of the breaking changes. When we eventually deprecate the email+key auth scheme, either `token` or `api_token` both mean the API Tokens.

Thus, sent another pull request: https://github.com/terraform-providers/terraform-provider-cloudflare/pull/450

Let me know what you think.

jacobbednarz:
Note to self, we'll need to update this branch now to account for the `api_token` addition in #450 to deprecate on the next major release (as opposed to just adding this).

jacobbednarz:
I've updated these to use the base branch of v2.0 so that we can test all of these changes together on a single branch and then just merge that one instead of managing merge conflicts all over the place.

jacobbednarz:
Just going to start a new branch for this. The addition of `api_token` made the merge resolution a mess.

PR #420, opened 2019-07-18
example of include = generates an error

briggleman:
the example above with the `include =` generates the following error: `An argument named "include" is not expected here. Did you mean to define a block of type "include"?` removing the extra `=` after include removes this error and correctly creates the policy.


jacobbednarz:
Thank you! This is awesome :rocket:

PR #421, opened 2019-07-18
Deprecation for `zone` in favour of explicit `zone_id`

jacobbednarz:
Zone names within Cloudflare are not universally unique which makes
performing a lookup based on just the name a little troublesome. Two
accounts can have the same zone name and both of these can be returned
if the lookups are not restricted to the zone ID or account ID.

To rememdy this, `zone` is being deprecated in favour of explicit
`zone_id` values. See the [discussion](https://github.com/terraform-providers/terraform-provider-cloudflare/issues/161#issuecomment-441128985) on this for full details.

Following Terraform's [guidelines for deprecations](https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-removal), we are putting in
a deprecation notice as part of a minor release and then in our next
major release, we will actually perform the schema and functionality
removal. This update will not prevent people from using `zone` however
use will trigger a warning that it's going away and they should migrate
themselves.

### DIFF
--- a/cloudflare/resource_cloudflare_access_rule.go
+++ b/cloudflare/resource_cloudflare_access_rule.go
@@ -22,10 +22,11 @@ func resourceCloudflareAccessRule() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"zone": {
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
-				Computed: true,
+				Type:       schema.TypeString,
+				Optional:   true,
+				ForceNew:   true,
+				Computed:   true,
+				Deprecated: "`zone` is deprecated in favour of explicit `zone_id` and will be removed in the next major release",
 			},
 			"zone_id": {
 				Type:     schema.TypeString,
--- a/cloudflare/resource_cloudflare_filter.go
+++ b/cloudflare/resource_cloudflare_filter.go
@@ -22,10 +22,11 @@ func resourceCloudflareFilter() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"zone": {
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
-				Computed: true,
+				Type:       schema.TypeString,
+				Optional:   true,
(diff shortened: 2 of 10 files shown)


jacobbednarz:
Confirmed this is working for a tested resource.

```
$ terraform apply

Warning: cloudflare_filter.wordpress: "zone": [DEPRECATED] `zone` is deprecated in favour of explicit `zone_id` and will be removed in the next major release


Warning: cloudflare_firewall_rule.wordpress: "zone": [DEPRECATED] `zone` is deprecated in favour of explicit `zone_id` and will be removed in the next major release
```

patryk:
🥽

SteveGoldthorpe-Work:
Think we've been a bit premature with this.
pagerule with zone specified:
```
Warning: "zone": [DEPRECATED] `zone` is deprecated in favour of explicit `zone_id` and will be removed in the next major release

  on pagerules_xxx.tf line 3, in resource "cloudflare_page_rule" "my_page_rule":
   3: resource "cloudflare_page_rule" "my_page_rule" {
```
pagerule with zone_id specified:
```
Error: Missing required argument

  on pagerules_xxx.tf line 3, in resource "cloudflare_page_rule" "my_page_rule":
   3: resource "cloudflare_page_rule" "my_page_rule" {

The argument "zone" is required, but no definition was found.
```
pagerule with both zone and zone_id specified:
```
Error: "zone_id": this field cannot be set

  on pagerules_xxx.tf line 3, in resource "cloudflare_page_rule" "my_page_rule":
   3: resource "cloudflare_page_rule" "my_page_rule" {
```
Looking in resource_cloudflare_page_rule.go, it seems zone is still Required and zone_id is computed (hence the above).  We probably shouldn't add deprecation warnings until we can take both the new AND old fields.  Maybe set them both to Optional and have a check at least one (or only one if both wouldn't be valid) is set?
Terraform v0.12.6
+ provider.cloudflare v1.17.0

patryk:
Good catch, @SteveGoldthorpe-Work. We should make `zone_id` settable. Forked to a new issue.

PR #423, opened 2019-07-18
Move zone tests under `cfapi.net`

jacobbednarz:
Updates the tests to not use `foo.net`, `baa.*` as they are real domains
and keep triggering anti-abuse warnings. Instead, use a dedicated test
domains which are under the `cfapi.net` domain.


jacobbednarz:
This has addressed the CI failures.

PR #428, opened 2019-07-23
Add a 'tcp' type monitor to cloudflare_load_balancer_monitor

### LINKED ISSUE #427: add a "tcp" type monitor to cloudflare_load_balancer_monitor
### Terraform Version
Terraform v0.12.5
+ provider.cloudflare v1.16.1

### Affected Resource(s)
cloudflare_load_balancer_monitor

### Feature request
I would like to create a cloudflare_load_balancer_monitor of type "tcp".
The resource currently supports "http and "https" only but "tcp" is available through the cloudflare api.

Here is my PR #428

tristan-weil:
This fix #427

I am not comfortable with go and terraform-provider dev, so I hope it's ok.

### DIFF
--- a/cloudflare/resource_cloudflare_load_balancer_monitor.go
+++ b/cloudflare/resource_cloudflare_load_balancer_monitor.go
@@ -147,9 +147,6 @@ func resourceCloudflareLoadBalancerPoolMonitorCreate(d *schema.ResourceData, met
 		Retries:  d.Get("retries").(int),
 	}
 
-	//
-	// common
-	//
 	if description, ok := d.GetOk("description"); ok {
 		loadBalancerMonitor.Description = description.(string)
 	}
@@ -203,9 +200,6 @@ func resourceCloudflareLoadBalancerPoolMonitorCreate(d *schema.ResourceData, met
 		}
 	}
 
-	//
-	// create
-	//
 	log.Printf("[DEBUG] Creating Cloudflare Load Balancer Monitor from struct: %+v", loadBalancerMonitor)
 
 	r, err := client.CreateLoadBalancerMonitor(loadBalancerMonitor)
@@ -235,9 +229,6 @@ func resourceCloudflareLoadBalancerPoolMonitorUpdate(d *schema.ResourceData, met
 		Retries:  d.Get("retries").(int),
 	}
 
-	//
-	// common
-	//
 	if description, ok := d.GetOk("description"); ok {
 		loadBalancerMonitor.Description = description.(string)
 	}
@@ -291,9 +282,6 @@ func resourceCloudflareLoadBalancerPoolMonitorUpdate(d *schema.ResourceData, met
 		}
 	}
 
-	//
-	// update
-	//
 	log.Printf("[DEBUG] Update Cloudflare Load Balancer Monitor from struct: %+v", loadBalancerMonitor)
 
 	_, err := client.ModifyLoadBalancerMonitor(loadBalancerMonitor)
--- a/cloudflare/resource_cloudflare_load_balancer_monitor.go
+++ b/cloudflare/resource_cloudflare_load_balancer_monitor.go
@@ -25,33 +25,12 @@ func resourceCloudflareLoadBalancerMonitor() *schema.Resource {
 
 		SchemaVersion: 0,
 		Schema: map[string]*schema.Schema{
-			"expected_body": {
-				Type:     schema.TypeString,
-				Required: true,
-			},
-
-			"expected_codes": {
-				Type:     schema.TypeString,
-				Required: true,
-			},
-
-			"method": {
+			//
+			// common
+			//
+			"description": {
 				Type:     schema.TypeString,
 				Optional: true,
-				Default:  "GET",
-			},
-
-			"timeout": {
-				Type:         schema.TypeInt,
-				Optional:     true,
-				Default:      5,
-				ValidateFunc: validation.IntBetween(1, 10),
-			},
-
-			"path": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Default:  "/",
 			},
 
 			"interval": {
@@ -61,6 +40,18 @@ func resourceCloudflareLoadBalancerMonitor() *schema.Resource {
 			},
 			// interval has to be larger than (retries+1) * probe_timeout:
 
+			"method": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"port": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				ValidateFunc: validation.IntBetween(0, 65535),
+			},
+
 			"retries": {
 				Type:         schema.TypeInt,
 				Optional:     true,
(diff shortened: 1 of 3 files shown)


jacobbednarz:
@tristan-weil Thanks for the PR! Looking at this, it seems as though the API documentation doesn't outline the functionality your building upon at https://api.cloudflare.com. Did you see the documentation for this elsewhere? If not, can you please raise a support ticket with Cloudflare to have this section of the documentation updated _before_ we merge this? We don't generally merge changes that aren't publicly documented due to the potential issues of building on an undocumented feature set.

tristan-weil:
Hello,

In the "type" parameter at https://api.cloudflare.com/#load-balancer-monitors-create-monitor:
`The protocol to use for the health check. Currently supported protocols are 'HTTP','HTTPS' and 'TCP'.`
(And of course, I saw it directly in the dashboard UI).

But I agree the doc could be confusing as "expected_codes" is in the "required parameters" block (even if it is stated that's only for http and https). 

I tried to open a ticket with Cloudflare but I got an error when submitting the ticket :(
I'll try later

jacobbednarz:
Nice one! Thanks for that @tristan-weil. I'll take a look at this shortly.

cloudflare/resource_cloudflare_load_balancer_monitor.go
```diff
@@ -25,20 +25,38 @@ func resourceCloudflareLoadBalancerMonitor() *schema.Resource {
 
 		SchemaVersion: 0,
 		Schema: map[string]*schema.Schema{
-			"expected_body": {
+			//
```
jacobbednarz: I don't think these comments are adding any value and we can do away with them. We should be able to use the schema (via `ValidateFunc` and `ConflictsWith`) itself to determine if we are providing the incorrect values.

cloudflare/resource_cloudflare_load_balancer_monitor.go
```diff
@@ -48,31 +66,44 @@ func resourceCloudflareLoadBalancerMonitor() *schema.Resource {
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"modified_on": {
 				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			//
```
jacobbednarz: Same comment as https://github.com/terraform-providers/terraform-provider-cloudflare/pull/428/files#r307064855

cloudflare/resource_cloudflare_load_balancer_monitor.go
```diff
@@ -135,20 +141,15 @@ func resourceCloudflareLoadBalancerPoolMonitorCreate(d *schema.ResourceData, met
...
-
-	if header, ok := d.GetOk("header"); ok {
-		loadBalancerMonitor.Header = expandLoadBalancerMonitorHeader(header)
+		Timeout:  d.Get("timeout").(int),
+		Type:     d.Get("type").(string),
+		Interval: d.Get("interval").(int),
+		Retries:  d.Get("retries").(int),
 	}
 
+	//
```
jacobbednarz: Similar comment to https://github.com/terraform-providers/terraform-provider-cloudflare/pull/428/files#r307064855

cloudflare/resource_cloudflare_load_balancer_monitor.go
```diff
@@ -157,14 +158,54 @@ func resourceCloudflareLoadBalancerPoolMonitorCreate(d *schema.ResourceData, met
...
+		}
+
+		if path, ok := d.GetOk("path"); ok {
+			loadBalancerMonitor.Path = path.(string)
+		} else {
+			loadBalancerMonitor.Path = "/"
+		}
 	}
 
+	//
```
jacobbednarz: This should be right to go as the methods are self explanatory.

cloudflare/resource_cloudflare_load_balancer_monitor.go
```diff
@@ -187,21 +228,16 @@ func resourceCloudflareLoadBalancerPoolMonitorUpdate(d *schema.ResourceData, met
...
-	if header, ok := d.GetOk("header"); ok {
-		loadBalancerMonitor.Header = expandLoadBalancerMonitorHeader(header)
+		ID:       d.Id(),
+		Timeout:  d.Get("timeout").(int),
+		Type:     d.Get("type").(string),
+		Interval: d.Get("interval").(int),
+		Retries:  d.Get("retries").(int),
 	}
 
+	//
```
jacobbednarz: Similar comment to https://github.com/terraform-providers/terraform-provider-cloudflare/pull/428/files#r307064855

cloudflare/resource_cloudflare_load_balancer_monitor.go
```diff
@@ -210,14 +246,54 @@ func resourceCloudflareLoadBalancerPoolMonitorUpdate(d *schema.ResourceData, met
...
 
-	if allowInsecure, ok := d.GetOk("allow_insecure"); ok {
-		loadBalancerMonitor.AllowInsecure = allowInsecure.(bool)
-	}
+	switch loadBalancerMonitor.Type {
+	case "tcp":
+		if method, ok := d.GetOk("method"); ok {
+			loadBalancerMonitor.Method = method.(string)
+		} else {
+			loadBalancerMonitor.Method = "connection_established"
```
jacobbednarz: This isn't a HTTP verb. Is it intended?

cloudflare/resource_cloudflare_load_balancer_monitor.go
```diff
@@ -210,14 +246,54 @@ func resourceCloudflareLoadBalancerPoolMonitorUpdate(d *schema.ResourceData, met
...
-	if followRedirects, ok := d.GetOk("follow_redirects"); ok {
-		loadBalancerMonitor.FollowRedirects = followRedirects.(bool)
+		if path, ok := d.GetOk("path"); ok {
+			loadBalancerMonitor.Path = path.(string)
+		} else {
+			loadBalancerMonitor.Path = "/"
+		}
 	}
 
+	//
```
jacobbednarz: This can go. The methods are self explanatory.

cloudflare/resource_cloudflare_load_balancer_monitor.go
tristan-weil: Yes, it's only valid when the "type" is "tcp" and this block is only is in the swtich/case "tcp" statement.

cloudflare/resource_cloudflare_load_balancer_monitor.go
tristan-weil: I have not been able to find a way to validate one parameter based on the content of another the parameter. For example, we don't want "expected_body" to be set if the content of "type" is "tcp".
That's why I put those comments in the schema.

But if you have a sample of code to check a parameter based on the content of another one or if you still think the comments are useless, I'll remove them.

cloudflare/resource_cloudflare_load_balancer_monitor.go
tristan-weil: removed

cloudflare/resource_cloudflare_load_balancer_monitor.go
tristan-weil: removed

cloudflare/resource_cloudflare_load_balancer_monitor.go
tristan-weil: removed

cloudflare/resource_cloudflare_load_balancer_monitor.go
tristan-weil: removed

jacobbednarz:
Thanks for the updates. Can we also ensure we update the website documentation in this PR too to reflect any of these changes? Source file can be found [here](https://github.com/terraform-providers/terraform-provider-cloudflare/blob/master/website/docs/r/load_balancer_monitor.html.markdown).

tristan-weil:
sorry!, I completely missed this.
Fixed now.

jacobbednarz:
I think we're ready to roll here. Thanks a bunch for your persistence here @tristan-weil! Your contribution is greatly appreciated 🍭

PR #437, opened 2019-08-01
provider: Use real Terraform version in UA header

radeksimko:
The existing implementation leverages Terraform's `version` package as Terraform currently provides the SDK for interacting with core and is therefore imported by the provider, including the `version` package.

The side effect of this is that provider reports version of vendored Terraform (i.e. plugin SDK), not the version which actually talks to the provider's gRPC server and parses the configuration.

This PR is leveraging recent changes introduced in 0.12 to fix that and report accurate version.

----

This also changes the header from the following

```
Terraform/X.Y.Z terraform-provider-cloudflare/X.Y.Z
```
to
```
HashiCorp Terraform/X.Y.Z (+https://www.terraform.io) terraform-provider-cloudflare/X.Y.Z
```

to align UA string with https://github.com/hashicorp/terraform/pull/22272

### DIFF
--- a/cloudflare/provider.go
+++ b/cloudflare/provider.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"strings"
 
 	"github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/go-cleanhttp"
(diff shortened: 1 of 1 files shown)


jacobbednarz:
@radeksimko Looks like hashicorp/terraform#22272 has landed, is this ready for merge now?

radeksimko:
We are in the process of resolving some upstream dependency issues, before we cut the next version and update this PR, please hold.

https://github.com/hashicorp/terraform/pull/22490

radeksimko:
@patryk I have updated the description of the PR so you can understand the full impact which may not be immediately obvious + I rebased to the latest `master`.

Could you please confirm you're ok with it and happy for me to merge?

patryk:
Thanks @radeksimko, great to see it in!

PR #443, opened 2019-08-14
Implement Workers KV Namespaces

ashnewmanjones:
Implemented Workers KV Namespace

Tested with Terraform v0.12.6

```
provider "cloudflare" {
  email = "${var.cloudflare_email}"
  token = "${var.cloudflare_token}"
  org_id = "${var.cloudflare_org}"
}

resource "cloudflare_workers_kv_namespace" "namespace" {
  title = "example"
}
```

This is my first contribution to an OSS Go project so I apologise for any non-idiomatic code, more than happy to fix any suggestions.


cloudflare/resource_cloudflare_workers_kv_namespace_test.go
```diff
@@ -0,0 +1,84 @@
...
+					resource.TestCheckResourceAttr(
+						resourceName, "title", rnd,
+					),
+				),
+			},
+		},
+	})
+}
+
+func testAccCloudflareWorkersKVNamespaceDesroy(s *terraform.State) error {
```
jacobbednarz:
```suggestion
func testAccCloudflareWorkersKVNamespaceDestroy(s *terraform.State) error {
```

cloudflare/resource_cloudflare_workers_kv_namespace_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+func TestAccCloudflareWorkersKVNamespace_Basic(t *testing.T) {
+	t.Parallel()
+	var namespace cloudflare.WorkersKVNamespace
+	rnd := generateRandomResourceName()
+	resourceName := "cloudflare_workers_kv_namespace." + rnd
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCloudflareWorkersKVNamespaceDesroy,
```
jacobbednarz:
```suggestion
		CheckDestroy: testAccCloudflareWorkersKVNamespaceDestroy,
```

website/docs/r/workers_kv_namespace.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+
+# cloudflare_workers_kv_namespace
+
+Provides a Workers KV Namespace
+
+## Example Usage
+
+```hcl
+resource "cloudflare_workers_kv_namespace" "example" {
+  title        = "test-namespace"
```
jacobbednarz:
```suggestion
  title = "test-namespace"
```

website/docs/r/workers_kv_namespace.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `title` - (Required) The name of the namespace you wish to create
```
jacobbednarz:
```suggestion
* `title` - (Required) The name of the namespace you wish to create.
```

website/docs/r/workers_kv_namespace.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+
+
+## Import
+
+KV Namespace settings can be imported using the it's title e.g.
```
jacobbednarz: I think we should be using IDs here as they will be unique whereas title values are not guaranteed. What do you think?

ashnewmanjones: Namespace titles are unique per Cloudflare account so I thought it may be simpler to just use its title rather than ID.

jacobbednarz: That's fair enough. Have you considered use cases where a title contains spaces? How would we import them using this method?

ashnewmanjones: Good point will update it shortly.

adam-janis:
can this be merged? :)

jimsorock:
Would this also close https://github.com/terraform-providers/terraform-provider-cloudflare/issues/277 ?

adam-janis:
> Would this also close #277 ?

It will not, this one only adds the `cloudflare_workers_kv_namespace` resource

adam-janis:
@jacobbednarz do you think this can be merged? we have another MR to create for the KV namespace binding in `cloudflare_worker` resource and the `cloudflare_workers_kv_namespace` resource is needed in order to have tests passing :)

patryk:
Looking good. Thanks!

PR #445, opened 2019-08-16
Added Priority field to cloudflare_zone_lockdown schema

mbrowne:
The ZoneLockdown API has a `Priority` field that was previously undocumented. They have since documented it, and I created a PR to add it to cloudflare-go, which was merged:
https://github.com/cloudflare/cloudflare-go/pull/266

This PR adds the Priority field to the `cloudflare_zone_lockdown` resource in this provider.


jacobbednarz:
@mbrowne Just thinking about this, as the priority field is optional, we should add another test where this isn't present to exercise that code path.

mbrowne:
Ok, I added a test that passes only the required fields (`paused` and `description` are also optional fields).

jacobbednarz:
Awesome! Thanks @mbrowne! You rock 🤘

PR #448, opened 2019-08-19
Fix expected_body not accepting empty string

### LINKED ISSUE #447: cloudflare_load_balancer_monitor rejects empty string for expected_body
Hi there,

### Terraform Version
```
Terraform v0.12.6
+ provider.aws v2.24.0
+ provider.cloudflare v1.17.1
+ provider.template v2.1.2
```

### Affected Resource(s)

```hcl
resource "cloudflare_load_balancer_monitor" "k8s_ingress" {
  type             = "https"
  method           = "GET"
  path             = "/healthz"
  port             = var.k8s_ingress_monitor_port
  expected_codes   = "200"
  expected_body    = ""
  timeout          = 3
  interval         = 60
  retries          = 3
  description      = "${var.project_name} Kubernetes Ingress Check"
  allow_insecure   = true
  follow_redirects = false
}
```

### Debug Output
Skipping this, can add if needed, this is not an internal issue.

### Panic Output
N/A

### Expected Behavior
Empty string is a valid body to expect, so this should have cleanly created the resource.

### Actual Behavior
Whether creating or modifying, this value is rejected
```
module.the_watch_platform.module.cloudflare_lb_monitor.cloudflare_load_balancer_monitor.k8s_ingress: Modifying... [id=3a0704af92a48e7d677627a4f53833d4]

Error: expected_body must be set

  on ../modules/cloudflare/lb_monitor/main.tf line 16, in resource "cloudflare_load_balancer_monitor" "k8s_ingress":
  16: resource "cloudflare_load_balancer_monitor" "k8s_ingress" {

```

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
N/A

### References

This code here is the problem:

https://github.com/terraform-providers/terraform-provider-cloudflare/blob/master/cloudflare/resource_cloudflare_load_balancer_monitor.go#L170-L174

GetOk checks for a "non zero" value. But the only zero value possible in strings is the empty string: `""` 

You will need to mark this value as required, and use `.Get()`, to get the intended effect of requiring it to be set to something, but allowing empty string.

SpamapS:
This string is perfectly valid in the CloudFlare API but is rejected
because it counts as a "zero value" in the Terraform schema libraries.

Closes #447

### DIFF
--- a/cloudflare/resource_cloudflare_load_balancer_monitor.go
+++ b/cloudflare/resource_cloudflare_load_balancer_monitor.go
@@ -93,7 +93,7 @@ func resourceCloudflareLoadBalancerMonitor() *schema.Resource {
 
 			"expected_body": {
 				Type:     schema.TypeString,
-				Optional: true,
+				Required: true,
 			},
 
 			"expected_codes": {
@@ -163,14 +163,11 @@ func resourceCloudflareLoadBalancerPoolMonitorCreate(d *schema.ResourceData, met
 			loadBalancerMonitor.Method = "connection_established"
 		}
 	case "http", "https":
+		expectedBody := d.Get("expected_body")
+		loadBalancerMonitor.ExpectedBody = expectedBody.(string)
+
 		if allowInsecure, ok := d.GetOk("allow_insecure"); ok {
 			loadBalancerMonitor.AllowInsecure = allowInsecure.(bool)
-		}
-
-		if expectedBody, ok := d.GetOk("expected_body"); ok {
-			loadBalancerMonitor.ExpectedBody = expectedBody.(string)
-		} else {
-			return fmt.Errorf("expected_body must be set")
 		}
 
 		if expectedCodes, ok := d.GetOk("expected_codes"); ok {
--- a/cloudflare/resource_cloudflare_load_balancer_monitor_test.go
+++ b/cloudflare/resource_cloudflare_load_balancer_monitor_test.go
@@ -265,6 +265,15 @@ resource "cloudflare_load_balancer_monitor" "test" {
 }`
 }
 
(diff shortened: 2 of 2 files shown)


cloudflare/resource_cloudflare_load_balancer_monitor_test.go
```diff
@@ -265,6 +265,15 @@ resource "cloudflare_load_balancer_monitor" "test" {
 }`
 }
 
+func testAccCheckCloudflareLoadBalancerMonitorConfigEmptyExpectedBody() string {
```
jacobbednarz: This test doesn't actually assert anything as is. We'll need to create a test case similar to https://github.com/terraform-providers/terraform-provider-cloudflare/pull/448/files#diff-ff2414310a6fa415074bb1da98684b38R91-R104 for asserting that this resource is created and looks like we'd expect it to.

SpamapS: Thanks! Will try and push a true assertion there soon.

cloudflare/resource_cloudflare_load_balancer_monitor.go
```diff
@@ -93,7 +93,7 @@ func resourceCloudflareLoadBalancerMonitor() *schema.Resource {
 
 			"expected_body": {
 				Type:     schema.TypeString,
-				Optional: true,
+				Required: true,
```
tristan-weil: this will raise an error when creating a tcp monitor because 'expected_body' is only valid with http(s) monitors

Evesy:
@SpamapS Are you still looking at finishing this PR off? We have hit a similar issue

jacobbednarz:
@Evesy feel free to pick this up and run with it. It’s been sitting for long enough.

jacobbednarz:
Superseded by #539

PR #451, opened 2019-08-21
Update to cloudflare-go 0.10.0

### LINKED ISSUE #227: Breaking changes for a major release

jacobbednarz:
Pulls in the deprecation of `Organization` in favour of `Account`.

Closes #227


jacobbednarz:
Pending Hashicorp adding `CLOUDFLARE_ACCOUNT_ID` environment variable for the tests to run correctly.

jacobbednarz:
CI environment variable added and suite is green.

patryk:
Let's not rename fields yet, just update internal fields to accommodate breaking change in cloudflare-go. Let's put aside field renames for major version release.

jacobbednarz:
@patryk this PR is intended to be that breaking change for the next major release 😛 Along with #417, this rounds out the changes we are doing for the next major release. I’ll also now need to remove the `zone` stuff too as you’ve merged that one.

jacobbednarz:
I've updated these to use the base branch of v2.0 so that we can test all of these changes together on a single branch and then just merge that one instead of managing merge conflicts all over the place.

PR #459, opened 2019-08-27
Rename "token" to "api_key"

jacobbednarz:
API "token" is currently used for HTTP authentication requests where the
`X-Auth-Key` needs to be set. Until recently, it was fine to use the
wrong term for this authentication method (as there was only the one)
however there is now a feature from Cloudflare which is actually called
API tokens and it requires a different authentication procedure.

The functionality for API tokens has been released (v1.18.0) and this
cleans up the old references from "token" to "api_key" to better reflect
it's purpose.

This attempt supersedes #417 where we were attempting to rename "token"
and introduce the new value at the same time. Instead, we're leaving
"api_token" in place for now and just renaming "token".

Note: This is merging into a release branch, not `master`. The release branch
will be merged and a release cut when we confirm all the behaviour is working
for the breaking changes.


jacobbednarz:
@garrettgalow @hunts This is an updated version of #417 whereby we just rename "token" instead of try to introduce the API token use as well.

At the moment, I don't really mind if we leave `api_token` as is 🤷‍♂

jacobbednarz:
integration test suite is green 🍏

patryk:
My preference goes to rename `key` to `api_key` and leave `api_token` as is. This would make naming consistent.

jacobbednarz:
While consistent, I do think the `api_` prefix is superfluous and the only reason we have `api_token` was to handle getting in API token support without a breaking change (context in  #450). 

I'm not sure I can buy into the consistency argument here as we then have `email` that is out on it's own. Personally, I'd like to drop `api_` prefixes entirely to make the keys (`key`/`token`) explain their purpose however I'm not sure if that is too much change given that we've just introduced `api_token` 🤷‍♂

jacobbednarz:
You've twisted my arm. Let's go with `api_token` and `api_key` for explicitness 🙂

PR #463, opened 2019-09-03
Update example resources to v0.12 syntax, and fix depends_on for worker_route

benjamin-maynard:
This PR updates the resource examples in the documentation to be more Terraform 0.12 friendly by removing the unnecessary use of interpolation `${}` when not required.

It also fixes an incorrect `depends_on` meta-argument in the cloudflare_worker_route resource. The resource in the `depends_on` meta-argument should not be in quotes.

`depends_on = ["cloudflare_worker_script.my_script"]` therefore becomes `depends_on = [cloudflare_worker_script.my_script]`

### DIFF
--- a/website/docs/r/access_policy.html.markdown
+++ b/website/docs/r/access_policy.html.markdown
@@ -42,7 +42,7 @@ resource "cloudflare_access_policy" "test_policy" {
   }
 
   require = {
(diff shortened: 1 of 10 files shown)


jacobbednarz:
Thank you!

PR #467, opened 2019-09-16
Update cloudflare page rule attribute

syed-awais-ali:
Since `zone` is deprecated in favour of explicit `zone_id` which is settable so I have updated the docs to reflect the same change 

https://github.com/terraform-providers/terraform-provider-cloudflare/pull/452


syed-awais-ali:
Hey @jacobbednarz can I have a review from you on this? Thanks

patryk:
Thanks @syed-awais-ali !

PR #468, opened 2019-09-16
Remove use_account_from_zone for v2.0 refactor

patryk:
Removes Account ID discovery logic, turning it into an explicit choice.
Part of breaking change track for v2.0.


patryk:
Then it goes! :)

PR #469, opened 2019-09-17
Missing = sign in the configuration block

macedogm:
Missing = sign in the configuration block. Otherwise Terraform v0.12.3 and CloudFlare provider v1.18.1 will complain about "Error: Missing required argument" and "Error: Unsupported block type"

`Error: Missing required argument

  on firewall_access_rules.tf line 1, in resource "cloudflare_access_rule" "omitted":
   1: resource "cloudflare_access_rule" "omitted" {

The argument "configuration" is required, but no definition was found.

Error: Unsupported block type

  on firewall_access_rules.tf line 9, in resource "cloudflare_access_rule" "omitted":
   9:   configuration {

Blocks of type "configuration" are not expected here. Did you mean to define
argument "configuration"? If so, use the equals sign to assign it a value.

`


patryk:
Thanks @macedogm !

PR #472, opened 2019-09-20
SDK-80: Zone to zone_id refactoring

### LINKED ISSUE #470: Refactor Workers resources to remove single-script code
Breaking change for v2.0

Multi-script support is open to everyone, let's remove single-script code paths.

patryk:
Closes #470

### DIFF
--- a/cloudflare/resource_cloudflare_waf_rule.go
+++ b/cloudflare/resource_cloudflare_waf_rule.go
@@ -26,16 +26,9 @@ func resourceCloudflareWAFRule() *schema.Resource {
 				Required: true,
 			},
 
-			"zone": {
-				Type:       schema.TypeString,
-				Optional:   true,
-				Deprecated: "`zone` is deprecated in favour of explicit `zone_id` and will be removed in the next major release",
-			},
-
 			"zone_id": {
 				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
+				Required: true,
 			},
 
 			"package_id": {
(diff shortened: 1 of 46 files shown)


jacobbednarz:
I’m about to fly back home so I’ll take a look at this first thing Monday and do some testing with our internal setup.

website/docs/r/waf_rule.html.markdown
```diff
@@ -34,13 +34,12 @@ The following arguments are supported:
...
 * `package_id` - The ID of the WAF Rule Package that contains the rule.
 
 ## Import
 
-Rules can be imported using a composite ID formed of zone name and the WAF Rule ID, e.g.
+Rules can be imported using a composite ID formed of zone ID and the WAF Rule ID, e.g.
 
 ```
-$ terraform import cloudflare_waf_rule.100000 example.com/100000
+$ terraform import cloudflare_waf_rule.100000 ae36f999674d196762efcc5abb06b345/100000
```
XaF: Why not keep this as the only case where the zone name could be used ?
This is user-facing, and it's easier for an user to provide the zone name.
In the rest of the code, it's possible to use a `zone` resource or `zones` data to fetch the zone ID to use in other blocks. Running `terraform import` however is usually manual.

jacobbednarz: Please see #227 and https://github.com/terraform-providers/terraform-provider-cloudflare/issues/161#issuecomment-441128985 for full details on why this has been done.

XaF: I totally understand why in the code side (i.e. terraform resources and all), but the `terraform import` command is user-facing and should be easy to use. How can I grab the zone ID from the UI ? Do I need to do an API call to have access to it ? If so, then it's not adapted to `terraform import` !

I'm not arguing with the removal of the zone name elsewhere, just about the specific import method.

jacobbednarz: The reasoning is identical to why it’s being deprecated for the internals; ambiguity in which zone name is intended to be used. 

The zone ID is available from the overview UI of a Cloudflare zone (right hand side navigation bar).

cloudflare/resource_cloudflare_waf_rule.go
```diff
@@ -169,18 +144,13 @@ func resourceCloudflareWAFRuleImport(d *schema.ResourceData, meta interface{}) (
 
 	// split the id so we can lookup
 	idAttr := strings.SplitN(d.Id(), "/", 2)
-	var zoneName string
+	var zoneID string
```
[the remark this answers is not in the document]
jacobbednarz: As mentioned in the thread at https://github.com/terraform-providers/terraform-provider-cloudflare/pull/472#discussion_r326809534, we are fully moving away from zone name so this isn’t something we’d be looking to support.

jacobbednarz:
The integration suite is now pretty green. Just waiting on `CLOUDFLARE_ALT_ZONE_ID` to be added to clean up the last of the issues.

patryk:
Great, after we merge this one, I'll start working on Workers (sic!), need to branch from `v2.0` to avoid merge conflicts. Also, we need to write migration guide.

patryk:
Thanks all!

PR #474, opened 2019-09-20
Upgrade to cloudflare-go v0.10.1

XaF:
This is required for two other pull requests to include support for WAF Packages and WAF Groups.

PR #475, opened 2019-09-20
Add support for WAF packages

XaF:
Depends on #474

### DIFF
--- a/cloudflare/resource_cloudflare_waf_package.go
+++ b/cloudflare/resource_cloudflare_waf_package.go
@@ -5,8 +5,8 @@ import (
 	"strings"
 
 	cloudflare "github.com/cloudflare/cloudflare-go"
-	"github.com/hashicorp/terraform/helper/schema"
-	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 )
 
 func resourceCloudflareWAFPackage() *schema.Resource {
--- a/cloudflare/resource_cloudflare_waf_package.go
+++ b/cloudflare/resource_cloudflare_waf_package.go
@@ -1,0 +1,187 @@
+package cloudflare
+
+import (
+	"fmt"
+	"strings"
+
+	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+)
+
+func resourceCloudflareWAFPackage() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceCloudflareWAFPackageCreate,
+		Read:   resourceCloudflareWAFPackageRead,
+		Update: resourceCloudflareWAFPackageUpdate,
+		Delete: resourceCloudflareWAFPackageDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: resourceCloudflareWAFPackageImport,
+		},
+
+		SchemaVersion: 0,
+		Schema: map[string]*schema.Schema{
+			"package_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"zone_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"sensitivity": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "high",
+				ValidateFunc: validation.StringInSlice([]string{"high", "medium", "low", "off"}, false),
+			},
+
+			"action_mode": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "challenge",
+				ValidateFunc: validation.StringInSlice([]string{"simulate", "block", "challenge"}, false),
+			},
+		},
+	}
+}
+
+func resourceCloudflareWAFPackageRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	packageID := d.Get("package_id").(string)
+	zoneID := d.Get("zone_id").(string)
+
+	pkg, err := client.WAFPackage(zoneID, packageID)
+	if err != nil {
+		return (err)
+	}
+
+	d.Set("sensitivity", pkg.Sensitivity)
+	d.Set("action_mode", pkg.ActionMode)
+	d.SetId(pkg.ID)
+
+	return nil
+}
+
+func resourceCloudflareWAFPackageCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	packageID := d.Get("package_id").(string)
+	zoneID := d.Get("zone_id").(string)
+	sensitivity := d.Get("sensitivity").(string)
+	actionMode := d.Get("action_mode").(string)
+
+	pkg, err := client.WAFPackage(zoneID, packageID)
+	if err != nil {
+		return fmt.Errorf("Unable to find WAF Package %s", packageID)
+	}
+
+	d.Set("zone_id", zoneID)
+	d.Set("package_id", packageID)
+	d.Set("sensitivity", sensitivity)
+	d.Set("action_mode", actionMode)
+
+	// Set the ID to the package_id parameter passed in from the user.
+	// All WAF packages already exist so we already know the package_id.
+	//
+	// This is a work around as we are not really "creating" a WAF Package,
+	// only associating it with our terraform config for future updates.
+	d.SetId(packageID)
+
+	if pkg.Sensitivity != sensitivity || pkg.ActionMode != actionMode {
+		err = resourceCloudflareWAFPackageUpdate(d, meta)
+		if err != nil {
+			d.SetId("")
+			return err
+		}
+	}
+
+	return nil
+}
+
+func resourceCloudflareWAFPackageDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	packageID := d.Get("package_id").(string)
+	zoneID := d.Get("zone_id").(string)
+
+	pkg, err := client.WAFPackage(zoneID, packageID)
+	if err != nil {
+		return err
+	}
+
+	// Can't delete WAF Package so instead reset it to default
+	schema := resourceCloudflareWAFPackage().Schema
+	defaultSensitivity := schema["sensitivity"].Default.(string)
+	defaultActionMode := schema["action_mode"].Default.(string)
+
+	if pkg.Sensitivity != defaultSensitivity || pkg.ActionMode != defaultActionMode {
+		options := cloudflare.WAFPackageOptions{
+			Sensitivity: defaultSensitivity,
+			ActionMode:  defaultActionMode,
+		}
+
+		_, err = client.UpdateWAFPackage(zoneID, packageID, options)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func resourceCloudflareWAFPackageUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	packageID := d.Get("package_id").(string)
+	zoneID := d.Get("zone_id").(string)
+	sensitivity := d.Get("sensitivity").(string)
+	actionMode := d.Get("action_mode").(string)
+
+	options := cloudflare.WAFPackageOptions{
+		Sensitivity: sensitivity,
+		ActionMode:  actionMode,
+	}
+
+	_, err := client.UpdateWAFPackage(zoneID, packageID, options)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func resourceCloudflareWAFPackageImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	client := meta.(*cloudflare.API)
+
+	// split the id so we can lookup
+	idAttr := strings.SplitN(d.Id(), "/", 2)
+	var zoneID string
+	var packageID string
+	if len(idAttr) == 2 {
+		zoneID = idAttr[0]
+		packageID = idAttr[1]
+	} else {
+		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"zoneID/PackageID\" for import", d.Id())
+	}
+
+	pkg, err := client.WAFPackage(zoneID, packageID)
+	if err != nil {
+		return nil, err
+	}
+
+	d.Set("package_id", pkg.ID)
+	d.Set("zone_id", zoneID)
+	d.Set("sensitivity", pkg.Sensitivity)
+	d.Set("action_mode", pkg.ActionMode)
+
+	d.SetId(pkg.ID)
+
+	return []*schema.ResourceData{d}, nil
+}
--- a/cloudflare/resource_cloudflare_waf_package_test.go
+++ b/cloudflare/resource_cloudflare_waf_package_test.go
@@ -1,0 +1,117 @@
+package cloudflare
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccCloudflareWAFPackage_CreateThenUpdate(t *testing.T) {
+	t.Parallel()
+	zoneID := os.Getenv("CLOUDFLARE_ZONE_ID")
+	packageID, err := testAccGetWAFPackage(zoneID)
+	if err != nil {
+		t.Errorf(err.Error())
+	}
+
+	rnd := generateRandomResourceName()
+	name := fmt.Sprintf("cloudflare_waf_package.%s", rnd)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckCloudflareWAFPackageDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckCloudflareWAFPackageConfig(zoneID, packageID, "medium", "simulate", rnd),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(name, "package_id", packageID),
+					resource.TestCheckResourceAttr(name, "zone_id", zoneID),
+					resource.TestCheckResourceAttr(name, "sensitivity", "medium"),
+					resource.TestCheckResourceAttr(name, "action_mode", "simulate"),
+				),
+			},
+			{
+				Config: testAccCheckCloudflareWAFPackageConfig(zoneID, packageID, "low", "block", rnd),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(name, "package_id", packageID),
+					resource.TestCheckResourceAttr(name, "zone_id", zoneID),
+					resource.TestCheckResourceAttr(name, "sensitivity", "low"),
+					resource.TestCheckResourceAttr(name, "action_mode", "block"),
+				),
+			},
+		},
+	})
+}
+
+func testAccGetWAFPackage(zoneID string) (string, error) {
+	config := Config{}
+	if apiToken, ok := os.LookupEnv("CLOUDFLARE_API_TOKEN"); ok {
+		config.APIToken = apiToken
+	} else if apiKey, ok := os.LookupEnv("CLOUDFLARE_API_KEY"); ok {
+		config.APIKey = apiKey
+		if email, ok := os.LookupEnv("CLOUDFLARE_EMAIL"); ok {
+			config.Email = email
+		} else {
+			return "", fmt.Errorf("Cloudflare email is not set correctly")
+		}
+	} else {
+		return "", fmt.Errorf("Cloudflare credentials are not set correctly")
+	}
+
+	client, err := config.Client()
+	if err != nil {
+		return "", err
+	}
+
+	pkgList, err := client.ListWAFPackages(zoneID)
+	if err != nil {
+		return "", fmt.Errorf("Error while listing WAF packages: %s", err)
+	}
+
+	for _, pkg := range pkgList {
+		if pkg.DetectionMode == "anomaly" {
+			return pkg.ID, nil
+		}
+	}
+
+	return "", fmt.Errorf("No anomaly package found")
+}
+
+func testAccCheckCloudflareWAFPackageDestroy(s *terraform.State) error {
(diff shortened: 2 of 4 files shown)


jacobbednarz:
Please remove the dependency update from this PR, we'll do that separately.

cloudflare/resource_cloudflare_waf_package.go
```diff
@@ -0,0 +1,255 @@
...
+				Computed: true,
+			},
+
+			"package_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"zone": {
```
jacobbednarz: We aren't supporting `zone` for old resources as of 2.0 so let's just drop it here entirely and only support zone ID going forward.

XaF:
> Please remove the dependency update from this PR, we'll do that separately.

I can't as it depends on methods that are only available in the latest `cloudflare-go`

cloudflare/resource_cloudflare_waf_package.go
```diff
@@ -0,0 +1,255 @@
...
+			"zone_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"sensitivity": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "high",
```
jacobbednarz: Let's remove the defaults here and instead provide validation that the sensitivity is one of the accepted values.

cloudflare/resource_cloudflare_waf_package.go
```diff
@@ -0,0 +1,255 @@
...
+			},
+
+			"action_mode": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "challenge",
```
jacobbednarz: Similar to https://github.com/terraform-providers/terraform-provider-cloudflare/pull/475/files#r326930670, let's move towards better validation without a default.

cloudflare/resource_cloudflare_waf_package.go
```diff
@@ -0,0 +1,255 @@
...
+	client := meta.(*cloudflare.API)
+
+	packageName := d.Get("package_name").(string)
+	packageID := d.Get("package_id").(string)
+	zone := d.Get("zone").(string)
+	zoneID := d.Get("zone_id").(string)
+	sensitivity := d.Get("sensitivity").(string)
+	actionMode := d.Get("action_mode").(string)
+
+	// While we are deprecating `zone`, we need to perform the validation
```
jacobbednarz: This block can go as we aren't supporting zone names for these resources.

cloudflare/resource_cloudflare_waf_package.go
```diff
@@ -0,0 +1,255 @@
...
+			"zone": {
+				Type:       schema.TypeString,
+				Optional:   true,
+				Computed:   true,
+				Deprecated: "`zone` is deprecated in favour of explicit `zone_id` and will be removed in the next major release",
+			},
+
+			"zone_id": {
+				Type:     schema.TypeString,
+				Optional: true,
```
jacobbednarz:
```suggestion
				Required: true,
```

cloudflare/resource_cloudflare_waf_package.go
```diff
@@ -0,0 +1,255 @@
...
+
+	// split the id so we can lookup
+	idAttr := strings.SplitN(d.Id(), "/", 2)
+	var zoneName string
+	var packageNameOrID string
+	if len(idAttr) == 2 {
+		zoneName = idAttr[0]
+		packageNameOrID = idAttr[1]
+	} else {
+		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"zoneName/PackageNameOrID\" for import", d.Id())
```
jacobbednarz: As we're only supporting zone ID we can get rid of the zone name references.

```suggestion
		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"zoneID/PackageNameOrID\" for import", d.Id())
```

cloudflare/resource_cloudflare_waf_package.go
XaF: Can't we simply do both? We need to know the default anyway to be able to act properly on the `delete` action, might as well use it as the default value if not provided.
I will add the Enum of valid values too, but thought it would be something that the API can directly check instead of forcing it here (doing that means any new value added for those in the API in the future will require an update to that object)

cloudflare/resource_cloudflare_waf_package.go
```diff
@@ -0,0 +1,255 @@
...
+func resourceCloudflareWAFPackageImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	client := meta.(*cloudflare.API)
+
+	// split the id so we can lookup
+	idAttr := strings.SplitN(d.Id(), "/", 2)
+	var zoneName string
+	var packageNameOrID string
+	if len(idAttr) == 2 {
+		zoneName = idAttr[0]
+		packageNameOrID = idAttr[1]
```
jacobbednarz: What cases isn't a package ID available? Ideally we don't use names for references in case they ever change inside of Cloudflare and that doesn't reflect in the terraform state.

XaF: Packages ID do not seem easily available through the interface (same for groups in the other PR), which is why I allowed matching through the name.
WAF rules IDs are easily accessible (you can see them right when clicking on a WAF Group), but WAF Groups IDs and WAF Packages IDs are not?

cloudflare/resource_cloudflare_waf_package.go
jacobbednarz: This isn't the provided default that a customer gets out of the box so we're making an assumption here to make the default high which comes with trade offs and not something we want to force on people. 

If you want to leave this, it needs to be the actual default that comes for a customer. 

> We need to know the default anyway to be able to act properly on the delete action, might as well use it as the default value if not provided.

We don't _have_ to have delete options if the resource doesn't support them. Once it's in Terraform, it can just be managed for `Update`s. If this complicates it, just noop the delete with an explanation in the method.

cloudflare/resource_cloudflare_waf_package.go
jacobbednarz: I don't think it's really viable. The name has spaces and is subject to change if the package ends up renamed. The IDs can be pulled from the API so perhaps we just include the documentation for fetching it from there for importing? The [`cloudflare/cf-terraforming`](https://github.com/cloudflare/cf-terraforming/) project might add support for this and look after those calls for the user too.

XaF: As you wish, I can change it.
This will make the terraform configuration less easy to read though, but if there's tooling around it, why not

cloudflare/resource_cloudflare_waf_package.go
XaF: > This isn't the provided default that a customer gets out of the box so we're making an assumption here to make the default high which comes with trade offs and not something we want to force on people. 

I was just following the API defaults: https://api.cloudflare.com/#waf-rule-packages-edit-firewall-package
If the defaults are different, please tell me what they are so I can change that.

> We don't have to have delete options if the resource doesn't support them. Once it's in Terraform, it can just be managed for Updates. If this complicates it, just noop the delete with an explanation in the method.

This would be weak and would be a different behavior than what is used for WAF Rules, which would be very confusing (unless we prefer disabling the delete from WAF Rules too ?). I think it makes more sense to have sort of a "reset to default", but I totally agree that the "reset to default" should reset to the actual default.

XaF:
Updated to only include those changes, but of course it won't pass the tests for now

XaF:
Rebased on master

XaF:
Fixed small mistakes I made when adding the `ValidateFunc`

XaF:
Updated the `ValidateFunc` to use the helper as suggested by @jacobbednarz !

XaF:
@patryk @jacobbednarz is there anything I could do so we can move forward with this PR and the other one for WAF groups ?

XaF:
Added documentation for the change

jacobbednarz:
I've just kicked off the integration tests for these and will report back shortly! Thanks for following through here @XaF!

jacobbednarz:
Looks like there is a failure in the test suite.

```
------- Stdout: -------
=== RUN   TestAccCloudflareWAFPackage_CreateThenUpdate
=== PAUSE TestAccCloudflareWAFPackage_CreateThenUpdate
=== CONT  TestAccCloudflareWAFPackage_CreateThenUpdate
--- FAIL: TestAccCloudflareWAFPackage_CreateThenUpdate (0.00s)
    testing.go:569: Step 0 error: Missing newline after argument: On /opt/teamcity-agent/temp/buildTmp/tf-test483636842/main.tf line 3: An argument definition must end with a newline.
FAIL
```

I'm able to replicate this running locally too.

XaF:
Not sure I understand, isn't it saying it's a tf issue?
Do you have any pointer as to what might be happening? Happy to fix but I'm not sure how to reproduce locally?

jacobbednarz:
The path in the error is a bit misleading (it's just where the ephemeral file is created and ran from).

The issue is in the `TestAccCloudflareWAFPackage_CreateThenUpdate` which you've added. To reproduce it locally, you need to run the integration tests which there are some instructions in the [README](https://github.com/terraform-providers/terraform-provider-cloudflare#developing-the-provider). I generally run it with the environment variables inline.

```
$ CLOUDFLARE_API_TOKEN="<CLOUDFLARE_API_TOKEN>" \
  CLOUDFLARE_ACCOUNT_ID="<CLOUDFLARE_ACCOUNT_ID>" \
  CLOUDFLARE_DOMAIN="<CLOUDFLARE_ZONE_NAME>" \
  CLOUDFLARE_ZONE_ID="<CLOUDFLARE_ZONE_ID>" \
  TF_ACC=1 \
  go test -count=1 -timeout 30s \
  github.com/terraform-providers/terraform-provider-cloudflare/cloudflare -run "^TestAccCloudflareWAFPackage_CreateThenUpdate" -v
```

I haven't dug into this specific error before so I can't give you any pointers off the top of my head. I'd need to dig in.

XaF:
I think I caught it:

Replaced:
```
  zone_id = %[1]s
```

By
```
  zone_id = "%[1]s"
```

Wondering if the non-numeric ID was the issue here.
Not on my computer right now so I'll see if I can try running the test tomorrow.

jacobbednarz:
Nice find! I've run this locally again and now there is a different error.

```
=== RUN   TestAccCloudflareWAFPackage_CreateThenUpdate
=== PAUSE TestAccCloudflareWAFPackage_CreateThenUpdate
=== CONT  TestAccCloudflareWAFPackage_CreateThenUpdate
--- FAIL: TestAccCloudflareWAFPackage_CreateThenUpdate (0.26s)
    testing.go:569: Step 0 error: errors during apply:
        
        Error: Unable to find WAF Package a25a9a7e9c00afc1fb2e0245519d725b
        
          on /opt/teamcity-agent/temp/buildTmp/tf-test574147229/main.tf line 2:
          (source code not available)
        
        
FAIL
```

Looking at the WAF packages endpoint responses from two different zones, it appears as though the IDs are unique to the zones. To combat this, in your test you may need to fetch all the WAF packages and then use the first one in the response to assert against instead of this static value.

XaF:
Oh! I thought we had the same ID for the same package in all zones. I guess that might change a bit the logic too as we might be able to automatically deduce the zone_id from the package_id when the former is not specified.

Will check that tomorrow and update! :)

XaF:
@jacobbednarz do you have any pointer as to how to use the client to call `ListWAFPackages` before starting the tests ?

i.e. how can I use the provider to grab the package ID before the tests are initialized and thus have configured the provider... ?

jacobbednarz:
I would add it as a part of the `PreCheck` step for the test.

```diff
--- a/cloudflare/resource_cloudflare_waf_package_test.go
+++ b/cloudflare/resource_cloudflare_waf_package_test.go
@@ -18,7 +18,10 @@ func TestAccCloudflareWAFPackage_CreateThenUpdate(t *testing.T) {
        name := fmt.Sprintf("cloudflare_waf_package.%s", rnd)

        resource.Test(t, resource.TestCase{
-               PreCheck:     func() { testAccPreCheck(t) },
+               PreCheck:     func() {
+                       testAccPreCheck(t)
+                       setupWAFPackage()
+               },
                Providers:    testAccProviders,
                CheckDestroy: testAccCheckCloudflareWAFPackageDestroy,
                Steps: []resource.TestStep{
@@ -77,3 +80,17 @@ func testAccCheckCloudflareWAFPackageConfig(zoneID, packageID, sensitivity, acti
                                        action_mode = "%[4]s"
                                }`, zoneID, packageID, sensitivity, actionMode, name)
 }
+
+func setupWAFPackage() {
+       client, clientErr := sharedClient()
+       if clientErr != nil {
+               // handle error
+       }
+
+       packages, err := client.ListWAFPackages()
+       if err != nil {
+               // handle error
+       }
+
+       // ...
+}
```

I haven't tested this out but it should put you on the right path.

XaF:
> I would add it as a part of the `PreCheck` step for the test.

The problem with that approach is that I need to know the `packageID` to prepare the config, which is evaluated at the time we're creating the test, so if we grab the actual `packageID` only when running the tests, we cannot act properly.

I thus went and created a new function that creates a new client specifically to grab the information needed, as you can see in the updated code. Please tell me what you think about that @jacobbednarz and if you have any better way to implement this ?

For now, the acceptance tests are passing:
```
=== RUN   TestAccCloudflareWAFPackage_CreateThenUpdate
=== PAUSE TestAccCloudflareWAFPackage_CreateThenUpdate
=== CONT  TestAccCloudflareWAFPackage_CreateThenUpdate
2019/10/08 17:26:36 [INFO] Cloudflare Client configured for user:
--- PASS: TestAccCloudflareWAFPackage_CreateThenUpdate (3.65s)
PASS
ok  	github.com/terraform-providers/terraform-provider-cloudflare/cloudflare	3.676s
```

cloudflare/resource_cloudflare_waf_package_test.go
```diff
@@ -0,0 +1,117 @@
...
+					resource.TestCheckResourceAttr(name, "action_mode", "block"),
+				),
+			},
+		},
+	})
+}
+
+func testAccGetWAFPackage(zoneID string) (string, error) {
+	config := Config{}
+	if apiToken, ok := os.LookupEnv("CLOUDFLARE_API_TOKEN"); ok {
```
jacobbednarz: I don't really like this in here as we're duplicating the logic we already have in the provider setup. If that ever changes, we now have somewhere that depends on it directly. Was there a reason you couldn't use the [`sharedClient`](https://github.com/terraform-providers/terraform-provider-cloudflare/blob/7df730622f105a6737afedd2c24453c855b32322/cloudflare/cloudflare_sweeper_test.go#L17)? We're already using that in sweepers and it leverages the existing methods.

jacobbednarz: If we can't use the `shareClient` functionality, let's just copy the function and use it here. Don't worry about the logic as we can swap it all when/if we move to API tokens for the test suite.

Another drawback to having this logic here is that `testAccPreCheck` doesn't check for the API token, only the older method of authenticating.

jacobbednarz:
@XaF are you able to avoid rebasing until we're ready to merge? It's quite difficult to review the individual changes you're making. 

We can squash it when it's ready to merge if you'd prefer that.

XaF:
> @XaF are you able to avoid rebasing until we're ready to merge? It's quite difficult to review the individual changes you're making.

Yes, sorry for that... I try to push fixup commits usually but forgot this time!

cloudflare/resource_cloudflare_waf_package_test.go
XaF: > Another drawback to having this logic here is that testAccPreCheck doesn't check for the API token, only the older method of authenticating which means use of API tokens isn't enforced at all.

Note that `testAccPreCheck` is still run just after, so any check it does will be done here, so that should be fine ? But true that we're listing the packages before having done that check...

> I don't really like this in here as we're duplicating the logic we already have in the provider setup. If that ever changes, we now have somewhere that depends on it directly. Was there a reason you couldn't use the sharedClient? We're already using that in sweepers and it leverages the existing methods.

I get this when using the `sharedClient`:

```
    resource_cloudflare_waf_package_test.go:18: Error while listing WAF packages: error from makeRequest: HTTP status 400: content "{\"success\":false,\"errors\":[{\"code\":6003,\"message\":\"Invalid request headers\",\"error_chain\":[{\"code\":6103,\"message\":\"Invalid format for X-Auth-Key header\"}]}],\"messages\":[],\"result\":null}"
```

Which means the `client.ListWAFPackages` is failing. Any idea what would be missing ?

jacobbednarz: > `sharedClient` currently seems to be forcing to use API Key & email instead of API token, is this expected ?

Yep, we'll eventually move this over to API tokens when it's supported fully.

WRT to your error, are you setting the `CLOUDFLARE_API_KEY` and `CLOUDFLARE_EMAIL` per the client initialisation when you're running the tests? As you mentioned, the PreCheck usually handles this but because we're running before it, it could be missed.

XaF: I did not want to go fetch an API key so I used `CLOUDFLARE_API_KEY="blah"` for my tests above, which is why I had that error (this is also why I corrected my message after the fact, realizing the error was related to that).

I think my last push (fixup this time, sorry again) should be good to go ? (using `sharedClient`)

jacobbednarz: Gotcha, that makes total sense then 😄

jacobbednarz:
Great stuff. I've just kicked off the integration suite again and we'll see how we go this time.

XaF:
🤞

jacobbednarz:
Looks like the unit tests are failing to build - https://travis-ci.org/terraform-providers/terraform-provider-cloudflare/jobs/595347342#L234

Integration tests are also failing in a similar fashion.

```

[09:22:28]
cloudflare/resource_cloudflare_waf_package.go:8:2: cannot find package "github.com/hashicorp/terraform/helper/schema" in any of:
[09:22:28]
	/opt/teamcity-agent/work/962a4ee8807683e3/src/github.com/terraform-providers/terraform-provider-cloudflare/vendor/github.com/hashicorp/terraform/helper/schema (vendor tree)
[09:22:28]
	/opt/goenv/versions/1.12.9/src/github.com/hashicorp/terraform/helper/schema (from $GOROOT)
[09:22:28]
	/opt/teamcity-agent/work/962a4ee8807683e3/src/github.com/hashicorp/terraform/helper/schema (from $GOPATH)
[09:22:28]
cloudflare/resource_cloudflare_waf_package.go:9:2: cannot find package "github.com/hashicorp/terraform/helper/validation" in any of:
[09:22:28]
	/opt/teamcity-agent/work/962a4ee8807683e3/src/github.com/terraform-providers/terraform-provider-cloudflare/vendor/github.com/hashicorp/terraform/helper/validation (vendor tree)
[09:22:28]
	/opt/goenv/versions/1.12.9/src/github.com/hashicorp/terraform/helper/validation (from $GOROOT)
[09:22:28]
	/opt/teamcity-agent/work/962a4ee8807683e3/src/github.com/hashicorp/terraform/helper/validation (from $GOPATH)
[09:22:28]
Process exited with code 1
```

XaF:
Probably because it's based on a previous version that did not use the same packages in the headers (the path seem to have changed).
Updated with the new paths, but I'm not able to test locally anymore unless I rebase (and thus `push -f`, so will let things run in here...

jacobbednarz:
> Updated with the new paths, but I'm not able to test locally anymore unless I rebase (and thus `push -f`, so will let things run in here...

I don't understand why this is? Are you able to let me know what issues you're hitting? You should always be able to run the tests locally when developing the provider.

XaF:
Oh, failed again as I forgot to update all of them.

It's just that my local branch is not `on top of the current master`, and to get on top of the current master I need to rebase.
The errors I get are kind of the opposite really:

> ==> Checking that code complies with gofmt requirements...
go install -ldflags="-X github.com/terraform-providers/terraform-provider-cloudflare/version.ProviderVersion=v2.0.0-8-g43ed656"
cloudflare/resource_cloudflare_waf_package.go:8:2: cannot find package "github.com/hashicorp/terraform-plugin-sdk/helper/schema" in any of:
	/Users/xaf/src/github.com/terraform-providers/terraform-provider-cloudflare/vendor/github.com/hashicorp/terraform-plugin-sdk/helper/schema (vendor tree)
	/usr/local/Cellar/go/1.12.4/libexec/src/github.com/hashicorp/terraform-plugin-sdk/helper/schema (from $GOROOT)
	/Users/xaf/src/github.com/hashicorp/terraform-plugin-sdk/helper/schema (from $GOPATH)
cloudflare/resource_cloudflare_waf_package.go:9:2: cannot find package "github.com/hashicorp/terraform-plugin-sdk/helper/validation" in any of:
	/Users/xaf/src/github.com/terraform-providers/terraform-provider-cloudflare/vendor/github.com/hashicorp/terraform-plugin-sdk/helper/validation (vendor tree)
	/usr/local/Cellar/go/1.12.4/libexec/src/github.com/hashicorp/terraform-plugin-sdk/helper/validation (from $GOROOT)
	/Users/xaf/src/github.com/hashicorp/terraform-plugin-sdk/helper/validation (from $GOPATH)

jacobbednarz:
> It's just that my local branch is not `on top of the current master`, and to get on top of the current master I need to rebase.

You can still merge in the upstream master and it will have the same effect without needing to rebase or force push changes. We're not fussed if there is merge commits or not so that is up to you.

XaF:
>     resource_cloudflare_waf_package_test.go:18: invalid credentials: key & email must not be empty
>     testing.go:462: Acceptance tests skipped unless env 'TF_ACC' set

Ok, might need to manually test for `TF_ACC` before trying to fetch the package!

XaF:
> All checks have passed

🎉

jacobbednarz:
Looks like the integration tests are happy too! I'll take another look over this and we'll get it merged.

XaF:
🎉

Let me know if you want me to `rebase -i --autosquash` all those fixups.
Once you confirm everything is fine, I'll also push the changes to the "Add support for WAF Groups" one!

website/docs/r/waf_package.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `zone_id` - (Required) The DNS zone ID to apply to.
+* `package_id` - (Required) The WAF Package ID.
+* `sensitivity` - (Required) The sensitivity of the package, can be one of ["high", "medium", "low", "off"].
```
jacobbednarz: This documentation doesn't match the schema. Do you intend for this to be optional (to match the schema)? Same with `action_mode`.

XaF: You're right, should read `Optional`

XaF: Updated

jacobbednarz:
Thanks for your patience with this! If you can apply the same fixes to #476, we'll get that merged too.

PR #476, opened 2019-09-20
Add support for WAF groups

### LINKED ISSUE #325: Add WAF Packages to Cloudflare Terraform
For WAF rules to be enabled or disabled via terraform the package needs to be enabled on Cloudflare. Packaged include for example Cloudflare Specials, Cloudflare Plone etc. 

Currently, you have to enable these rulesets via API or UI before you can apply a configuration for the WAF rules.

Each package has a package ID for example:

`{
  "result": {
    "id": "c8485e461f935bead889fa0fe87dd76e",
    "name": "Cloudflare Php",
    "description": "This ruleset should only be enabled if PHP is used for this domain. It contains additional rules that complement the technology-specific protections provided by similar rules in the OWASP ruleset.",
    "mode": "on",
    "package_id": "1e334934fd7ae32ad705667f8c1057aa",
    "rules_count": 13,
    "modified_rules_count": 13
  },
  "success": true,
  "errors": [],
  "messages": []
}`

so would imagine something like the following in the .tf file:

`resource "cloudflare_waf_rule" "identifier_org_WP_Package" {
    package_id = "1e334934fd7ae32ad705667f8c1057aa"
    zone = "domain.com"
    mode = "on"
}`

is required

XaF:
Depends on #474

Closes #325


jacobbednarz:
Please remove the dependency update from this PR, we'll do that separately.

XaF:
> Please remove the dependency update from this PR, we'll do that separately.

I can't as it depends on methods that are only available in the latest `cloudflare-go`

jacobbednarz:
You can run the newer dependency version locally, just don't commit it. Once the upgrade lands in master, we can then update these branches to include the changes. 

We split it to only review what is actually being changed in the intended PR. Also helps keep the reviews smaller.

XaF:
Updated to only include those changes, but of course it won't pass the tests for now

XaF:
Rebased on master

XaF:
Fixed small mistakes I made when adding the `ValidateFunc`

cloudflare/resource_cloudflare_waf_group.go
```diff
@@ -0,0 +1,210 @@
...
+				Default:  "on",
+				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
+					switch value := val.(string); value {
+					case "on", "off":
+						break
+					default:
+						errs = append(errs, fmt.Errorf("%q must be either 'on' or 'off', got: %s", key, value))
+					}
+					return
+				},
```
jacobbednarz: There is a built in helper for this, [`validation.StringInSlice()`](https://github.com/hashicorp/terraform/blob/00a76d019a3e5b4d9d040a3e28c9e0b4e56ad951/helper/validation/validation.go#L127-L147) which simplifies the reading of this validation.

XaF: Updated!

XaF:
Added documentation for the change

XaF:
Updated

XaF:
> Please also add new documentation page to TOC: https://github.com/terraform-providers/terraform-provider-cloudflare/blob/master/website/cloudflare.erb

Done! I also added the one for the `waf_package` resource as the other PR has been merged :)

XaF:
Anything else I can do for this one @patryk @jacobbednarz ? :)

cloudflare/resource_cloudflare_waf_group.go
```diff
@@ -0,0 +1,203 @@
...
+		pkgList = append(pkgList, cloudflare.WAFPackage{ID: packageID})
+	}
+
+	for _, pkg := range pkgList {
+		var err error
+		var group cloudflare.WAFGroup
+
+		group, err = client.WAFGroup(zoneID, pkg.ID, groupID)
+
+		if err == nil {
```
jacobbednarz: Instead of nesting here, let's flip the check and return early if there is an error. Make the flow a little more readable.

```go
if err != nil {
  // handle error
}

d.Set("group_id", group.ID)

// ...
```

cloudflare/resource_cloudflare_waf_group.go
```diff
@@ -0,0 +1,203 @@
...
+			d.Set("group_id", group.ID)
```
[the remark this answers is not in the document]
XaF: Not sure about the majority, as `group_id` and `package_id` and `zone_id` are expected to exist, only the `mode` is left and the call to `d.SetId`, and if there is any difference to the `mode` we want to set it to the value required by the terraform config, hence changing it ourselves and then using `d.Set` too. I'm not sure if calling `Read` would be useful here ? (seems like it would be an extra API call for nothing)

Note also that I created this file based on the logic used by `resource_cloudflare_waf_rule.go`.

cloudflare/resource_cloudflare_waf_group.go
XaF: Updated

cloudflare/resource_cloudflare_waf_group.go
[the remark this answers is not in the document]
XaF: Currently, we have:

```golang
                group, err = client.WAFGroup(zoneID, pkg.ID, groupID)
                if err != nil {
                        continue
                }

                d.Set("group_id", group.ID)
                d.Set("zone_id", zoneID)
                d.Set("package_id", pkg.ID)
                d.Set("mode", mode)

                // Set the ID to the group_id parameter passed in from the user.
                // All WAF Groups already exist so we already know the group_id
                //
                // This is a work around as we are not really "creating" a WAF Group,
                // only associating it with our terraform config for future updates.
                d.SetId(group.ID)

                if group.Mode != mode {
                        err = resourceCloudflareWAFGroupUpdate(d, meta)
                        if err != nil {
                                d.SetId("")
                                return err
                        }
                }

                return nil
```

And with your suggestion, we would have:

```golang
                group, err = client.WAFGroup(zoneID, pkg.ID, groupID)
                if err != nil {
                        continue
                }

                d.Set("group_id", group.ID)
                d.Set("zone_id", zoneID)
                d.Set("package_id", pkg.ID)

                if group.Mode != mode {
                        err = resourceCloudflareWAFGroupUpdate(d, meta)
                        if err != nil {
                                d.SetId("")
                                return err
                        }
                }

                return resourceCloudflareWAFGroupRead(d, meta) // Extra call to client.WAFGroup to grab the data we just grabbed above
```

Not sure it's worth it ?

XaF:
Added the `CHANGELOG.md` line so you don't need to add it in a separate commit after merging this

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 2.0.1 (Unreleased)
 
+* **New Resource:** `cloudflare_waf_group` ([#476](https://github.com/terraform-providers/terraform-provider-cloudflare/issues/476))
```
jacobbednarz: we usually add this directly on master once it's merged, you don't need to worry about that in this PR.

cloudflare/resource_cloudflare_waf_group.go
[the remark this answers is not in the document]
jacobbednarz: > Not sure about the majority

There are two bits of documentation from Hashicorp on using this approach. The first is about [capturing state in `Read`](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#capture-all-state-in-read) and the second is [updating state following modifications](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#update-state-after-modification). I'd prefer we follow these best practices unless we have a pressing reason not to.

CHANGELOG.md
XaF: I'll remove this then!

cloudflare/resource_cloudflare_waf_group.go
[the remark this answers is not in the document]
XaF: > I'd prefer we follow these best practices unless we have a pressing reason not to.

I felt an extra API call was a good reason given there might be rate limits?
But I can change it with the code offered above.

jacobbednarz:
thanks for getting this one over the line! you rock 🤘

XaF:
Thanks a lot for your time to review and get this done!

PR #478, opened 2019-09-25
SDK-81: Refactor Workers resources to remove single-script code.

### LINKED ISSUE #387: worker_script doesnt work with multi-script for non enterprise accounts
Cloudflare launched multi script support for non enterprise accounts some week ago.
It's not possible to create a multiscript for non enterprise accounts using terraform plugin.

### Terraform Version
2:29 $ terraform -v
Terraform v0.12.2
+ provider.cloudflare (v1.16.0)

### Affected Resource(s)
- cloudflare_worker_script
- cloudflare_worker_route (?)

### Debug Output
```tf
module.maps.cloudflare_worker_script.script: Creating...

Error: error creating worker script: organization ID required for enterprise only request

  on terraform-modules/worker/script.tf line 6, in resource "cloudflare_worker_script" "script":
   6: resource "cloudflare_worker_script" "script" {
```

resource (module):
```tf
resource "cloudflare_worker_script" "script" {
  name    = "${var.name}"
  content = file("dist/${var.name}.js")
}
resource "cloudflare_worker_route" "script-route" {
  zone       = "${var.zone}"
  pattern    = "${var.pattern}"
  enabled    = var.enabled
  depends_on = [cloudflare_worker_script.script]
}
```

### Expected Behavior
scripts and routes should have been created

### Actual Behavior
nothing is created.

https://blog.cloudflare.com/just-write-code-improving-developer-experience-for-cloudflare-workers/#multiscript-for-all

### LINKED ISSUE #470: Refactor Workers resources to remove single-script code

patryk:
Closes #470
Closes #387

### DIFF
--- a/cloudflare/resource_cloudflare_worker_script.go
+++ b/cloudflare/resource_cloudflare_worker_script.go
@@ -46,15 +46,13 @@ func getScriptData(d *schema.ResourceData, client *cloudflare.API) (ScriptData, 
 	scriptName := d.Get("name").(string)
 
 	var params cloudflare.WorkerRequestParams
-	var id string
 
 	params = cloudflare.WorkerRequestParams{
 		ScriptName: scriptName,
 	}
-	id = "name:" + scriptName
 
 	return ScriptData{
-		id,
+		scriptName,
 		params,
 	}, nil
 }
--- a/cloudflare/resource_cloudflare_worker_script.go
+++ b/cloudflare/resource_cloudflare_worker_script.go
@@ -21,20 +21,10 @@ func resourceCloudflareWorkerScript() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"zone_id": {
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
-				// zone_id is used for single-script, name is used for multi-script
-				ConflictsWith: []string{"name"},
-			},
-
 			"name": {
 				Type:     schema.TypeString,
-				Optional: true,
+				Required: true,
 				ForceNew: true,
-				// zone_id is used for single-script, name is used for multi-script
-				ConflictsWith: []string{"zone_id"},
 			},
 
 			"content": {
@@ -53,26 +43,15 @@ type ScriptData struct {
 }
 
 func getScriptData(d *schema.ResourceData, client *cloudflare.API) (ScriptData, error) {
-	zoneID := d.Get("zone_id").(string)
 	scriptName := d.Get("name").(string)
-	if zoneID == "" && scriptName == "" {
-		return ScriptData{}, fmt.Errorf("either `zone_id` or `name` field must be set")
-	}
 
 	var params cloudflare.WorkerRequestParams
 	var id string
 
-	if scriptName != "" {
-		params = cloudflare.WorkerRequestParams{
-			ScriptName: scriptName,
-		}
-		id = "name:" + scriptName
-	} else {
-		params = cloudflare.WorkerRequestParams{
-			ZoneID: zoneID,
-		}
-		id = "zone:" + zoneID
+	params = cloudflare.WorkerRequestParams{
+		ScriptName: scriptName,
 	}
+	id = "name:" + scriptName
 
 	return ScriptData{
 		id,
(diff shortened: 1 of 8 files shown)


cloudflare/resource_cloudflare_worker_script.go
```diff
@@ -53,26 +43,15 @@ type ScriptData struct {
...
-		id = "name:" + scriptName
-	} else {
-		params = cloudflare.WorkerRequestParams{
-			ZoneID: zoneID,
-		}
-		id = "zone:" + zoneID
+	params = cloudflare.WorkerRequestParams{
+		ScriptName: scriptName,
 	}
+	id = "name:" + scriptName
```
jRiest: We could just have the `scriptName` be the ID now that we don't need to differentiate between single-script and multi-script resources. However, I'm not sure what the backwards-compat concerns are around changing the ID format. It will work fine even if we leave the `name:` prefix.

cloudflare/resource_cloudflare_worker_script.go
```diff
@@ -184,24 +163,8 @@ func resourceCloudflareWorkerScriptDelete(d *schema.ResourceData, meta interface
...
-	}
-
-	if scriptType == "name" {
-		d.Set("name", scriptId)
-	} else if scriptType == "zone" {
-		d.Set("zone_id", scriptId)
-	} else {
-		return nil, fmt.Errorf("invalid scriptType (\"%s\") specified, should be either \"name\" or \"zone\"", scriptType)
-	}
+	scriptID := d.Id()
```
jRiest: We may have a mismatch about ID format here. If the IDs follow the `name:{scriptName}` format, then would we still need people to specify `name:{scriptName}` when they import it?

cloudflare/resource_cloudflare_worker_script.go
jacobbednarz: Seeing how this is going into 2.0, we can remove this and make a backwards incompatible change as long as we document the change and what people need to do to keep it working. Something we might want to consider though is a way for existing scripts to migrate to this though. In the past I've used [state migrations](https://www.terraform.io/docs/extend/resources/state-migration.html) however I'm not sure if they are different in 0.12.x now.

website/docs/r/worker_script.html.markdown
```diff
@@ -38,30 +24,15 @@ resource "cloudflare_worker_script" "my_script" {
...
-* `ae36f999674d196762efcc5abb06b345` - the zone ID
-
-### multi-script
-
-To import a script from a multi-script account, use an id like `name:script_name`
+To import a script, use a script name, e.g. `script_name`
 
 ```
-$ terraform import cloudflare_worker_script.default name:script_name
+$ terraform import cloudflare_worker_script.default script_name
```
jacobbednarz: I usually shy away from using user provided names for things as they can change or be reused whereas the IDs that Cloudflare generate do not. Not a big deal but something worth raising I think to save us some hassle later.

patryk: Workers API has made a decision that script name is actually an ID too.
e.g.
```
    {
      "id": "example-script",
      "etag": "830924823da98de8429c9e9596ee79525aa2914fe32a19ca7e7aef77bfee2166",
      "modified_on": "2019-09-23T22:59:55.617676Z",
      "created_on": "2019-09-23T22:31:26.487806Z"
    },
```
becomes on API:
https://api.cloudflare.com/client/v4/accounts/cf0df04d99b840a890163aa457d8ee3d/workers/scripts/example-script

cloudflare/resource_cloudflare_worker_script.go
patryk: Removed.

cloudflare/resource_cloudflare_worker_script.go
patryk: Dropped prefix.

patryk:
CI is all green. Please have a look at the changes.

website/docs/r/worker_script.html.markdown
jacobbednarz: :cool: this is great to know!

jacobbednarz:
This also closes #387

PR #493, opened 2019-10-02
APILIB-901: h2_prioritization setting not able to be set in bulk

patryk:
Because the API is fronted by API Gateway which routes various endpoints into different upstreams, `h2_prioritization` cannot be set through general `/zone/settings` endpoint. We need to call it directly as a separate API call.

### DIFF
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -665,13 +665,3 @@ func resourceCloudflarePageRuleImport(d *schema.ResourceData, meta interface{}) 
 
 	return []*schema.ResourceData{d}, nil
 }
-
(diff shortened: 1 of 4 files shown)


jacobbednarz:
Have we got a way to assert that this is working? Would an integration test cover the expectations here?

jacobbednarz:
An alternative option: Should we just send all settings to their direct endpoint instead of using the bulk endpoint to minimise the code complexity? It would be more HTTP requests but the big upside is that we would have less code paths in the provider code.

patryk:
It is expected all zone settings have their own endpoints, so yeah, it's entirely logical we can streamline the code this way and get rid of edge case logic.

Also, integration test would be a nice to have and would validate the issue fixed here.

PR #498, opened 2019-10-07
addon zone settings info about default values

eliasdorneles:
Update zone settings docs with information about default values,
as documented in:
https://api.cloudflare.com/#zone-settings-properties


patryk:
Great, thank you @eliasdorneles !

PR #505, opened 2019-10-09
Adding example with cloudflare_zones

ruzickap:
It's handy to know how to use `cloudflare_zones` with `cloudflare_record`, because it's not obvious that you need to use `lookup`.


jacobbednarz:
Thanks for this but we already have a documented example of using the output from `cloudflare_zone` data source into another resource. See https://www.terraform.io/docs/providers/cloudflare/d/zones.html. As the resource in question doesn’t really matter here, I don’t think we need to update the data source documentation.

PR #508, opened 2019-10-09
Add data source for WAF Groups

XaF:


jacobbednarz:
Are you able to provide a description of what the use case is for this? I can't think of any and would like to make sure we have a use case before just adding these new datasources.

XaF:
We want to be able to use the values returned by those data sources to use `for_each` on a resource element after and run bulk options for "most of" the groups/packages (for the other PR).

For instance, something like:

```terraform
resource "cloudflare_waf_group" "bulk" {
  for_each = {
    for _, group in data.cloudflare_waf_groups.selected_bulk.groups :
    group['id'] => group
    if group['id'] != xxxx // If I want to exclude this one here because I did a singular action on it
  }

  zone_id = data.cloudflare_waf_groups.selected_bulk.zone_id
  group_id = each.key
  package_id = each.value['package_id']
  mode = "off"
}
```

cc @sbfaulkner

cloudflare/data_source_waf_groups.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+func dataSourceCloudflareWAFGroups() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceCloudflareWAFGroupsRead,
+
+		Schema: map[string]*schema.Schema{
+			"zone_id": {
+				Type:     schema.TypeString,
+				Required: true,
```
patryk: This field is required, yet not documented.

XaF: Fixed! As well as for the `package_id` field.

XaF:
> Please also add new documentation page to TOC: https://github.com/terraform-providers/terraform-provider-cloudflare/blob/master/website/cloudflare.erb

Done!

patryk:
Thanks!

XaF:
Thank you! :)

PR #511, opened 2019-10-13
Account for pending zone setups

### LINKED ISSUE #488: provider ignores pending zone config
### Terraform Version
```
Terraform v0.11.14
+ provider.cloudflare v1.18.1
```

### Affected Resource(s)
- cloudflare_zone

```hcl
resource "cloudflare_zone" "myzone_com" {
  zone   = "myzone.com"
  plan   = "enterprise"
}
```

### Debug Output
```
  ~ cloudflare_zone.myzone_com
      plan:       "free" => "enterprise"
```

### Panic Output
n/a

### Expected Behavior
When a zone has been set to `plan = "enterprise"` (or, presumably, anything other than "free"), the provider should recognize that (based on "pending changes" returned by the API) even **when the zone has not been activated yet**.

### Actual Behavior
When configuring a zone via Terraform, the provider doesn't return the correct state (based on pending zone changes) to Terraform **when the zone has not been activated yet**. As a result, Terraform will want to make the above change (see debug output) over and over... until the zone is activated.

When a zone has not been activated yet, changing the plan will be reflected under "pending changes" (rather than "plan") in the API. However, the provider does not seem to look at pending changes when computing the state of the zone resource and will keep returning `plan = "free"` to Terraform. Terraform then correctly concludes that a change has to be made (even though that change has already been made and is in the "pending" state on the Cloudflare side).

### Steps to Reproduce
1. Create Cloudflare zone (via terraform or by hand and then import into terraform)
2. Do **not** update the registrar with Cloudflare's auth NS servers for the newly created zone
3. Set `plan = "enterprise"` (or, most likely, anything other than "free")
4. Run `terraform apply` multiple times (and actually _apply_ changes)

Terraform will want to make the same change over and over.

### Important Factoids
You obviously need an Enterprise account (or Pro or some other pay-for account) in order to replicate this.

### References
n/a

jacobbednarz:
During the creation of a zone, the plan name is always "free" despite the desired end result of a paid plan while it is in a "pending" state. In most API scenarios this doesn't really matter as the plan value shifts on it's own behind the scenes. However as Terraform keeps track of the input vs the remote state, this becomes problematic as it sees the difference in the desired state.

To account for this, we need to check if the state is pending and then dig into the `plan_pending` object for the real plan in order to do the comparison.

Fixes #488


cloudflare/resource_cloudflare_zone.go
```diff
@@ -162,21 +171,34 @@ func resourceCloudflareZoneRead(d *schema.ResourceData, meta interface{}) error
 		return fmt.Errorf("Error finding Zone %q: %s", d.Id(), err)
 	}
 
+	// In the cases where the zone isn't completely setup yet, we need to
+	// check the `status` field and should it be pending, use the `Name`
+	// from `zone.PlanPending` instead to account for paid plans.
+	var plan string
+	if status, ok := d.GetOk("status"); ok {
+		if status == "pending" && zone.PlanPending.Name != "" {
+			plan = zone.PlanPending.Name
```
jacobbednarz: This is currently failing at https://github.com/terraform-providers/terraform-provider-cloudflare/blob/1a55f4f97636d5663f81ddb39f69d4f17042befa/cloudflare/resource_cloudflare_zone.go#L160 

For some reason, the second time `Read` is called, this is the struct.

```
PlanPending: (cloudflare.ZonePlan) {
  ZonePlanCommon: (cloudflare.ZonePlanCommon) {
   ID: (string) "",
   Name: (string) "",
   Price: (int) 0,
   Currency: (string) "",
   Frequency: (string) ""
  },
  IsSubscribed: (bool) false,
  CanSubscribe: (bool) false,
  LegacyID: (string) "",
  LegacyDiscount: (bool) false,
  ExternallyManaged: (bool) false
 }
```

I'll continue digging into this one.

jacobbednarz: Looking a the `zone` on it's own is 👌 

```
 Plan: (cloudflare.ZonePlan) {
  ZonePlanCommon: (cloudflare.ZonePlanCommon) {
   ID: (string) (len=32) "0feeeeeeeeeeeeeeeeeeeeeeeeeeeeee",
   Name: (string) (len=12) "Free Website",
   Price: (int) 0,
   Currency: (string) (len=3) "USD",
   Frequency: (string) ""
  },
  IsSubscribed: (bool) true,
  CanSubscribe: (bool) false,
  LegacyID: (string) (len=4) "free",
  LegacyDiscount: (bool) false,
  ExternallyManaged: (bool) false
 },
 PlanPending: (cloudflare.ZonePlan) {
  ZonePlanCommon: (cloudflare.ZonePlanCommon) {
   ID: (string) (len=32) "94f3b7b768b0458b56d2cac4fe5ec0f9",
   Name: (string) (len=18) "Enterprise Website",
   Price: (int) 0,
   Currency: (string) (len=3) "USD",
   Frequency: (string) (len=7) "monthly"
  },
  IsSubscribed: (bool) false,
  CanSubscribe: (bool) true,
  LegacyID: (string) (len=10) "enterprise",
  LegacyDiscount: (bool) false,
  ExternallyManaged: (bool) true
 },
 Status: (string) (len=7) "pending",
 ... 
```

However using `zone.PlanPending` directly looks like an empty struct.

```
 PlanPending: (cloudflare.ZonePlan) {
  ZonePlanCommon: (cloudflare.ZonePlanCommon) {
   ID: (string) "",
   Name: (string) "",
   Price: (int) 0,
   Currency: (string) "",
   Frequency: (string) ""
  },
  IsSubscribed: (bool) false,
  CanSubscribe: (bool) false,
  LegacyID: (string) "",
  LegacyDiscount: (bool) false,
  ExternallyManaged: (bool) false
 },
```

🤔

jacobbednarz: Have taken this up with Cloudflare Engineers as there is some stuff that isn't working in the API as expected.

jacobbednarz: Looks like we'll need to swap to using the [zone subscription API](https://api.cloudflare.com/#zone-subscription-update-zone-subscription) to get this working as expected.

jacobbednarz: Addressing the underlying issue in cloudflare/cloudflare-go#362 and once that lands, we will be good to continue shipping this.

cloudflare/resource_cloudflare_zone.go
```diff
@@ -226,28 +254,19 @@ func flattenMeta(d *schema.ResourceData, meta cloudflare.ZoneMeta) map[string]in
...
-		}
-	}
-	return nil, fmt.Errorf("plan '%s' not found amongst the available plans", planID)
+	// Due to the async delivery of the subscription service, there is
+	// potential that the update is made and we query the endpoint
+	// before it's populated in other systems. This is an arbitary number
+	// and it would be preferrable to work out a better way of polling
+	// for changes without having to do it in every system that sets the
+	// rate plan but it works.
+	time.Sleep(3000 * time.Millisecond)
```
jacobbednarz: @patryk any thoughts on this random wait time? Do you have a better approach or insights here that might allow us to do away with a static sleep?

patryk: Progressively longer sleep times, as you implemented in newer commit, is good, we can keep it.

PR #512, opened 2019-10-14
Update cloudflare_zone_settings_override attributes list (remove read only ones)

knapo:
Some of `cloudflare_zone_settings_override` attributes are readonly and should not be listed in doc.

When they are set cloudflare complains and return 
`Error: invalid zone setting "true_client_ip_header" (value: off) found - cannot be set as it is read only`


jacobbednarz:
Some of these attributes are read only based on your plan type. When creating a zone override resource, you should only be putting in the attributes you intend to override, not all of them.

PR #517, opened 2019-10-16
Bump `cloudflare-go` to v0.10.6

jacobbednarz:
Pulls in the changes related to using the subscription API for updating
zone rate plans.


jacobbednarz:
Due to the method signature change, I'm going to bundle this with the work that updates the method use in #511

PR #521, opened 2019-10-17
Add support for Access Service Tokens

### LINKED ISSUE #209: Missing service token support for Cloudflare access
### Affected Resource(s)
- No resource for Cloudflare access "service token" creation
- cloudflare_access_policy 

### Expected Behavior
- A resource to create Cloudflare access service token should be available.
- cloudflare_access_policy should have a valid `non_identity` "decision" attribute that enables "include" attribute to have an object assigned with `access_service_token` or `any_valid_access_service_token` or `common name` (should be defined in `policyOptionElement`) attributes.

### Actual Behavior
We can't create Cloudflare access service tokens.
Existing tokens can't be added to a policy.

### Important Factoids
Service tokens would be commonly used by any application that needs to access protected content.

### References
- https://developers.cloudflare.com/access/connecting-to-apps/service-token/

jacobbednarz:
Introduces the ability to manage Cloudflare Access Service Tokens via
Terraform.

See https://developers.cloudflare.com/access/service-auth/service-token/

Closes #209

### Screenshots

**Inactive sidebar** 

**Active sidebar**

**Documentation**

### DIFF
--- a/website/docs/r/access_service_token.html.markdown
+++ b/website/docs/r/access_service_token.html.markdown
@@ -1,0 +1,56 @@
+---
+layout: "cloudflare"
+page_title: "Cloudflare: cloudflare_access_service_token"
+sidebar_current: "docs-cloudflare-resource-access-service-token"
+description: |-
+  Provides a Cloudflare Access Service Token resource.
+---
+
+# cloudflare_access_service_token
+
+Access Service Tokens are used for service-to-service communication
+when an application is behind Cloudflare Access.
+
+## Example Usage
+
+```hcl
+resource "cloudflare_access_service_token" "my_app" {
+  account_id = "d41d8cd98f00b204e9800998ecf8427e"
+  name       = "CI/CD app"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `account_id` - (Required) The ID of the account where the Access
+Service is being created.
+* `name` - (Required) Friendly name of the token's intent.
+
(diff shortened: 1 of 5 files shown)


website/docs/r/access_service_token.html.markdown
```diff
@@ -0,0 +1,56 @@
...
+## Attributes Reference
+
+The following attributes are exported:
+
+* `client_id` - UUID client ID associated with the Service Token.
+* `client_secret` - A secret for interacting with Access protocols.
+
+## Import
+
+~> **Important:** If you are importing an Access Service Token you will
```
jacobbednarz: 🚨 this is really important 🚨 if you're intending on importing, this should be well known.

patryk: Maybe just skip import if not useful?

jacobbednarz: I wasn't sure if it would be completely useless as if you've already got a system that is using the token but you just want to manage everything in code; this is handy to have. The only caveat to remember is that you can't reference it in other resources. 

I wonder if on import, it should be set to a value that clearly states this incase someone is trying to reference it 🤔

patryk: Better not to set any value, and maybe emit a warning.

jacobbednarz:
Just waiting on the permissions to be updated for the CI user to use Access to get the tests passing.

PR #506, opened 2019-10-09
Add data source for WAF Packages

XaF:

PR #525, opened 2019-10-24
Add data source for WAF Rules

XaF:
Following up on #506 and #508, and as explained in comment https://github.com/terraform-providers/terraform-provider-cloudflare/pull/508#issuecomment-540235472, we need to be able to bulk enable/disable rules too. As the `resource` already existed I did not pay attention that we also needed to create the `data source` previously... But realized that missing part while updating our terraform to the last 2.0.1 release!

This thus adds the new data source and related documentation and acceptance tests.

```
=== RUN   TestAccCloudflareWAFRules_NoFilter
=== PAUSE TestAccCloudflareWAFRules_NoFilter
=== RUN   TestAccCloudflareWAFRules_MatchDescription
=== PAUSE TestAccCloudflareWAFRules_MatchDescription
=== RUN   TestAccCloudflareWAFRules_MatchMode
=== PAUSE TestAccCloudflareWAFRules_MatchMode
=== CONT  TestAccCloudflareWAFRules_NoFilter
=== CONT  TestAccCloudflareWAFRules_MatchMode
=== CONT  TestAccCloudflareWAFRules_MatchDescription
--- PASS: TestAccCloudflareWAFRules_MatchMode (6.40s)
--- PASS: TestAccCloudflareWAFRules_NoFilter (6.66s)
--- PASS: TestAccCloudflareWAFRules_MatchDescription (6.69s)
PASS
ok  	github.com/terraform-providers/terraform-provider-cloudflare/cloudflare	6.720s
```

### DIFF
--- a/cloudflare/data_source_waf_rules.go
+++ b/cloudflare/data_source_waf_rules.go
@@ -162,7 +162,7 @@ func dataSourceCloudflareWAFRulesRead(d *schema.ResourceData, meta interface{}) 
 		return fmt.Errorf("Error setting WAF rules: %s", err)
 	}
 
-	d.SetId("WAFRules " + time.Now().UTC().String())
+	d.SetId(fmt.Sprintf("WAFRules_%s", time.Now().UTC().String()))
 	return nil
 }
 
--- a/cloudflare/data_source_waf_rules.go
+++ b/cloudflare/data_source_waf_rules.go
@@ -1,0 +1,205 @@
+package cloudflare
+
+import (
+	"fmt"
+	"log"
+	"regexp"
+	"time"
+
+	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceCloudflareWAFRules() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceCloudflareWAFRulesRead,
+
+		Schema: map[string]*schema.Schema{
+			"zone_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"package_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"filter": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"description": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"mode": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"group_id": {
+							Type:     schema.TypeString,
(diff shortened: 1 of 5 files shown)


cloudflare/data_source_waf_rules.go
```diff
@@ -0,0 +1,204 @@
...
+			break
+		}
+	}
+
+	err = d.Set("rules", ruleDetails)
+	if err != nil {
+		return fmt.Errorf("Error setting WAF rules: %s", err)
+	}
+
+	d.SetId("WAFRules " + time.Now().UTC().String())
```
jacobbednarz: Using this as the non-unique ID here isn't my ideal as it doesn't match the other formats but we can move forward with it providing that we swap the space for an underscore to make it possible to lookup without any spaces.
```suggestion
	d.SetId(fmt.Sprintf("WAFRules_%s", time.Now().UTC().String()))
```

website/docs/d/waf_rules.html.md
```diff
@@ -0,0 +1,61 @@
...
+
+- `zone_id` - (Required) The ID of the DNS zone in which to search for the WAF Rules.
+- `package_id` - (Optional) The ID of the WAF Rule Package in which to search for the WAF Rules.
+- `filter` - (Optional) One or more values used to look up WAF Rules. If more than one value is given all
+values must match in order to be included, see below for full list.
+
+**filter**
+
+- `description` - (Optional) A regular expression matching the description of the WAF Rules to lookup.
+- `mode` - (Optional) Mode of the WAF Rules to lookup. Valid values: on and off.
```
jacobbednarz:
```suggestion
- `mode` - (Optional) Mode of the WAF Rules to lookup. Valid values: `"on"` and `"off"`.
```

cloudflare/data_source_waf_rules_test.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func testAccCloudflareWAFRulesConfig(zoneID string, filters map[string]string, name string) string {
+	filters_str := make([]string, 0, len(filters))
+	for k, v := range filters {
+		filters_str = append(filters_str, fmt.Sprintf(`%[1]s = "%[2]s"`, k, v))
+	}
+
+	return fmt.Sprintf(`
+				data "cloudflare_waf_rules" "%[1]s" {
```
jacobbednarz: Can you fix the JSON formatting here? It's all over the place.

XaF: Happy to fix it, but could you explain what you mean?

jacobbednarz: Sure, it appears as though this JSON is indented with tabs and currently sits about 35-40 columns in. It would be great if it was readable and didn't pad with additional tabs.

XaF: Done!

PR #539, opened 2019-11-15
Allow empty string for expected_body in Load Balancer Monitors

### LINKED ISSUE #447: cloudflare_load_balancer_monitor rejects empty string for expected_body

Evesy:
Fixes #447

### DIFF
--- a/cloudflare/resource_cloudflare_load_balancer_monitor_test.go
+++ b/cloudflare/resource_cloudflare_load_balancer_monitor_test.go
@@ -69,7 +69,8 @@ func TestAccCloudflareLoadBalancerMonitor_FullySpecified(t *testing.T) {
 func TestAccCloudflareLoadBalancerMonitor_EmptyExpectedBody(t *testing.T) {
 	t.Parallel()
 	var loadBalancerMonitor cloudflare.LoadBalancerMonitor
-	name := "cloudflare_load_balancer_monitor.test"
+	rnd := generateRandomResourceName()
+	name := fmt.Sprintf("cloudflare_load_balancer_monitor.%s", rnd)
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
@@ -77,7 +78,7 @@ func TestAccCloudflareLoadBalancerMonitor_EmptyExpectedBody(t *testing.T) {
 		CheckDestroy: testAccCheckCloudflareLoadBalancerMonitorDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccCheckCloudflareLoadBalancerMonitorConfigEmptyExpectedBody(),
+				Config: testAccCheckCloudflareLoadBalancerMonitorConfigEmptyExpectedBody(rnd),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckCloudflareLoadBalancerMonitorExists(name, &loadBalancerMonitor),
 					// checking empty string value passes all validations and created
@@ -307,13 +308,13 @@ resource "cloudflare_load_balancer_monitor" "test" {
 }`
 }
 
-func testAccCheckCloudflareLoadBalancerMonitorConfigEmptyExpectedBody() string {
-	return `
-resource "cloudflare_load_balancer_monitor" "test" {
+func testAccCheckCloudflareLoadBalancerMonitorConfigEmptyExpectedBody(resourceName string) string {
+	return fmt.Sprintf(`
+resource "cloudflare_load_balancer_monitor" "%[1]s" {
   expected_body = ""
   expected_codes = "2xx"
   description = "we don't want to check for a given body"
-}`
+}`, resourceName)
 }
 
 func testAccCheckCloudflareLoadBalancerMonitorConfigTcpFullySpecified() string {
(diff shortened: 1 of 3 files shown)


Evesy:
@jacobbednarz If you don't mind, would appreciate your opinion on what approach to take for this.

Right now I've just taken the approach of defaulting `expected_body` to `""`, so even though a user-specified value of `""` will not pass the non-zero validation, it will be set accordingly anyway.

The object in the Cloudflare API after creating a Load Balancer Monitor with minimal config sets `expected_body` to `""` so this falls in line with the existing Cloudflare behaviour.

I can't run the full test suite right now so I can't validate this currently works as is. Was hoping you could feedback with a general direction to head before I sink too much time into it (First time touching a Terraform provider, want to make sure I'm doing the right thing!)

jacobbednarz:
Thanks for this @Evesy! The approach looks fine to me. I don't have any reservations about the `""` being set as that's how we're handling it in other places. The main thing here is that is it only used for HTTP/HTTPS monitors and doesn't get applied or prevent a TCP monitor being created.

To run the unit and acceptance tests, you can follow the documentation in the [README](https://github.com/terraform-providers/terraform-provider-cloudflare#developing-the-provider). I've kicked off an integration test run now and will report back on the results when it's finished.

jacobbednarz:
We have a new failure

```
TestAccCloudflareLoadBalancerMonitor_NoRequired

------- Stdout: -------
=== RUN   TestAccCloudflareLoadBalancerMonitor_NoRequired
=== PAUSE TestAccCloudflareLoadBalancerMonitor_NoRequired
=== CONT  TestAccCloudflareLoadBalancerMonitor_NoRequired
--- FAIL: TestAccCloudflareLoadBalancerMonitor_NoRequired (0.97s)
    testing.go:616: Step 0, no error received, but expected a match to:
        
        expected_body must be set
        
        
FAIL
```

Evesy:
Thanks, I've updated that test assertion now since `expected_body` will no longer be required user input.

I'll have to look at getting an isolated environment set up with our Enterprise account in order to run the full acceptance tests as my personal account is on the Free plan

cloudflare/resource_cloudflare_load_balancer_monitor_test.go
```diff
@@ -59,6 +59,29 @@ func TestAccCloudflareLoadBalancerMonitor_FullySpecified(t *testing.T) {
...
+				),
+			},
+		},
+	})
+}
+
+func TestAccCloudflareLoadBalancerMonitor_EmptyExpectedBody(t *testing.T) {
+	t.Parallel()
+	var loadBalancerMonitor cloudflare.LoadBalancerMonitor
+	name := "cloudflare_load_balancer_monitor.test"
```
jacobbednarz: For the name, let's swap it to a randomly generated ID to avoid any potential collisions.

```suggestion
      rnd := generateRandomResourceName()
	name := fmt.Sprintf("cloudflare_load_balancer_monitor.%s", rnd)
```

jacobbednarz:
> I'll have to look at getting an isolated environment set up with our Enterprise account in order to run the full acceptance tests as my personal account is on the Free plan

That's 👌, don't stress too much about it. We have dedicated CI environments setup for this type of thing if you don't/can't get access to a particular type of account.

jacobbednarz:
CI is green now. Once we make that resource name change, I'm happy to land this.

cloudflare/resource_cloudflare_load_balancer_monitor_test.go
jacobbednarz: fixed in a712f502ef7896e94b15b80c0aa38c0e02e1ab05

jacobbednarz:
thanks for this @Evesy! you rock 🤘

Issue #188, opened 2019-01-09
cloudflare-access_policy policyOptionElement is missing access groups


AshleyPoole:
### Affected Resource(s)
- cloudflare_access_policy

### Expected Behavior
Within the cloudflare_access_policy resource, I expect that the "include" option (policyOptionElement) should also support "access groups" as per the UI.

### Actual Behavior
Currently not supported. Only email, email domain, IP and everyone is supported.

### Important Factoids
This is important for people who use access groups across multiple policies.

### References
https://api.cloudflare.com/#access-policy-create-access-policy

Currently supported via API:
Match a specific email - Can be done via Terraform
Match an entire email domain  - Can be done via Terraform
Match everyone  - Can be done via Terraform
Match an IP address block  - Can be done via Terraform
Match an Access Group - Cannot be done via Terraform. Proposing, at least this one option is added.
Match an Azure Group - Cannot be done via Terraform.
Match a Github Organization - Cannot be done via Terraform.
Match a GSuite Group - Cannot be done via Terraform.
Match an Okta Group - Cannot be done via Terraform.
Match a SAML Group - Cannot be done via Terraform.

jacobbednarz:
The access groups API endpoints aren't publicly available yet so this will be pending those APIs being exposed. When they are available, a new resource will be introduced as `access_group` or similar.

msheiny:
> The access groups API endpoints aren't publicly available

OHhhh is that what that means @jacobbednarz ? I was perplexed why when we visit that API page the examples aren't showing up. I even opened [this ticket](https://community.cloudflare.com/t/access-api-examples-not-showing-up-on-api-cloudflare-com/69984) on Cloudflare's discourse forum trying to figure out what is wrong. If description elements are empty on the cloudflare API that historically just means its not fully exposed yet?

grrywlsn:
@jacobbednarz is it still the case that the access group API endpoints aren't publicly available? and if not, is there a timeline for when they might be?

jacobbednarz:
hey @grrywlsn I've just come back from vacation and seen this along with service tokens are now in the public API so this is ready to be implemented. We use Access quite a bit internally so we'll be looking to add support soonish.

purva1192:
I have similar question but for Okta Groups. Is there a way to support Okta groups in the access_policy? Should I open another ticket for this?

jacobbednarz:
[...]

PR #541, opened 2019-11-19
Add `non_identity` to allowed decision schema

jacobbednarz:
This is required for Access Service Token support to allow non-humans to
communicate with the Access endpoints.


erickellerek1:
Hi there,
this new schema is currently not documented.
I would like to know if there are some plan to support all non_identity fields:

Access Groups
IP Ranges
Common Name
Valid Certificates
Access Service Token
Any Valid Access Service Token

since we can currently easily create new access group but there is no reference as include in the documentation... https://www.terraform.io/docs/providers/cloudflare/r/access_policy.html#include

What is missing?

jacobbednarz:
See #188 and #597 for details.

PR #542, opened 2019-11-19
cloudflare_zone_settings_override: Add support for 0rtt and http3

### LINKED ISSUE #528: Support http3 and 0rtt settings
0rtt and http3 are new on/off settings. They seem like things cloudflare_zone_settings_override could support.

jacobbednarz:
Adds documentation and functionality to support the use of `0rtt` and
`http3` for zone configurations.

Closes #528

### DIFF
--- a/cloudflare/resource_cloudflare_zone_settings_override.go
+++ b/cloudflare/resource_cloudflare_zone_settings_override.go
@@ -415,9 +415,17 @@ var resourceCloudflareZoneSettingsSchema = map[string]*schema.Schema{
 	},
 
 	"http2": {
-		Type: schema.TypeString, ValidateFunc: validation.StringInSlice([]string{"on", "off"}, false),
-		Optional: true,
-		Computed: true,
+		Type:         schema.TypeString,
+		ValidateFunc: validation.StringInSlice([]string{"on", "off"}, false),
+		Optional:     true,
+		Computed:     true,
+	},
+
+	"http3": {
+		Type:         schema.TypeString,
+		ValidateFunc: validation.StringInSlice([]string{"on", "off"}, false),
+		Optional:     true,
+		Computed:     true,
 	},
 
 	"pseudo_ipv4": {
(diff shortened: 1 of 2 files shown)


cloudflare/resource_cloudflare_zone_settings_override.go
```diff
@@ -415,9 +415,17 @@ var resourceCloudflareZoneSettingsSchema = map[string]*schema.Schema{
 	},
 
 	"http2": {
-		Type: schema.TypeString, ValidateFunc: validation.StringInSlice([]string{"on", "off"}, false),
-		Optional: true,
-		Computed: true,
+		Type:         schema.TypeString,
+		ValidateFunc: validation.StringInSlice([]string{"on", "off"}, false),
```
jacobbednarz: This puts `ValidateFunc` on it's own line instead of inline. Best review this with `?w=1` for actual change.

cloudflare/resource_cloudflare_zone_settings_override.go
```diff
@@ -467,6 +475,13 @@ var resourceCloudflareZoneSettingsSchema = map[string]*schema.Schema{
 		Optional:     true,
 		Computed:     true,
 	},
+
+	"0rtt": {
```
dantman: Aren't attributes supposed to be identifiers that don't start with a number?

I may not be on the version that this was added. But I get an `An argument or block definition is required here.` error at the start of `0rtt = "off"` when including it in an cloudflare_zone_settings_override.

jacobbednarz: I'll have to have a look. The alpha characters restrictions for other things are associated with resource naming/use but I haven't dug into whether that applies for the schema.

sean-: It does.  This attribute needs to be renamed to something like `zero_rtt` instead.  See the [HCL scanner code](https://github.com/hashicorp/hcl/blob/master/hcl/scanner/scanner.go#L165-L166) for details.

PR #547, opened 2019-12-02
Implement Origin CA certificate resource

### LINKED ISSUE #433: Add Origin CA Certificate resource
As per [this comment](https://github.com/terraform-providers/terraform-provider-cloudflare/issues/224#issuecomment-509979257) on #224, I'd like to be able to provision origin certificates from Terraform. This would enable end-to-end setup for Cloudflare's 'Full' encryption offering.

As per [the API](https://api.cloudflare.com/#origin-ca-properties), I'd imagine syntax would be similar to the following:

```hcl
resource "cloudflare_origin_ca_certificate" "my_certificate" {
  zone = "${cloudflare_zone.my_site.zone}"
  hostnames = [
    "example.com",
    "*.example.com",
  ]
  requested_validity = 5475
  request_type = "origin-rsa" # optional
  csr = "-----BEGIN CERTIFICATE REQUEST----- ..." # optional
}
```

player-two:
Fixes #433
Depends on https://github.com/cloudflare/cloudflare-go/pull/387

Before the sdk pr is merged, this can only be tested via
```
go mod edit -replace github.com/cloudflare/cloudflare-go=/path/to/cloudflare-go
make build
make testacc TESTARGS='-run TestAccCloudflareOriginCACertificate_Basic'
```

I'm not terribly confident about the `api_user_service_key` - is there ever a case where it would be set to something other than the Origin CA key?  If so, I think we'd need a way to save both and have this resource's methods create a new client.

### DIFF
--- a/cloudflare/config.go
+++ b/cloudflare/config.go
@@ -8,10 +8,11 @@ import (
 )
 
 type Config struct {
-	Email    string
-	APIKey   string
-	APIToken string
-	Options  []cloudflare.Option
+	Email             string
+	APIKey            string
+	APIUserServiceKey string
+	APIToken          string
+	Options           []cloudflare.Option
 }
 
 // Client returns a new client for accessing cloudflare.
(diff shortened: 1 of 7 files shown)


patryk:
I've bumped the `cloudflare-go` release, can you rebase this PR?

player-two:
Yep will do!  I'll have to wait on https://github.com/cloudflare/cloudflare-go/pull/392 first

player-two:
That last commit will have to change, either with a legit version number once the library is bumped or removed entirely if dependencies should be handled in a separate PR.  I just added it for now to see if CI would pass.

jacobbednarz:
@player-two you should be able to remove the last commit and just wait on the version bump. We'll merge it once it's landed in cloudflare-go.

PR #550, opened 2019-12-03
Support allowed modes for WAF Rules

XaF:
This was causing an issue when using WAF Rules that did not use the
'default' mode but the on/off approach. This fixes that by switching
the default value to reset depending in the available values in the
`AllowedModes` field.

This also returns the `AllowedModes` when using the WAF rules data source, as the `allowed_modes` parameter in each item of the `rules` list, in order for users to know which modes can be used.


patryk:
Thanks @XaF !

PR #552, opened 2019-12-05
Improvements to the `cloudflare_waf_rule` resource

XaF:
As part of the setup we are currently doing, I required to be able to grab the Group ID in which a rule is. The data source I wrote a few weeks back provides that parameter, but when I'm using the resource, it is not available. This thus includes a fix for that to provide the `group_id` parameter.

This also improves slightly the resource, inspired from things I did in the `cloudflare_waf_group` resource:
- Allow to provide the `package_id`: if I'm currently bulk-applying something and I already know the package ID, providing it allows to avoid 1 API call per `cloudflare_waf_rule` (which is used to list then find the package needed). If the `package_id` is not provided, it will be determined as it was before.
- Call `resourceCloudflareWAFRuleRead(d, meta)` on _Create_ and _Update_; this is a recommendation from terraform that was not applied in that resource, but now is.

None of those changes are breaking.


jacobbednarz:
Thanks @XaF !

PR #556, opened 2019-12-09
Fix typo in import command's resource

NicholasPCole:
A simple fix for a typo I noticed while reading the docs.


jacobbednarz:
thank you! 🎉

PR #562, opened 2019-12-17
update Spectrum application resource to use string enum for proxy_pro…

acmanderson:
…tocol field

Opening this PR against the renovate branch since this depends on the version bump in that branch. Once that's merged this can be retargeted to master if desired. This should fix the CI build for that PR as well.


T4cC0re:
@jacobbednarz Can we get a release cut out for this? Right now the latest release fails with this message:

```
Error: error creating spectrum application for zone: error unmarshalling the JSON response: json: cannot unmarshal string into Go struct field SpectrumApplication.proxy_protocol of type bool
```

jacobbednarz:
@T4cC0re v2.3.0 has now been cut however we were waiting for a couple of other features to land with it, hence the delays.

PR #564, opened 2019-12-20
Handle errors properly in Logpush resource

pawelkowalak:
I was unable to create a Logpush resource in Cloudflare and error is not printed out. Here is example apply:

```
Error: error creating logpush job

  on cloudflare-log-push.tf line 34, in resource "cloudflare_logpush_job" "cloudflare_logpush":
  34: resource "cloudflare_logpush_job" "cloudflare_logpush" {
```

This is resource I'm trying to create:

```
// To generate ownership_challenge see: https://developers.cloudflare.com/logs/logpush/logpush-configuration-api/understanding-logpush-api/#ownership
resource "cloudflare_logpush_job" "cloudflare_logpush" {
  count               = var.cloudflare_logpush_challenge == "" ? 0 : 1
  zone_id             = var.cloudflare_zone_id_map[var.environment]
  name                = "Logpush Job for ${var.environment}"
  destination_conf    = "s3://${module.cloudflare_logpush_bucket.name}/{DATE}?region=${var.region}&sse=AES256"
  logpull_options     = "fields=RayID,ClientIP,EdgeStartTimestamp&timestamps=rfc3339"
  ownership_challenge = var.cloudflare_logpush_challenge
}
```

I've noticed that `error creating logpush job` log line is swallowing original error, probably from Cloudflare API, so I've created this PR to fill the gaps. Also fixed some other error messages.

### DIFF
--- a/cloudflare/resource_cloudflare_logpush_job.go
+++ b/cloudflare/resource_cloudflare_logpush_job.go
@@ -49,9 +49,8 @@ func resourceCloudflareLogpushJob() *schema.Resource {
 
 func getJobFromResource(d *schema.ResourceData) cloudflare.LogpushJob {
 	id, err := strconv.Atoi(d.Id())
-
 	if err != nil {
-		fmt.Errorf("Could not extract Logpush job from resource - invalid identifier: %+v", id)
+		fmt.Errorf("could not extract Logpush job from resource - invalid identifier (%s): %v", d.Id(), err)
 	}
 
 	job := cloudflare.LogpushJob{
@@ -68,9 +67,11 @@ func getJobFromResource(d *schema.ResourceData) cloudflare.LogpushJob {
 func resourceCloudflareLogpushJobRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*cloudflare.API)
 	jobId, err := strconv.Atoi(d.Id())
+	if err != nil {
+		fmt.Errorf("could not extract Logpush job from resource - invalid identifier (%s): %v", d.Id(), err)
+	}
 
 	job, err := client.LogpushJob(d.Get("zone_id").(string), jobId)
-
 	if err != nil {
 		if strings.Contains(err.Error(), "404") {
 			log.Printf("[INFO] Could not find LogpushJob with id: %q", jobId)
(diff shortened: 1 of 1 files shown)


cloudflare/resource_cloudflare_logpush_job.go
```diff
@@ -84,12 +85,21 @@ func resourceCloudflareLogpushJobRead(d *schema.ResourceData, meta interface{})
 		return nil
 	}
 
-	d.Set("name", job.Name)
-	d.Set("enabled", job.Enabled)
-	d.Set("logpull_options", job.LogpullOptions)
-	d.Set("destination_conf", job.DestinationConf)
-	d.Set("ownership_challenge", d.Get("ownership_challenge"))
-
+	if err := d.Set("name", job.Name); err != nil {
```
jacobbednarz: I’m not sure whether this is adding any value as `d.Set` is a provider internal method which manages the state and the conditions of which it errors get into some pretty gnarly edge cases that we can’t solve here.

cloudflare/resource_cloudflare_logpush_job.go
```diff
@@ -100,11 +110,9 @@ func resourceCloudflareLogpushJobCreate(d *schema.ResourceData, meta interface{}
 	log.Printf("[DEBUG] Creating Cloudflare Logpush Job from struct: %+v", job)
 
 	j, err := client.CreateLogpushJob(d.Get("zone_id").(string), job)
-
 	if err != nil {
-		return fmt.Errorf("error creating logpush job")
+		return fmt.Errorf("error creating logpush job: %v", err)
```
pawelkowalak: @jacobbednarz this is the most important change in this PR, currently this error is swallowed so we don't know why Logpush job was not created. I haven't tested this change yet, but I think this fix is obvious. I can remove other fixes from this PR (but I'm not convinced that `d.Set()` errors are safe to ignore). Anyway, I'll try testing build with this change in my environment to check what is the actual problem when creating this Logpush job and I'll get back to you.

jacobbednarz: That’s fine, I don’t have any issues expanding this error like you’ve done. 

As I mentioned above, the `Set`s are the only thing that I’m questioning whether we’re adding any value as almost every other instance of it’s usage (including outside of this provider) do not include error handling like you’ve added and I’d be looking to find the root cause as opposed to just logging like we’ve done here.

Mystou:
Hello,
I come here because I've issue with Logpush Job and terraform.
I don't know if @pawelkowalak have managed to create a logpush ressource, but I don't think the cloudflare_logpush_job can work in state.

There is a missing part in the cloudflare_logpush_job resource: you can't validate the challenge.

I think the easiest way to do that, is to call "GetLogpushOwnershipChallenge" when there is no challenge provided (and so, make the challenge non required), and when you update the ressource with the challenge, call the "ValidateLogpushOwnershipChallenge".

pawelkowalak:
Hi @Mystou, I didn't have time to debug it more or even check what error is hidden. We've created Logpush manually for now. We'll get back to this at some point probably. I'll update this PR according to @jacobbednarz comments.

jacobbednarz:
Thanks for this @pawelkowalak :tada: !

PR #587, opened 2020-01-22
Add attributes reference for cloudflare_access_application

dhoeric:
Find myself difficult to cross-reference the attributes from the doc, until I find the attributes from source code.

ref. https://github.com/terraform-providers/terraform-provider-cloudflare/blob/master/cloudflare/resource_cloudflare_access_application.go


jacobbednarz:
Thank you @dhoeric!

Issue #565, opened 2019-12-21
Add Workers KV data resource


danielnaab:
The provider supports adding Worker KV namespaces. It would also be very useful to have a resource to add values to the namespace, particularly because Worker KV is a recommended method for providing secrets to Service Workers.

### Related resources
- cloudflare_worker_script
- cloudflare_workers_kv_namespace

### References
- https://github.com/terraform-providers/terraform-provider-cloudflare/issues/164
- https://github.com/terraform-providers/terraform-provider-cloudflare/issues/277
- https://github.com/terraform-providers/terraform-provider-cloudflare/pull/443

PR #595, opened 2020-02-12
Add worker kv

nrf110:
Support Workers KV Pairs.  This should resolve [#565](https://github.com/terraform-providers/terraform-provider-cloudflare/issues/565)

Proposed resource syntax:
`resource "cloudflare_workers_kv" "example_kv" {
  namespace_id = cloudflare_workers_kv_namespace.example_ns.id
  key = "test-key"
  value = "some value"
}`

### DIFF
--- a/cloudflare/resource_cloudflare_worker_kv.go
+++ b/cloudflare/resource_cloudflare_worker_kv.go
@@ -1,0 +1,127 @@
+package cloudflare
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strings"
+
+	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/pkg/errors"
+)
+
+func resourceCloudflareWorkerKV() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceCloudflareWorkersKVCreate,
+		Read:   resourceCloudflareWorkersKVRead,
+		Update: resourceCloudflareWorkersKVUpdate,
+		Delete: resourceCloudflareWorkersKVDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceCloudflareWorkersKVImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"key": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"namespace_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"value": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+		},
+	}
+}
+
+func resourceCloudflareWorkersKVRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	namespaceID, key := parseId(d)
+
+	value, err := client.ReadWorkersKV(context.Background(), namespaceID, key)
+	if err != nil {
+		return errors.Wrap(err, "error reading workers kv")
+	}
+
+	if value == nil {
+		d.SetId("")
+		return nil
+	}
+
+	d.Set("value", string(value))
+	return nil
+}
+
+func resourceCloudflareWorkersKVCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	namespaceID := d.Get("namespace_id").(string)
+	key := d.Get("key").(string)
+	value := d.Get("value").(string)
+
+	_, err := client.WriteWorkersKV(context.Background(), namespaceID, key, []byte(value))
+	if err != nil {
+		if err != nil {
+			return errors.Wrap(err, "error creating workers kv")
+		}
+	}
+
+	d.SetId(fmt.Sprintf("%s_%s", namespaceID, key))
+
+	log.Printf("[INFO] Cloudflare Workers KV Namespace ID: %s", d.Id())
+
+	return nil
+}
+
+func resourceCloudflareWorkersKVUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	namespaceID := d.Get("namespace_id").(string)
+	key := d.Get("key").(string)
+	value := d.Get("value").(string)
+
+	_, err := client.WriteWorkersKV(context.Background(), namespaceID, key, []byte(value))
+	if err != nil {
+		if err != nil {
+			return errors.Wrap(err, "error creating workers kv")
+		}
+	}
+
+	return nil
+}
+
+func resourceCloudflareWorkersKVDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	namespaceID, key := parseId(d)
+
+	log.Printf("[INFO] Deleting Cloudflare Workers KV with id: %+v", d.Id())
+
+	_, err := client.DeleteWorkersKV(context.Background(), namespaceID, key)
+	if err != nil {
+		return errors.Wrap(err, "error deleting workers kv")
+	}
+
+	return nil
+}
+
+func resourceCloudflareWorkersKVImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	client := meta.(*cloudflare.API)
+	namespaceID, key := parseId(d)
+	value, err := client.ReadWorkersKV(context.Background(), namespaceID, key)
+
+	if err != nil {
+		return nil, fmt.Errorf("error finding workers kv namespace %q: %s", d.Id(), err)
+	}
+
+	d.Set("value", string(value))
+	d.SetId(d.Id())
+
+	return []*schema.ResourceData{d}, nil
+}
+
+func parseId(d *schema.ResourceData) (string, string) {
+	parts := strings.SplitN(d.Id(), "_", 2)
+	return parts[0], parts[1]
+}
--- a/cloudflare/resource_cloudflare_worker_kv.go
+++ b/cloudflare/resource_cloudflare_worker_kv.go
@@ -28,6 +28,7 @@ func resourceCloudflareWorkerKV() *schema.Resource {
 			},
 			"namespace_id": {
 				Type:     schema.TypeString,
+				ForceNew: true,
 				Required: true,
 			},
 			"value": {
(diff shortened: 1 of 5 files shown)


jacobbednarz:
Looks like integration tests are failing for this one:

```
------- Stdout: -------
=== RUN   TestAccCloudflareWorkersKV_Basic
=== PAUSE TestAccCloudflareWorkersKV_Basic
=== CONT  TestAccCloudflareWorkersKV_Basic
--- FAIL: TestAccCloudflareWorkersKV_Basic (0.00s)
    testing.go:654: Step 0 error: 4 problems:
        
        - Invalid multi-line string: On /opt/teamcity-agent/temp/buildTmp/tf-test760257884/main.tf line 3: Quoted strings may not be split over multiple lines. To produce a multi-line string, either use the \n escape to represent a newline character or use the "heredoc" multi-line template syntax.
        - Invalid multi-line string: On /opt/teamcity-agent/temp/buildTmp/tf-test760257884/main.tf line 4: Quoted strings may not be split over multiple lines. To produce a multi-line string, either use the \n escape to represent a newline character or use the "heredoc" multi-line template syntax.
        - Invalid multi-line string: On /opt/teamcity-agent/temp/buildTmp/tf-test760257884/main.tf line 5: Quoted strings may not be split over multiple lines. To produce a multi-line string, either use the \n escape to represent a newline character or use the "heredoc" multi-line template syntax.
        - Unterminated template string: On /opt/teamcity-agent/temp/buildTmp/tf-test760257884/main.tf line 3: No closing marker was found for the string.
FAIL
```

cloudflare/resource_cloudflare_worker_kv.go
```diff
@@ -0,0 +1,127 @@
...
+			},
+		},
+	}
+}
+
+func resourceCloudflareWorkersKVRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	namespaceID, key := parseId(d)
+
+	value, err := client.ReadWorkersKV(context.Background(), namespaceID, key)
```
jacobbednarz: All of the Worker KV methods require setting [`api.AccountID`](https://github.com/cloudflare/cloudflare-go/blob/4a4622ed71d4de360d64a887e3cc3417d41ed299/workers_kv.go#L203) however we're not setting it or documenting it as a required configuration parameter. How does it work at the moment?

cloudflare/resource_cloudflare_worker_kv.go
```diff
@@ -0,0 +1,127 @@
...
+		if err != nil {
+			return errors.Wrap(err, "error creating workers kv")
+		}
+	}
+
+	d.SetId(fmt.Sprintf("%s_%s", namespaceID, key))
+
+	log.Printf("[INFO] Cloudflare Workers KV Namespace ID: %s", d.Id())
+
+	return nil
```
jacobbednarz: All methods that modify a remote resource should also perform a `Read` to ensure the remote state is maintained with the Terraform state. Check out the [filters](https://github.com/terraform-providers/terraform-provider-cloudflare/blob/master/cloudflare/resource_cloudflare_filter.go) resource for prior art. The [Terraform documentation](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#update-state-after-modification) also has a small write up on this.

cloudflare/resource_cloudflare_worker_kv.go
```diff
@@ -0,0 +1,127 @@
...
+
+func resourceCloudflareWorkersKVUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	namespaceID := d.Get("namespace_id").(string)
+	key := d.Get("key").(string)
+	value := d.Get("value").(string)
+
+	_, err := client.WriteWorkersKV(context.Background(), namespaceID, key, []byte(value))
+	if err != nil {
+		if err != nil {
```
jacobbednarz: I don't quite understand what this achieves. Are you able to elaborate a little for me please?

cloudflare/resource_cloudflare_worker_kv.go
```diff
@@ -0,0 +1,127 @@
...
+		return errors.Wrap(err, "error deleting workers kv")
+	}
+
+	return nil
+}
+
+func resourceCloudflareWorkersKVImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	client := meta.(*cloudflare.API)
+	namespaceID, key := parseId(d)
+	value, err := client.ReadWorkersKV(context.Background(), namespaceID, key)
```
jacobbednarz: Generally in `Import` methods, you don't pull from the remote source and instead, you set the required fields that are passed in and then use the `Read` method to synchronise the latest values.

cloudflare/resource_cloudflare_worker_kv_test.go
```diff
@@ -0,0 +1,91 @@
...
+func TestAccCloudflareWorkersKV_Basic(t *testing.T) {
+	t.Parallel()
+	var kvPair cloudflare.WorkersKVPair
+	name := generateRandomResourceName()
+	key := generateRandomResourceName()
+	value := generateRandomResourceName()
+	resourceName := "cloudflare_workers_kv." + name
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
```
jacobbednarz: If we're testing account level functionality, you'll also want to use [`testAccPreCheckAccount`](https://github.com/terraform-providers/terraform-provider-cloudflare/blob/57ef25dc04a6605d8ec8b4088f1a62f507be5f63/cloudflare/provider_test.go#L62-L66) to ensure the account ID is also defined.

cloudflare/resource_cloudflare_worker_kv.go
nrf110: The nested if isn't meant to be here, looks like I fat-fingered something and didn't notice it.

cloudflare/resource_cloudflare_worker_kv_test.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+	}
+
+	return nil
+}
+
+func testAccCheckCloudflareWorkersKV(rName string, key string, value string) string {
+	return testAccCheckCloudflareWorkersKVNamespace(rName) + fmt.Sprintf(`
+resource "cloudflare_workers_kv" "%[1]s" {
+	namespace_id = "${cloudflare_workers_kv_namespace.%[1]s.id}"
```
jacobbednarz: This test isn't going to be 0.11 and 0.12 compatible due to the change in the way variables are referenced. We have a couple of options here:

- Only add test cases for 0.12.x (as that is what we're running in CI); or
- Don't use variables and instead explicitly reference the resources with hardcoded strings.

cloudflare/resource_cloudflare_worker_kv.go
```diff
@@ -0,0 +1,127 @@
...
+	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/pkg/errors"
+)
+
+func resourceCloudflareWorkerKV() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceCloudflareWorkersKVCreate,
+		Read:   resourceCloudflareWorkersKVRead,
+		Update: resourceCloudflareWorkersKVUpdate,
```
jacobbednarz: I see your create + update methods are pretty similar. In some resources where this is also the case, the `Create` and `Update` just share the one method. Up to you as I don't mind either way.

cloudflare/resource_cloudflare_worker_kv.go
```diff
@@ -0,0 +1,127 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"key": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"namespace_id": {
+				Type:     schema.TypeString,
+				Required: true,
```
jacobbednarz: I'm wondering if we want a `ForceNew: true` on `namespace_id` to recreate in the event of it changing 🤔

cloudflare/resource_cloudflare_worker_kv.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, fmt.Errorf("error finding workers kv namespace %q: %s", d.Id(), err)
+	}
+
+	d.Set("value", string(value))
+	d.SetId(d.Id())
+
+	return []*schema.ResourceData{d}, nil
+}
+
+func parseId(d *schema.ResourceData) (string, string) {
```
jacobbednarz: Given that you don't need the whole `*schema.ResourceData` that you're passing in here, could changing the method signature to accept a string and spit out two strings suffice instead?

```go
// function 
func parseId(id string) (string, string) {
        parts := strings.SplitN(id, "_", 2)
	return parts[0], parts[1]
}

// usage
namespaceID, key := parseId(d.Id())
```

website/docs/r/workers_kv.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+  key = "test-key"
+  value = "test value"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `namespace_id` - (Required) The id of the namespace in which you want to create the KV pair
```
jacobbednarz:
```suggestion
* `namespace_id` - (Required) The ID of the Workers KV namespace in which you want to create the KV pair
```

cloudflare/resource_cloudflare_worker_kv.go
nrf110: None of the methods have a parameter for the account ID, it is set on the cloudflare.API client when the provider configures it.  This is identical to how the existing cloudflare_workers_kv_namespace resource works, but I can add it to the documentation if needed.

cloudflare/resource_cloudflare_worker_kv.go
nrf110: Sure

cloudflare/resource_cloudflare_worker_kv.go
nrf110: That seems reasonable.  I can't think of any downside to doing it.

cloudflare/resource_cloudflare_worker_kv.go
nrf110: Is there any harm in calling `d.SetId` in an `Update` method?  That's the main difference between them.  I'm all for saving some code if that isn't an issue.

nrf110: These have been combined into one method.

cloudflare/resource_cloudflare_worker_kv.go
nrf110: Added a note in the docs that setting `account_id` on the provider, or the `CLOUDFLARE_ACCOUNT_ID` environment variable is necessary to use this.  Does that cover the requirement?

nrf110:
@jacobbednarz I think I've addressed all of the feedback so far.  Anything else we need before this can be approved?

patryk:
Thanks @nrf110 for the PR!

PR #598, opened 2020-02-13
Include service to service communication example

jacobbednarz:
This got asked about today and I forgot this works as is so I added an example for it.


jacobbednarz:
Actually, this is wrong. I need to add another param otherwise everyone gets in.

PR #599, opened 2020-02-16
Apply Single Zone Settings on Delete

tootedom:
The singleZoneSettings, for example h2_prioritization, are not reverted in the Delete function.
The result being if you run terratest (or terraform acceptance tests) you end up with the error:
```
Error: errors during apply: error from makeRequest: HTTP status 400: content "{\"success\":false,\"errors\":[{\"code\":1006,\"message\":\"Unrecognized zone setting name\"}],\"messages\":[],\"result\":null}"
```

As h2_prioritization, and the other singleZoneSettings are trying to be applied in the `UpdateZoneSettings` client request, in the delete function.   The cloudflare api isn't allowing the single zone settings of h2_prioritization to be set in UpdateZoneSettings, but requires the to be applied in `ZoneSingleSetting`.  

This has been catered for in all other methods, but the Delete.  The result is it causes issues if we try to run terratest or `TF_ACC=1`, where the zone settings are deleted (reverted) at the end of the tests.

The tests have been updated to read in the singleZoneSettings.


tootelldels:
Fixes:
- https://github.com/terraform-providers/terraform-provider-cloudflare/issues/531

jacobbednarz:
@tootelldels is this ready for review now? I'm unsure what the status is from your side.

tootedom:
Hi @jacobbednarz ,

All ready from my side.  Please review away.

Cheers
/dom

PR #606, opened 2020-02-27
document `cache_level` settings with their meaning

hoexter:
Sadly the Cloudflare WebUI and documentation (e.g. in
https://support.cloudflare.com/hc/en-us/articles/200168256) uses
different terminoly, compared to the settings on the API level.
To spare everyone the confusion and task to look up what is what,
add the short description to the documentation here.


patryk:
Changed a bit, but otherwise good idea. Thanks!

PR #610, opened 2020-03-04
adds validation to only allow valid DNS record types

joel-g:
Current version allows user to submit any string to API as a `Record`'s `Type`.

Anything not from this list will return an error from the API: A, AAAA, CNAME, TXT, SRV, LOC, MX, NS, SPF, CERT, DNSKEY, DS, NAPTR, SMIMEA, SSHFP, TLSA, URI 

This validation will prompt the user before attempting to create resources.

### DIFF
--- a/cloudflare/resource_cloudflare_record.go
+++ b/cloudflare/resource_cloudflare_record.go
@@ -9,6 +9,7 @@ import (
 
 	cloudflare "github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 )
 
 func resourceCloudflareRecord() *schema.Resource {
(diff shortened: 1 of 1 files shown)


rdeleon-zebra:
So, why not `CAA`?

jacobbednarz:
@rdeleon-zebra please see the latest commits on master - CAA was added within hours of identifying this oversight.

rdeleon-zebra:
Yup, saw it here: https://github.com/terraform-providers/terraform-provider-cloudflare/commit/2b1b4fc5a815a16e459bee1b33e84c878e4eac21

Any ETA on this release fix? Curious... we down-graded back to 2.3.0 b/c of this.

jacobbednarz:
No ETA at this point. We aim for semi regular releases but will let 2.3.0 sit for a couple of days before releasing a patch release to address in case anything else pops up. The issue isn’t a major one that requires an immediate fix.

T4cC0re:
> The issue isn’t a major one that requires an immediate fix.

@jacobbednarz It might not be for you, but it is for us. This MR caught us off-guard and caused all of our terraform to collapse since we use this provider's resources in a module, used across our code. Some parts of which want to create CAA records, which of course fail now.

This patch broke something and is thus IMO worth a dedicated patch release asap. To wait `in case anything else pops up` is not what I look for when there already is a fix for an issue, that was introduced as a regression.

We have since pinned this provider to an older version, but I don't think we should have to, for the reasons outlined above.

PR #612, opened 2020-03-05
Support service_token for cloudflare_access_policy

dhoeric:
Integrate it since https://github.com/cloudflare/cloudflare-go/pull/339 is already available!

### DIFF
--- a/cloudflare/resource_cloudflare_access_policy.go
+++ b/cloudflare/resource_cloudflare_access_policy.go
@@ -86,6 +86,13 @@ var policyOptionElement = &schema.Resource{
 				Type: schema.TypeString,
 			},
 		},
+		"service_token": {
+			Type:     schema.TypeList,
+			Optional: true,
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+		},
 		"everyone": {
 			Type:     schema.TypeBool,
 			Optional: true,
@@ -260,6 +267,10 @@ func buildAccessPolicyCondition(options map[string]interface{}) []interface{} {
 					policy = append(policy, cloudflare.AccessPolicyIP{IP: struct {
 						IP string `json:"ip"`
 					}{IP: value.(string)}})
+				case "service_token":
+					policy = append(policy, cloudflare.AccessPolicyServiceToken{ServiceToken: struct {
+						ID string `json:"token_id"`
+					}{ID: value.(string)}})
 				}
 			}
 		}
--- a/cloudflare/resource_cloudflare_access_policy.go
+++ b/cloudflare/resource_cloudflare_access_policy.go
@@ -93,7 +93,7 @@ var policyOptionElement = &schema.Resource{
 				Type: schema.TypeString,
 			},
 		},
-		"any_valid_service_tokens": {
+		"any_valid_service_token": {
 			Type:     schema.TypeBool,
 			Optional: true,
 		},
@@ -256,9 +256,9 @@ func buildAccessPolicyCondition(options map[string]interface{}) []interface{} {
 				log.Printf("[DEBUG] values for everyone %s", values)
 				policy = append(policy, cloudflare.AccessPolicyEveryone{})
 			}
-		} else if accessPolicyType == "any_valid_service_tokens" {
+		} else if accessPolicyType == "any_valid_service_token" {
 			if values == true {
-				log.Printf("[DEBUG] values for any_valid_service_tokens %s", values)
+				log.Printf("[DEBUG] values for any_valid_service_token %s", values)
 				policy = append(policy, cloudflare.AccessPolicyAnyValidServiceToken{})
 			}
 		} else {
--- a/cloudflare/resource_cloudflare_access_policy.go
+++ b/cloudflare/resource_cloudflare_access_policy.go
@@ -93,6 +93,10 @@ var policyOptionElement = &schema.Resource{
 				Type: schema.TypeString,
 			},
 		},
+		"any_valid_service_tokens": {
+			Type:     schema.TypeBool,
+			Optional: true,
+		},
 		"everyone": {
 			Type:     schema.TypeBool,
 			Optional: true,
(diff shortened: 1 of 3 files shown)


cloudflare/resource_cloudflare_access_policy.go
```diff
@@ -86,6 +86,13 @@ var policyOptionElement = &schema.Resource{
 				Type: schema.TypeString,
 			},
 		},
+		"service_token": {
```
jacobbednarz: I'm torn on this as to whether or not to pluralise it because it accepts multiple values. Without pluralising it and iterating over the values, you'll need to provide multiple of these blocks to achieve the same thing.

```hcl
resource "cloudflare_access_policy" "test_policy" {
  application_id = "cb029e245cfdd66dc8d2e570d5dd3322"
  zone_id        = "d41d8cd98f00b204e9800998ecf8427e"
  name           = "staging policy"
  precedence     = "1"
  decision       = "allow"

  include {
    service_token = ["abcd1234"]
    service_token = ["4321dcba"]
  }
}
```

vs

```hcl
resource "cloudflare_access_policy" "test_policy" {
  application_id = "cb029e245cfdd66dc8d2e570d5dd3322"
  zone_id        = "d41d8cd98f00b204e9800998ecf8427e"
  name           = "staging policy"
  precedence     = "1"
  decision       = "allow"

  include {
    service_tokens = [
      "abcd1234",
      "4321dcba"
    ]
  }
}
```

website/docs/r/access_policy.html.markdown
```diff
@@ -77,6 +77,8 @@ conditions which can be applied. The conditions are:
   `email = ["test@example.com"]`
 * `email_domain` - (Optional) A list of email domains. Example:
   `email_domain = ["example.com"]`
+* `service_token` - A list of service token ids. Example:
+  `service_token = ["${cloudflare_access_service_token.demo.id}"]`
```
jacobbednarz: This example should be using the 0.12 syntax for all new documentation.

jacobbednarz: (or a simple string value for the purposes of examples)

dhoeric: How about we create a separate issue for migrating 0.12 syntax?

cloudflare/resource_cloudflare_access_policy.go
dhoeric: The naming convention is just following the syntax as in other params like email / email_domain. For me is clearer to not putting an s as a param.

Furthermore, it works with both 
```
  include {
    service_token = ["abcd1234"]
    service_token = ["4321dcba"]
  }
```

or

```
  include {
    service_token = [
      "abcd1234",
      "4321dcba"
    ]
  }
```

dhoeric:
> I've added some notes inline but we will also need to add integration tests for this new functionality before it can land.
>
> I'm on the fence whether we also want to roll in support for "any valid service token" as opposed to just a specific one given that we are already adding for service tokens to be used here.

Sure, let me add them as well.

website/docs/r/access_policy.html.markdown
jacobbednarz: What would the driver be for not doing it in this given that we’re introducing the change now and we aim to document 0.12 for the examples as that is the latest supported by Terraform?

cloudflare/resource_cloudflare_access_policy.go
jacobbednarz: If it can be multiple, it makes more sense that we pluralise the directive to convey the intention.

cloudflare/resource_cloudflare_access_policy.go
```diff
@@ -86,6 +86,17 @@ var policyOptionElement = &schema.Resource{
...
+			Type:     schema.TypeList,
+			Optional: true,
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+		},
+		"any_valid_service_tokens": {
```
jacobbednarz:
```suggestion
		"any_valid_service_token": {
```

website/docs/r/access_policy.html.markdown
dhoeric: Good to know that we need to proceed for 0.12 syntax but this PR is just about to add a new parameter. 

If we are gonna to switch to 0.12 syntax, how about we do that as one shot and update on all the documents?
That could keep the document consistent and cleaner to read.

The idea is just to keep the PR simple.

cloudflare/resource_cloudflare_access_policy.go
dhoeric: So is that implies to change other params (email, email_domain, ip) to be plural as well? 
I am afraid it could introduce backward incapabilities.

jacobbednarz: 🤔 that raises a good point; I didn’t realise you were referring to this resource. 👍 for following what is present for the others. we can always rename later if needed.

website/docs/r/access_policy.html.markdown
jacobbednarz: I’m not sure I follow here. The documentation should already be 0.12 compatible as a sweep was done following the 2.x release. If there is something that was missed, please drop it below as we will need to fix it.

dhoeric: Ok, updated to 0.12 syntax.

dhoeric:
Added test for access policy on service_token.

```shell
CLOUDFLARE_DOMAIN=xxx CLOUDFLARE_API_KEY=xxx CLOUDFLARE_EMAIL=xxx CLOUDFLARE_ACCOUNT_ID=xxx CLOUDFLARE_ZONE_ID=xxx make testacc TEST=./cloudflare TESTARGS='-run=TestAccAccessPolicy'
TF_ACC=1 go test ./cloudflare -v -run=TestAccAccessPolicy -timeout 120m
=== RUN   TestAccAccessPolicyServiceToken
--- PASS: TestAccAccessPolicyServiceToken (6.32s)
=== RUN   TestAccAccessPolicyAnyServiceToken
--- PASS: TestAccAccessPolicyAnyServiceToken (5.69s)
PASS
ok      github.com/terraform-providers/terraform-provider-cloudflare/cloudflare
```

jacobbednarz:
Thanks for working through this one with me @dhoeric! Greatly appreciated ⭐

PR #630, opened 2020-03-22
Fix firewall rule products support which was missing handling

filipowm:
Resource `cloudflare_firewall_rule` contained action `bypass` and field `products`, however there was no handling implemented what was resulting in error:
```
products must be specified for the 'bypass' action
```

### DIFF
--- a/cloudflare/resource_cloudflare_firewall_rule_test.go
+++ b/cloudflare/resource_cloudflare_firewall_rule_test.go
@@ -84,3 +84,50 @@ func testFirewallRuleConfig(resourceID, zoneID, paused, description, expression,
 		}
 		`, resourceID, zoneID, paused, description, expression, action, priority)
 }
+
+func TestFirewallRuleBypass(t *testing.T) {
+	rnd := generateRandomResourceName()
+	name := "cloudflare_firewall_rule." + rnd
+	zoneID := os.Getenv("CLOUDFLARE_ZONE_ID")
+	domain := os.Getenv("CLOUDFLARE_DOMAIN")
+
+	filterQuoted := `(http.host eq \"` + domain + `\")`
+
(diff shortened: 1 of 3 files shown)


cloudflare/resource_cloudflare_firewall_rule_test.go
```diff
@@ -84,3 +84,49 @@ func testFirewallRuleConfig(resourceID, zoneID, paused, description, expression,
 		}
 		`, resourceID, zoneID, paused, description, expression, action, priority)
 }
+
+func TestFirewallRuleBypass(t *testing.T) {
+	rnd := generateRandomResourceName()
+	name := "cloudflare_firewall_rule." + rnd
+	zoneID := os.Getenv("CLOUDFLARE_ZONE_ID")
+
+	filterQuoted := `(http.host eq \"` + rnd + `\")`
```
jacobbednarz: this rule is invalid but i'm unsure why it's not failing when creating it in the tests (as the rule is parsed for validity on creation).

```suggestion
	filterQuoted := `(http.host eq \"terraform.cfapi.net\")`
```

(or pull it from the env using `CLOUDFLARE_DOMAIN` if you'd rather)

filipowm: I checked on CF and any string works.

But I've changed it to proposed `CLOUDFLARE_DOMAIN`,

jacobbednarz:
Thanking you @filipowm 🍰

thommy101:
Hi, any chance this will be fixed in a new patch release any time soon? We would love to use this feature.

filipowm:
@jacobbednarz Any idea when will be the next release?

patryk:
We usually release ~ once a month, but it's not a fixed schedule. I'll see about releasing today.

PR #647, opened 2020-04-05
Travis CI migration GitHub actions

jacobbednarz:
Moves all CI to GitHub Actions from Travis CI. I'm also going to be looking at moving the acceptance tests here too as TeamCity isn't publicly accessible and I want those automatically run.


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,32 @@
...
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        go: [1.13, 1.14]
+        os: [ubuntu-latest, macos-latest]
```
jacobbednarz, line 11: don't test on windows atm, `gofmt` chokes due to line ending issues on almost every file 😞

jacobbednarz:
For acceptance tests, we need to add the secrets and ensure that we have a way for forks/PRs to trigger them as the [API documentation suggests this isn't possible](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow).

PR #649, opened 2020-04-07
resource/logpush_job: Add support for "dataset" parameter

### LINKED ISSUE #641: cloudflare_logpush_job API should contain "dataset" argument
Here's the documented API for cloudflare_logpush_job: https://www.terraform.io/docs/providers/cloudflare/r/logpush_job.html

It allows the following arguments:
- name
- zone_id 
- destination_conf
- ownership_challenge
- logpull_options 
- enable

Cloudflare's API is documented here: https://api.cloudflare.com/#logpush-jobs-create-logpush-job and also contains the "dataset" parameter. This is important for choosing between HTTP requests, Spectrum events, and soon other logs like Firewall and Browser Insight.

There is additional documentation about Logpush jobs here: https://developers.cloudflare.com/logs/logpush/logpush-configuration-api/understanding-logpush-api/

jacobbednarz:
Updates the `logpush_job` resource to support the "dataset" parameter
which allows different configurations for different types of data.

See https://developers.cloudflare.com/logs/logpush/logpush-configuration-api/understanding-logpush-api/

Closes #641

### DIFF
--- a/cloudflare/resource_cloudflare_logpush_job.go
+++ b/cloudflare/resource_cloudflare_logpush_job.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 )
 
 func resourceCloudflareLogpushJob() *schema.Resource {
@@ -33,6 +34,11 @@ func resourceCloudflareLogpushJob() *schema.Resource {
 			"name": {
 				Type:     schema.TypeString,
 				Optional: true,
+			},
+			"dataset": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice([]string{"http_requests", "spectrum_events"}, false),
 			},
 			"logpull_options": {
 				Type:     schema.TypeString,
@@ -60,6 +66,7 @@ func getJobFromResource(d *schema.ResourceData) cloudflare.LogpushJob {
 		ID:                 id,
 		Enabled:            d.Get("enabled").(bool),
 		Name:               d.Get("name").(string),
+		Dataset:            d.Get("dataset").(string),
 		LogpullOptions:     d.Get("logpull_options").(string),
 		DestinationConf:    d.Get("destination_conf").(string),
 		OwnershipChallenge: d.Get("ownership_challenge").(string),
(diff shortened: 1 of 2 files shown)


cloudflare/resource_cloudflare_logpush_job.go
```diff
@@ -34,6 +35,11 @@ func resourceCloudflareLogpushJob() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"dataset": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice([]string{"http_requests", "spectrum_events"}, false),
```
jacobbednarz, line 41: The [API documentation](https://api.cloudflare.com/#logpush-jobs-properties) has the validation has alphanumeric characters only however only "http_requests" and "spectrum_events" are listed in the developer guide. This leads me to believe that there are more data sets coming but for now we can be specific here and catch some user error before it gets too far into the processing to identify the issue.

jacobbednarz:
I also suspect in the future we may need to update some of the API calls to use the `/datasets/:dataset/fields` endpoints instead of the current blanket one too.

jamesarosen:
Is it possible that this was a breaking change? If it's a new _required_ field, it should go out in 3.0.

We have this resource:
```tf
resource "cloudflare_logpush_job" "everlane-com" {
  enabled             = true
  zone_id             = cloudflare_zone.example-com.id
  name                = "request-logs"
  logpull_options     = "fields=${join(",", local.logpush_fields)}&timestamps=rfc3339&sample=${local.logpush_sample_rate}"
  destination_conf    = "s3://${aws_s3_bucket.example-cloudflare-logs.bucket}/logs/?region=us-east-1"
  ownership_challenge = "this value is irrelevant after initial setup, but required for Terraform"
}
```

On my computer, with `terraform-provider-cloudflare_v2.4.0_x4`, `terraform plan` works fine.

On @antonhalim's computer, with `terraform-provider-cloudflare_v2.8.0_x4`, `terraform plan` results in
```
Error: Missing required argument
  on cloudflare_logs.tf line 84, in resource "cloudflare_logpush_job" "everlane-com":
  84: resource "cloudflare_logpush_job" "everlane-com" {
The argument "dataset" is required, but no definition was found.
```

jacobbednarz:
I don't see this as a backwards-incompatible change; we didn't remove/alter functionality that would prevent systems from working. The change in question was always the implicit value however the API now offers multiple values so an explicit definition is required. We could have opted to make this optional however I feel that would have introduced more bugs as people accidentally creating logpush jobs with the wrong dataset value. The definition itself is not that widely used in Terraform as it is as we query by resource ID instead but that won't always be the casew.

If you cannot add the resource parameter, you can pin to an older version of the provider until you have the capability to upgrade to support this parameter.

PR #651, opened 2020-04-08
Remove workaround for Access not supporting API tokens

jacobbednarz:
The Access APIs now support API tokens so unsetting and working around
that restriction is no longer required.


jacobbednarz:
Service tokens are still requiring API keys and can be revisited at the beginning of May.

PR #656, opened 2020-04-17
Fix cloudflare_access_group.service_token example to mention a list of ids

knapo:
The description says `A list of service token ids`, but the example mentions a single string.


jacobbednarz:
Thanks @knapo 🙏

PR #658, opened 2020-04-18
Adds cloudflare_universal_ssl resource controlling USSL status

cehrig:
This is a new resource `cloudflare_universal_ssl` talking to `zones/:zone_identifier/ssl/universal/settings` via `client.UniversalSSLSettingDetails` and `client.EditUniversalSSLSetting`.

It would have fit into the zone settings override resource but since USSL works on a different API I decided to create a new resource.

Usage
```
resource "cloudflare_universal_ssl" "example" {
    settings {
        status = "on"
    }
}
```

The resource supports importing.

Test Case:
```
cehrig@che-box ~/go/src/github.com/terraform-providers/terraform-provider-cloudflare $ go test -v -timeout 30s github.com/terraform-providers/terraform-provider-cloudflare/cloudflare -run TestAccCloudflareUniversalSSL
=== RUN   TestAccCloudflareUniversalSSL
=== PAUSE TestAccCloudflareUniversalSSL
=== CONT  TestAccCloudflareUniversalSSL
2020/04/18 16:05:18 DestroyEdgeTransformer: pruning unused resource node cloudflare_universal_ssl.veihdokfpq (prepare state)
--- PASS: TestAccCloudflareUniversalSSL (5.57s)
PASS
ok      github.com/terraform-providers/terraform-provider-cloudflare/cloudflare 5.579s
```


jacobbednarz:
Thanks for taking the time to put this one together!

I’m on the fence as to whether this should be it’s own resource or whether it should be an option on `cloudflare_zone_override` as you cannot have multiples of this (as you would a traditional resource). By adding it to `cloudflare_zone_override` you would also be following the same pattern for other zone level settings.

cehrig:
That'd be fine I guess. I'm just concerned as this would be the first setting in zone overrides not using `/zones/:zone_identifier/settings`. Shall we call it `settings.0.universal_ssl` ?

jacobbednarz:
Sounds pretty reasonable. I don’t mind too much about it being the only one as it is portrayed as a zone setting so the association for the end user is still in tact.

cehrig:
@jacobbednarz I would appreciate if you could double check the approach. Since most of the implementation builds on top of `cloudflare.ZoneSetting` I'm wrapping the USSL status in a struct of this type. This way I can keep the update pretty simple.

jacobbednarz:
This seems like a good approach Are you able to add some test coverage?

cehrig:
Sure. I'va added to the existing test

```cehrig@che-box ~/go/src/github.com/terraform-providers/terraform-provider-cloudflare $ go test -v -timeout 120s github.com/terraform-providers/terraform-provider-cloudflare/cloudflare -run TestAccCloudflareZoneSettingsOverride_Full
2020/04/20 00:33:24 DestroyEdgeTransformer: pruning unused resource node cloudflare_zone_settings_override.test (prepare state)
--- PASS: TestAccCloudflareZoneSettingsOverride_Full (35.13s)
PASS
ok      github.com/terraform-providers/terraform-provider-cloudflare/cloudflare 35.140s```

jacobbednarz:
Thanks for this @cehrig 👏

PR #660, opened 2020-04-19
Add missing 'firewall_events' dataset to cloudflare_logpush_job.

prdonahue:
Also realized that we need a logstream boolean too. I'll add that now in a follow up PR.


prdonahue:
Thanks for the review and merge!

Actually looks like we may not need the logstream bool anymore. Did that not make the final cut. for the API, @tlozoot?

PR #666, opened 2020-04-23
logpush_ownership_challenge: Add new resource

jacobbednarz:
This introduces a new resource (`logpush_ownership_challenge`) that
allows the end user to either automate the creation of the Logpush jobs
through requesting the validation, fetching it (using a resource outside
of the Cloudflare provide) or manually obtaining it, and then plug it
into the `logpush_job` ready for creation and management.

`logpush_ownership_challenge` is required as the Logpush service expects
the setup to first prove ownership of the destination by providing a
unique identifier in the creation of the Logpush Job. Without this new
resource it creates a chicken/egg problem which doesn't lend itself to
automation. With this new resource, it allows management in the
following ways:

- Create ownership challenge (this provider).
- Fetch the ownership challenge (outside of Cloudflare provider) by
  either manually looking it up or using another provider to
  automatically get the contents (AWS/GCP/Sumologic).
- Create the `logpush_job` referencing this new challenge proof.

### Screenshots

**Updated `logpush_job` explaining how to use it**

**New resource documentation**


jacobbednarz:
@prdonahue This PR has split the existing single resource into two which should now lend itself to automation. Would you mind eyeballing and providing some feedback on wording and whether this helps alleviate the manual step you were previously needing?

cloudflare/resource_cloudflare_logpush_ownership_challenge_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+)
+
+func TestAccCloudflareLogpushOwnershipChallenge(t *testing.T) {
+	rnd := generateRandomResourceName()
+	name := "cloudflare_logpush_ownership_challenge." + rnd
+	zoneID := os.Getenv("CLOUDFLARE_ZONE_ID")
+
+	destinationConf := `s3://jacob-tf-provider-testing-cf-logpush?region=us-east-1`
```
jacobbednarz, line 16: @patryk Do you have a S3 or GCP bucket you'd prefer here? This bucket has a lifecycle policy on it to ensure automatic cleanup but up to you.

patryk: We can use/create GCP bucket we use for acceptance tests.

jacobbednarz: Thanks, totally forgot about this but updated in 5831f35c.

patryk:
Awesome!

PR #677, opened 2020-05-07
Update import example command to reflect reality

jeffbyrnes:
Update the `terraform import` command to reflect the instructions provided alongside. I.e., to use the zone name, instead of the zone ID.


patryk:
Actually, `Zone ID` here is correct one, but there was a slightly outdated wording in docs anyway. Thanks for pointing out.

Issue #682, opened 2020-05-11
Access policy gsuite, okta providers fail to handle `identity_provider_id` update


wolfmd:
In the process of importing existing Cloudflare access policies to a tfstate, an issue occurred when importing gsuite or okta. When the `identity_provider_id` value is set per https://www.terraform.io/docs/providers/cloudflare/r/access_group.html#gsuite
```
resource "cloudflare_access_policy" "app_foo_policy" {
  application_id = cloudflare_access_application.app.id
  zone_id        = var.zone_id
  name           = "foo policy"
  decision       = "allow"
  precedence     = 2

  include {
    gsuite {
      name                 = "<gsuite-email>"
      identity_provider_id = cloudflare_access_identity_provider.gsuite.id
    }
  }
}
```
the object returned from the api changes from 
```
          {
            "gsuite": {
              "id": "<id>",
              "email": "<email>",
              "connection_id": "<connection_id>"
            }
          },
```
to
```
          {
            "gsuite": {
              "id": "<id>",
              "email": "<email>",
              "identity_provider_id": "<connection_id>"
            }
          },
```
When this is applied to the gsuite or okta include block, I'm no longer able to log in via these connections.

The same issue occurs when trying to add a new access policy that wasn't previously imported from terraform or when tainting and recreating the access policy

Investigation in the UI provides the following error message when trying to update the access policy
![Screen Shot 2020-05-11 at 2 31 37 PM](https://user-images.githubusercontent.com/8314397/81614938-a3753880-9395-11ea-82bd-89fa6333cd07.png)
 which perhaps implies that the gsuite and okta provider don't support the `identity_provider_id`. This may not be a bug with the terraform module, but I would like to report it in case the behavior of renaming the `connection_id` value to `identity_provider_id` is unintended

wolfmd:
Updated since my ability to login via okta previously seemed to be a fluke

wolfmd:
I hacked together a local struct with ConnectionID instead of IdentityProviderID. Running it against my staging zone it seems to clear this issue up. I'll put together a proper PR for this tomorrow assuming my fix seems correct.

analytically:
Also hitting this issue.

Justin-Holmes:
As stated here https://github.com/cloudflare/cloudflare-go/pull/471, the Access team is deprecating `connection_id` in favor of `identity_provider_id`. The UI has been fixed to allow the `identity_provider_id` field and should no longer show the errors!

[...]

PR #684, opened 2020-05-12
Update the Okta and Gsuite access groups to use `connection_id` rather than `identity_provider_id`

wolfmd:
As noted in https://github.com/terraform-providers/terraform-provider-cloudflare/issues/682, the API returns 'connection_id' rather than 'identity_provider_id' for this field

Okta and Gsuite are confirmed to use this schema. Other providers are left as is. This is dependent on this update to the vendored cloudflare-go library https://github.com/cloudflare/cloudflare-go/pull/458

### DIFF
--- a/cloudflare/resource_cloudflare_access_group.go
+++ b/cloudflare/resource_cloudflare_access_group.go
@@ -346,11 +346,11 @@ func BuildAccessGroupCondition(options map[string]interface{}) []interface{} {
 			for _, v := range values.([]interface{}) {
 				gsuiteCfg := v.(map[string]interface{})
 				group = append(group, cloudflare.AccessGroupGSuite{Gsuite: struct {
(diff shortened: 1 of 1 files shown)


analytically:
@jacobbednarz

jacobbednarz:
See the discussion upstream at cloudflare/cloudflare-go#459 and #682. This isn’t going to be actioned until we get answer as the API allows both fields so this change isn’t the issue as it stands.

analytically:
And https://github.com/cloudflare/cloudflare-go/pull/471

jacobbednarz:
Please see the conversation in https://github.com/terraform-providers/terraform-provider-cloudflare/issues/682#issuecomment-639139477. This isn't needed and is working as is so we don't need this PR.

PR #686, opened 2020-05-14
resource: page rule: Add support for cache key fields action

XaF:
Cache Key Fields action was not terraformable so far. This adds
the elements necessary to allow to define the Cache Key Fields
action in a page rule, along with the necessary documentation.

### DIFF
--- a/cloudflare/resource_cloudflare_page_rule.go
+++ b/cloudflare/resource_cloudflare_page_rule.go
@@ -217,6 +217,152 @@ func resourceCloudflarePageRule() *schema.Resource {
 							ValidateFunc: validation.StringInSlice([]string{"bypass", "basic", "simplified", "aggressive", "cache_everything"}, false),
 						},
 
+						"cache_key_fields": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MinItems: 1,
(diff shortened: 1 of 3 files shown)


jacobbednarz:
Thanks for doing this @XaF but it looks like we've already got a PR for this one in #685. I'll have a chat with @patryk which way we want to go with this as while his PR is more verbose, it doesn't group the fields together in a way where value changes would require a big refactoring.

XaF:
Oh well, seems that we've crossed beams!

patryk:
That would be bad!

Well, I don't know what to do from this place. Maybe we could move your simplified transform code to my PR, because I'm not proud of large amount of type casting?

XaF:
I'm good with anything you decide really! We just need that feature as we're terraforming a zone for which we want to configure the cache key fields, and I wrote this when I figured it wasn't supported yet, so as long as the feature becomes available soon I'm good! :)

If you want to take chunks of my code to put in your PR that works for me!

patryk:
I have simplified the transform code, inspired by your PR here and merged. Thanks!

XaF:
Thank you @patryk !

PR #689, opened 2020-05-19
Update docs to denote plan-dependent arguments

lukiffer:
Updates documentation with a note on plan based features and side effects of enabling them when they are within the plan entitlements.


jacobbednarz:
Thanks! I updated the PR description to reflect the actual changes made here.

PR #698, opened 2020-06-03
docs/r/zone_settings_override: remove obsolete page settings attributes

ddseanc:
A more extensive audit is probably required, but I just noticed these while upgrading a module internally.


jacobbednarz:
What makes you say these are deprecated? All three of these are controllable via an override setting.

jacobbednarz:
Closed due to lack of activity.

PR #706, opened 2020-06-11
SDK-95: cache_ttl_by_status support for Page Rules

patryk:
This PR adds support for Cache TTL by status code.

https://support.cloudflare.com/hc/en-us/articles/360043842472-Configuring-cache-TTL-by-status-code

### DIFF
--- a/cloudflare/resource_cloudflare_page_rule_test.go
+++ b/cloudflare/resource_cloudflare_page_rule_test.go
@@ -438,6 +438,27 @@ func TestAccCloudflarePageRuleCacheKeyFields2(t *testing.T) {
 	})
 }
 
+func TestAccCloudflarePageRuleCacheTTLByStatus(t *testing.T) {
+	var pageRule cloudflare.PageRule
+	domain := os.Getenv("CLOUDFLARE_DOMAIN")
+	zoneID := os.Getenv("CLOUDFLARE_ZONE_ID")
+	target := fmt.Sprintf("test-cache-ttl-by-status.%s", domain)
+
+	resource.Test(t, resource.TestCase{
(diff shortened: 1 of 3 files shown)


jacobbednarz:
Closes #711

cloudflare/resource_cloudflare_page_rule_test.go
```diff
@@ -438,6 +438,27 @@ func TestAccCloudflarePageRuleCacheKeyFields2(t *testing.T) {
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckCloudflarePageRuleDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckCloudflarePageRuleConfigCacheTTLByStatus(zoneID, target),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCloudflarePageRuleExists("cloudflare_page_rule.test", &pageRule),
```
jacobbednarz, line 455: Instead of hardcoding this to "test" which may collide with other resources, could we use a random string and pass it into the configuration instead?

Similar to https://github.com/terraform-providers/terraform-provider-cloudflare/blob/639c5c8404f2cc2ef12fe44d9510156fba1036dc/cloudflare/resource_cloudflare_access_group_test.go#L44-L68

jacobbednarz: fixed this in 04fa7ca8

jacobbednarz:
🍻 thanks @patryk!

PR #710, opened 2020-06-14
Add support for worker text bindings

blakeembrey:
Closes https://github.com/terraform-providers/terraform-provider-cloudflare/issues/615, closes https://github.com/terraform-providers/terraform-provider-cloudflare/issues/164. Not entirely sure on the terraform idioms so feel free to correct anything I'm doing incorrectly.

I wanted to use separate keys in `binding` because they have different requirements for things like `sensitive` and possibly values in the future. The key in the schema determines the "type" of the binding, e.g. `secret_text`, `plain_text` or `kv_namespace_id`. Roughly follows https://github.com/terraform-providers/terraform-provider-cloudflare/issues/615#issuecomment-598801105 but didn't see the need for specifying `type` since every binding is different enough so far they can be differentiated easily by the key set.

### DIFF
--- a/cloudflare/resource_cloudflare_worker_script.go
+++ b/cloudflare/resource_cloudflare_worker_script.go
@@ -10,24 +10,47 @@ import (
 	"github.com/pkg/errors"
 )
 
-var bindingResource = &schema.Resource{
+const concealedString = "**********************************"
+
+var kvNamespaceBindingResource = &schema.Resource{
 	Schema: map[string]*schema.Schema{
 		"name": {
 			Type:     schema.TypeString,
 			Required: true,
 		},
-		"kv_namespace_id": {
+		"namespace_id": {
 			Type:     schema.TypeString,
-			Optional: true,
+			Required: true,
 		},
-		"plain_text": {
+	},
+}
+
+var plainTextBindingResource = &schema.Resource{
+	Schema: map[string]*schema.Schema{
+		"name": {
 			Type:     schema.TypeString,
-			Optional: true,
+			Required: true,
 		},
-		"secret_text": {
+		"text": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+	},
+}
+
+var secretTextBindingResource = &schema.Resource{
+	Schema: map[string]*schema.Schema{
+		"name": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+		"text": {
 			Type:      schema.TypeString,
+			Required:  true,
 			Sensitive: true,
-			Optional:  true,
+			StateFunc: func(val interface{}) string {
+				return concealedString
+			},
 		},
 	},
 }
@@ -52,27 +75,20 @@ func resourceCloudflareWorkerScript() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
-			"binding": {
+			"plain_text_binding": {
 				Type:     schema.TypeSet,
 				Optional: true,
-				Elem:     bindingResource,
+				Elem:     plainTextBindingResource,
+			},
+			"secret_text_binding": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     secretTextBindingResource,
 			},
 			"kv_namespace_binding": {
 				Type:     schema.TypeSet,
 				Optional: true,
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"name": {
-							Type:     schema.TypeString,
-							Required: true,
-						},
-						"namespace_id": {
-							Type:     schema.TypeString,
-							Required: true,
-						},
-					},
-				},
-				Deprecated: "`kv_namespace_binding` has been deprecated in favour of using `binding.kv_namespace_id` instead.",
+				Elem:     kvNamespaceBindingResource,
 			},
 		},
 	}
@@ -116,16 +132,6 @@ func getWorkerScriptBindings(scriptName string, client *cloudflare.API) (ScriptB
 }
 
 func parseWorkerBindings(d *schema.ResourceData, bindings ScriptBindings) error {
-	for _, rawData := range d.Get("binding").(*schema.Set).List() {
-		data := rawData.(map[string]interface{})
-		name := data["name"].(string)
-		binding := parseWorkerBinding(data)
-		if binding == nil {
-			return fmt.Errorf("binding must specify `kv_namespace_id`, `plain_text` or `secret_text`: %s", name)
-		}
-		bindings[name] = binding
-	}
-
 	for _, rawData := range d.Get("kv_namespace_binding").(*schema.Set).List() {
 		data := rawData.(map[string]interface{})
 		bindings[data["name"].(string)] = cloudflare.WorkerKvNamespaceBinding{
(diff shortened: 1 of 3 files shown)


cloudflare/resource_cloudflare_worker_script.go
```diff
@@ -27,11 +27,36 @@ func resourceCloudflareWorkerScript() *schema.Resource {
...
+						},
+						"plain_text": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"secret_text": {
+							Type:      schema.TypeString,
+							Sensitive: true,
+							Optional:  true,
+						},
```
blakeembrey, line 55: I wanted to use `ExactlyOneOf` schema check here but couldn't get it to work. Tried to search for examples in other providers like AWS but also see they only use it for the top-level schema and not within a `TypeSet`. Let me know if there's a simple way to achieve this validation, otherwise we'll be relying on documentation.

jacobbednarz: `Set`s are pretty annoying to work with for these cases. I think we start with documentation and see how we go before spending too much time on solving it technically.

jacobbednarz:
I've kicked off an integration test run for this one and will report the findings back shortly.

jacobbednarz:
Looks like we've got a new failure.

```
------- Stdout: -------
--- FAIL: TestAccCloudflareWorkerScript_MultiScriptEnt (7.54s)
testing.go:684: Step 2 error: errors during apply:
Error: error updating worker script: NamespaceID for binding "PLAIN_TEXT" cannot be empty
on /opt/teamcity-agent/temp/buildTmp/tf-test036547402/main.tf line 6:
(source code not available)
FAIL
```

blakeembrey:
@jacobbednarz Thanks, I think I might be misunderstanding the map provided by terraform from the `TypeSet` list method. I'll take a look now.

blakeembrey:
@jacobbednarz can you try it again? If it doesn’t pass this time I’ll put some more work into figuring out how to it myself properly

blakeembrey:
I have things working locally but appear to have run into https://github.com/hashicorp/terraform/issues/21901 on `binding`.

blakeembrey:
@jacobbednarz I've resolved the secret text issue by using the text value from the existing resource during read. This seems to be how other providers work, i.e. AWS resources with passwords. The bindings only cause a change in two cases now:

1. When the value of the binding changes (from the read or previous state in the case of `secret_text`)
2. When a secret text binding has been removed from terraform but not Cloudflare

For 2, I didn't see the APIs available to handle it within the cloudflare package or documented on their site. Having the diff show up each time is probably enough to prompt that it's not being deleted properly, but happy to document this behavior and how to delete it from the cloudflare UI.

jacobbednarz:
@blakeembrey That seems like a reasonable fix. We do have other cases of write-once-never-seen-again APIs in Cloudflare however their values are just masked so we write the masked value to state instead of pulling it from the schema.

Your second point is also fine for now 👍 If it becomes a bigger issue than we can poke Cloudflare to improve the UX somehow.

blakeembrey:
Sweet, I ported all my local code to use this and it's been working great so far. By masked value do you mean an empty string or a value like `xxx`?

jacobbednarz:
Masked value of asterisks

https://github.com/terraform-providers/terraform-provider-cloudflare/blob/dc74b3b0d164d212249d2f4678bf0d1e0533ce62/cloudflare/resource_cloudflare_access_identity_provider.go#L13

https://github.com/terraform-providers/terraform-provider-cloudflare/blob/dc74b3b0d164d212249d2f4678bf0d1e0533ce62/cloudflare/resource_cloudflare_access_identity_provider.go#L76-L84

I'll kick off an integration test run and 🤞 we're ready to roll!

jacobbednarz:
Integration is looking good! I'm going to leave this for today and make sure that no one else has any thoughts before we land this one.

Thanks for the PR so far 🙇

blakeembrey:
@jacobbednarz Any opinion on using the singular `binding` versus many bindings? For example:

```
text_binding {
  name = "TEXT"
  value = "TEXT"
}

secret_binding {
  name = "SECRET"
  value = "SECRET"
}
```

This would make the change backward compatible (no need to deprecate the existing binding) and makes something like the concealed `StateFunc` easier to implement.

blakeembrey:
@jacobbednarz Done, though I couldn't get the concealed text behavior to easily work within the set so I fell back on the existing logic of looking up the previous value to diff.

cloudflare/resource_cloudflare_worker_script_test.go
```diff
@@ -74,24 +74,29 @@ resource "cloudflare_worker_script" "%[1]s" {
...
-	namespace_id = cloudflare_workers_kv_namespace.%[1]s.id
+    name         = "KV_NAMESPACE"
+    namespace_id = cloudflare_workers_kv_namespace.%[1]s.id
   }
 
-  kv_namespace_binding {
-	name = "%[1]s-copy"
-	namespace_id = cloudflare_workers_kv_namespace.%[1]s.id
+  plain_text_binding {
+    name = "PLAIN_TEXT"
```
jacobbednarz, line 93: Just to confirm here,  PLAIN_TEXT is just the arbitrary name of the key, right? If that’s the case, can we change the example to something more example like? That way we know it’s definitely an arbitrary key.

blakeembrey: Is this comment about documentation or the test case? I updated the documentation, I can also change the test cases if it helps.

jacobbednarz: Both. Ideally I’d like to remove any question whether this is a special value (such as a constant) or whether the field can be something arbitrary

blakeembrey: Do you want to specify the behavior? I don't want to make it any more confusing, it seemed clearest to me because these are the keys checked in the test suite so if one is missing you know which it is easily. If this is about the casing, I can change it, but that's the standard style I've seen and used in workers because it binds to a constant in JavaScript.

blakeembrey: Can you comment on whether the documentation change is clear enough?

website/docs/r/worker_script.html.markdown
```diff
@@ -21,11 +21,21 @@ resource "cloudflare_workers_kv_namespace" "my_namespace" {
 resource "cloudflare_worker_script" "my_script" {
   name = "script_1"
   content = file("script.js")
-  
+
   kv_namespace_binding {
-    name = "my_binding"
+    name = "DB"
```
jacobbednarz, line 26:
```suggestion
    name = "MY_EXAMPLE_KV_NAMESPACE_KEY"
```

jacobbednarz, line 26:
```suggestion
    name = "MY_EXAMPLE_KV_NAMESPACE_KEY"
```

website/docs/r/worker_script.html.markdown
```diff
@@ -21,11 +21,21 @@ resource "cloudflare_workers_kv_namespace" "my_namespace" {
...
     namespace_id = cloudflare_workers_kv_namespace.my_namespace.id
   }
+
+  plain_text_binding {
+    name = "ENVIRONMENT"
+    text = "staging"
```
jacobbednarz, line 32:
```suggestion
    name = "MY_EXAMPLE_PLAINTEXT_KEY"
    text = "foobar"
```

website/docs/r/worker_script.html.markdown
```diff
@@ -21,11 +21,21 @@ resource "cloudflare_workers_kv_namespace" "my_namespace" {
...
+  }
+
+  secret_text_binding {
+    name = "SENTRY_DSN"
+    text = var.sentry_dsn
```
jacobbednarz, line 37:
```suggestion
    name = "MY_EXAMPLE_SECRET_KEY"
    text = var.secret_foo_value
```

cloudflare/resource_cloudflare_worker_script_test.go
jacobbednarz: Tests I don't mind but I've just proposed (hopefully) very obvious example values to not confuse the documentation readers.

blakeembrey: Ok, I've updated the documentation with your suggestions minus the `_KEY` suffix. I removed it because it's not really a key to something, but a direct binding in the script as a variable. I can add it back if you think it's still helpful though.

blakeembrey:
Awesome! Really excited to see it released 😄

PR #714, opened 2020-06-17
adds support for CustomSSLOptions GeoRestrictions to be nil

trjstewart:
This is a minor change to account for underlying API changes in `cloudflare-go`.

Based on https://github.com/cloudflare/cloudflare-go/pull/480 the `GeoRestrictions` property for `CustomSSLOptions` can now possibly be `nil`. This simply checks if that's the case before flattening the `Label` property.

Without this change; when `GeoRestrictions` is `nil` it results in a segfault as `flattenCustomSSLOptions` is trying to access a property of an undefined property.


jacobbednarz:
Change looks good; do you mind adding a test case exercising this new code path?

trjstewart:
Hey @jacobbednarz. My colleague has added a couple of tests to check both code paths here.

jacobbednarz:
@trjstewart @bas-papegaaij Thanks for adding those however they are closer to what I'd expect to see in `cloudflare-go` in the form of unit tests. Within the Terraform provider, we pretty heavily rely on integration tests such as [this one](https://github.com/terraform-providers/terraform-provider-cloudflare/blob/37e4ff26ba3f4c5b22dba73a8a459035465cba2c/cloudflare/resource_cloudflare_custom_ssl_test.go#L14-L45) to confirm the Terraform state is correct.

The concept of your tests are spot on however instead we should be looking to assert against the configuration and state generated from the resources. Here is a non-working example but demonstrates what I'd expect.

```go
func TestAccCloudflareCustomSSLWithEmptyGeoRestrictions(t *testing.T) {
  t.Parallel()
  var customSSL cloudflare.ZoneCustomSSL
  zoneID := os.Getenv("CLOUDFLARE_ZONE_ID")
  resourceName := "cloudflare_custom_ssl." + rnd
  resource.Test(t, resource.TestCase{
    PreCheck:     func() { testAccPreCheck(t) },
    Providers:    testAccProviders,
    CheckDestroy: testAccCheckCloudflareCustomSSLDestroy,
    Steps: []resource.TestStep{
      {
        // Will need to change this to match the setup of the geo restrictions you provide
        Config: testAccCheckCloudflareCustomSSLCertBasic(zoneID, rnd),
          testAccCheckCloudflareCustomSSLExists(resourceName, &customSSL),
          // .. snip
          resource.TestCheckNoResourceAttr(resourceName, "custom_ssl_options.geo_restrictions"),
        ),
      },
    },
  })
}
```

Is this something you feel you could take a pass at? If not, I'm happy to add them to show you what I mean.

trjstewart:
@jacobbednarz if you have the capacity to add them that would be greatly appreciated. If not, let us know and one of us can take a crack at it. Thanks!

jacobbednarz:
I've got a couple of other tasks in front of this one so depending on how important this is to you, I'd recommend you take a crack at it otherwise I can look at it when I get to it.

jacobbednarz:
Thanks for the work here @trjstewart and @bas-papegaaij. Feel free to check out 7956ce0 if you're curious on the test case that ended up confirming this behaviour.

PR #716, opened 2020-06-21
Restrict zone sweeper to account ID

### LINKED ISSUE #648: Restrict test sweepers to account ID provided
While porting the integration test suite I managed to delete our non-critical production domains as the [`ListZones`](https://github.com/terraform-providers/terraform-provider-cloudflare/blob/master/cloudflare/data_source_zones_test.go#L27) method used by the [`cloudflare_zone_sweeper`](https://github.com/terraform-providers/terraform-provider-cloudflare/blob/master/cloudflare/data_source_zones_test.go#L21-L52) doesn't restrict the zones returned by the `CLOUDFLARE_ACCOUNT_ID` provided.

The sweeper should only use the `CLOUDFLARE_ACCOUNT_ID` like other resources and fail if it isn't provided. I suspect we will need to update `cloudflare/cloudflare-go` to support this properly.

jacobbednarz:
Updates the test zone sweeper to use the `withZoneFilters` method and
restrict the API fetching of zones to a single account ID. This
eliminates the risk of accidentally running this in a production account
should you provide the account ID in the CLOUDFLARE_ACCOUNT_ID
environment variable (which you must now).

Fixes #648

### DIFF
--- a/cloudflare/data_source_zones_test.go
+++ b/cloudflare/data_source_zones_test.go
@@ -1,12 +1,16 @@
 package cloudflare
 
 import (
+	"context"
 	"fmt"
 	"log"
+	"os"
 	"strconv"
 	"testing"
 
+	"github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/pkg/errors"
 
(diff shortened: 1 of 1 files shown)


jacobbednarz:
Confirmed existing functionality is still working and cannot delete outside of the provided account ID.

PR #724, opened 2020-06-26
Update access resources to use account_id over zone_id

Justin-Holmes:
This PR changes Access resources to use `account_id` over `zone_id` since the `/zones` routes have been deprecated in favor of `/accounts`. I'll update the `cloudflare-go` version once https://github.com/cloudflare/cloudflare-go/pull/486 is merged.

### DIFF
--- a/cloudflare/utils.go
+++ b/cloudflare/utils.go
@@ -8,6 +8,7 @@ import (
 	"regexp"
 	"strings"
 
+	cloudflare "github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	errors "github.com/pkg/errors"
 )
@@ -136,4 +137,17 @@ func stringFromBool(status bool) string {
 		return "on"
 	}
 	return "off"
+}
+
+func getAccountIDFromZoneID(d *schema.ResourceData, client *cloudflare.API) (string, error) {
+	accountID := d.Get("account_id").(string)
+	if accountID == "" {
+		zoneID := d.Get("zone_id").(string)
+		zone, err := client.ZoneDetails(zoneID)
+		if err != nil {
+			return "", fmt.Errorf("error retrieving zone for zone_id %q: %s", zoneID, err)
+		}
+		accountID = zone.Account.ID
+	}
+	return accountID, nil
 }
--- a/cloudflare/utils.go
+++ b/cloudflare/utils.go
@@ -149,5 +149,7 @@ func getAccountIDFromZoneID(d *schema.ResourceData, client *cloudflare.API) (str
 		}
 		accountID = zone.Account.ID
 	}
+
+	d.Set("account_id", accountID)
 	return accountID, nil
 }
--- a/cloudflare/resource_cloudflare_access_application.go
+++ b/cloudflare/resource_cloudflare_access_application.go
@@ -24,11 +24,13 @@ func resourceCloudflareAccessApplication() *schema.Resource {
 			"account_id": {
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
 			},
 			"zone_id": {
 				Deprecated: "This field will be removed in version 3 and replaced with the account_id field.",
 				Type:       schema.TypeString,
 				Optional:   true,
+				Computed:   true,
 			},
 			"aud": {
 				Type:     schema.TypeString,
@@ -120,6 +122,7 @@ func resourceCloudflareAccessApplicationCreate(d *schema.ResourceData, meta inte
 	if err != nil {
 		return err
 	}
+
 	allowedIDPList := expandInterfaceToStringList(d.Get("allowed_idps"))
 
 	newAccessApplication := cloudflare.AccessApplication{
@@ -146,6 +149,7 @@ func resourceCloudflareAccessApplicationCreate(d *schema.ResourceData, meta inte
 	}
 
 	d.SetId(accessApplication.ID)
+	d.Set("account_id", accountID)
 
 	return resourceCloudflareAccessApplicationRead(d, meta)
 }
--- a/cloudflare/resource_cloudflare_access_application.go
+++ b/cloudflare/resource_cloudflare_access_application.go
@@ -21,9 +21,14 @@ func resourceCloudflareAccessApplication() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"zone_id": {
+			"account_id": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
+			},
+			"zone_id": {
+				Deprecated: "This field will be removed in version 3 and replaced with the account_id field.",
+				Type:       schema.TypeString,
+				Optional:   true,
 			},
 			"aud": {
 				Type:     schema.TypeString,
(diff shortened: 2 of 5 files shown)


patryk:
This is a breaking change. We cannot merge it now as-is, unless the underlying API is removed. Good candidate for v3.0, though.

jacobbednarz:
I think we can still do this without incurring a breaking change right now but still migrate to the accounts API. 

My thinking

- add `account_id` to schema
- add deprecation notice on `zone_id`
- if the `zone_id` value is provided, lookup the associated `account_id` from it and make the request
- if `account_id` is provided, pass it straight through to the request

Then, the breaking change of removing `zone_id` can come in 3.x. This will incur an additional zone lookup from the API but it is a smoother user experience _and_ we get off the old endpoint. 

Thoughts?

patryk:
Sounds good. I'm OK with it.

Justin-Holmes:
I added the `zone_id` back in the schema with the deprecation warning and the `account_id` lookup. Let me know if you have any feedback!

cloudflare/utils.go
```diff
@@ -137,3 +138,17 @@ func stringFromBool(status bool) string {
 	}
 	return "off"
 }
+
+func getAccountID(d *schema.ResourceData, client *cloudflare.API) (string, error) {
+	accountID := d.Get("account_id").(string)
+	if accountID == "" {
+		zoneID := d.Get("zone_id").(string)
+		var err error
```
jacobbednarz, line 146: I don't think we need to instantiate `err` here as we're instantiating and assigning in the next line for the `zoneDetails` call.

cloudflare/utils.go
```diff
@@ -137,3 +138,17 @@ func stringFromBool(status bool) string {
 	}
 	return "off"
 }
+
+func getAccountID(d *schema.ResourceData, client *cloudflare.API) (string, error) {
```
jacobbednarz, line 142: Bit of a nitpick but would this make more sense as `getAccountIDFromZoneID`? Explicitly calling out the source and expected returned data?

jacobbednarz:
This looks pretty great! Thank you! Along with the inline comments, I feel it would be beneficial to add a test case to each that exercises the `getAccountID` path where only the `zone_id` is set in the schema but assert we get the correct account ID, etc. This way, we are also testing the upgraded path condition when people already have the `zone_id` set and want to swap to the `account_id` value later. May help with identitying any regressions later?

cloudflare/utils.go
Justin-Holmes: Good catch, this was vestige from before it was pulled into a separate function.

cloudflare/utils.go
Justin-Holmes: I agree, `getAccountIDFromZoneID` is a better name for this.

Justin-Holmes:
@jacobbednarz I added regression tests to the Access application and policy resources. I'm explicitly testing create and update actions since (from my understanding) the read and delete actions get tested implicitly -- let me know if my assumption on this is wrong!

jacobbednarz:
Thanks for the updates @Justin-Holmes 🌟 I've just kicked off the integration test runs and once we get a green run on there, we're good to merge this in.

jacobbednarz:
The CI tests are throwing some weird permission errors which I initially suspected as being caused by API token/API key differences however that doesn’t seem to be the case. 

I’m chatting with @patryk who is chasing this up internally to get an idea on the cause.

jacobbednarz:
I just found out why this is failing on permission errors. The upstream change to use the account endpoints hasn't been merged yet and the account ID is being used as a zone ID.



I'm not sure why I thought this should work however to protect this PR from future Jacob getting excited to merge, I'll explicitly say it now. We're still pending cloudflare/cloudflare-go#486 for the CI suite to be green on this one.

jacobbednarz:
@Justin-Holmes you should be good to pull in the latest master, resolve these conflicts and we can get this one shipped! v0.13.1 of `cloudflare-go` has been cut.

Justin-Holmes:
@jacobbednarz This should be good to go now! We may want to fast-track this PR since Access resources will be broken until this gets merged in. Thanks again for the help.

jacobbednarz:
Integration suite is ✅

```
=== RUN   TestAccCloudflareAccessApplicationBasic
--- PASS: TestAccCloudflareAccessApplicationBasic (5.34s)
=== RUN   TestAccCloudflareAccessApplicationWithCORS
--- PASS: TestAccCloudflareAccessApplicationWithCORS (5.43s)
=== RUN   TestAccCloudflareAccessApplicationWithAutoRedirectToIdentity
--- PASS: TestAccCloudflareAccessApplicationWithAutoRedirectToIdentity (5.41s)
=== RUN   TestAccCloudflareAccessApplicationWithADefinedIdps
--- PASS: TestAccCloudflareAccessApplicationWithADefinedIdps (8.09s)
=== RUN   TestAccCloudflareAccessApplicationWithZoneID
--- PASS: TestAccCloudflareAccessApplicationWithZoneID (8.40s)
=== RUN   TestAccCloudflareAccessGroupConfig_Basic
--- PASS: TestAccCloudflareAccessGroupConfig_Basic (4.91s)
=== RUN   TestAccCloudflareAccessGroup_Exclude
--- PASS: TestAccCloudflareAccessGroup_Exclude (5.54s)
=== RUN   TestAccCloudflareAccessGroup_Require
--- PASS: TestAccCloudflareAccessGroup_Require (6.01s)
=== RUN   TestAccCloudflareAccessGroup_FullConfig
--- PASS: TestAccCloudflareAccessGroup_FullConfig (4.85s)
=== RUN   TestAccCloudflareAccessGroup_Updated
--- PASS: TestAccCloudflareAccessGroup_Updated (8.04s)
=== RUN   TestAccCloudflareAccessGroup_CreateAfterManualDestroy
--- PASS: TestAccCloudflareAccessGroup_CreateAfterManualDestroy (8.35s)
=== RUN   TestAccCloudflareAccessIdentityProviderOneTimePin
=== PAUSE TestAccCloudflareAccessIdentityProviderOneTimePin
=== RUN   TestAccCloudflareAccessIdentityProviderOAuth
=== PAUSE TestAccCloudflareAccessIdentityProviderOAuth
=== RUN   TestAccCloudflareAccessIdentityProviderOAuthWithUpdate
=== PAUSE TestAccCloudflareAccessIdentityProviderOAuthWithUpdate
=== RUN   TestAccCloudflareAccessIdentityProviderSAML
=== PAUSE TestAccCloudflareAccessIdentityProviderSAML
=== RUN   TestAccAccessPolicyServiceToken
--- PASS: TestAccAccessPolicyServiceToken (8.80s)
=== RUN   TestAccAccessPolicyAnyServiceToken
--- PASS: TestAccAccessPolicyAnyServiceToken (8.62s)
=== RUN   TestAccAccessPolicyWithZoneID
--- PASS: TestAccAccessPolicyWithZoneID (13.08s)
=== RUN   TestAccAccessRuleASN
--- PASS: TestAccAccessRuleASN (2.45s)
=== RUN   TestAccAccessServiceTokenCreate
--- PASS: TestAccAccessServiceTokenCreate (2.46s)
=== RUN   TestAccAccessServiceTokenUpdate
--- PASS: TestAccAccessServiceTokenUpdate (4.94s)
=== RUN   TestAccAccessServiceTokenDelete
--- PASS: TestAccAccessServiceTokenDelete (2.80s)
=== CONT  TestAccCloudflareAccessIdentityProviderOneTimePin
=== CONT  TestAccCloudflareAccessIdentityProviderSAML
=== CONT  TestAccCloudflareAccessIdentityProviderOAuthWithUpdate
=== CONT  TestAccCloudflareAccessIdentityProviderOAuth
--- PASS: TestAccCloudflareAccessIdentityProviderOneTimePin (2.61s)
--- PASS: TestAccCloudflareAccessIdentityProviderSAML (2.65s)
--- PASS: TestAccCloudflareAccessIdentityProviderOAuth (2.71s)
--- PASS: TestAccCloudflareAccessIdentityProviderOAuthWithUpdate (4.60s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/cloudflare	118.155s
```

Issue #721, opened 2020-06-24
Add support for `auto_redirect_to_identity` feature on Cloudflare Access


xens:
Goal of this issue is to track the implementation of this missing feature `auto_redirect_to_identity` inside the Cloudflare Terraform provider - Access Application resource.

For now this feature is not yet supported inside the cloudflare-go library (WIP -> cloudflare/cloudflare-go#483). When it'll be done we'll be able to work on the Cloudflare Terraform provider implementation.

Ref:
* https://api.cloudflare.com/#access-applications-properties

xens:
@jacobbednarz I saw that https://github.com/cloudflare/cloudflare-go/issues/483 was merged lately. I'm not super familiar with the whole process, I guess that I should be able to pull the master branch of cloudflare-go and start working on it locally right ?

jacobbednarz:
I would wait until a release is cut and it is automatically pulled into this repository. You can start locally if you want but we'll still be blocked until it lands in a release.

xens:
Ok I'll wait then and check from time to time the new cloudflare-go releases.

jacobbednarz:
Closed by #730

PR #725, opened 2020-07-01
resource/access_application: Add CORS support

jacobbednarz:
Updates `cloudflare_access_application` to define the CORS settings
associated with it.

Depends on cloudflare/cloudflare-go#492 to allow
`AccessApplicationCorsHeaders` to be conditionally sent within the
request payload.

Closes #473

PR #730, opened 2020-07-09
Access: add support for auto_redirect_to_identity

xens:
Goal of this commit is to add support for the `auto_redirect_to_identity` feature of Cloudflare Access applications. #721

### DIFF
--- a/cloudflare/resource_cloudflare_access_application.go
+++ b/cloudflare/resource_cloudflare_access_application.go
@@ -90,6 +90,11 @@ func resourceCloudflareAccessApplication() *schema.Resource {
 							Optional:     true,
 							ValidateFunc: validation.IntBetween(-1, 86400),
 						},
+						"auto_redirect_to_identity": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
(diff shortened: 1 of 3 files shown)


jacobbednarz:
This change looks good! We will need to update the associated tests though to ensure we're taking this into account for the existing tests.

xens:
Hi @jacobbednarz yes for sure, believe it or not I wasn't able to find them earlier... I'm wondering if I'm missing something but I couldn't find any unittests related to ```resource_cloudflare_access_application.go``` .

jacobbednarz:
Doesn't look like we have a test file for it in `master` just yet. #725 introduces it for CORS changes so you're welcome to use that as a starting point and we can resolve the conflicts from there once either one lands if you'd like?

xens:
Yes make sense, I added the tests and updated the doc at the same time, let me know.

jacobbednarz:
You should be good to resolve the conflicts now and we can get this re-reviewed.

xens:
rebased + fixed conflicts.

jacobbednarz:
Thanking you @xens! 🌟

PR #733, opened 2020-07-13
resource/waf_package: Only modify anomaly resources

### LINKED ISSUE #723: Bug creating/updating non-anomaly WAF packages
Hi there,

I'm currently having issues creating/updating/managing WAF Packages through Terraform, in particular, the **Cloudflare Managed Ruleset**. This is blocking me from fully managing my organisation's Cloudflare resources with Terraform.

### Terraform Version
```
Terraform v0.12.27
+ provider.cloudflare v2.7.0
```

### Affected Resource(s)
`cloudflare_waf_package`

```hcl
resource "cloudflare_waf_package" "cf_managed_ruleset" {
    zone_id = cloudflare_zone.my_zone.id
    package_id = "1e334934fd7ae32ad***************"
}
```

#### When importing
I imported the WAF package and it worked fine.

#### When running apply

```
cloudflare_waf_package.cf_managed_ruleset: Modifying... [id=1e334934fd7ae32ad***************]

Error: error from makeRequest: HTTP status 400: content "{\"success\":false,\"errors\":[{\"code\":1027,\"message\":\"Only anomaly WAF packages can have their package attributes updated\"}],\"messages\":[],\"result\":null}"

  on main.tf line 182, in resource "cloudflare_waf_package" "cf_managed_ruleset":
 182: resource "cloudflare_waf_package" "cf_managed_ruleset" {
```

It fails with `Only anomaly WAF packages can have their package attributes updated` even though I'm not updating anything, just trying to check my configuration files are all good after the import.

### Expected Behavior
No errors.

### Steps to Reproduce
You'll need a Cloudflare account with a zone and the Cloudflare managed ruleset activated on your zone.
Then try to import/apply the resource above

### Important Factoids
I used the code in the [cloudflare_waf_packages data source documentation page](https://www.terraform.io/docs/providers/cloudflare/d/waf_packages.html) to fetch the WAF Packages I had in my account. It returned :
```

data "cloudflare_waf_packages" "test" {
    id       = "WAFPackages 2020-06-26 15:10:58.438902 +0000 UTC"
    packages = [
        {
            action_mode    = ""
            description    = ""
            detection_mode = "traditional"
            id             = "1e334934fd7ae32ad***************"
            name           = "CloudFlare"
            sensitivity    = ""
        },
    ]
    zone_id  = "**REDACTED**"

    filter {
        detection_mode = "traditional"
    }
}
```
If I set `action_mode = ""` and `sensitivity = ""` in the `cloudflare_waf_package` resource, apply fails because empty strings are not valid values for those keys.

Just in case, I'm also managing the OWASP Core ruleset with Terraform and that's working great so far.

### Any help?

Any help on this issue is greatly appreciated! I'm available for any additional info.
Thanks

jacobbednarz:
Within WAF packages, there are a few types of resources based on their
detection mode; user, anomaly and traditional. However, only packages
with the detection mode set to "anomaly" are able to be modified by the
end user. Attempting to update the others results in the following API
exception.

```
{
  "success": false,
  "errors": [
    {
      "code": 1027,
      "message": "Only anomaly WAF packages can have their package attributes updated"
    }
  ],
  "messages": [],
  "result": null
}
```

This issue surfaces when you attempt to manage non-anomaly rule types in
Terraform as the existing `Default` values set on the schema would
differ from the assumption that all packages had these properties and it
would attempt to get them into sync.

Instead, we need to conditionally apply updates to anomaly packages and
leave the other types to be managed as a read only source. To do this,
we can look at the `detection_type` and restrict the `Update` calls to
the anomaly based WAF packages.

Fixes #723

### DIFF
--- a/cloudflare/resource_cloudflare_waf_package.go
+++ b/cloudflare/resource_cloudflare_waf_package.go
@@ -11,6 +11,11 @@ import (
 
 const CLOUDFLARE_INVALID_OR_REMOVED_WAF_PACKAGE_ID_ERROR = 1002
 
+var (
+	defaultSensitivity = "high"
+	defaultActionMode  = "challenge"
+)
+
 func resourceCloudflareWAFPackage() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceCloudflareWAFPackageCreate,
@@ -39,14 +44,12 @@ func resourceCloudflareWAFPackage() *schema.Resource {
 			"sensitivity": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				Default:      "high",
 				ValidateFunc: validation.StringInSlice([]string{"high", "medium", "low", "off"}, false),
 			},
 
 			"action_mode": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				Default:      "challenge",
 				ValidateFunc: validation.StringInSlice([]string{"simulate", "block", "challenge"}, false),
 			},
 		},
@@ -75,8 +78,16 @@ func resourceCloudflareWAFPackageRead(d *schema.ResourceData, meta interface{}) 
 		return err
 	}
 
-	d.Set("sensitivity", pkg.Sensitivity)
-	d.Set("action_mode", pkg.ActionMode)
+	if pkg.DetectionMode == "anomaly" {
+		if pkg.Sensitivity != "" {
+			d.Set("sensitivity", pkg.Sensitivity)
+		}
+
+		if pkg.ActionMode != "" {
+			d.Set("action_mode", pkg.ActionMode)
+		}
+	}
+
 	d.SetId(pkg.ID)
 
 	return nil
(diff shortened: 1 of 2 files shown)


jacobbednarz:
cc @XaF as the original implementor of these resources

XaF:
I don't see any issue with the code.
Since with this modification I could try changing the `sensitivity` and `action_mode` without success in my terraform configuration, without understanding why, I would suggest that, if this is moving forward, this should fail terraform if there is a manually set `sensitivity` or `action_mode` for a non-anomaly package.

However, could you explain the use-case where you would want to use a `resource` for something that is read-only instead of using the `data` element ? The `resource` failing to update something that is not updatable seems expected to me, since this is a read-only element, not an actual resource that I can tweak through terraform. If the package could be updated otherwise I would understand avoiding the fail when other options can be tweaked, but here we're just having a resource that's not a resource, this does not seem really fitting with terraform concepts.

jacobbednarz:
#723 outlines the specific use case however I was thinking more broadly about custom rulesets (which are now largely deprecated with Firewall Filters and Workers). After digging back into this, the ones I was thinking about aren't a WAF package but a WAF ruleset within a package which would make this change moot. 

I can't envision Cloudflare adding more non-anomaly packages but perhaps it's safer this way? I'll get back to the original issue raiser and see what about the managing read-only resources bit too.

jacobbednarz:
Closing as we haven't heard from the original issue raiser.

PR #739, opened 2020-07-16
Update IdP identifiers to accept a list in access group schema

Justin-Holmes:
This PR should clean up redundancy in our IdP policy rules and create an API that's consistent with our dashboard UI. Here is an example of the proposed change:

**Current**
```terraform
include {
  gsuite {
    email                = "test1@example.com"
    identity_provider_id = "xxxx-xxx-xxxx"
  }
  gsuite {
    email                = "test2@example.com"
    identity_provider_id = "xxxx-xxx-xxxx"
  }
}
```

**New**
```terraform
include {
  gsuite {
    email                = ["test1@example.com", "test2@example.com"]
    identity_provider_id = "xxxx-xxx-xxxx"
  }
}
```

Unfortunately, I wasn't able to run the acceptance tests but I was able to create resources in my local environment without any issues.


jacobbednarz:
The current test suite is all green however we should probably a new test case to excercise multiple values in https://github.com/terraform-providers/terraform-provider-cloudflare/blob/master/cloudflare/resource_cloudflare_access_group_test.go.

Justin-Holmes:
@jacobbednarz Sure thing. Do you know if those tests create/destroy actual resources? If so, I think these tests will require an IdP to be created beforehand (along with their respective IdP groups) and I'm not sure I know the best way to go about that.

jacobbednarz:
Yep, the integration tests do create/destroy real resources however I think we can get away with dummy values like we already do at https://github.com/terraform-providers/terraform-provider-cloudflare/blob/55c2d9946b52853ebb4decaa542842ed72544eb9/cloudflare/resource_cloudflare_access_identity_provider_test.go#L174-L185

Within the test, you can try creating the fake IdP and then attaching the groups to it. I know I had issues with it in the past but I think that was for actual reproduction test cases and not integration testing (as we're really only asserting the Terraform state). Let me know if you have issues though and we can either rely on a mocked values or I'll setup a real SSO provider for it.

Justin-Holmes:
@jacobbednarz it looks like the acceptance tests are failing because of missing credentials. Let me know if there is anything I can do on my end to help!

jacobbednarz:
That's all good and thanks for the offer. Myself and @patryk (mainly @patryk though :p) moved the acceptance tests into GitHub Actions and forks don't get the credentials it needs as secrets aren't shared. 

I'll run these locally shortly.

cloudflare/resource_cloudflare_access_group.go
```diff
@@ -345,47 +354,56 @@ func BuildAccessGroupCondition(options map[string]interface{}) []interface{} {
...
 				githubCfg := v.(map[string]interface{})
-				group = append(group, cloudflare.AccessGroupGitHub{GitHubOrganization: struct {
-					Name               string `json:"name"`
-					Team               string `json:"team,omitempty"`
-					IdentityProviderID string `json:"identity_provider_id"`
-				}{
-					Name:               githubCfg["name"].(string),
-					IdentityProviderID: githubCfg["identity_provider_id"].(string),
-				}})
+				for _, team := range githubCfg["teams"].([]interface{}) {
```
jacobbednarz, line 370: @Justin-Holmes looks like the `teams` attribute is missing from the schema.

```
--- FAIL: TestAccCloudflareAccessGroupWithIDP (0.02s)
    testing.go:684: Step 0 error: config is invalid: Unsupported argument: An argument named "teams" is not expected here.
```

cloudflare/resource_cloudflare_access_group_test.go
```diff
@@ -120,6 +120,35 @@ func TestAccCloudflareAccessGroup_FullConfig(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCloudflareAccessGroupWithIDP(accountID, rnd, team),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCloudflareAccessGroupExists(groupName, &accessGroup),
+					resource.TestCheckResourceAttr(groupName, "account_id", accountID),
+					resource.TestCheckResourceAttr(groupName, "name", rnd),
+					resource.TestCheckResourceAttr(groupName, "include.0.github.account_id", accountID),
+					resource.TestCheckResourceAttr(groupName, "include.0.github.name", idpName),
+					resource.TestCheckResourceAttr(groupName, "include.0.github.team.0", team),
```
jacobbednarz, line 145: Once this is in the schema, it will be `teams`

```suggestion
					resource.TestCheckResourceAttr(groupName, "include.0.github.teams.0", team),
```

Justin-Holmes:
@jacobbednarz thank you, good catch!

jacobbednarz:
we're green! thanks!

analytically:
This PR breaks Cloudflare <> Okta for us. 'Error: missing expected [' even though the expected string to string array conversion is there.

image

jacobbednarz:
@analytically are you able to provide a new issue with a reproduction case and I can take a look?

groodt:
I see the same issue on Okta.

jacobbednarz:
@groodt do you mind opening an issue with a reproduction test case and I'll dig into it?

analytically:
Sorry don't have time.

PR #745, opened 2020-07-23
Support Spectrum applications with port ranges

player-two:
See the Port Ranges section of https://developers.cloudflare.com/spectrum/getting-started/configuration-options/
The API docs should be updated at some point today.

Requires https://github.com/cloudflare/cloudflare-go/pull/506, at which point this will no longer be a draft PR.


jacobbednarz:
The failing `webite-build` check is unrelated to these changes and will be fixed in another PR.

PR #757, opened 2020-08-10
Add CH Fallback Origin resources

cdloh:
I've added what unit tests I can, but I don't have an account I'm able to run the full suite across. Wasn't sure if you'd want update tests for this as well.

I've set the id to be `zoneid`-custom_hostname_fallback_origin as the API doesn't return a nice ID to use.

Not sure if I should set ForceNew for zoneid?

anyway thoughts / comments.


patryk:
Hi @cdloh, yes, `zone_id` should have `ForceNew`, as the config cannot be simply updated across zones, so it needs to be destroyed and created again when targetted zone changes.

cdloh:
@patryk should it also be doing that for custom hostnames then? At present the module isn't unless it's doing it in a different manner.

cloudflare/resource_cloudflare_custom_hostname_fallback_origin.go
```diff
@@ -0,0 +1,123 @@
...
+	fallbackOrigin := cloudflare.CustomHostnameFallbackOrigin{
+		Origin: origin,
+	}
+
+	_, err := client.UpdateCustomHostnameFallbackOrigin(zoneID, fallbackOrigin)
+	if err != nil {
+		return errors.Wrap(err, "failed to create custom hostname fallback origin")
+	}
+
+	d.SetId(zoneID + "-custom_hostname_fallback_origin")
```
jacobbednarz, line 83: Can we pass this through to [`stringChecksum`](https://github.com/cloudflare/terraform-provider-cloudflare/blob/f28c2bb96e78247690f690e65b9ebe4589ccd275/cloudflare/utils.go#L60-L66) to maintain consistent ID formats? It will return an MD5 hash and match the other identifiers we have in the state.

cloudflare/resource_cloudflare_custom_hostname_fallback_origin.go
```diff
@@ -0,0 +1,123 @@
...
+	fallbackOrigin := cloudflare.CustomHostnameFallbackOrigin{
+		Origin: origin,
+	}
+
+	_, err := client.UpdateCustomHostnameFallbackOrigin(zoneID, fallbackOrigin)
+	if err != nil {
+		return errors.Wrap(err, "failed to update custom hostname fallback origin")
+	}
+
+	d.SetId(zoneID + "-custom_hostname_fallback_origin")
```
jacobbednarz, line 102: You shouldn't need to set the ID in an `Update`. The ID should only be set within the `Create`.

cloudflare/resource_cloudflare_custom_hostname_fallback_origin.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"zoneID/origin\"", d.Id())
+	}
+
+	zoneID, origin := idAttr[0], idAttr[1]
+
+	log.Printf("[DEBUG] Importing Cloudflare Custom Hostname Fallback Origin: origin %s for zone %s", origin, zoneID)
+
+	d.Set("zone_id", zoneID)
+	d.Set("origin", origin)
+	d.SetId(zoneID + "-custom_hostname_fallback_origin")
```
jacobbednarz, line 120: Same comment as https://github.com/cloudflare/terraform-provider-cloudflare/pull/757/files#r468187578

cloudflare/resource_cloudflare_custom_hostname_fallback_origin_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+)
+
+func TestAccCloudflareCustomHostnameFallbackOrigin(t *testing.T) {
```
jacobbednarz, line 11: We're only testing the `Create` method here. Should we also be adding coverage for `Update`, `Delete` and `Import`?

jacobbednarz:
> should it also be doing that for custom hostnames then? At present the module isn't unless it's doing it in a different manner.

Yep, if that's been missed, we can add it as a condition.

cloudflare/resource_cloudflare_custom_hostname_fallback_origin.go
cdloh: Done.

cloudflare/resource_cloudflare_custom_hostname_fallback_origin_test.go
cdloh: I've added tests for Update. There hasn't been any other Delete or Import tests that I could see in the rest of the codebase. I'll have a punt at writing a delete one but not sure where to start for Import.

cdloh: I added checkdestroy to all of the tests.

jacobbednarz:
Hey @cdloh, thanks for all this! I've gone ahead and pushed up a few changes as the integration test suite wasn't consistently working and these got it to that happy place. Feel free to step through the individual commits however the summary is:

- We can't use parallel tests for this one as it's a single resource and they clobber each other if not run sequentially
- Some general housekeeping for the tests themselves to make it a bit more debuggable, readable and flexible in the event we ever change the test domain names
- Retry logic to account for eventual consistency issues in the remote API when creating/updating this resource

I'm going to give this another look over in a few hours after I've stepped away but I think we're nearly ready to merge this one.

jacobbednarz:
Thanks for your persistence here @cdloh! This is great! 🌟

Issue #702, opened 2020-06-09
Insufficient permissions error with Zone-specific API tokens


tornyairoland:
Hi!

I was directed here from [this](https://github.com/pulumi/pulumi-cloudflare/issues/82) pulumi-cloudflare issue.

While trying to limit the access of our Cloudflare API Tokens, I have come across an error.
I have created a new Token, added permissions to "Edit" the "Zone Settings" and "Included" only that "Specific Zone".
![kép](https://user-images.githubusercontent.com/48458705/84165420-f8c76700-aa73-11ea-86f5-73eafaafec04.png)

### Terraform Version
The Pulumi Cloudflare provider is using the Cloudflare Terraform provider, v2.7.0, and ` github.com/terraform-providers/terraform-provider-cloudflare v1.18.2-0.20200520193353-7dbdc13b06b6`.

### Affected Resource(s)
I tried it with `cloudflare_zone_settings_override`, but other `cloudflare_*` resources could be affected.

### Terraform Configuration Files
Based on the pulumi code I tried to run:
```hcl
resource "cloudflare_zone_settings_override" "settings" {
    zone_id = var.cloudflare_zone_id
    settings {
        ssl = "flexible"
        min_tls_version = "1.2"
    }
}
```
I tried getting the cloudflare_zone_id programmatically and copying the Zone ID manually from Cloudflare as well.

### Expected Behavior
Zone settings should have been applied as the token used for this has the permissions to do that.

### Actual Behavior
Got the following error from Pulumi:
```
Error: invocation of cloudflare:index/getZones:getZones returned an error: invoking cloudflare:index/getZones:getZones: error listing Zone: error from makeRequest: HTTP status 403: insufficient permissions
```

### Steps to Reproduce
1. Create an API Token limited to just a specific Zone, nothing Account related.
2. Try to apply some setting to just the specified Zone.

### Important Factoids
1. Using the same token and calling the Cloudflare API with curl works:
```
$ curl -X GET "https://api.cloudflare.com/client/v4/zones?name=<zone name>" -H "Authorization: Bearer <api token>" -H "Content-Type:application/json"

{"result":[{"id":"be2[...]","name":"<zone name>","status":"active",[...]}
```
2. Using the same curl command, but without the `?name=<zone name>` part returns an error from the Cloudflare API:
```
[...]

PR #760, opened 2020-08-11
Add more info on cloudflare_zones data source breaking change

tomaspinho:


jacobbednarz:
I appreciate the effort you've taken in an attempt to clarify this. Unfortunately, I don't think it conveys the message any differently or specifically to what is already there. In fact, it may be misleading as the release itself addressed a couple of edge cases where calling `/zones` for a single endpoint was returning an error (see #702).

I've provided a suggestion (removing the sentence about invalid permissions) and if we're happy with that, I'll merge this update.

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@
...
 * `resource/access_application`: Add CORS support ([#725](https://github.com/cloudflare/terraform-provider-cloudflare/issues/725)) 
 * `resource/cloudflare_custom_ssl`: Allow `geo_restrictions` to be `nil` and not included in the request payload ([#714](https://github.com/cloudflare/terraform-provider-cloudflare/issues/714)) 
-* `datasource/cloudflare_zones`: Filtering is now performed on the server side and the `name` parameter is no longer a regex. Instead, `name` is a string to match on and `match` is a regex. See the website documentation for more examples and updated references ([#708](https://github.com/cloudflare/terraform-provider-cloudflare/issues/708)) 
+* `datasource/cloudflare_zones`: Filtering is now performed on the server side and the `name` parameter is no longer a regex. Instead, `name` is a string to match on and `match` is a regex. This may pose a breaking change in your codebases resulting in bogus CloudFlare API permission errors. See the website documentation for more examples and updated references ([#708](https://github.com/cloudflare/terraform-provider-cloudflare/issues/708)) in order to make your code compatible with this release.
```
jacobbednarz, line 17:
```suggestion
* `datasource/cloudflare_zones`: Filtering is now performed on the server side and the `name` parameter is no longer a regex. Instead, `name` is a string to match on and `match` is a regex. See the website documentation for more examples and updated references ([#708](https://github.com/cloudflare/terraform-provider-cloudflare/issues/708)) in order to make your code compatible with this release.
```

tomaspinho: That API permission error sent me on a wild goose chase that made me think it was an API issue rather than a compatibility issue with the recently release version of the provider - hence why I didn't check the CHANGELOG the first time. Others may go through the same.

If that's something you want out of the CHANGELOG entry, I would be ok with removing it, however I think it devalues the change.

jacobbednarz: I understand what you're saying. The thing I want to avoid here is people coming across this changelog and start attributing permission errors to this single change. Unfortunately, edge cases causing permission errors are slightly more common since the launch of API tokens due to the number of configurations and options established around API keys and I wouldn't want to muddy the waters with a statement like the proposed one here.

tomaspinho: Ok. commited the proposed change. Hopefully SEO brings people to this PR if they face the same issue :smile:

jacobbednarz:
Cheers @tomaspinho! This is greatly appreciated!

PR #761, opened 2020-08-12
ForceNew if zone changes on CustomHostname resource

cdloh:
Custom Hostnames cannot be moved across zones. ForceNew if the zone changes.

### DIFF
--- a/cloudflare/provider_test.go
+++ b/cloudflare/provider_test.go
@@ -53,7 +53,9 @@ func testAccPreCheckAltDomain(t *testing.T) {
 	if v := os.Getenv("CLOUDFLARE_ALT_DOMAIN"); v == "" {
 		t.Fatal("CLOUDFLARE_ALT_DOMAIN must be set for this acceptance test")
 	}
+}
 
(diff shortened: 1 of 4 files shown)


jacobbednarz:
Thanks for this one @cdloh! Can you please add a test exercising this new condition? https://github.com/cloudflare/terraform-provider-cloudflare/blob/f61f3ad5355dc8150d4e39589fba71ea6bd99b43/cloudflare/resource_cloudflare_custom_hostname_test.go#L188-L224 is an example of where we do this elsewhere and confirm the values are indeed changing and Terraform is doing the expected thing.

cdloh:
@jacobbednarz is there another zoneid within the environment to test with? If there isn't another zone I'm unsure how to test it moving across zones.

jacobbednarz:
Yep, use [`CLOUDFLARE_ALT_ZONE_ID`](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/.github/workflows/acceptance-tests.yml#L16) and confirm the `plan` isn't empty.

PR #766, opened 2020-08-16
Adds new resource cloudflare_ip_list

cehrig:
This PR adds a new resource for managing account-level IP Lists (https://blog.cloudflare.com/introducing-ip-lists/)

WIP:
~~- Write acceptance tests~~
~~- Update cloudflare-go @jacobbednarz, i might need a hand here~~
~~- Documentation~~

Sample resource block
```
resource "cloudflare_ip_list" "example" {
  account_id = "d41d8cd98f00b204e9800998ecf8427e"
  name = "example_list"
  kind = "ip"
  description = "list description"

  item {
    value = "192.0.2.0"
    comment = "one"
  }

  item {
    value = "192.0.2.1"
    comment = "two"
  }
}
```


jacobbednarz:
Does IP Lists have public API support yet? If not, I'd probably wait for that to see what the schema needs to support.

cehrig:
> Does IP Lists have public API support yet? If not, I'd probably wait for that to see what the schema needs to support.

@jacobbednarz That's IP Lists aka Rules Lists we put to cloudflare-go some weeks ago: https://github.com/cloudflare/cloudflare-go/pull/507

jacobbednarz:
Ahhh, I remember that PR but didn't associate it with the issue I raised and thought was still open. Good to see!

jacobbednarz:
As for the updated dependency, once a new release is created an automatic PR will be opened here. If it's about time for a new release, we can ping @patryk to take a look.

cehrig:
Sounds good. I honestly was expecting the checks to fail because `vendor/` is not yet updated. So either I don't fully understand Go modules or the checks are not building the final binary and they are passing because Unit Tests are not yet written.

cloudflare/resource_cloudflare_ip_list.go
```diff
@@ -0,0 +1,170 @@
...
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("error creating IP List Items"))
+		}
+	}
+
+	return resourceCloudflareIPListRead(d, meta)
+}
+
+func resourceCloudflareIPListImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	listID := d.Id()
```
jacobbednarz, line 87: I suspect we will need the ID to be a composite ID of `accountID/listID` to allow people with access to multiple accounts to determine which one they would like to manage.

cloudflare/resource_cloudflare_ip_list.go
```diff
@@ -0,0 +1,170 @@
...
+	return &schema.Resource{
+		Create: resourceCloudflareIPListCreate,
+		Read:   resourceCloudflareIPListRead,
+		Update: resourceCloudflareIPListUpdate,
+		Delete: resourceCloudflareIPListDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceCloudflareIPListImport,
+		},
+
+		Schema: map[string]*schema.Schema{
```
jacobbednarz, line 23: Do we need `account_id` in here somewhere to tell the provider which accont to operate one? Or do you have a different setup in mind?

jacobbednarz (commented):
Looks pretty reasonable so far 👍 A question around multiple account usage but we'll also need some tests when you're ready.

cloudflare/resource_cloudflare_ip_list.go
cehrig: Oh thanks for raising this. I was a victim to `DefaultFunc: schema.EnvDefaultFunc("CLOUDFLARE_ACCOUNT_ID", nil),` in provider.go ... I'll get the whole PR ready by - hopefully - tomorrow!

jacobbednarz: No rush! I just finally had some time this morning to review some PRs and yours was at the top of the list 🙂

cehrig: This is fixed

cloudflare/resource_cloudflare_ip_list.go
cehrig: I'm not having a strong opinion on this. I would have reused the account_id from the provider block here since it is configuring the API client already. On the other hand I can see account_id being configured on the resource-level for some Access related things. (`cloudflare_access_service_tokens` for example). What do you think?

cehrig: Just read your last comment, Multi-Account usage makes totally sense. I've added this to the resource.

cehrig:
Okay, I think this is in a decent shape. Reviews appreciated :-)

jacobbednarz:
I've pushed 57942341 as the test account has a limit of 1 list at a time and the parallel tests step on top of one another.

jacobbednarz (approved):
🙇‍♂️ thanks heaps for this @cehrig!

PR #780, opened 2020-08-30
Add support for webassembly_binding in cloudflare_worker_script

### LINKED ISSUE #178: Support for wasm attachments to Cloudflare Workers
Hi! I'm looking into attaching a WASM module to my Cloudflare Worker, but I don't see support for it in the documentation. It looks like the functionality piggybacks on the [script upload](https://developers.cloudflare.com/workers/api/resource-bindings/webassembly-modules/) API endpoint.

I was wondering: are there plans to bring this functionality into the terraform cloudflare provider? (Would it be hard to do so?) No worries at all if support isn't planned for the time being! (I'd offer to help implement, but I'm barely conversant in Go — apologies!) Thanks for your work!

ldesgoui:
This remains a draft for now because I have not tested beyond running `make test`, I will attempt to make it run with terraform, I am quite new with the tool, so it may take some time.

Closes #178

cc @jacobbednarz

### DIFF
--- a/cloudflare/resource_cloudflare_worker_script.go
+++ b/cloudflare/resource_cloudflare_worker_script.go
@@ -1,7 +1,9 @@
 package cloudflare
 
 import (
+	"encoding/base64"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"strings"
 
@@ -50,6 +52,19 @@ var secretTextBindingResource = &schema.Resource{
 	},
 }
 
+var webAssemblyBindingResource = &schema.Resource{
+	Schema: map[string]*schema.Schema{
+		"name": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+		"module": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+	},
+}
+
 func resourceCloudflareWorkerScript() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceCloudflareWorkerScriptCreate,
@@ -84,6 +99,11 @@ func resourceCloudflareWorkerScript() *schema.Resource {
 				Type:     schema.TypeSet,
 				Optional: true,
 				Elem:     kvNamespaceBindingResource,
+			},
+			"webassembly_binding": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     webAssemblyBindingResource,
 			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


ldesgoui:
I found how to run acceptance tests:

```
[nix-shell:~/terraform-provider-cloudflare]$ TF_ACC=1 go test -v github.com/cloudflare/terraform-provider-cloudflare/cloudflare -run TestAccCloudflareWorkerScript
=== RUN   TestAccCloudflareWorkerScript_Import
--- PASS: TestAccCloudflareWorkerScript_Import (18.05s)
--- PASS: TestAccCloudflareWorkerScript_MultiScriptEnt (37.41s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/cloudflare	55.469s
```

ldesgoui:
Should I update CHANGELOG.md?

cloudflare/resource_cloudflare_worker_script.go
```diff
@@ -50,6 +52,20 @@ var secretTextBindingResource = &schema.Resource{
...
+var webAssemblyBindingResource = &schema.Resource{
+	Schema: map[string]*schema.Schema{
+		"name": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+		"base64": {
+			Type:      schema.TypeString,
+			Required:  true,
+			Sensitive: true,
```
jacobbednarz, line 64: What is the reasoning behind the base64 contents being marked as sensitive here?

cloudflare/resource_cloudflare_worker_script.go
```diff
@@ -50,6 +52,20 @@ var secretTextBindingResource = &schema.Resource{
 	},
 }
 
+var webAssemblyBindingResource = &schema.Resource{
+	Schema: map[string]*schema.Schema{
+		"name": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+		"base64": {
```
jacobbednarz, line 61: I find the schema name here a little lacking; what are your thoughts on using something more aligned with the existing schema fields but still convey it's base64 encoded, like `base64_encoded_text`? We could also reuse `text` here and note in the documentation it is the bas64 encoded value?

jacobbednarz:
Thanks for the quick turn around @ldesgoui! No need to update the CHANGELOG, we'll drop that in once it lands to avoid resolving merge conflicts for it within your branch.

cloudflare/resource_cloudflare_worker_script.go
ldesgoui: Woops, it shouldn't be, mistake when copy-pasting.

cloudflare/resource_cloudflare_worker_script.go
ldesgoui: Hm, I'm not particularly happy with the notion of "text" either, what's actually supposed to be here is a wasm module, which is binary data. `module` probably works best, changing it.

jacobbednarz: How about `data`? `module` leaves some ambiguity for a novice like me.

ldesgoui: It feels like `data` would be more ambiguous than `module`, the latter being defined in spec to be this exact thing

ldesgoui: Is this still a point of contention? I can deal with it being data `data`

jacobbednarz: No point of contention from me, just want this make sense to those that are going to be using it. I have little experience here so happy to leave as-is for your wording.

jacobbednarz (approved):
integration suite is all green 🍏 thanks for this!

PR #781, opened 2020-09-01
Remove AOP API Tokens gate as it's now supported in production

dhaynespls:
Essentially reverts https://github.com/cloudflare/terraform-provider-cloudflare/pull/772


dhaynespls:
These permissions allow you to interface with all AOP API endpoints through API tokens 
image

jacobbednarz (approved):

PR #789, opened 2020-09-06
resource/healthcheck: Retry hostname resolution errors

### LINKED ISSUE #786: Health checks are not able to see newly created DNS resources
Hi there,

### Terraform Version
0.12.28

### Affected Resource(s)
cloudflare_record
cloudflare_healthcheck

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Expected Behavior
If I create a DNS record, and then want to create a health check for that site, it should create fine

### Actual Behavior
module.www-dns.cloudflare_record.record: Creating...
module.www-dns.cloudflare_record.record: Creation complete after 1s [id=056f8d5ef3f06a9a2ce4c0ed1299f15f]
module.www-dns.cloudflare_healthcheck.health_check: Creating...

Error: error creating standalone healthcheck: error from makeRequest: HTTP status 400: content "{\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 1002,\n      \"message\": \"lookup www.mysite.com on 10.32.0.10:53: no such host\"\n    }\n  ],\n  \"messages\": []\n}\n"

### Steps to Reproduce

resource "cloudflare_record" "record" {
  zone_id = lookup(data.cloudflare_zones.zone.zones[0], "id")
  name    = var.name
  value   = var.value
  type    = var.type
  proxied = var.proxy_enabled
}

resource "cloudflare_healthcheck" "health_check" {
  zone_id                      = lookup(data.cloudflare_zones.zone.zones[0], "id")
  name                         = "${var.env}-${var.name}"
  description                  = var.description
  address                      = cloudflare_record.record.hostname
  suspended                    = false
  check_regions                = ["WNAM","ENAM"]
  notification_suspended       = false
  notification_email_addresses = ["ibinder@binderworld.com"]
  type                         = "HTTPS"
  port                         = "443"
  method                       = "GET"
  path                         = "/"
  expected_codes               = ["2xx"]
  follow_redirects             = true
  allow_insecure               = false
  timeout                      = 10
  retries                      = 2
  interval                     = 60
  consecutive_fails            = 3
  consecutive_successes        = 2
}

jacobbednarz:
This implements the retry logic which will allow leeway when a
Cloudflare DNS record is created at the same time as the healthcheck and
the former hasn't yet propogated to other systems.

By implementing this, we do create a much longer feedback loop to the
user when the record in use is indeed broken. It will go from ~ 1s to a
maximum of 30s. Should someone be debugging this scenario, they will
still receive the error in the Terraform logs but it will make
discoverying the underlying cause harder.

Fixes #786

### DIFF
--- a/cloudflare/resource_cloudflare_healthcheck.go
+++ b/cloudflare/resource_cloudflare_healthcheck.go
@@ -2,10 +2,12 @@ package cloudflare
 
 import (
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	"log"
 	"strings"
 	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 
 	"github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
@@ -162,6 +164,9 @@ func resourceCloudflareHealthcheck() *schema.Resource {
 				Computed: true,
 			},
 		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(30 * time.Second),
+		},
 	}
 }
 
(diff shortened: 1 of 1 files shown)


jacobbednarz:
@patryk thoughts as a maintainer? @cehrig as the original implementor (and assuming you're using it a bit), do you have any thoughts? This is a pretty significant slowdown if you do hit this scenario due to a broken hostname being added in.

cehrig:
Interesting, I did not come across this issue yet. It pretty much leaves us to retrying since the API is missing a bypass of the hostname check. :+1:  for the implementation and TIL that there's a timeout capability.

jacobbednarz:
Had an out of band conversation with @patryk on this one and we're willing to give it a go despite the potential drawback of sitting there for 30s if you input the hostname incorrectly.

patryk:
Sorry for not responding here. I think the delay that happens most often should be limited to single seconds, mainly related to underlying replication lag in the storage layer. We could tune it later if needed.

Issue #785, opened 2020-09-03
Health Checks do not handle state changes outside of Terraform if removed


ianbinder:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
0.12.28

### Affected Resource(s)
Please list the resources as a list, for example:
cloudflare_healthcheck

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Expected Behavior
If a health check is removed via console or API call outside of Terraform, Terraform should re-create the health check, just like it would any other resource. 

### Actual Behavior
During plan, returned this error: 

Error: error reading healthcheck information for "32c07cf03f68d972ff71c301c4b7a417": error from makeRequest: HTTP status 404: content "{\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 1001,\n      \"message\": \"object does not exist\"\n    }\n  ],\n  \"messages\": []\n}\n"

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Create a health check in terraform
2. Delete the health check via console
3. Run a terraform plan

jacobbednarz:
Thanks for raising this one @ianbinder. I've raised #787 which handles this scenario by handling when the call to `client.Healthcheck` returns an error mentioning the resource no longer exists at Cloudflare. By handling that specific case, you no longer get the error but instead prompted to create the resource and the old one will be cleaned up from your state.

Feel free to grab that PR and build it locally to confirm it also works for your use case.

PR #794, opened 2020-09-08
resource/ip_list: handle missing remote IP List and recreate

cehrig:
This is fixing pretty much the same issue as with #785 but for IP Lists

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 **Fixes**
 
 * `resource/cloudflare_healthcheck`: Handle resource deletion outside of Terraform ([#787](https://github.com/cloudflare/terraform-provider-cloudflare/issues/787)) 
+* `resource/cloudflare_custom_hostname`: Ensure `Import` sets hostname to prevent recreation ([#788](https://github.com/cloudflare/terraform-provider-cloudflare/issues/788)) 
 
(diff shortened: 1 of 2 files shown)


cloudflare/resource_cloudflare_healthcheck.go
```diff
@@ -225,14 +230,20 @@ func resourceCloudflareHealthcheckCreate(d *schema.ResourceData, meta interface{
...
+			}
 
-	d.SetId(hc.ID)
+			return resource.NonRetryableError(errors.Wrap(err, fmt.Sprintf("error creating standalone healthcheck")))
+		}
 
-	return resourceCloudflareHealthcheckRead(d, meta)
+		d.SetId(hc.ID)
+
+		return resource.NonRetryableError(resourceCloudflareHealthcheckRead(d, meta))
```
jacobbednarz: Did you intend to bring in #789? Or could we pick this one out of this PR?

jacobbednarz: Heh, speedy recovery 😛

jacobbednarz (approved):

cloudflare/resource_cloudflare_healthcheck.go
cehrig: Local git weirdness. Should be fixed with the latest merge

PR #796, opened 2020-09-10
resouce/ip_list: Remove `item.id` from schema

jacobbednarz:
Despite the `item` ID being marked as Computed, it is showing
unnecessary changes when running `apply` without any changes. For
whatever reason, the ID being returned is changing and Terraform is
trying to keep it in sync.

As we aren't using it, let's get rid of it and avoid the hassle of
keeping it in sync.

### DIFF
--- a/cloudflare/resource_cloudflare_ip_list.go
+++ b/cloudflare/resource_cloudflare_ip_list.go
@@ -3,13 +3,14 @@ package cloudflare
 import (
 	"context"
 	"fmt"
+	"log"
+	"regexp"
+	"strings"
+
 	"github.com/cloudflare/cloudflare-go"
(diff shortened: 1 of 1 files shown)


jacobbednarz:
@cehrig When this was added, was there any particular reason the item ID was required? Are we 👌 to do away with it?

cehrig:
Yeap, it's OK to remove it atm. since the resource is replacing the current list when making any changes to the items. That very likely is the reason why Item IDs are changing. 

If I find some time, I'll try to find a way to explicitly remove / add / update items by their ID. When I initially created the resource there were some very strange ordering side effect when working on the Item IDs, so I decided to go the full-replace route.

jacobbednarz:
Awesome, thanks for the confirmation. We can always add it back in if we have a way of updating the list items without replacing the entire list.

Issue #795, opened 2020-09-09
deleting a cloudflare_logpush_job when web-ui has no objects should not cause an error


olenm:
### Terraform Version
```
terraform -v
Terraform v0.13.2
+ provider registry.terraform.io/cloudflare/cloudflare v2.10.1
```

### Affected Resource(s)
- cloudflare_logpush_job

### Terraform Configuration Files
referenced docs page https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/logpush_job#example-usage-with-aws-provider

```hcl
resource "cloudflare_logpush_ownership_challenge" "ownership_challenge" {
  zone_id = "d41d8cd98f00b204e9800998ecf8427e"
  destination_conf = "s3://my-bucket-path?region=us-west-2"
}

data "aws_s3_bucket_object" "challenge_file" {
  bucket = "my-bucket-path"
  key    = cloudflare_logpush_ownership_challenge.ownership_challenge.ownership_challenge_filename
}

resource "cloudflare_logpush_job" "example_job" {
  enabled = true
  zone_id = "d41d8cd98f00b204e9800998ecf8427e"
  name = "My-logpush-job"
  logpull_options = "fields=RayID,ClientIP,EdgeStartTimestamp&timestamps=rfc3339"
  destination_conf = "s3://my-bucket-path?region=us-west-2"
  ownership_challenge = aws_s3_bucket_object.challenge_file.body
  dataset = "http_requests"
}
```

Then in the web-ui, delete the logpush configuration the above exists, re-run `terraform plan/apply`

### Debug Output
https://gist.github.com/olenm/72d2bc8db6d8e9e90e268ed95d4fbaa4

### Expected Behavior
Terraform queries said resources do not exist, and resources get unmanaged/deleted from terraform statefile

### Actual Behavior
```
# on terraform apply
cloudflare_logpush_job.cloudflare_logpush_job: Destroying... [id=12345]
Error: error deleting logpush job: 12345
```

### Steps to Reproduce
1. create/configure/enable terraform code to deploy a `cloudflare_logpush_job`
2. delete the configuration via the web-ui
3. delete the hcl code from step 1
4. re-run terraform plan/apply

### Current workaround
before plan, utilize `terraform state rm cloudflare_logpush_job.example_job`

jacobbednarz:
@olenm can you please provide your debug output with sensitive details redacted?

olenm:
@jacobbednarz debug output gist updated

PR #798, opened 2020-09-10
set logpush job id to "" when not found

### LINKED ISSUE #797: Logpush Job not removed from state when deleted from UI manually
### Terraform Version
v0.12.29 - however this isn't version specific

### Affected Resource(s)
- logpush_job

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

```hcl
resource "cloudflare_logpush_ownership_challenge" "ownership_challenge" {
  destination_conf = "s3://my-bucket-path?region=us-west-2"
}

data "aws_s3_bucket_object" "challenge_file" {
  bucket = "my-bucket-path"
  key    = cloudflare_logpush_ownership_challenge.ownership_challenge.ownership_challenge_filename
}

resource "cloudflare_logpush_job" "example_job" {
  enabled = true
  name = "My-logpush-job"
  logpull_options = "fields=RayID,ClientIP,EdgeStartTimestamp&timestamps=rfc3339"
  destination_conf = "s3://my-bucket-path?region=us-west-2"
  ownership_challenge = aws_s3_bucket_object.challenge_file.body
  dataset = "http_requests"
}
```

### Expected Behavior
The provider should replace the resource if it has been manually deleted outside of Terraform.

### Actual Behavior
The provider catches the 404 and logs an INFO message, however it does not mark the resource for removal.

### Steps to Reproduce
1. `terraform apply`
2. Delete the `example_job` Logpush Job via the Cloudflare Dashboard or UI.
3. `terraform plan`
4. result is no change detected.

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

trjstewart:
fixes #797 

There are currently no tests for the `logpush_job` resource that I could find so I haven't added or amended any. I assume that is because it requires resources external to Cloudflare?

Either way I have build the provider with this change and tested it ourselves and it's working as intended.

### DIFF
--- a/cloudflare/resource_cloudflare_logpush_job.go
+++ b/cloudflare/resource_cloudflare_logpush_job.go
@@ -90,6 +90,7 @@ func resourceCloudflareLogpushJobRead(d *schema.ResourceData, meta interface{}) 
 	if err != nil {
 		if strings.Contains(err.Error(), "404") {
 			log.Printf("[INFO] Could not find LogpushJob with id: %q", jobID)
+			d.SetId("")
 			return nil
 		}
 		return fmt.Errorf("error finding logpush job %q: %s", jobID, err)
--- a/cloudflare/resource_cloudflare_logpush_job.go
+++ b/cloudflare/resource_cloudflare_logpush_job.go
@@ -164,10 +164,17 @@ func resourceCloudflareLogpushJobDelete(d *schema.ResourceData, meta interface{}
 
 	deleteErr := client.DeleteLogpushJob(d.Get("zone_id").(string), job.ID)
 	if deleteErr != nil {
+		if strings.Contains(err.Error(), "job not found") {
+			log.Printf("[INFO] Could not find logpush job with id: %q", job.ID)
+			d.SetId("")
+			return nil
+		}
 		return fmt.Errorf("error deleting logpush job: %+v", job.ID)
 	}
 
-	return resourceCloudflareLogpushJobRead(d, meta)
+	d.SetId("")
+
+	return nil
 }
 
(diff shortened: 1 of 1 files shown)


jacobbednarz:
Thanks for putting in the time to fix this one 🍰 As the scenario we're addressing here is when the logpush job isn't in the remote but still in the state, we should include a fix for #795 where attempting to delete the resource fails due to it not being present in the remote. This should then cover off all the cases where the remote is out of sync with the local state.

trjstewart:
Good catch @jacobbednarz, I must be blind because I completely missed that issue. For now this small patch solved our immediate issue and we're just running a custom build of this branch internally for the time being. I won't likely have the time in the next few days to address anything additional for #795 so anyone else is more than welcome to.

cloudflare/resource_cloudflare_logpush_job.go
```diff
@@ -163,10 +164,17 @@ func resourceCloudflareLogpushJobDelete(d *schema.ResourceData, meta interface{}
...
 	if deleteErr != nil {
+		if strings.Contains(err.Error(), "job not found") {
+			log.Printf("[INFO] Could not find logpush job with id: %q", job.ID)
+			d.SetId("")
+			return nil
+		}
 		return fmt.Errorf("error deleting logpush job: %+v", job.ID)
 	}
 
-	return resourceCloudflareLogpushJobRead(d, meta)
```
jacobbednarz, line 169: We don't need to call `Read` as there isn't anything to "sync" after we `d.SetId("")`.

PR #800, opened 2020-09-16
Change cloudflare_certificate_pack hosts to TypeSet

### LINKED ISSUE #799: cloudflare_custom_certificate resources are being replaced unnecessarily
I believe I've found a bug which causes the `cloudflare_custom_certificate` resource to be unnecessarily replaced each time you run a `terraform apply`, even if there were no changes to the underlying resource.

### Terraform Version
```
Terraform v0.13.2
+ provider registry.terraform.io/terraform-providers/cloudflare v2.11.0
```

### Affected Resource(s)
- `cloudflare_custom_certificate`

```hcl
terraform {
  required_providers {
    cloudflare = {
      source = "terraform-providers/cloudflare"
    }
  }
  required_version = "= 0.13.2"
}

variable "api_token" {}
variable "account_id" {}
variable "zone_id" {}
variable "domain_name" {}

provider "cloudflare" {
  version    = "= 2.11"
  api_token  = var.api_token
  account_id = var.account_id
}

resource "cloudflare_certificate_pack" "testcert" {
  zone_id = var.zone_id
  type    = "dedicated_custom"
  hosts   = ["*.foo.${var.domain_name}", "foo.${var.domain_name}", "${var.domain_name}"]
}
```

### Expected Behavior
When I run `terraform apply` twice, I expected the first apply to create the certificate (which it did), but the second apply should not change anything, since the certificate was already created.

### Actual Behavior
The first time I ran the apply, it created a certificate, and saved the state like this:

```
$ terraform show                                      
# cloudflare_certificate_pack.testcert:
resource "cloudflare_certificate_pack" "testcert" {
    hosts   = [
        "foo.example.com",
        "*.foo.example.com",
        "example.com",
    ]
    id      = "REDACTED"
    type    = "dedicated_custom"
    zone_id = "REDACTED"
}
```

_Note that the hosts are ordered differently than they are in the .tf file._

The second time I ran the apply, it tried to destroy and recreate the resource because it was trying to reorder the hosts:

```

  # cloudflare_certificate_pack.testcert must be replaced
-/+ resource "cloudflare_certificate_pack" "testcert" {
      ~ hosts   = [ # forces replacement
          - "foo.example.com",
            "*.foo.example.com",
          + "foo.example.com",
            "example.com",
        ]
      ~ id      = "REDACTED" -> (known after apply)
        type    = "dedicated_custom"
        zone_id = "REDACTED"
    }
```

Even if I accept this change, then it will still prompt me to recreate the resource again when I run it a third time.

### Steps to Reproduce
1. Save the HCL file above into main.tf
2. `terraform init`
2. `terraform apply`
3. `terraform apply` (I expected this to complete with no changes, but it prompts to recreate the resource)

### Important Factoids
I am running this on an Enterprise Cloudflare account with SSL for SaaS.

msmith:
Fixes #799

The ordering of `hosts` within a Certificate Pack does not matter and it's not consistent. I observed that the ordering can change when the certificate status transitions from "pending_deployment" to "active".

Changing it to a TypeSet prevents the Certificate Pack from being unnecessarily replaced when the ordering changes.

### DIFF
--- a/cloudflare/resource_cloudflare_certificate_pack.go
+++ b/cloudflare/resource_cloudflare_certificate_pack.go
@@ -35,7 +35,7 @@ func resourceCloudflareCertificatePack() *schema.Resource {
 				ValidateFunc: validation.StringInSlice([]string{"custom", "dedicated_custom", "advanced"}, false),
 			},
 			"hosts": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
 				Required: true,
 				ForceNew: true,
 				Elem: &schema.Schema{
@@ -73,7 +73,7 @@ func resourceCloudflareCertificatePackCreate(d *schema.ResourceData, meta interf
 	client := meta.(*cloudflare.API)
 	zoneID := d.Get("zone_id").(string)
 	certificatePackType := d.Get("type").(string)
-	certificateHostnames := expandInterfaceToStringList(d.Get("hosts").([]interface{}))
+	certificateHostSet := d.Get("hosts").(*schema.Set)
 	certificatePackID := ""
 
 	if certificatePackType == "advanced" {
@@ -84,7 +84,7 @@ func resourceCloudflareCertificatePackCreate(d *schema.ResourceData, meta interf
 
 		cert := cloudflare.CertificatePackAdvancedCertificate{
 			Type:                 "advanced",
-			Hosts:                certificateHostnames,
+			Hosts:                expandInterfaceToStringList(certificateHostSet.List()),
 			ValidationMethod:     validationMethod,
 			ValidityDays:         validityDays,
 			CertificateAuthority: ca,
@@ -98,7 +98,7 @@ func resourceCloudflareCertificatePackCreate(d *schema.ResourceData, meta interf
 	} else {
 		cert := cloudflare.CertificatePackRequest{
 			Type:  certificatePackType,
-			Hosts: certificateHostnames,
+			Hosts: expandInterfaceToStringList(certificateHostSet.List()),
 		}
 		certPackResponse, err := client.CreateCertificatePack(zoneID, cert)
 		if err != nil {
@@ -122,7 +122,7 @@ func resourceCloudflareCertificatePackRead(d *schema.ResourceData, meta interfac
 	}
 
 	d.Set("type", certificatePack.Type)
-	d.Set("hosts", flattenStringList(certificatePack.Hosts))
+	d.Set("hosts", expandStringListToSet(certificatePack.Hosts))
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


jacobbednarz:
Nice! Are you able to add a regression test for this one? Certificates are a hard one to do due to needing pretty specific setup and scenarios so let me know if you have any issues with it.

msmith:
I'll give it a try. Do the regression tests run against a Cloudflare account that already has the appropriate entitlements?

jacobbednarz:
Yes and no :) The acceptance tests do but they don't get triggered on forks. I can pull it locally and confirm as well.
GitHub <notifications@github.com> wrote:
“I'll give it a try. Do the regression tests run against a Cloudflare account that already has the appropriate entitlements?”

msmith:
Ah, that makes sense. Now that I have the acceptance tests running locally against my own account, I've realized that this change breaks some of them. For example:

```
TestAccCertificatePackDedicatedCustom: testing.go:684: Step 0 error: Check failed: Check 3/4 error: cloudflare_certificate_pack.nrtveqkdho: Attribute 'hosts.0' not found
```

If I'm understanding correctly, the attribute key is now something like `hosts.123456`, since `hosts` is now a TypeSet instead of a TypeList. So I'll have to update those acceptance tests to check for the right attribute key.

I'm struggling to come up with a good way to write a regression test that demonstrates the bug that I initially found. I believe the bug only appears if some time passes between the first apply and the second one, because the host order doesn't change until the certificate state transitions from "pending_deployment" to "active". Maybe there's an elegant way to test this with the Terraform testing framework, but I haven't discovered it yet.

Here's what I think I can do:
1. Update the existing `TestAccCertificatePack` tests so that they all pass with my changes
2. Add a new acceptance test which demonstrates that re-ordering of hosts does not result in a change.

jacobbednarz:
> If I'm understanding correctly, the attribute key is now something like `hosts.123456`, since `hosts` is now a TypeSet instead of a TypeList. So I'll have to update those acceptance tests to check for the right attribute key.

Yep, when using a `TypeSet`, you don't assert the contents but instead the number of elements within it. An example is https://github.com/cloudflare/terraform-provider-cloudflare/blob/114b1a52ef89199874e4fee539e71e8ba58171a5/cloudflare/resource_cloudflare_firewall_rule_test.go#L108

This leads into your second part, I don't think there is a great way to inspect the elements of the `TypeSet` due to it being dynamically indexed. If your assumption is correct about the need for time to lapse and change state, it's probably not something we can test for without mocking the endpoint with a time delay/reshuffle of the data.

In most cases, `TypeSet`s are the way to go (for exactly this reason) so I'd 👍 the change without a regression test providing the others are back to passing with this change.

msmith:
I just pushed `2c0afcf`, which *should* make it so that the existing acceptance tests pass. I was only able to verify the `TestAccCertificatePackDedicatedCustom` test, because my account is not entitled for Advanced Certificate Manager.

It feels like I'm making the test coverage worse than it was before. Do you have any other suggestions for how I can improve this?

jacobbednarz:
Why do you think the test coverage has decreased? While this isn't testing specific values, it still maintains the coverage that changing functionality in this section will trigger the acceptance tests to be run and identify any potential issues from a higher level.

I'm happy proceeding with this if you are. I'll pull this locally and confirm it against our integration account.

msmith:
If you're able to verify that the tests still pass, then I'm happy for it to be merged. Thanks for the help!

jacobbednarz (approved):
LGTM and tests still work nicely, thanks! 🍭

PR #813, opened 2020-10-01
Possible typo in cloudflare_certificate_pack example

cpick:
I suspect that it should be "lets_encrypt" instead of "lets_encrypot".


jacobbednarz:
Cheers!

PR #818, opened 2020-10-08
Fix to consider that a custom hostname fallback origin can move to active during a retry

marc-barry:
We have been trying to use the new custom hostname fallback origin creation functionality of the provider and consistently hit the following error:

```
Error: expected custom hostname fallback to be created but was active
```

I traced the issue to the following condition:

```go
if fallbackHostname.Status != "pending_deployment" {	
    return resource.RetryableError(fmt.Errorf("expected custom hostname fallback to be created but was %s", fallbackHostname.Status))
}
```

The issue is that our custom hostname fallback origins from the https://api.cloudflare.com/#custom-hostname-fallback-origin-for-a-zone-properties API calls end up in the `active` state during the retry period. The retry logic didn't account for that state change and thus our custom domain fallback origin creation fails.

### DIFF
--- a/cloudflare/resource_cloudflare_custom_hostname_fallback_origin.go
+++ b/cloudflare/resource_cloudflare_custom_hostname_fallback_origin.go
@@ -89,10 +89,9 @@ func resourceCloudflareCustomHostnameFallbackOriginCreate(d *schema.ResourceData
 		}
 
 		// Address an eventual consistency issue where deleting a fallback hostname
-		// and then adding it _may_ cause some issues. We don't expect "active" here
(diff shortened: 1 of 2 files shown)


jacobbednarz:
Thanks for this one @marc-barry! Are you able to remove `go.sum` and we can get this one merged?

marc-barry:
> Thanks for this one @marc-barry! Are you able to remove `go.sum` and we can get this one merged?

@jacobbednarz I reverted the `go.sum` back to its original state in def1b37cd896fe5a7262da8ee4bac420ab49efbb.

Issue #812, opened 2020-10-01
Revert access applications to use zone-id instead of account-id


piizei:
Hi,

Please revert changes introduced in https://github.com/cloudflare/terraform-provider-cloudflare/pull/724 so that access resources prefer zone_id instead of account_id. 

Zones path deprecation has been cancelled due the fact that you cannot create granular enough access-tokens in large enterprises without the zones. Currently we are unable to work with provider versions 2.10 forward.

Br,
Petteri

jacobbednarz:
@piizei Are you able to please elaborate on the issue you're facing here? A production test case would be great to fully demonstrate the issue.

@Justin-Holmes do you know anything about the deprecation being reverted?

Justin-Holmes:
Hello, 

@piizei is correct that we reverted our decision to deprecate zone-level routes. In terms of next steps, I was thinking we'd support both `zone_id` and `account_id` for Access resources and direct the request to the proper route. This will require us to add support for the zone-level endpoints in [cloudflare-go](https://github.com/cloudflare/cloudflare-go) before we can make this change. Assuming this approach sounds good to you @jacobbednarz, I can get started on this tomorrow!

piizei:
Hi,
@jacobbednarz the problem is that after 2.10.1 release this does not work anymore:
```hcl
provider "cloudflare" {
  api_token = var.api_token
}
resource "cloudflare_access_application" "app" {
  zone_id          = var.dns_zone_id
  name             = var.target_dns
  domain           = var.target_dns
  session_duration = "30m"
  auto_redirect_to_identity = true
  allowed_idps = var.allowed_idps
}
```
Reason being that our api_tokens don't have account level access. Change introduced in 2.10.1 tries to use level access even if just zone_id is provided. This results 403 for us when terraform runs.

PR #819, opened 2020-10-08
Add support for zone-level routes to Access resources

Justin-Holmes:
Requires a new cloudflare-go version with https://github.com/cloudflare/cloudflare-go/pull/538 before this can be merged.

Closes https://github.com/cloudflare/terraform-provider-cloudflare/issues/812.

### DIFF
--- a/cloudflare/resource_cloudflare_access_application.go
+++ b/cloudflare/resource_cloudflare_access_application.go
@@ -23,15 +23,16 @@ func resourceCloudflareAccessApplication() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"account_id": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"zone_id"},
 			},
 			"zone_id": {
-				Deprecated: "This field will be removed in version 3 and replaced with the account_id field.",
-				Type:       schema.TypeString,
-				Optional:   true,
-				Computed:   true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"account_id"},
 			},
 			"aud": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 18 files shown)


Justin-Holmes:
@jacobbednarz would love your opinion on this approach before I implement this for the remaining resources!

cloudflare/resource_cloudflare_access_application.go
```diff
@@ -28,10 +28,9 @@ func resourceCloudflareAccessApplication() *schema.Resource {
 				Computed: true,
 			},
 			"zone_id": {
-				Deprecated: "This field will be removed in version 3 and replaced with the account_id field.",
-				Type:       schema.TypeString,
-				Optional:   true,
-				Computed:   true,
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
```
jacobbednarz, line 33: Should we add some schema level validation here via `ConflictsWith: []string{"other_attr"}` to make sure that we only ever have `zone_id` or `account_id`?

jacobbednarz (commented):
I'm on board with this approach 👍 My initial concerns were that this is going to get unwieldy with these conditionals in each CRUD method however I can't think a better way of getting it without sacrificing readability and introducing some ✨ magic methods ✨ but it will make life pretty hard to debug when the time comes so let's give this a whirl.

patryk (changes requested):
Need documentation changes and remove go.mod/sum from the PR, then we're good to merge.

jacobbednarz:
A thought for the documentation - It would be great if we could explicitly call out which uses cases fit `account_id` and `zone_id` so that we can have that as a reference point when debugging or digging into any issues that pop up.

Justin-Holmes:
@jacobbednarz Sure thing! Do you have a particular place in the documentation you'd like to me to add the notes?

jacobbednarz:
Nothing specific in mind so happy to defer to your best judgement on what explains it best. I'm just after an easy reference/guide for determining which one should be used. Some examples that come to mind are things like:

- Full HCL examples;
- Note in the attributes reference section; 
- Callout on the documentation (like https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/certificate_pack). Available callouts are documented at https://github.com/hashicorp/terraform-website#callouts

patryk (approved):

PR #844, opened 2020-10-26
Update `TestAccCloudflareAccessApplicationWithZoneID` to match assertion

### LINKED ISSUE #834: TestAccCloudflareAccessApplicationWithZoneID is failing
`TestAccCloudflareAccessApplicationWithZoneID` is failing with the following exception

```
 === RUN   TestAccCloudflareAccessApplicationWithZoneID
    testing.go:705: Step 0 error: Check failed: Check 2/2 error: cloudflare_access_application.ujaypcfsbx: Attribute 'account_id' not found
```

Link to run: https://github.com/cloudflare/terraform-provider-cloudflare/runs/1295364502?check_suite_focus=true#step:6:88

jacobbednarz:
Updates the assertion inside of
`TestAccCloudflareAccessApplicationWithZoneID` to check the `zone_id`
property to match the resource that is being created.

Closes #834

### DIFF
--- a/cloudflare/resource_cloudflare_access_application_test.go
+++ b/cloudflare/resource_cloudflare_access_application_test.go
@@ -291,14 +291,14 @@ func TestAccCloudflareAccessApplicationWithZoneID(t *testing.T) {
 				Config: testAccessApplicationWithZoneID(rnd, zone, zoneID),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(name, "name", rnd),
-					resource.TestCheckResourceAttr(name, "account_id", accountID),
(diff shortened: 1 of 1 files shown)


jacobbednarz:
@Justin-Holmes it looks like this was introduced in #819 however I'm not entirely certain on the intention. Was the intention of this test to test `zone_id` and it was a copy/paste error? Or should the `initIdentifier` method be setting the `account_id` and `zone_id` regardless of which is defined by the end user? Both seem plausible 🤷‍♂️

patryk (approved):

patryk:
Yeah, I think that might be just a copy pasta error.

PR #847, opened 2020-10-27
test/custom_ssl: Automate test certificate creation

### LINKED ISSUE #524: Automate creation of certificates for tests
While doing some unrelated cleanup, I encountered the following failure.

```
=== RUN   TestAccCloudflareCustomSSL_Basic
=== PAUSE TestAccCloudflareCustomSSL_Basic
=== CONT  TestAccCloudflareCustomSSL_Basic
--- FAIL: TestAccCloudflareCustomSSL_Basic (0.28s)
    testing.go:569: Step 0 error: errors during apply:

        Error: Failed to create custom ssl cert: error from makeRequest: HTTP status 400: content "{\"success\":false,\"errors\":[{\"code\":1206,\"message\":\"The certificate will expire within soon. Please supply a newer one\"}],\"messages\":[],\"result\":null}"

          on /opt/teamcity-agent/temp/buildTmp/tf-test656714404/main.tf line 2:
          (source code not available)
```

In one of our tests, we use a static TLS certificate (issued by Let's Encrypt) and it's expiry is approaching. I went ahead and generated a new one via #523 however it's not sustainable to keep doing this every N months. Instead, we should look to automate creation of these certificates during the test run so we never need to update this static reference again.

Something to consider is that this may be slow and error prone.

jacobbednarz:
Updates the test cases for `cloudflare_custom_ssl` resources to use the
Terraform ACME and TLS providers instead of storing static certificates.
This approach instead uses _real_ Let's Encrypt certificates which are
requested in real time, validated using DNS and then plugged into the
resource.

Closes #524


patryk:
The idea in general is correct, but we need to think about [Let's Encrypt's rate limits](https://letsencrypt.org/docs/rate-limits/).

jacobbednarz:
The main limitation I can see is hitting is the duplicate certificates (seeing how we're requesting the same hostnames). Is this inline with your thinking? If so, we may be able to work around this by appending an additional hostname which won't count towards a renewal. We may still hit the certificates per registered domain however I think that is less common given how often the CI suite runs. 

I initially thought about trying to store the certificates somewhere however I kept running into the issue that we would then need to introduce the logic for handling the renewal in the provider test suite. Example for initial run would be:

- generate certificate 
- run existing custom SSL resource tests
- upload to GCP

A renewal would then need to:

- pull certificate from GCP
- check certificate expiration 
- regenerate certificate if needed
- upload the new certificate to GCP
- run custom SSL resource tests

This wouldn't be an issue if we maintained the state somewhere as the ACME provider would automatically handle the refreshing of certificates. Do you have any other ideas to try here?

jacobbednarz:
coming back to this one, it looks like the ACME provider now (?) handles [renewal automatically](https://registry.terraform.io/providers/vancluever/acme/latest/docs/resources/certificate#certificate-renewal). we just need to make it work for our use case.

jacobbednarz:
superseded by #1698

Issue #821, opened 2020-10-12
Creation of CAA records


stack72:
Hi,

I am investigating the creation of CAA records with this Terraform Provider. It was suggested this was added in #97 where I could specify this as follows:

```
resource "cloudflare_record" "caa" {
  domain  = "mydomain.com"
  name    = "mydomain.com"
  data    = {
    flags = "0"
    tag   = "issue"
    value = "letsencrypt.org"
  }
  type    = "CAA"
  ttl     = 1
}
```

I cannot seem to find where the `value` parameter as part of the `data` map was added. Without this `value` parameter, no details can be passed as part of the record. 

I tried to set both the data and value blocks together but both are conflicting

```
if valueOk == dataOk {
		return fmt.Errorf(
			"either 'value' (present: %t) or 'data' (present: %t) must be provided",
			valueOk, dataOk)
	}
```

Please can you help with an example implementation here?

Paul

jacobbednarz:
Using the following setup, I'm unable to reproduce the issue.

```hcl
variable "cloudflare_email" {}
variable "cloudflare_api_key" {}
variable "cloudflare_zone_id" {}
variable "cloudflare_domain" {}

provider "cloudflare" {
  email      = var.cloudflare_email
  api_key    = var.cloudflare_api_key
}

resource "cloudflare_record" "caa" {
  zone_id  = var.cloudflare_zone_id
  name    = "caa-test.${var.cloudflare_domain}"
  data    = {
    flags = "0"
    tag   = "issue"
    value = "letsencrypt.org"
  }
  type    = "CAA"
  ttl     = 600
}
```

```
$ dig caa-test.terraform.cfapi.net -t caa +short
0 issue "letsencrypt.org"
```

You shouldn't be using `domain` attribute as it is deprecated in favour of `zone_id` and the `ttl` cannot be 1 for a non-proxied record.

stack72:
Hi @jacobbednarz 

Thanks for confirming here - just a follow up question on this - how does this even work when value isn't a parameter available on the data struct?

Paul

jacobbednarz:
It's kind of two part. The first is that nested or complex types in Terraform sometimes allow weird behaviour like not catching unexpected attributes (the case here) and the fact the provider iterates over the `d.GetOk("data")` to build the request.

[...]

PR #866, opened 2020-11-10
Adding specific CAA record fields

stack72:
Relying on the map feature means it could have changed in the future

### DIFF
--- a/cloudflare/resource_cloudflare_record_test.go
+++ b/cloudflare/resource_cloudflare_record_test.go
@@ -179,7 +179,8 @@ func TestAccCloudflareRecord_CAA(t *testing.T) {
 	var record cloudflare.DNSRecord
 	domain := os.Getenv("CLOUDFLARE_DOMAIN")
 	zoneID := os.Getenv("CLOUDFLARE_ZONE_ID")
-	resourceName := fmt.Sprintf("cloudflare_record.foobar")
+	rnd := generateRandomResourceName()
+	resourceName := fmt.Sprintf("cloudflare_record.%s", rnd)
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
@@ -187,7 +188,7 @@ func TestAccCloudflareRecord_CAA(t *testing.T) {
 		CheckDestroy: testAccCheckCloudflareRecordDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccCheckCloudflareRecordConfigCAA(zoneID, fmt.Sprintf("tf-acctest-caa.%s", domain)),
+				Config: testAccCheckCloudflareRecordConfigCAA(rnd, zoneID, fmt.Sprintf("tf-acctest-caa.%s", domain)),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckCloudflareRecordExists(resourceName, &record),
 					resource.TestCheckResourceAttr(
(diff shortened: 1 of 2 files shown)


cloudflare/resource_cloudflare_record_test.go
```diff
@@ -174,6 +174,30 @@ func TestAccCloudflareRecord_SRV(t *testing.T) {
 	})
 }
 
+func TestAccCloudflareRecord_CAA(t *testing.T) {
+	t.Parallel()
+	var record cloudflare.DNSRecord
+	domain := os.Getenv("CLOUDFLARE_DOMAIN")
+	zoneID := os.Getenv("CLOUDFLARE_ZONE_ID")
+	resourceName := fmt.Sprintf("cloudflare_record.foobar")
```
jacobbednarz, line 182: instead of using static values, let's prefer dynamic and random resource names to prevent potential collisions.

```suggestion
	rnd := generateRandomResourceName()
	name := fmt.Sprintf("cloudflare_record.%s", rnd)
```

I know most of these other resources don't adhear to this however it's how we're looking to move forward with newer tests.

jacobbednarz: This will also require you to update the rest of the test to pass through these values.

stack72: Hey @jacobbednarz 

I just made another commit that passes this through - I hope this is what you mean

Paul

jacobbednarz: yep, looks great, thanks!

cloudflare/resource_cloudflare_record_test.go
```diff
@@ -174,6 +174,31 @@ func TestAccCloudflareRecord_SRV(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckCloudflareRecordDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckCloudflareRecordConfigCAA(rnd, zoneID, fmt.Sprintf("tf-acctest-caa.%s", domain)),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCloudflareRecordExists(resourceName, &record),
+					resource.TestCheckResourceAttr(
```
jacobbednarz, line 194: this test doesn't fail if i comment out the changes you've made to `cloudflare/resource_cloudflare_record.go`. is that because we're already saving it to state through the existing map?

jacobbednarz: i suppose the case we want to protect against here is if that map was to go missing in favour of individual assignment so this is expected? 🤔

stack72: Correct - this was always working but it was never actually guaranteed that it would work. This is only calling out specific values like the lead from other resource types

stack72: This is as per the discussion in #821

jacobbednarz (approved):

stephanepham-sib:
hello @stack72 

Does it mean now we can pass the data block as null? and use value field?

Regards,

stack72:
I have not tested that specific usecase I'm afraid

stephanepham-sib:
Ok, because I have an issue for now where I can't pass data = null and set a value.
I'm using a terraform module to create either CAA record or A record, when it's CAA I want to set value to null and I fill the data block but when I create a A record I want to set value field and set to null the data block.

do you get my point?

stack72:
No functionality has changed - this specific PR has only added a little more to the schema to make it explicit

Issue #764, opened 2020-08-12
TF13: data.cloudflare_zones reports changes on every plan


threemachines:
### Terraform Version
Terraform v0.13.0

### Affected Resource(s)
- cloudflare_zones

### Terraform Configuration Files
```hcl
terraform {
  required_providers {
    cloudflare = {
      source = "terraform-providers/cloudflare"
    }
  }
  required_version = ">= 0.13"
}

provider "cloudflare" {
  version = "= 2.9.0"
  email   = "redacted"
  api_key = "redacted"
}

data "cloudflare_zones" "test" {
  filter {
    name = "somedomain.com"
  }
}

```

### Expected Behavior
Repeated plans should report no changes.

### Actual Behavior
Every plan shows a change for the data source:
```
Terraform will perform the following actions:

  # data.cloudflare_zones.test will be read during apply
  # (config refers to values not yet known)
 <= data "cloudflare_zones" "test"  {
      ~ id    = "2020-08-12 22:31:04.158252 +0000 UTC" -> "2020-08-12 22:31:04.758952 +0000 UTC"
        zones = [
            {
                id   = "redacted"
                name = "somedomain.com"
            },
        ]

        filter {
            lookup_type = "exact"
            name        = "somedomain.com"
            paused      = false
        }
    }

Plan: 0 to add, 0 to change, 0 to destroy.
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform init`
2. `terraform plan -out terraform.plan`
3. `terraform apply terraform.plan`
4. `terraform plan`

### References
This looks like basically the same issue as https://github.com/terraform-providers/terraform-provider-aws/issues/14579?

jacobbednarz:
Thanks for this; For now, let's see if this gets addressed in 0.13.x core instead of each provider and if it doesn't we can look to update this within the provider itself.

ionosphere80:
Same issue with Terraform 0.13.2.

Rude-Monkey:
Hi! If I'm not mistaken, they patched something similar on the core side in June (https://github.com/hashicorp/terraform/pull/25302). 
Sounds like they are expecting further fixes to be made on the provider side.

trjstewart:
+1 We're running into this too.

pecigonzalo:
[...]

PR #869, opened 2020-11-12
Replace ids with checksum of the resource ids

pecigonzalo:
Fixes: https://github.com/cloudflare/terraform-provider-cloudflare/issues/764

I was not sure if you would prefer to use `sha1` or some other function, but I think this should get the job done.

I chose to leave `generateShaId` as a separate function but repeat `sort` and `.join` because I thought there could be non list ids we could want to produce a sha for, but then noticed we do not. I think leaving it reusable its not bad, but let me know if this is not your preferred choice.

### DIFF
--- a/cloudflare/data_source_waf_groups.go
+++ b/cloudflare/data_source_waf_groups.go
@@ -135,6 +135,7 @@ func dataSourceCloudflareWAFGroupsRead(d *schema.ResourceData, meta interface{})
 				"modified_rules_count": group.ModifiedRulesCount,
 				"package_id":           pkg.ID,
 			})
+			groupIds = append(groupIds, group.ID)
 		}
 	}
 
--- a/cloudflare/data_source_waf_groups.go
+++ b/cloudflare/data_source_waf_groups.go
@@ -4,7 +4,8 @@ import (
 	"fmt"
 	"log"
 	"regexp"
-	"time"
+	"sort"
+	"strings"
 
 	cloudflare "github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
@@ -110,6 +111,7 @@ func dataSourceCloudflareWAFGroupsRead(d *schema.ResourceData, meta interface{})
 	}
 
 	log.Printf("[DEBUG] Reading WAF Groups")
+	groupIds := make([]string, 0)
 	groupDetails := make([]interface{}, 0)
 	for _, pkg := range pkgList {
 		groupList, err := client.ListWAFGroups(zoneID, pkg.ID)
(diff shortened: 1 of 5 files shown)


pecigonzalo:
>I chose to leave generateShaId as a separate function but repeat sort and .join because I thought there could be non list ids we could want to produce a sha for, but then noticed we do not. I think leaving it reusable its not bad, but let me know if this is not your preferred choice.

Decided this made no sense and merged the sort and sha.

jacobbednarz:
Thanks but we already have `stringChecksum` for this purpose. Are you able to update this PR to use that function instead?

pecigonzalo:
For sure, I did not notice there was one already, ill update this to use
that function.

pecigonzalo:
@jacobbednarz I reused the checksum function but kept the wrapper function to make the list -> checksum, in order to avoid repeating the `join`, etc. Would you prefer to remove that as well or maybe generalize it.

cloudflare/data_source_waf_groups.go
```diff
@@ -110,6 +109,7 @@ func dataSourceCloudflareWAFGroupsRead(d *schema.ResourceData, meta interface{})
 	}
 
 	log.Printf("[DEBUG] Reading WAF Groups")
+	groupIds := make([]string, 0)
```
jacobbednarz, line 112: I can't see where we are `append`ing to this slice. Am I missing it?

cloudflare/data_source_waf_packages.go
```diff
@@ -96,6 +95,7 @@ func dataSourceCloudflareWAFPackagesRead(d *schema.ResourceData, meta interface{
 	}
 
 	log.Printf("[DEBUG] Reading WAF Packages")
+	packageIds := make([]string, 0)
```
jacobbednarz, line 98: Same as https://github.com/cloudflare/terraform-provider-cloudflare/pull/869/files#r525448639

cloudflare/data_source_waf_rules.go
```diff
@@ -119,6 +118,7 @@ func dataSourceCloudflareWAFRulesRead(d *schema.ResourceData, meta interface{})
 	}
 
 	log.Printf("[DEBUG] Reading WAF Rules")
+	ruleIds := make([]string, 0)
```
jacobbednarz, line 121: Same as https://github.com/cloudflare/terraform-provider-cloudflare/pull/869/files#r525448639

jacobbednarz:
I think the dedicated `stringListChecksum` function is 👌 to avoid repeating it in each call site 👍 I've just got a couple of questions about appending to the slices which I can only seem to find a single place where we actually build up the slice.

pecigonzalo:
@jacobbednarz Good catch, I completely missed it when copy/pasting this from one data source to another. I believe I covered them all now.

jacobbednarz:
Awesome, thanks @pecigonzalo! You rock 🤘

rcny:
Can you release a patch version with this please?

jacobbednarz:
Probably not right away - we generally release on a monthly basis unless there are pressing fixes included in it. There are a couple of PRs in flight we'll be looking to include in the next release.

Issue #849, opened 2020-10-28
Worker secret resources


helenatuominen:
### Current Terraform version
0.13.5

### Description
The cloudflare API and cloudflare-go both expose the capability of directly adding secrets to workers non-destructively, having this available in Terraform would bring parity and allow further flexibility around IaC and worker CI/CD pipelines. Currently terraform-provider-cloudflare only allows secrets to be added at time of provision, requiring the script to accompany the worker.
I'm personally happy to work on a PR for this.

### Use-cases
A good use case we have is routes, workers and secrets (ie. cloud access keys) provisioned via Terraform, then subsequent upload of scripts to workers via a separate build-push pipeline. An option immediately pops out:
1. Allow worker resources with empty or no script value; and optionally
2. A distinct Terraform resource for worker secrets

### Potential Terraform configuration
1.
```
resource "cloudflare_worker_script" "my_script" {
  name = "script_1" # Note the absence of a script file.

  secret_text_binding {
    name = "MY_EXAMPLE_SECRET_TEXT"
    text = var.secret_foo_value
  }
}
```
2.
```
resource "cloudflare_worker_script" "my_script" {
  name = "script_1"
}

resource "cloudflare_worker_secret" "my_script_secret" {
  script_name = "script_1"
  name = "MY_EXAMPLE_SECRET_TEXT"
  text = var.secret_foo_value
}
```

### Community note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
  to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull
  request, please leave a comment

helenatuominen:
@jacobbednarz should I follow through with this PR?

jacobbednarz:
@helenatuominen Apologies for not looking at this one sooner but digging into the Workers side of things with Terraform/cloudflare-go translation is always alot of context to carry on and I just haven't had time. 

Looking at the Terraform resources and the Cloudflare documentation, it hints using the resources as we currently model:

> Resource bindings allow you to attach resources, such as secrets, or KV namespaces to your Worker script. If you are including Resources in your Worker, you need to specify their Bindings as a part of your upload
>
> -- https://developers.cloudflare.com/workers/platform/scripts#resource-bindings

> Currently terraform-provider-cloudflare only allows secrets to be added at time of provision, requiring the script to accompany the worker.

[...]

PR #870, opened 2020-11-16
Helenatuominen/cloudflare worker secret

helenatuominen:
RE:[Issue 849](https://github.com/cloudflare/terraform-provider-cloudflare/issues/849)
Hi this is work complete for the implementation of worker_secret distinct from work_script. The two are to be used mutually exclusively in a plan. Please let me know how it is :)

### DIFF
--- a/cloudflare/resource_cloudflare_worker_secret.go
+++ b/cloudflare/resource_cloudflare_worker_secret.go
@@ -81,7 +81,7 @@ func resourceCloudflareWorkerSecretCreate(d *schema.ResourceData, meta interface
 		return errors.Wrap(err, "error creating worker secret")
 	}
 
-   d.SetId(stringChecksum(fmt.Sprintf("%s/%s", scriptName, name)))
+	d.SetId(stringChecksum(fmt.Sprintf("%s/%s", scriptName, name)))
 
 	log.Printf("[INFO] Cloudflare Workers Secret ID: %s", d.Id())
 
--- a/cloudflare/resource_cloudflare_worker_secret.go
+++ b/cloudflare/resource_cloudflare_worker_secret.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"strings"
 
 	cloudflare "github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
@@ -12,9 +13,8 @@ import (
 
 func resourceCloudflareWorkerSecret() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceCloudflareWorkerSecretUpdate,
+		Create: resourceCloudflareWorkerSecretCreate,
 		Read:   resourceCloudflareWorkerSecretRead,
-		Update: resourceCloudflareWorkerSecretUpdate,
 		Delete: resourceCloudflareWorkerSecretDelete,
 		Importer: &schema.ResourceImporter{
 			State: resourceCloudflareWorkersSecretImport,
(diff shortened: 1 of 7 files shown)


cloudflare/resource_cloudflare_worker_secret.go
```diff
@@ -0,0 +1,119 @@
...
+		},
+	}
+}
+
+func resourceCloudflareWorkerSecretRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	scriptName := d.Get("script_name").(string)
+	name := d.Get("name").(string)
+
+	secrets, err := client.ListWorkersSecrets(context.Background(), scriptName)
```
jacobbednarz, line 52: depending on what we are carrying here, `context.TODO` may be more appropriate. do you recall what the underlying library is wanting here?

cloudflare/resource_cloudflare_worker_secret.go
```diff
@@ -0,0 +1,119 @@
...
+		return errors.Wrap(err, "error reading worker secrets")
+	}
+
+	for _, secret := range secrets.Result {
+		if secret.Name == name {
+			return nil
+		}
+	}
+
+	d.SetId("")
```
jacobbednarz, line 63: calling [`d.SetId("")` removes the resource from the state](https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_data.go#L255-L272) which i don't think you want here. typically in the `Read` operation, we sync all the values from the remote call to ensure we have the latest values. e.g.

```golang
d.Set("script_name", secrets.ScriptName)
d.Set("name", secrets.Name)
d.SetId(stringChecksum(....)) // maybe combine the name + script name?
```

cloudflare/resource_cloudflare_worker_secret.go
```diff
@@ -0,0 +1,119 @@
...
+		Text: secretText,
+		Type: cloudflare.WorkerSecretTextBindingType,
+	}
+
+	_, err := client.SetWorkersSecret(context.Background(), scriptName, &request)
+	if err != nil {
+		return errors.Wrap(err, "error creating worker secret")
+	}
+
+	d.SetId(fmt.Sprintf("%s/%s", scriptName, name))
```
jacobbednarz, line 84: lets use [`stringChecksum`](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/utils.go#L61-L67) here to output a MD5 and match the other resource IDs

 ```suggestion
	d.SetId(stringChecksum(fmt.Sprintf("%s/%s", scriptName, name)))
```

File: cloudflare/resource_cloudflare_worker_secret.go [hunk 4]
```diff
@@ -0,0 +1,119 @@
...
+
+	log.Printf("[INFO] Cloudflare Workers Secret ID: %s", d.Id())
+
+	return nil
```
jacobbednarz, line 88: at the end of the `Create` it's [a good practice to call `Read`](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#update-state-after-modification) to make sure the state is in sync

cloudflare/resource_cloudflare_worker_secret.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func resourceCloudflareWorkerSecretDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	scriptName, name := parseSecretId(d.Id())
```
jacobbednarz, line 93: if we're cleaning up `d.Id()` we'll need to use `d.Get` here instead to populate the values.

cloudflare/resource_cloudflare_worker_secret.go
```diff
@@ -0,0 +1,119 @@
...
+
+	log.Printf("[INFO] Deleting Cloudflare Workers secret with id: %+v", d.Id())
+
+	_, err := client.DeleteWorkersSecret(context.Background(), scriptName, name)
+	if err != nil {
+		return errors.Wrap(err, "error deleting worker secret")
+	}
+
```
jacobbednarz, line 101: in the `Delete` we should either call `Read` (and handle the missing script) or `d.SetId("")` to remove the resource from being managed by Terraform.

cloudflare/resource_cloudflare_worker_secret_test.go
```diff
@@ -0,0 +1,165 @@
...
+			{
+				Config: testAccCheckCloudflareWorkerSecretWithWorkerScript(workerSecretTestScriptName, name, secretText),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCloudflareWorkerSecretExists(workerSecretTestScriptName, name, secretText),
+				),
+			},
+		},
+	})
+
+	deleteWorker()
```
jacobbednarz, line 44: we shouldn't need to explicitly call `deleteWorker` here as the test harness calls the `Delete` operation on our behalf.

cloudflare/resource_cloudflare_worker_secret_test.go
```diff
@@ -0,0 +1,165 @@
...
+		if len(secretResponse.Result) > 0 {
+			return fmt.Errorf("Worker secret with name %s still exists against Work Script %s", secretResponse.Result[0].Name, params.ScriptName)
+		}
+	}
+
+	return nil
+}
+
+func testAccCheckCloudflareWorkerSecretWithWorkerScript(scriptName string, name string, secretText string) string {
+	err := createWorker()
```
jacobbednarz, line 70: the way that we usually do this is create the resource as a part of the config where you include all the resources and from there, reference the values in the dependant resources. this ensures that we're using the resources under test and not using custom code (like we have here) where the codepaths may be different. check out Access resource tests for some examples of this one.

cloudflare/resource_cloudflare_worker_secret_test.go
```diff
@@ -0,0 +1,165 @@
...
+	_, err = client.DeleteWorker(&scriptRequestParams)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func createCloudflareClient() (*cloudflare.API, error) {
```
jacobbednarz, line 142: we don't need to re-create an API client, the provider test harness does this out of the box - you will need to provide it with the correct credentials via environment variables though.

jacobbednarz:
Solid start with some comments inline to update with the Terraform-isms. Thanks for putting this up!

cloudflare/resource_cloudflare_worker_secret.go
helenatuominen: I don't understand. The library is listing the worker secrets by issuing a get request against the server.

cloudflare/resource_cloudflare_worker_secret.go
jacobbednarz: Sure, the value cannot be retrieved however it should still be persisted to state. Without it, every Terraform operation will re-create this.

cloudflare/resource_cloudflare_worker_secret_test.go
jacobbednarz: From my understanding, that test case shouldn't require to be manually configuring this like this - even if they are mutually exclusive as they need to be able to co-exist in configuration together. Can you put up a small reproduction case to show the issues you had that lead here and we can see what we can do about it.

cloudflare/resource_cloudflare_worker_secret_test.go
jacobbednarz: Gotcha, that to me is an indication that we're holding something wrong as the provider credential configuration should be the entry point for all test cases. Let's continue this in the comment above as I think that will unlock some more ideas.

cloudflare/resource_cloudflare_worker_secret.go
jacobbednarz: `context.TODO()` and `context.Background()` [share the same underlying functionality](https://github.com/golang/go/blob/master/src/context/context.go#L204-L218) however the way I read the intended usage is that `TODO()` is when we aren't explicitly expecting to use the context for deadlines, etc whereas `Background()` is when are explicitly expecting to use it in that way. My question is mainly around are we intentionally doing this or is it to just satisfy the function signature? In which case, `context.TODO()` may make more sense here. I'm not fussed too much  but do think we want to clarify.

cloudflare/resource_cloudflare_worker_secret.go
jacobbednarz: yep - that will get us into a place where the `Read` centralises the state syncing and we can use that in a few places.

cloudflare/resource_cloudflare_worker_secret.go
helenatuominen: AFAICT context.TODO() is used when there is intent in the codebase to _eventually_ provide a background in future work. That is not the case here as we won't be passing a Context into this function in the future, so I'm happy with it as is :)

cloudflare/resource_cloudflare_worker_secret.go
helenatuominen: I've decided that the value must always be recreated as the existing secret cannot be checked against what is being presented to the resource.

helenatuominen: Unfortunately as there is no way to retrieve the secret values from Cloudflare, the drift cannot be detected, and so the secrets must be recreated/overwritten each time the plan is executed. This is as per the resource provider for the Hashicorp Vault provider: https://github.com/hashicorp/terraform-provider-vault/blob/master/vault/resource_generic_secret.go

jacobbednarz:
@helenatuominen is this still something you're interested in persuing?

jacobbednarz:
closing this due to going stale. if there is still interest, let's respawn this thread when ready.

lboynton:
Is this something we can resurrect? Happy to help contribute unless there is a reason why this wasn't added?

Issue #871, opened 2020-11-19
Error 1002 while trying to create a TCP monitor


thecodeassassin:
### Terraform configuration files

```hcl
resource "cloudflare_load_balancer_monitor" "tcp_monitor" {
  type = "tcp"
  method = "connection_established"
  timeout = 7
  interval = 60
  retries = 5
  description = "example tcp load balancer"
}
```

### Debug output

https://gist.github.com/thecodeassassin/0aa038f39e96562046ab22f22a6eb83f

### Expected behavior

Monitor should have been created

### Actual behavior

```json
Error: error creating load balancer monitor: error from makeRequest: HTTP status 400: content "{\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 1002,\n      \"message\": \"interval is not in range [0, 0]: validation failed\"\n    }\n  ],\n  \"messages\": []\n}\n"
 ```

### Steps to reproduce

1. Make a clean TF file like this:

```hcl
provider "cloudflare" {
  email   = ""
  api_key = ""
}


resource "cloudflare_load_balancer_monitor" "tcp_monitor" {
  type = "tcp"
  method = "connection_established"
  timeout = 7
  interval = 60
  retries = 5
  description = "example tcp load balancer"
}
```

^^^ This comes from the documentation here => https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/load_balancer_monitor

terraform init
terraform plan
terraform apply

jacobbednarz:
@thecodeassassin Are you able to provide a full reproduction test case as well as a full `TF_LOG=DEBUG` output? The resource and reproduction case you've provided don't match up. One is for `cloudflare_load_balancer_monitor` but the gist is for a failed spectrum application.

thecodeassassin:
@jacobbednarz I updated the gist, these are logs from a clean terraform file as described above. All I did was copy paste the example and I was able to reproduce the issue.

jacobbednarz:
Looking at that gist, the API response seems to erroneous as the validation field is looking for a range of 0,0 despite the [API documentation allowing integers](https://api.cloudflare.com/#load-balancer-monitors-create-monitor). Would you mind opening a support ticket and posting the # back here to track it?

thecodeassassin:
@jacobbednarz the ticket number is  #2029359.

thecodeassassin:
@jacobbednarz 

[...]

PR #879, opened 2020-12-01
corrected docs for load balancer TCP monitor

thecodeassassin:
What was updated?

The docs did not mention that port is required when configuring a TCP load balancer monitor.

This fixes https://github.com/cloudflare/terraform-provider-cloudflare/issues/871

### DIFF
--- a/website/docs/r/load_balancer_monitor.html.markdown
+++ b/website/docs/r/load_balancer_monitor.html.markdown
@@ -10,7 +10,7 @@ description: |-
 
 If you're using Cloudflare's Load Balancing to load-balance across multiple origin servers or data centers, you configure one of these Monitors to actively check the availability of those servers over HTTP(S) or TCP.
 
--> **Note:** When creating a monitor, you have to pass `account_id` to the provider configuration.
+-> **Note:** When creating a monitor, you have to pass `account_id` to the provider configuration in order to create account level resources. Otherwise, by default, it will be a user level resource.
 
 ## Example Usage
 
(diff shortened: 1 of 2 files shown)


website/docs/r/load_balancer_monitor.html.markdown
```diff
@@ -10,6 +10,9 @@ description: |-
 
 If you're using Cloudflare's Load Balancing to load-balance across multiple origin servers or data centers, you configure one of these Monitors to actively check the availability of those servers over HTTP(S) or TCP.
 
+*Note*
+When creating a monitor, you have to pass `account_id` to the provider configuration.
```
jacobbednarz, line 14: There is some special Terraform website markup we can use here to make this a callout

```suggestion
-> **Note:** When creating a monitor, you have to pass `account_id` to the provider configuration.
```

See https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/access_application for an example of it in action.

website/docs/r/load_balancer_monitor.html.markdown
```diff
@@ -10,6 +10,8 @@ description: |-
 
 If you're using Cloudflare's Load Balancing to load-balance across multiple origin servers or data centers, you configure one of these Monitors to actively check the availability of those servers over HTTP(S) or TCP.
 
+-> **Note:** When creating a monitor, you have to pass `account_id` to the provider configuration.
```
jacobbednarz, line 13:
```suggestion
-> **Note:** When creating a monitor, you have to pass `account_id` to the provider configuration in order to create account level resources. Otherwise, by default, it will be a user level resource.
```

jacobbednarz: Expanded on this a little more to clarify that you don't _have_ to set it if the intention is to use the user level resources.

jacobbednarz (approved):
thanks for this one!

thecodeassassin:
> thanks for this one!

thanks to you as well!

PR #885, opened 2020-12-07
Update interpolation syntax

putz612:
As of 0.11 interpolation-only expressions are deprecated.  Updating examples to use newer syntax.

### DIFF
--- a/website/docs/d/zones.html.md
+++ b/website/docs/d/zones.html.md
@@ -87,7 +87,7 @@ resource "cloudflare_zone_lockdown" "endpoint_lockdown" {
 resource "cloudflare_zone" "example" {
   zone_id     = lookup(data.cloudflare_zones.test.zones[0], "id")
   name        = "www"
-  value       = "192.168.0.11"
+  value       = "203.0.113.1"
   type        = "A"
   proxied     = false
 }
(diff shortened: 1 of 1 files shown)


website/docs/d/zones.html.md
```diff
@@ -83,6 +83,14 @@ resource "cloudflare_zone_lockdown" "endpoint_lockdown" {
     value = "198.51.100.4"
   }
 }
+
+resource "cloudflare_zone" "example" {
+  zone_id     = lookup(data.cloudflare_zones.test.zones[0], "id")
+  name        = "www"
+  value       = "192.168.0.11"
```
jacobbednarz, line 90: Can we please use the reserved IP ranges for documentation from [rfc5737](https://tools.ietf.org/html/rfc5737)? 192.0.2.1 or 203.0.113.1 for example.

jacobbednarz (changes requested):

website/docs/d/zones.html.md
putz612: Sure, updated.  I was just copying the examples from docs/resources/record

jacobbednarz:
Thanks!

PR #887, opened 2020-12-08
more accurate cloudflare_load_balancer example

theRealWardo:
I am a silly user that copy pastes examples like a pro. I'd propose a more accurate example to help silly users like me.

also sorry for not opening an issue for these 11 characters. hopefully we can all agree we don't need one for a small docs patch.

```
* cloudflare_load_balancer.manifestlb: error creating load balancer for zone: error from makeRequest: HTTP status 400: content "{\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 1002,\n      \"message\": \"invalid hostname: subdomain-here: validation failed\"\n    }\n  ],\n  \"messages\": []\n}\n"
```

### DIFF
--- a/website/docs/r/load_balancer.html.markdown
+++ b/website/docs/r/load_balancer.html.markdown
@@ -18,7 +18,7 @@ Provides a Cloudflare Load Balancer resource. This sits in front of a number of 
 # Within each pop or region we can define multiple pools in failover order
 resource "cloudflare_load_balancer" "bar" {
   zone_id = "d41d8cd98f00b204e9800998ecf8427e"
-  name = "example-load-balancer"
+  name = "example-load-balancer.mysite.com"
   fallback_pool_id = cloudflare_load_balancer_pool.foo.id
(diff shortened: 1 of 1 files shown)


website/docs/r/load_balancer.html.markdown
```diff
@@ -18,7 +18,7 @@ Provides a Cloudflare Load Balancer resource. This sits in front of a number of
 # Within each pop or region we can define multiple pools in failover order
 resource "cloudflare_load_balancer" "bar" {
   zone_id = "d41d8cd98f00b204e9800998ecf8427e"
-  name = "example-load-balancer"
+  name = "example-load-balancer.mysite.com"
```
jacobbednarz, line 21:
```suggestion
  name = "example-load-balancer.example.com"
```

jacobbednarz:
thanks! i've confirmed the attribute description is accurate so this is all that was needed.

PR #895, opened 2020-12-17
Add custom deny fields to Access application

Justin-Holmes:
Requires a new cloudflare-go version with https://github.com/cloudflare/cloudflare-go/pull/566 before this can be merged.

### DIFF
--- a/cloudflare/resource_cloudflare_access_application.go
+++ b/cloudflare/resource_cloudflare_access_application.go
@@ -119,6 +119,14 @@ func resourceCloudflareAccessApplication() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"custom_deny_message": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"custom_deny_url": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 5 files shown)


cloudflare/resource_cloudflare_access_application.go
```diff
@@ -134,6 +142,8 @@ func resourceCloudflareAccessApplicationCreate(d *schema.ResourceData, meta inte
 		SessionDuration:        d.Get("session_duration").(string),
 		AutoRedirectToIdentity: d.Get("auto_redirect_to_identity").(bool),
 		EnableBindingCookie:    d.Get("enable_binding_cookie").(bool),
+		CustomDenyMessage:      d.Get("custom_deny_message").(string),
```
jacobbednarz, line 145: do we always want this appended to send zero values? or should we conditionally add it in?

Justin-Holmes: We don't want to append zero values. I'm using the `omitempty` tag on those fields https://github.com/Justin-Holmes/cloudflare-go/blob/justin/access_custom_deny_fields/access_application.go#L26-L27, so I believe they'll be omitted from the request.

jacobbednarz (approved):
wicked! thanks 🙇

Issue #653, opened 2020-04-09
Remove edge_cache_ttl zone setting


garrettgalow:
This setting is not officially supported, and doesn't actually change any edge behavior. It should not have been included in Terraform and should be removed. The proper way to control this is via page rules

garrettgalow:
in tf docs it is currently documented here: https://www.terraform.io/docs/providers/cloudflare/r/zone_settings_override.html#edge_cache_ttl

PR #899, opened 2020-12-28
Adds ciphers to zone settings override

riuvshyn:
Adds zone settings override for `ciphers`: https://api.cloudflare.com/#zone-settings-change-ciphers-setting

Also I've noticed that `client/v4/zones/.../settings` returns some more options that are not available in terraform:
```
{
  "result": [

...

    {
      "id": "ciphers",
      "value": [],
      "modified_on": null,
      "editable": true
    },

...

    {
      "id": "edge_cache_ttl",
      "value": 7200,
      "modified_on": null,
      "editable": true
    },

...

    {
      "id": "orange_to_orange",
      "value": "off",
      "modified_on": null,
      "editable": true
    },

...

    {
      "id": "visitor_ip",
      "value": "on",
      "modified_on": null,
      "editable": true
    },

...

  ],
  "success": true,
  "errors": [],
  "messages": []
}
```

Should I also add `visitor_ip`, `orange_to_orange`, `edge_cache_ttl` settings? Also I do not see anything for these settings in the docs...

### DIFF
--- a/cloudflare/resource_cloudflare_zone_settings_override_test.go
+++ b/cloudflare/resource_cloudflare_zone_settings_override_test.go
@@ -62,6 +62,8 @@ func TestAccCloudflareZoneSettingsOverride_Full(t *testing.T) {
 						name, "settings.0.zero_rtt", "off"),
 					resource.TestCheckResourceAttr(
 						name, "settings.0.universal_ssl", "off"),
+					resource.TestCheckResourceAttr(
+						name, "settings.0.ciphers", "[\"ECDHE-RSA-AES128-GCM-SHA256\", \"AES128-SHA\"]"),
 				),
 			},
 		},
@@ -245,6 +247,7 @@ resource "cloudflare_zone_settings_override" "test" {
 	zone_id = "%s"
 	settings {
 		brotli = "on"
+		ciphers = ["ECDHE-RSA-AES128-GCM-SHA256", "AES128-SHA"]
 		challenge_ttl = 2700
 		security_level = "high"
 		opportunistic_encryption = "on"
--- a/cloudflare/resource_cloudflare_zone_settings_override.go
+++ b/cloudflare/resource_cloudflare_zone_settings_override.go
@@ -126,6 +126,15 @@ var resourceCloudflareZoneSettingsSchema = map[string]*schema.Schema{
 			86400, 604800, 2592000, 31536000}),
 	},
 
+	"ciphers": {
+		Type:     schema.TypeList,
+		Optional: true,
+		Computed: true,
+		Elem: &schema.Schema{
+			Type: schema.TypeString,
+		},
+	},
+
 	"development_mode": {
 		Type:         schema.TypeString,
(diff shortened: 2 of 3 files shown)


jacobbednarz:
Can you please split this into two separate PRs? The changes are not related and should be reviewed separately.

riuvshyn:
sure.

riuvshyn:
@jacobbednarz sure, thanks for looking into this
probe_zone change is here: #903

Also can you please comment on my question in description about `visitor_ip`, `orange_to_orange`, `edge_cache_ttl` attributes that returned by API? I am struggling to find any documentation related to these

jacobbednarz:
> Should I also add `visitor_ip`, `orange_to_orange`, `edge_cache_ttl` settings?

I'll confirm about these first two however `edge_cache_ttl` was removed in #653. I'm assuming they are enterprise only features too which will make the pool of who will be using them smaller again.

cloudflare/resource_cloudflare_zone_settings_override_test.go
```diff
@@ -62,6 +62,8 @@ func TestAccCloudflareZoneSettingsOverride_Full(t *testing.T) {
 						name, "settings.0.zero_rtt", "off"),
 					resource.TestCheckResourceAttr(
 						name, "settings.0.universal_ssl", "off"),
+					resource.TestCheckResourceAttr(
```
jacobbednarz, line 65: This test is failing due to the setting not being persisted to state.

```
    testing.go:705: Step 1 error: Check failed: Check 8/8 error: cloudflare_zone_settings_override.test: Attribute 'settings.0.ciphers' not found
--- FAIL: TestAccCloudflareZoneSettingsOverride_Full (21.01s)
FAIL
FAIL	github.com/cloudflare/terraform-provider-cloudflare/cloudflare	21.386s
FAIL
```

Check out the README for running these integration tests if you haven't already. 

Instead of looking for specific values (which I don't think `TypeList` handles too well), just check the number of elements. This line would become `resource.TestCheckResourceAttr(name, "settings.0.ciphers.#", "2")` however if you swap it to that, you'll see it's empty without any elements too.

jacobbednarz (changes requested):
we'll need to address the failing test on this one before merging

cloudflare/resource_cloudflare_zone_settings_override_test.go
riuvshyn: mmm now it fails with:

```
Error: error from makeRequest: HTTP status 400: content "{\"success\":false,\"errors\":[{\"code\":1023,\"message\":\"Advanced Certificate Manager is required to set custom cipher suites\"}],\"messages\":[],\"result\":null}"
```
seems it requires some extra configuration tobe enabled...

elovelan:
@riuvshyn, are you still looking into this? would you like me to take it over since I have a need for this?

jacobbednarz:
@elovelan if you want this functionality, i'd recommend grabbing a new PR and submitting as this one is now stagnent.

PR #904, opened 2021-01-05
Adds import docs for LB pools and monitors

riuvshyn:
Adds import docs for LB pools and monitors

I've actually tested it.

### DIFF
--- a/website/docs/r/load_balancer_monitor.html.markdown
+++ b/website/docs/r/load_balancer_monitor.html.markdown
@@ -78,3 +78,15 @@ The following attributes are exported:
 * `id` - Load balancer monitor ID.
 * `created_on` - The RFC3339 timestamp of when the load balancer monitor was created.
 * `modified_on` - The RFC3339 timestamp of when the load balancer monitor was last modified.
+
+## Import
+
+Load balancer monitor can be imported using a monitor ID
+
+```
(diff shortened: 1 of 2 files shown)


jacobbednarz (changes requested):
While this works (due to passing through the ID and `schema.ImportStatePassthrough`) it's ambiguous as to which level of load balancer resource it should create (user, zone, account). I'm hesitant to document this as doing so as I suspect it will catch people out.

To get this over the line, we'll need to update the `Import` methods to determine which level it is using (probably by including it in the import string identifier) before attempting to import the state.

riuvshyn:
@jacobbednarz are LB pools and monitors can be user/zone levels? from what I can see these are only account level at least  with current state of provider.
I do not see how to create zone specific LB pool or monitor, i see there is an API for user level but where that API is used?

jacobbednarz:
Yep, the CRUD methods call the `cloudflare-go` library which use a helper method (`api.userBaseURL` https://github.com/cloudflare/cloudflare-go/blob/master/load_balancing.go#L160) checking for an attribute on the API client. The Terraform side of things set this when the `account_id` attribute is included in the provider configuration. There is a simpler notice on the registry documentation for load balancer monitors too - https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/load_balancer_monitor

jacobbednarz:
closing due to going stale

PR #905, opened 2021-01-06
Add new resource `cloudflare_argo_tunnel`

### LINKED ISSUE #603: Feature: Add resources for Argo Tunnel
Currently when using argo tunnel one must manually login with `cloudflared login` on each server you want to tunnel with Argo. This prevents using Argo tunnel at scale.

The `cert.pem` file created by `cloudflared login` consists of the CA Origin Cert, Private Key, and an `ARGO TUNNEL TOKEN`. 

The existing resource [origin_ca_certificate](https://www.terraform.io/docs/providers/cloudflare/r/origin_ca_certificate.html) takes care of the origin cert and private key, but we need an additional resource to provision argo tunnel tokens.

### Terraform Version

terraform v0.12.20

jacobbednarz:
Creates a new resource for managing Argo Tunnels.

Depends on cloudflare/cloudflare-go#572 for V4 response structure.

Closes #603

### DIFF
--- a/cloudflare/resource_cloudflare_argo_tunnel.go
+++ b/cloudflare/resource_cloudflare_argo_tunnel.go
@@ -1,0 +1,105 @@
+package cloudflare
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/pkg/errors"
+)
+
+func resourceCloudflareArgoTunnel() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceCloudflareArgoTunnelCreate,
+		Read:   resourceCloudflareArgoTunnelRead,
+		Update: resourceCloudflareArgoTunnelUpdate,
+		Delete: resourceCloudflareArgoTunnelDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceCloudflareArgoTunnelImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"account_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"secret": {
+				Type:      schema.TypeString,
+				Required:  true,
+				Sensitive: true,
+			},
+		},
+	}
+}
+
+func resourceCloudflareArgoTunnelCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	accID := d.Get("account_id").(string)
+	name := d.Get("name").(string)
+	secret := d.Get("secret").(string)
+
+	tunnel, err := client.CreateArgoTunnel(context.Background(), accID, name, secret)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("failed to create Argo Tunnel"))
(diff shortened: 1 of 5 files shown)


cloudflare/resource_cloudflare_argo_tunnel.go
```diff
@@ -0,0 +1,105 @@
...
+
+	return resourceCloudflareArgoTunnelRead(d, meta)
+}
+
+func resourceCloudflareArgoTunnelRead(d *schema.ResourceData, meta interface{}) error {
+	return nil
+}
+
+func resourceCloudflareArgoTunnelUpdate(d *schema.ResourceData, meta interface{}) error {
+	return nil
```
jacobbednarz, line 62: need to do _something_ here to stop in place updates from being a possibility. will have a dig around and see what options we have.

cloudflare/resource_cloudflare_argo_tunnel.go
```diff
@@ -0,0 +1,105 @@
...
+		return errors.Wrap(err, fmt.Sprintf("failed to create Argo Tunnel"))
+	}
+
+	d.SetId(tunnel.ID)
+
+	return resourceCloudflareArgoTunnelRead(d, meta)
+}
+
+func resourceCloudflareArgoTunnelRead(d *schema.ResourceData, meta interface{}) error {
+	return nil
```
jacobbednarz, line 58: resource is read only once created so 🤷‍♂️ nothing to sync

patryk (approved):

website/docs/r/argo_tunnel.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `account_id` - (Required) The Cloudflare account ID that you wish to manage the Argo Tunnel on.
+* `name` - (Required) A user-friendly name chosen when the tunnel is created. Cannot be empty.
+* `secret` - (Required) 32 or more bytes, encoded as a base64 string. The Create Argo Tunnel endpoint sets this as the tunnel's password. Anyone wishing to run the tunnel needs this password.
```
abeluck, line 29: Just a heads up..I'm not sure how the cloudflare-go library handles this, but the HTTP API does not have the requirement that the secret be a base64 string (it might not even have the requirement that it be 32 bytes or more, but all my scripts use 32 bytes or more so I can't say off the top of my head).

When the secret is auto generated by the cloudflared cli tool, it generates a base64 encoded string, but I think that is just their quick way of generating a random string.

I've used long strings generated with password generators with no problem.

jacobbednarz: The requirement is [documented](https://api.cloudflare.com/#argo-tunnel-create-argo-tunnel) (also where I stole this line from) and is enforced by the API schema.

In terms of `cloudflare-go` handles this, it only passes on the string it is sent. It does not check for base64 encoding or the length. Instead, it will error out if the API errors out. Same principle here where we don't actually enforce it and instead offload that to the API itself.

abeluck: Huh, interesting. Thanks for the link @jacobbednarz. Looks like a mistake on our side, however I can report that the CF API (at least before christmas, when I last poked it) doesn't error out if you pass an invalid base64 string. Will update our code though to comply with the docs in the event it starts being enforced.

jacobbednarz:

Issue #738, opened 2020-07-15
Handle '@' in record name and values


innovate-invent:
### Terraform Version
Terraform v0.12.26

### Affected Resource(s)
- cloudflare_record

### Expected Behavior
Automatically resolve '@' to root domain before comparing to state

### Actual Behavior
Marks resource as changed as cloudflare resolves '@' to the root domain in the resulting state.
Resource was not actually changed.

### Steps to Reproduce
Use '@' in the name or value field of a cloudflare_record

jacobbednarz:
Given the following configuration on provider v2.8, I'm unable to reproduce the issue you've raised. 

```hcl
variable "cloudflare_email" {}
variable "cloudflare_api_key" {}
variable "cloudflare_zone_id" {}

provider "cloudflare" {
  email   = var.cloudflare_email
  api_key = var.cloudflare_api_key
}

resource "cloudflare_record" "foo" {
  zone_id = var.cloudflare_zone_id
  name    = "@"
  value   = "test txt value"
  type    = "TXT"
  ttl     = "600"
}
```

```
$ terraform init && terraform apply -var "cloudflare_email=$CF_EMAIL" -var "cloudflare_api_key=$CF_API_KEY" -var "cloudflare_zone_id=$CF_ZONE_ID"

Initializing the backend...

Initializing provider plugins...

The following providers do not have any version constraints in configuration,
so the latest version was installed.

To prevent automatic upgrades to new major versions that may contain breaking
changes, it is recommended to add version = "..." constraints to the
corresponding provider blocks in configuration, with the constraint strings
suggested below.

* provider.cloudflare: version = "~> 2.8"

Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # cloudflare_record.foo will be created
  + resource "cloudflare_record" "foo" {
      + created_on  = (known after apply)
      + hostname    = (known after apply)
      + id          = (known after apply)
      + metadata    = (known after apply)
      + modified_on = (known after apply)
      + name        = "@"
      + proxiable   = (known after apply)
      + proxied     = false
[...]

PR #909, opened 2021-01-11
Test @ as record value

innovate-invent:
This PR adds a test that demonstrates #738


jacobbednarz:
Closing per my feedback in #738.

PR #910, opened 2021-01-12
Add support for all allowed session_duration values

Justin-Holmes:
We're currently limited in what values we can use for session duration; however, the API allows any value that is parsable from https://golang.org/pkg/time/#ParseDuration.

Cloudflare's API documentation will be updated with these changes shortly.


jacobbednarz:
Thanks! 🍭

jacobbednarz:

PR #911, opened 2021-01-13
Adds rate limit match.response.header attribute

riuvshyn:
Adds rate limit response.headers attribute

### DIFF
--- a/cloudflare/resource_cloudflare_rate_limit.go
+++ b/cloudflare/resource_cloudflare_rate_limit.go
@@ -526,7 +526,15 @@ func flattenRateLimitResponseMatcher(cfg cloudflare.RateLimitResponseMatcher) []
 	}
 
 	if len(cfg.Headers) > 0 {
-		data["headers"] = cfg.Headers
+		headers := make([]map[string]interface{}, len(cfg.Headers))
+		for i, header := range cfg.Headers {
+			headers[i] = map[string]interface{}{
+				"name":  header.Name,
+				"op":    header.Op,
+				"value": header.Value,
+			}
+		}
+		data["header"] = headers
 	}
 
 	if len(data) > 0 {
--- a/cloudflare/resource_cloudflare_rate_limit.go
+++ b/cloudflare/resource_cloudflare_rate_limit.go
@@ -147,6 +147,30 @@ func resourceCloudflareRateLimit() *schema.Resource {
 										Optional: true,
 										Computed: true,
 									},
+
+									"header": {
+										Type:     schema.TypeSet,
+										Optional: true,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"name": {
+													Type:     schema.TypeString,
+													Required: true,
+												},
+
+												"op": {
+													Type:         schema.TypeString,
+													Required:     true,
+													ValidateFunc: validation.StringInSlice([]string{"eq", "ne"}, false),
+												},
+
+												"value": {
+													Type:     schema.TypeString,
+													Required: true,
+												},
+											},
+										},
+									},
 								},
 							},
 						},
@@ -347,6 +371,20 @@ func expandRateLimitTrafficMatcher(d *schema.ResourceData) (matcher cloudflare.R
 			originTraffic := originIface.(bool)
 			responseMatcher.OriginTraffic = &originTraffic
 		}
+
+		if headersSet, ok := matchResp["header"]; ok {
+			headersArray := make([]cloudflare.RateLimitResponseMatcherHeader, headersSet.(*schema.Set).Len())
+			for i, entry := range headersSet.(*schema.Set).List() {
+				e := entry.(map[string]interface{})
+				headersArray[i] = cloudflare.RateLimitResponseMatcherHeader{
+					Name:  e["name"].(string),
+					Op:    e["op"].(string),
+					Value: e["value"].(string),
+				}
+			}
+			responseMatcher.Headers = headersArray
+		}
+
 		matcher.Response = responseMatcher
 	}
 	return
@@ -485,6 +523,10 @@ func flattenRateLimitResponseMatcher(cfg cloudflare.RateLimitResponseMatcher) []
 
 	if len(cfg.Statuses) > 0 {
 		data["statuses"] = schema.NewSet(IntIdentity, flattenIntList(cfg.Statuses))
+	}
+
+	if len(cfg.Headers) > 0 {
+		data["headers"] = cfg.Headers
 	}
 
 	if len(data) > 0 {
--- a/cloudflare/resource_cloudflare_rate_limit.go
+++ b/cloudflare/resource_cloudflare_rate_limit.go
@@ -148,28 +148,10 @@ func resourceCloudflareRateLimit() *schema.Resource {
 										Computed: true,
 									},
 
-									"header": {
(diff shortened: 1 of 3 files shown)


jacobbednarz (changes requested):
This is a good start but `headers` needs to accept the possibility of multiple HTTP header combinations. See the `cloudflare/cloudflare-go` struct: https://github.com/cloudflare/cloudflare-go/blob/a71461723e375f9fb4126b5d973b2619bb010f27/rate_limiting.go#L41

An example might be:

```hcl
resource "cloudflare_rate_limit" "example" {
  zone_id = var.cloudflare_zone_id
  # .. snip
  match {
    request {
      # .. snip
    }
    response {
      statuses = [200, 201, 202, 301, 429]
      origin_traffic = false
      headers = [
        {
          name  = "Host"
          op    = "eq"
          value = "localhost"
	      },
        {
          name  = "X-Example"
          op    = "neq"
          value = "my-value"
	      }
      ]
    }
  }
  # .. snip
}
```

riuvshyn:
@jacobbednarz oh! you absolutely right. initial intention was to make it as TypeSet which should allow to put multiple `headers{}` but then I was sidetracked and missed this. Will fix.

jacobbednarz:
The multiple `header` blocks don't feel right here. The API and dashboard represent them as a single construct but we're defining our own representation here. A map of objects is more annoying to work with internally for Terraform but it does allow us to map (no pun intended) to the existing API structure. Thoughts?

riuvshyn:
@jacobbednarz I was basically following zone_lockdown implementation with `configurations{}` and I find it pretty handy.

jacobbednarz:
I don't disagree that it's easier to use in the right cases; the issue here is that it doesn't match the API or dashboard representation which can cause confusion for those trying to simulate what they have already created in the other methods.

riuvshyn:
@jacobbednarz mmhmm for some reasons I can't get schema like 
```hcl
      headers = [
        {
          name  = "Host"
          op    = "eq"
          value = "localhost"
	},
      ]
```
to work... tf does not like it or maybe I am missing something... do you have any example of such schema implemented?
I can't find any List of maps example anywhere...

my schema looks like:

```
	"headers": {
		Type:     schema.TypeList,
		Optional: true,
		Elem: &schema.Resource{
	.....
	}
```

I am getting such error: 

```
headers = [

An argument named "headers" is not expected here. Did you mean to define a
block of type "headers"?
```

Any suggestion?

jacobbednarz:
check out `cloudflare_magic_firewall_ruleset`. there is a `rules` schema object which serves as a good example.

riuvshyn:
@jacobbednarz oki, I;ve just pushed changes, but I;ve noticed also that it is going to create `Cf-Cache-Status` even if it is not declared in TF. How do you think it should be handled?

jacobbednarz:
hmm 🤔 for now, it should sync to the state file (via the `Read` method) but let me check on why that is being set. IIRC, you can create rules without it and it shouldn't appear there.

riuvshyn:
@jacobbednarz it is coming from somewhere by default and it is not persisted in state so TF plan shows that it is going tobe deleted:

```
          ~ response {
              ~ headers        = [
                    {
                        "name"  = "Host"
                        "op"    = "eq"
                        "value" = "localhost"
                    },
                  - {
                      - "name"  = "Cf-Cache-Status"
                      - "op"    = "ne"
                      - "value" = "HIT"
                    },
                ]

```

jacobbednarz:
Looking at this with fresh eyes, it makes sense. If you're conditionally rate limiting a request, you need to be excluding the cached entries (`cf-cache-status ne "HIT"`). To apply caching to the cached resources, the `response.headers` section needs to be empty and `origin_traffic` set to false.

```
$ curl 'https://api.cloudflare.com/client/v4/zones/$ZONE_ID/rate_limits' \
  -H 'Authorization: $CF_API_TOKEN' \ 
  -H 'content-type: application/json' \
  --data-raw '{"match":{"request":{"methods":["_ALL_"],"schemes":["_ALL_"],"url":"example.com"},"response":{"origin_traffic":false,"headers":[]}},"action":{"mode":"ban","response":{"content_type":"default"},"timeout":3600},"login_protect":false,"description":"aaaa","threshold":10,"period":60,"disabled":false}'
```

riuvshyn:
@jacobbednarz ok so how should I handle that in provider? If I do nothing after apply next plan TF will want to delete `cf-cache-status` header because it is not declared in TF code, should I leave it as this or just skip this header in `read` method?

jacobbednarz:
As far as I can tell, this is expected and the lack of including the Cf-Cache-Status is a user error when the `origin_traffic` attribute isn't defined.

riuvshyn:
@jacobbednarz so I've updated PR does it makes sense now?

jacobbednarz:
All of the integration tests are passing and this looks good!

```
=== RUN   TestAccCloudflareRateLimit_Import
=== PAUSE TestAccCloudflareRateLimit_Import
=== RUN   TestAccCloudflareRateLimit_Basic
=== PAUSE TestAccCloudflareRateLimit_Basic
=== RUN   TestAccCloudflareRateLimitChallenge_Basic
=== PAUSE TestAccCloudflareRateLimitChallenge_Basic
=== RUN   TestAccCloudflareRateLimit_FullySpecified
=== PAUSE TestAccCloudflareRateLimit_FullySpecified
=== RUN   TestAccCloudflareRateLimit_Update
=== PAUSE TestAccCloudflareRateLimit_Update
=== RUN   TestAccCloudflareRateLimit_CreateAfterManualDestroy
=== PAUSE TestAccCloudflareRateLimit_CreateAfterManualDestroy
=== RUN   TestAccCloudflareRateLimit_WithoutTimeout
=== PAUSE TestAccCloudflareRateLimit_WithoutTimeout
=== RUN   TestAccCloudflareRateLimit_ChallengeWithTimeout
=== PAUSE TestAccCloudflareRateLimit_ChallengeWithTimeout
=== CONT  TestAccCloudflareRateLimit_Import
=== CONT  TestAccCloudflareRateLimit_CreateAfterManualDestroy
=== CONT  TestAccCloudflareRateLimitChallenge_Basic
=== CONT  TestAccCloudflareRateLimit_FullySpecified
=== CONT  TestAccCloudflareRateLimit_ChallengeWithTimeout
=== CONT  TestAccCloudflareRateLimit_Update
=== CONT  TestAccCloudflareRateLimit_WithoutTimeout
=== CONT  TestAccCloudflareRateLimit_Basic
--- PASS: TestAccCloudflareRateLimit_WithoutTimeout (0.14s)
--- PASS: TestAccCloudflareRateLimit_ChallengeWithTimeout (0.15s)
--- PASS: TestAccCloudflareRateLimit_Basic (4.69s)
--- PASS: TestAccCloudflareRateLimitChallenge_Basic (5.25s)
--- PASS: TestAccCloudflareRateLimit_FullySpecified (5.85s)
--- PASS: TestAccCloudflareRateLimit_Import (6.24s)
--- PASS: TestAccCloudflareRateLimit_CreateAfterManualDestroy (7.90s)
--- PASS: TestAccCloudflareRateLimit_Update (8.85s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/cloudflare	9.411s
```

Thank you!

jacobbednarz:

PR #912, opened 2021-01-13
Update access application documentation

Justin-Holmes:
Having a negative session duration doesn't make sense even though it's _technically_ considered a valid value. Better to remove that example and avoid the confusion altogether.


jacobbednarz:
👍

PR #921, opened 2021-01-26
cloudflare_access_rule: add validation to discover errors during planning phase

freddierice:
The goal of this PR is to discover preventable mistakes during a terraform plan rather than an apply. Specifically, preventing developers from attempting to apply non /24 and /32 `ip_range`s, resulting in partial changes and leaving them exposed. 

I didn't create an issue because writing the code took about the same time as describing the problem. I'm open to other solutions.

### DIFF
--- a/cloudflare/resource_cloudflare_access_rule.go
+++ b/cloudflare/resource_cloudflare_access_rule.go
@@ -290,7 +290,7 @@ func validateAccessRuleConfigurationIPRange(v string) (warnings []string, errors
 
 	if ip.To4() != nil {
 		ones, _ := ipNet.Mask.Size()
-		if ones != 24 && ones != 32 {
+		if ones != 16 && ones != 24 {
 			errors = append(errors, fmt.Errorf("ip_range with ipv4 address must be a /24 or /32, got a /%d", ones))
 			return warnings, errors
 		}
--- a/cloudflare/resource_cloudflare_access_rule.go
+++ b/cloudflare/resource_cloudflare_access_rule.go
@@ -3,6 +3,7 @@ package cloudflare
 import (
 	"fmt"
 	"log"
+	"net"
 	"strings"
 
 	cloudflare "github.com/cloudflare/cloudflare-go"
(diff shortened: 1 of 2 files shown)


jacobbednarz:
I don't mind the change, however, the test included here doesn't excercise the code path it intends to.

With the schema validation commented out, it still passes.

```diff
diff --git cloudflare/resource_cloudflare_access_rule.go cloudflare/resource_cloudflare_access_rule.go
index 910c84d..4ddd5a2 100644
--- cloudflare/resource_cloudflare_access_rule.go
+++ cloudflare/resource_cloudflare_access_rule.go
@@ -42,7 +42,7 @@ func resourceCloudflareAccessRule() *schema.Resource {
                                Required:         true,
                                ForceNew:         true,
                                DiffSuppressFunc: configurationDiffSuppress,
-                               ValidateFunc:     validateAccessRuleConfiguration,
+                               // ValidateFunc:     validateAccessRuleConfiguration,
                                Elem: &schema.Resource{
                                        Schema: map[string]*schema.Schema{
                                                "target": {
```

```
=== RUN   TestValidateAccessRuleConfigurationIPRange
--- PASS: TestValidateAccessRuleConfigurationIPRange (0.00s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/cloudflare	0.471s
```

cloudflare/resource_cloudflare_access_rule.go
```diff
@@ -258,3 +260,47 @@ func configurationDiffSuppress(k, old, new string, d *schema.ResourceData) bool
...
+	}
+
+	if ipNet == nil {
+		errors = append(errors, fmt.Errorf("ip_range must hold a range"))
+		return warnings, errors
+	}
+
+	if ip.To4() != nil {
+		ones, _ := ipNet.Mask.Size()
+		if ones != 24 && ones != 32 {
```
jacobbednarz, line 293: Shouldn't this be `if ones != 16 && ones != 24 {`? From the docs

> Limited to /16 and /24 for IPv4 and /32, /48 and /64 for IPv6.

freddierice: Yes! Good catch.

freddierice:
Thanks for the review! I fixed the bitmask bug and added a new test to make sure that commenting out the validator will result in a failed test.

jacobbednarz:
Nice one, thanks for this! 🥇

jacobbednarz:

PR #930, opened 2021-02-02
test/cloudflare_magic_firewall_ruleset: skip CI tests for default zone

jacobbednarz:
Like BYOIP, Magic Transit doesn't really lend itself to the typical CI
setup where we create and destroy resources on the fly. As such, we need
to update the test suite to work accordingly. The tests are very
valuable and we want them around however we shouldn't run them on a zone
we know isn't configured (in this case, able to be configured to do so).

This updates the acceptance tests to skip when the default zone ID is in
use however will allow everyone else to run the tests like normal.

Fixes https://github.com/cloudflare/terraform-provider-cloudflare/runs/1810568088?check_suite_focus=true#step:6:317

### DIFF
--- a/cloudflare/provider_test.go
+++ b/cloudflare/provider_test.go
@@ -122,3 +122,13 @@ func testAccPreCheckBYOIPPrefix(t *testing.T) {
 func generateRandomResourceName() string {
 	return acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
 }
+
+// skipMagicTransitTestForNonConfiguredDefaultZone will force an acceptance test
+// to skip instead of running and failing due to not having setup Magic Transit.
+// This will allow those who intentionally want to run the test to do so while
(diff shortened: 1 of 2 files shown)


cloudflare/provider_test.go
```diff
@@ -9,8 +9,23 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
 )
 
-var testAccProviders map[string]terraform.ResourceProvider
-var testAccProvider *schema.Provider
+var (
+	testAccProviders map[string]terraform.ResourceProvider
+	testAccProvider  *schema.Provider
+
+	// Integration test account ID
```
jacobbednarz, line 16: these are copied out of the GitHub workflow where these are defined.

jacobbednarz:
more of a FYI for you @cehrig but also open to your thoughts if you have any

Issue #878, opened 2020-12-01
Terraform resource for uploading Access Mutual TLS Root Certificates


jonathanjuursema:
### Current Terraform version

`0.12.29` at the time of writing, but we're going to upgrade to `0.13.x` as soon as we can. :)

### Description

As far as I can tell (and please do tell me if I've missed something) there is no support for uploading Mutual TLS Root Certificates even though there seem to be [API endpoint](https://api.cloudflare.com/#zone-level-access-mutual-tls-authentication-properties) for doing so.

Are there plans for supporting a `cloudflare_access_mutual_tls_root_certificate` resource? Or a reason I've overlooked for why this would be hard to achieve?

### Use-cases

I'm developing on an in-house deployment pipeline where internal teams can deploy infrastructure (mainly AWS, Cloudflare) using off-the-shelve "modules" we developed. These modules consist mainly of CloudFormation, Terraform and Kubernetes templates.

We already have a "module" that helps teams to deploy most of their Access configuration on Cloudflare. However, we have several customers that require us to support Mutual TLS. Right now, uploading Mututal TLS Root Certificates (for use in conjunction with `cloudflare_access_policy` resources) is a manual action in the Cloudflare dashboard that we would love for our teams to be able to do themselves. (Teams have read-only permission in the Cloudflare dashboard, and are supposed to deploy all their infrastructure via this pipeline.)

### Potential Terraform configuration

Something like this would do the job just fine for us:

```hcl
resource "cloudflare_access_mutual_tls_root_certificate" "some_certificate" {
  zone_id              = "9a7806061c88ada191ed06f989cc3dac"
  name                 = "Some mTLS Root Certificate"
  certificate          = "-----BEGIN CERTIFICATE-----\nMIIGAjCCA+qgAwIBAgIJAI7kymlF7CWT...N4RI7KKB7nikiuUf8vhULKy5IX10\nDrUtmu/B\n-----END CERTIFICATE-----"
  associated_hostnames = [
    "mtls.example.com"
  ]
}
```

### References

I've tried looking for similar feature requests but couldn't find any. My apologies if I overlooked some.

Cloudflare API: https://api.cloudflare.com/#zone-level-access-mutual-tls-authentication-properties

### Community note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
  to the original issue to help the community and maintainers prioritize this request
[...]

Issue #929, opened 2021-02-01
terraform import of zone specific access policy does not work


iserko:
### Terraform version

0.13.5

Cloudflare provider version 2.14.0

### Affected resource(s)

- cloudflare_access_policy

### Terraform configuration files

```hcl
resource "cloudflare_access_policy" "allow_policy" {
  application_id = cloudflare_access_application.my_app.id
  decision       = "allow"
  zone_id        = var.cf_zone_id
  name           = "Name of policy"
  precedence     = "1"

  include {
    group = [
      cloudflare_access_group.my_group.id
    ]
  }
}
```

### Debug output

https://gist.github.com/iserko/f45ea0153f2c913a8cd4bd68ff6dce0b

### Panic output

No panic

### Expected behavior

```
terraform import cloudflare_access_policy.allow_policy ZONE_ID/APPLICATION_ID/POLICY_ID
```

should work as described in https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/access_policy#import

> Access Policies can be imported using a composite ID formed of zone ID, application ID and policy ID.

### Actual behavior

Terraform provider attempts to import `ACCOUNT_ID/APPLICATION_ID/POLICY_ID` and fails doing so as that POLICY_ID does not exist in the account, only in the zone.

### Steps to reproduce

1. create an access policy in a zone
2. attempt to import it

### Important factoids

n/a

### References

n/a

### Community note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
  to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull
  request, please leave a comment

mdolinin:
Was on my way to open similar bug. That doc needs to be fixed, import works for me only with `ACCOUNT_ID/APPLICATION_ID/POLICY_ID`. But my policy was created via UI few months ago.

jacobbednarz:
Looks to be left over from when we attempted to move to account only access policies. To fix this, we'll need to update the identifier to include "zone" or "account". 

@iserko are you interested in submitting a PR for this?

PR #951, opened 2021-02-11
Add Access Mutual TLS certificate resource

Justin-Holmes:
Closes https://github.com/cloudflare/terraform-provider-cloudflare/issues/878


cloudflare/resource_cloudflare_access_mutual_tls_certificate.go
```diff
@@ -0,0 +1,195 @@
...
+		accessMutualTLSCert, err = client.ZoneAccessMutualTLSCertificate(identifier.Value, d.Id())
+	}
+
+	if err != nil {
+		if strings.Contains(err.Error(), "HTTP status 404") {
+			log.Printf("[INFO] Access Mutal TLS Certificate %s no longer exists", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error finding Access Mutual TLS Certificate %q: %s", d.Id(), err)
```
jacobbednarz, line 110: let's lowercase the exception message to align with Go's conventions

```suggestion
		return fmt.Errorf("error finding Access Mutual TLS Certificate %q: %s", d.Id(), err)
```

cloudflare/resource_cloudflare_access_mutual_tls_certificate.go
```diff
@@ -0,0 +1,195 @@
...
+	} else {
+		accessMutualTLSCert, err = client.UpdateZoneAccessMutualTLSCertificate(identifier.Value, d.Id(), updatedAccessMutualTLSCert)
+	}
+	if err != nil {
+		return fmt.Errorf("error updating Access Mutual TLS Certificate for %s %q: %s", identifier.Type, identifier.Value, err)
+	}
+
+	if accessMutualTLSCert.ID == "" {
+
+	}
```
jacobbednarz, line 148: i don't _think_ we need this for anything in the `Update`

cloudflare/resource_cloudflare_access_mutual_tls_certificate.go
```diff
@@ -0,0 +1,195 @@
...
+	if identifier.Type == AccountType {
+		err = client.DeleteAccessMutualTLSCertificate(identifier.Value, certID)
+	} else {
+		err = client.DeleteZoneAccessMutualTLSCertificate(identifier.Value, certID)
+	}
+	if err != nil {
+		return fmt.Errorf("error deleting Access Mutual TLS Certificate for %s %q: %s", identifier.Type, identifier.Value, err)
+	}
+
+	resourceCloudflareAccessMutualTLSCertificateRead(d, meta)
```
jacobbednarz, line 173: for a `Delete` we don't usually call the `Read` but instead just `d.SetId("")` as we don't have a need to perform another HTTP GET to attempt to retrieve the state; instead, just remove it. i see you are handling the missing resource in the `Read` as well.

cloudflare/resource_cloudflare_access_mutual_tls_certificate.go
```diff
@@ -0,0 +1,195 @@
...
+
+	if len(attributes) != 2 {
+		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"accountID/accessMutualTLSCertificateID\"", d.Id())
+	}
+
+	accountID, accessMutualTLSCertificateID := attributes[0], attributes[1]
+
+	log.Printf("[DEBUG] Importing Cloudflare Access Mutual TLS Certificate: id %s for account %s", accessMutualTLSCertificateID, accountID)
+
+	d.Set("account_id", accountID)
```
jacobbednarz, line 189: we have the ability to set the zone _and_ account - should we add another identitifer to the import ID to help differentiate which to use? i.e. `account/abc123/cert123` and `zone/def456/cert123`?

cloudflare/resource_cloudflare_access_mutual_tls_certificate_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+)
+
+func TestAccCloudflareAccessMutualTLSBasic(t *testing.T) {
+	rnd := generateRandomResourceName()
+	name := fmt.Sprintf("cloudflare_access_mutual_tls_certificate.%s", rnd)
+	cert := os.Getenv("CLOUDFLARE_CERTIFICATE")
```
jacobbednarz, line 16: as we've got a few certificate things in the test suite, would it make sense to be more specific with the naming?
```suggestion
	cert := os.Getenv("CLOUDFLARE_MUTUAL_TLS_CERTIFICATE")
```

website/docs/r/access_mutual_tls_certificate.html.markdown
```diff
@@ -0,0 +1,49 @@
...
+
+* `id` - ID of the Access Mutual TLS Certificate resource
+
+## Import
+
+Access Mutual TLS Certificate can be imported using a composite ID formed of the zone
+ID and the mutual TLS certificate ID.
+
+```
+$ terraform import cloudflare_access_mutual_tls_certificate.staging cb029e245cfdd66dc8d2e570d5dd3322/d41d8cd98f00b204e9800998ecf8427e
```
jacobbednarz, line 48: this will be pending my import ID value above

jacobbednarz (commented):
looks like we're 99% of the way there, just some nitpicks and questions from me! 🤜 🤛

cloudflare/resource_cloudflare_access_mutual_tls_certificate.go
Justin-Holmes: Ah, I think I forgot to clean this up. I noticed we did this in a few places such as https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/resource_cloudflare_access_application.go#L278-L280 and https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/resource_cloudflare_access_policy.go#L177-L179 but I don't think it's an actual concern.

cloudflare/resource_cloudflare_access_mutual_tls_certificate.go
Justin-Holmes: Yeah, good call. It looks like we need to do this with all of our Access resources -- should we do that in a separate PR?

jacobbednarz: I would fix this resource in this PR but the others are fine to go in another PR. We have #929 to track it as well - I just haven't got it to it just yet :p

jacobbednarz:
just need to add a usable mutual TLS certificate to the [`env`](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/.github/workflows/acceptance-tests.yml#L13) and we're good to merge. as it is not secret and actually in use, we might not need to store this as a secret.

PR #969, opened 2021-02-26
Fixing typo on cloudflare_zones data resource documentation

Asteao:
Changing example resource from `cloudflare_zone` to `cloudflare_record` as all the attributes of the resource match a record. I believe this is a typo.


jacobbednarz:
thanks @Asteao! 👏

PR #970, opened 2021-02-28
API Token error message matches validator regex.

bryson:
Signed-off-by: Bryson Holland <commits@bryson.io>


bryson:
I've never been more offended by a code review. 🤣

Issue #979, opened 2021-03-07
Expose "default_mode" of WAF Rule through cloudflare_waf_rules data source


JoseRoman:
### Current Terraform version
```
Terraform v0.14.5
cloudflare v2.18.0
```

### Description

I would like to have access to the default mode of a WAF rule via the cloudflare_waf_rules data source.

### Use-cases
Being able to modify the mode of 100+ rules via Terraform and leaving the ones that default to "Disabled" untouched requires having visibility into the "default_mode" of each rule through the data source.

### References

I have not found any other issues similar to this. 

I did notice that this is currently available in the cloudflare-go library
https://github.com/cloudflare/cloudflare-go/blob/master/waf.go#L81

### Community note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
  to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull
  request, please leave a comment

jacobbednarz:
please check out #1079 for this one. i've put up the PR and i'd recommend you build the provider locally (docs https://github.com/cloudflare/terraform-provider-cloudflare#building-the-provider) and test it out before it is merged.

PR #980, opened 2021-03-07
Exposing default_mode of WAF Rule through cloudflare_waf_rules data source

JoseRoman:
#979

### DIFF
--- a/cloudflare/data_source_waf_rules_test.go
+++ b/cloudflare/data_source_waf_rules_test.go
@@ -52,6 +52,27 @@ func TestAccCloudflareWAFRules_MatchDescription(t *testing.T) {
 	})
 }
 
+func TestAccCloudflareWAFRules_DefaultMode(t *testing.T) {
+	t.Parallel()
+	zoneID := os.Getenv("CLOUDFLARE_ZONE_ID")
+	rnd := generateRandomResourceName()
+	name := fmt.Sprintf("data.cloudflare_waf_rules.%s", rnd)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCloudflareWAFRulesConfig(zoneID, map[string]string{}, rnd),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCloudflareWAFRulesDataSourceID(name),
+					resource.TestCheckResourceAttrSet(name, "default_mode"),
+				),
+			},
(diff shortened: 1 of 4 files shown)


jacobbednarz:
seems reasonable 👍 can you please add test coverage for this new attribute?

JoseRoman:
@jacobbednarz Of course, I can add some tests for this.

- What is the process of running the acceptance tests locally? Do I have to create a test Cloudflare account to be able to run them? 
- This is a read-only field (similar to `allowed_modes`) that is not being filtered, do we still write tests for these?

jacobbednarz:
> What is the process of running the acceptance tests locally? Do I have to create a test Cloudflare account to be able to run them?

There is a section in the README on [developing the provider](https://github.com/cloudflare/terraform-provider-cloudflare#developing-the-provider) which will step you through testing. Yes, you'll need a Cloudflare account as the tests create real resources.

> This is a read-only field (similar to allowed_modes) that is not being filtered, do we still write tests for these?

https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/data_source_waf_rules_test.go#L13 is an example of an existing `data_source_waf_rules` test which you could copy, tweak and confirm your new fields are usable (either through confirming them being written to the state file or in another resource).

cloudflare/data_source_waf_rules_test.go
```diff
@@ -52,6 +52,27 @@ func TestAccCloudflareWAFRules_MatchDescription(t *testing.T) {
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCloudflareWAFRulesConfig(zoneID, map[string]string{}, rnd),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCloudflareWAFRulesDataSourceID(name),
+					resource.TestCheckResourceAttrSet(name, "rules.#"),
```
jacobbednarz, line 69: i'm afraid this assertion doesn't do assert anything useful for us here. this test is identical to some others and doesn't actually cover the functionality you've added. 

perhaps have a look at `testAccCheckCloudflareWAFRulesDataSourceID` and i suspect you'll need to programmatically loop through the resource rules and confirm the key is present in the state following this change.

jacobbednarz (changes requested):
we need to adjust the test case for this because as it stands, it is a duplicate of an existing test with a different name and doesn't cover the added fields being stored in state.

jacobbednarz:
closing in favour of #1079

PR #987, opened 2021-03-11
access_mutual_tls_certificate: fix tests and deletion

jacobbednarz:
Fixes the `Delete` operation to first remove the associated hostnames
and then blow away the certificate.


cloudflare/resource_cloudflare_access_mutual_tls_certificate.go
```diff
@@ -157,18 +159,45 @@ func resourceCloudflareAccessMutualTLSCertificateDelete(d *schema.ResourceData,
...
+		_, err = client.UpdateZoneAccessMutualTLSCertificate(context.Background(), identifier.Value, d.Id(), deletedCertificate)
 	}
+
 	if err != nil {
-		return fmt.Errorf("error deleting Access Mutual TLS Certificate for %s %q: %s", identifier.Type, identifier.Value, err)
+		return fmt.Errorf("error updating Access Mutual TLS Certificate for %s %q: %s", identifier.Type, identifier.Value, err)
 	}
 
-	d.SetId("")
+	return resource.Retry(d.Timeout(schema.TimeoutDelete), func() *resource.RetryError {
```
jacobbednarz, line 181: We implement a `Retry` here in the event that the service has lag removing the associations in order to check again after a small wait.

PR #999, opened 2021-03-19
Add firewall_rules data source

### LINKED ISSUE #997: Add Data Source for firewall_rule
### Current Terraform version

Terraform v0.13.6

### Description

Add a Data Source to query existing firewall rules.

### Use cases

When adding new firewall rules it is necessary to know which rules already exist to provide the correct priority to new rules.

```hcl
data "cloudflare_firewall_rules" "example" {
  zone_id = "<ZONE_ID>"
  filter {
    priority = 5
    match = "gte"
    paused = false
  }
}
```

Attribute reference:
- `firewall_rules`

`firewall_rules`:
- `action`
- `priority`
- `paused`

### References

_No response_

ldb:
This change adds a new data source `firewall_rules` which can query existing firewall rules based on their `paused` and `priority` attributes.

fixes #997

### DIFF
--- a/cloudflare/data_source_firewall_rule.go
+++ b/cloudflare/data_source_firewall_rule.go
@@ -1,0 +1,161 @@
+package cloudflare
+
+import (
+	"context"
+	"fmt"
+	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+)
+
+func dataSourceCloudflareFirewallRules() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceCloudflareFirewallRulesRead,
+
+		Schema: map[string]*schema.Schema{
+			"zone_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"filter": {
+				Type:     schema.TypeList,
+				Required: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"priority": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							ValidateFunc: validation.IntBetween(1, 2147483647),
+						},
+						"match_type": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							ValidateFunc: validation.StringInSlice([]string{"lt", "lte", "eq", "gte", "gt"}, false),
+						},
+						"paused": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
+					},
+				},
+			},
+
+			"firewall_rules": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"action": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"priority": {
+							Type:     schema.TypeInt,
+							Optional: true,
+						},
+						"paused": {
+							Type:     schema.TypeBool,
+							Optional: true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func dataSourceCloudflareFirewallRulesRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	filter, err := expandFilterFirewallRules(d.Get("filter"))
+	if err != nil {
+		return err
+	}
+
+	zoneID := d.Get("zone_id").(string)
+	rules, err := client.FirewallRules(context.Background(), zoneID, cloudflare.PaginationOptions{})
+	if err != nil {
+		return fmt.Errorf("error listing Firewall Rules: %s", err)
+	}
+
+	filteredRules := make([]map[string]interface{}, 0)
+	for _, v := range rules {
+		if filter.paused != v.Paused {
+			continue
+		}
+
+		rulePrio, ok := v.Priority.(int)
+		if !ok {
+			rulePrio = 2147483647
+		}
+		if !filter.matchPriority(rulePrio) {
+			continue
+		}
+
+		filteredRules = append(filteredRules, map[string]interface{}{
+			"id":       v.ID,
+			"action":   v.Action,
+			"priority": v.Priority,
+			"paused":   v.Paused,
+		})
+	}
+
+	err = d.Set("firewall_rules", filteredRules)
+	if err != nil {
+		return fmt.Errorf("error setting Firewall Rules: %s", err)
+	}
+
+	return nil
+}
+
+func expandFilterFirewallRules(d interface{}) (*firewallRulesFilter, error) {
+	cfg := d.([]interface{})
+	filter := &firewallRulesFilter{}
+
+	m := cfg[0].(map[string]interface{})
+	priority, ok := m["priority"]
+	if ok {
+		filter.priority = priority.(int)
+	}
+
+	matchType, ok := m["match_type"]
+	if ok {
+		filter.matchType = matchType.(string)
+	}
+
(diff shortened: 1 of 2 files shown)


jacobbednarz:
On hold pending use case discussion in #997.

jacobbednarz (changes requested):
On hold pending use case discussion in #997.

cloudflare/data_source_firewall_rule_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+)
+
+func TestAccCloudflareFirewallRulesMatchPaused(t *testing.T) {
```
jacobbednarz, line 11: These tests do not pass as is.

```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareFirewallRulesMatch" -timeout 120m -parallel 1
=== RUN   TestAccCloudflareFirewallRulesMatchPaused
=== PAUSE TestAccCloudflareFirewallRulesMatchPaused
=== RUN   TestAccCloudflareFirewallRulesMatchPriorityFilter
=== PAUSE TestAccCloudflareFirewallRulesMatchPriorityFilter
=== CONT  TestAccCloudflareFirewallRulesMatchPaused
    testing.go:705: Step 0 error: config is invalid: 4 problems:

        - Missing required argument: The argument "filter_id" is required, but no definition was found.
        - Missing required argument: The argument "filter_id" is required, but no definition was found.
        - Invalid data source: The provider provider.cloudflare does not support data source "cloudflare_firewall_rules".
        - Missing required argument: The argument "filter_id" is required, but no definition was found.
=== CONT  TestAccCloudflareFirewallRulesMatchPriorityFilter
--- FAIL: TestAccCloudflareFirewallRulesMatchPaused (0.02s)
=== CONT  TestAccCloudflareFirewallRulesMatchPriorityFilter
    testing.go:705: Step 0 error: config is invalid: 4 problems:

        - Invalid data source: The provider provider.cloudflare does not support data source "cloudflare_firewall_rules".
        - Missing required argument: The argument "filter_id" is required, but no definition was found.
        - Missing required argument: The argument "filter_id" is required, but no definition was found.
        - Missing required argument: The argument "filter_id" is required, but no definition was found.
--- FAIL: TestAccCloudflareFirewallRulesMatchPriorityFilter (0.02s)
FAIL
FAIL	github.com/cloudflare/terraform-provider-cloudflare/cloudflare	0.687s
FAIL
make: *** [testacc] Error 1
```

cloudflare/data_source_firewall_rule_test.go
```diff
@@ -0,0 +1,122 @@
...
+	t.Parallel()
```
jacobbednarz, line 12: You don't want to run these in parallel as they create real resources which if they overlap will taint the expected results.

cloudflare/data_source_firewall_rule_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+func testAccCloudflareFirewallRulesConfigMatchPriority(rnd string) string {
+	return fmt.Sprintf(`
+data "cloudflare_firewall_rules" "%[2]s" {
+  filter {
+    priority   = 2
+	match_type = "gte"
+    // This is an ordering fix to ensure that the test suite doesn't assert
+    // state before all the resources are available.
+    paused = "${cloudflare_firewall_rule.baa_net.paused}"
```
jacobbednarz, line 88: this looks like it has been copy/pasted from the cloudflare zones. can you please update the resource names to match what they are targetting here?

cloudflare/data_source_firewall_rule_test.go
```diff
@@ -0,0 +1,122 @@
...
+  }
+}
+
+%[1]s
+`, testRules, rnd)
+}
+
+const testRules = `resource "cloudflare_firewall_rule" "baa_com" {
```
jacobbednarz, line 96: similar to my comment above about renaming these appropriately but you're also missing `filters` here which firewall rules require as a reference via `filter_id`.

cloudflare/data_source_firewall_rule_test.go
```diff
@@ -0,0 +1,122 @@
...
+  zone_id  = "0da42c8d2132a9ddaf714f9e7c920711"
```
jacobbednarz, line 97: `zone_id` can be pointed at [`testAccCloudflareZoneID`](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/provider_test.go#L20) instead of a static value now that we have that defined.

cloudflare/data_source_firewall_rule.go
```diff
@@ -0,0 +1,161 @@
...
+	case "lt":
+		return match < f.priority
+	case "eq":
+		return match == f.priority
+	case "gt":
+		return match > f.priority
+	case "gte":
+		return match >= f.priority
+	default:
+		return true
```
jacobbednarz, line 159: i don't think this is a sensible default -- we should default to `false` if the operator isn't matched to prevent misguided typos.

```suggestion
		return false
```

cloudflare/data_source_firewall_rule.go
```diff
@@ -0,0 +1,161 @@
...
+
+func dataSourceCloudflareFirewallRulesRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	filter, err := expandFilterFirewallRules(d.Get("filter"))
+	if err != nil {
+		return err
+	}
+
+	zoneID := d.Get("zone_id").(string)
+	rules, err := client.FirewallRules(context.Background(), zoneID, cloudflare.PaginationOptions{})
```
jacobbednarz, line 82: you'll probably want to handle things like pagination here, otherwise, you'll only get the first 25 results - https://api.cloudflare.com/#firewall-rules-list-of-firewall-rules

cloudflare/data_source_firewall_rule.go
```diff
@@ -0,0 +1,161 @@
...
+
+	err = d.Set("firewall_rules", filteredRules)
+	if err != nil {
+		return fmt.Errorf("error setting Firewall Rules: %s", err)
+	}
+
+	return nil
+}
+
+func expandFilterFirewallRules(d interface{}) (*firewallRulesFilter, error) {
```
jacobbednarz, line 117: this isn't expanding filrewall rules but the firewall filters. we should update the name to match it's intention.

```suggestion
func expandFirewallFilters(d interface{}) (*firewallRulesFilter, error) {
```

cloudflare/data_source_firewall_rule.go
```diff
@@ -0,0 +1,161 @@
...
+			"paused":   v.Paused,
+		})
+	}
+
+	err = d.Set("firewall_rules", filteredRules)
+	if err != nil {
+		return fmt.Errorf("error setting Firewall Rules: %s", err)
+	}
+
+	return nil
```
jacobbednarz, line 114: to use this resource, you'll also need to `d.SetId` otherwise the values won't be persisted with an identifier to access. see https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/data_source_api_token_permission_groups.go#L46 for an example.

cloudflare/data_source_firewall_rule.go
```diff
@@ -0,0 +1,161 @@
...
+	for _, v := range rules {
+		if filter.paused != v.Paused {
+			continue
+		}
+
+		rulePrio, ok := v.Priority.(int)
+		if !ok {
+			rulePrio = 2147483647
+		}
+		if !filter.matchPriority(rulePrio) {
```
jacobbednarz, line 97:
```suggestion

		if !filter.matchPriority(rulePrio) {
```

jacobbednarz (changes requested):

jacobbednarz:
closing this off as it is now stagnent. please let me know if you want to pick this up again or take another pass.

PR #956, opened 2021-02-16
resource/access_policy: support importing zone and account level resources

jacobbednarz:
Updates the `Import` to support managing zone and account level
policies.

Closes #929

PR #1002, opened 2021-03-24
Correct access_application terraform import docs

tjefferson08:
The correct import key is the composite of _account ID_ (not zone ID) and application ID

[source for import](https://github.com/cloudflare/terraform-provider-cloudflare/blob/b55e604df70794bbf6dbfc5e425f8a2e30b7b6a1/cloudflare/resource_cloudflare_access_application.go#L318)

### DIFF
--- a/website/docs/r/access_application.html.markdown
+++ b/website/docs/r/access_application.html.markdown
@@ -93,7 +93,7 @@ The following additional attributes are exported:
 
 ## Import
 
-Access Applications can be imported using a composite ID formed of zone
+Access Applications can be imported using a composite ID formed of account
 ID and application ID.
 
 ```


jacobbednarz:
Thanks for this @tjefferson08. We can ignore the tfproviderlint failures here as it is unrelated to the documentation change.

Worth noting this fixes the documentation however the Import itself is still broken due to not supporting the Account and Zone level Access applications which we'll likely need to fix in another PR.

tjefferson08:
Happy to help!

>Worth noting this fixes the documentation however the Import itself is still broken due to not supporting the Account and Zone level Access applications which we'll likely need to fix in another PR.

@jacobbednarz as I recall, I was able to import once I swapped in the account ID (after peeking at the source)

But you're saying you need to support import by zone ID?

jacobbednarz:
Correct. Access resources are currently available for both zone + accounts however the `access_application` import doesn't support that. I've fixed this for `access_policy` via #956 in the past.

waynejnicklin:
So it seems that the documentation still says that only account id import is supported, not zone id?

PR #1025, opened 2021-04-12
Handles the case of an SSL cert ID changing after an update

joel-g:
checks if the SSL object ID changed and sets the new one to state if it did

### DIFF
--- a/cloudflare/resource_cloudflare_custom_ssl.go
+++ b/cloudflare/resource_cloudflare_custom_ssl.go
@@ -165,6 +165,9 @@ func resourceCloudflareCustomSslUpdate(d *schema.ResourceData, meta interface{})
 			updateErr = true
 		} else {
 			log.Printf("[DEBUG] Custom SSL set to: %s", res.ID)
+			if res.ID != certID {
+				d.SetId(res.ID)
+			}
 		}
 
 	}


garrettgalow:
replied over email but also here is a link to the docs: https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/certificate_pack

joel-g:
According to this document https://api.cloudflare.com/#custom-ssl-for-a-zone-ssl-configuration-details the cert ID will change after a patch. This PR will check if the ID has changed and will set it to state.

"Upload a new private key and/or PEM/CRT for the SSL certificate. Note: PATCHing a configuration for sni_custom certificates will result in a new resource id being returned, and the previous one being deleted."

jacobbednarz:
@joel-g are you able to expand on your use case or issue here a little more? there are a few things going on here and i'd like to make sure we're all on the same page with what problems we are solving as your change doesn't entirely cover what your PR title/description says it is aiming to.

joel-g:
The problem we are encountering is that we update a custom SSL certificate and it is successful. Then the next time the Terraform job runs it GETs the cert object from the API and gets a 404 because the cert ID has changed in the API but not in Terraform state. So then Terraform attempts to POST the certificate again.

jacobbednarz:
thanks for clarifying! i wanted to confirm as the ticket also didn't really line up with what we have here so i wanted to be sure before we start tackling this.

as for the change, right now the ID is only updated when the `custom_ssl_options` is updated (using [`d.HasChanges()`](https://pkg.go.dev/github.com/hashicorp/terraform/helper/schema#ResourceData.HasChange)) however i assume you'll also want this to change if you add a new host or should the priority change as well. note: if this is working without the `custom_ssl_options` being changed, i suspect we're encountering a bug with nested lists within terraform and should explicitly fix it to prevent any breakage in the future when/if this gets fixed.

along with the change, we need to be pretty careful with updating resource IDs within the CRUD operations like this due to be being used for an "anchor" for other library calls. i can't see anywhere that we access the ID field outside of `d.Id()` (i.e. `d.Get("id")` so we should be 👌 for normal operation. this _may_ however, cause issues if the Terraform operation is interrupted while this is updating and recovery will be quite messy so avoid that if you can 😅.

joel-g:
I can't speak on the occurrence of this issue happening within the other hook where priority changes happen because we never change the priority. We are adopting this change internally on our fork so I will monitor it for other inconsistencies.

jacobbednarz:
any updates from your side @joel-g? if you'd like to address those couple of inline comments and you haven't seen any issues with this, i'm happy to look at getting this merged for the next release.

owengo:
note that updating the certificate (ie: replace cert + private key with new ones ) without touching the "options" triggers the bug which is fixed by the patch, even if it's not obvious that d.HasChange("custom_ssl_options") would be true in this case.

jacobbednarz:
superseded by #1025

PR #1029, opened 2021-04-14
d/zones: fix attribute documentation

alexjurkiewicz:
To match [the code](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/data_source_zones.go#L51).


jacobbednarz:
thanking you!

PR #1047, opened 2021-04-26
MaxItems for ZoneLockdown to 250

SwissKid:
Based on the Cloudflare spec, this is the max and it cannot be exceeded by support requests (to my knowledge).
This would nicely make `terraform plan` fail.

### DIFF
--- a/cloudflare/resource_cloudflare_zone_lockdown.go
+++ b/cloudflare/resource_cloudflare_zone_lockdown.go
@@ -52,6 +52,7 @@ func resourceCloudflareZoneLockdown() *schema.Resource {
 			"configurations": {
 				Type:     schema.TypeSet,
 				MinItems: 1,
+				MaxItems: 250,
 				Required: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{


jacobbednarz (changes requested):
Thanks for this @SwissKid! Could you please add test coverage for this change? You should be able to do something like the below and attempt to create more than 250 blocks in order to trigger the exception to be called.

```go
package main

import (
	"fmt"
)

func main() {
	configBlock := `
configuration {
  target = "ip"
  value = "192.0.2.%d"
}
`
	i := 1
	for i <= 250 {
		fmt.Printf(configBlock, i)
		i = i + 1
	}
}
```

SwissKid:
Thanks! I'll definitely work on that.

https://support.cloudflare.com/hc/en-us/articles/115001595131-Understanding-Cloudflare-Zone-Lockdown

I am reading here and realizing that the limit is different depending on the account type. Is there a good pattern for identifying the account type? I also realized this limit may be arbitrary so I should talk to Cloudflare support and see if they increase this limit upon request (which could affect customers who increased their limit)

jacobbednarz:
🤔 this may be why we don't have it enforced. If that's that the case, I'm afraid we can't add this. There isn't a public approach to fetching entitlements.

jacobbednarz:
closing as we can't enforce this in the schema due to account entitlement differences

Issue #1010, opened 2021-03-31
Make `ownership_challenge` argument optional (or documented) for `cloudflare_logpush_job` resources


taurelius:
### Current Terraform version

0.14.8

### Description

Currently the `ownership_challenge` argument is required when creating a `cloudflare_logpush_job` resource. However, it is not a required field (in the API) when creating a Logpush job with a destination of either Datadog or Splunk.

[From the docs:](https://developers.cloudflare.com/logs/logpush/datadog)
> Note: Unlike configuring Logpush jobs for AWS S3, GCS, or Azure, there is no ownership challenge when configuring Logpush to Datadog.

I can see the thinking here, the majority of destinations do require an ownership challenge, though it means when creating a Datadog job a dummy value needs to be provided e.g. `0`.

I think either the argument should be made optional and a note added that it's required for the majority of use-cases, or left as is calling out the workaround for Datadog and Splunk. I'm more than happy to raise a PR for either of these.

The Cloudflare API docs for this also [seem to be incorrect](https://api.cloudflare.com/#logpush-jobs-create-logpush-job), again calling out the `destination_challenge` field as required (I've checked and verified it isn't) not matching up with the Datadog integration guide linked above.

### Use cases

Creating a `cloudflare_logpush_job` with a destination of either Datadog or Splunk

### Potential Terraform configuration

N/A

### References

- https://developers.cloudflare.com/logs/logpush/datadog
- https://developers.cloudflare.com/logs/logpush/splunk
- https://api.cloudflare.com/#logpush-jobs-create-logpush-job

jacobbednarz:
Sorry for the radio silence on this one @taurelius, it fell off my radar. 

I'm 👍 to making the schema attribute optional and instead enforcing the validation in the `Create`/`Update` operations themselves looking at the `destination_conf` value. Something like the following would suffice:

```go
destConf := d.Get("destination_conf").(string) 
ownershipChallenge := d.Get("ownership_challenge").(string) 
var re = regexp.MustCompile(`^(datadog|splunk)\:\/\/`)

if ownershipChallenge == "" && !re.MatchString(destConf) {
    // we will error out as the destination conf requires a value
} 
```

PR #1048, opened 2021-04-27
logpush_job: make ownership_challenge optional

taurelius:
Makes `ownership_challenge` argument optional when the `destination_conf` is Datadog, Splunk or S3-Compatible (non-AWS).

closes cloudflare/terraform-provider-cloudflare#1010

### DIFF
--- a/cloudflare/resource_cloudflare_logpush_job.go
+++ b/cloudflare/resource_cloudflare_logpush_job.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -51,7 +52,7 @@ func resourceCloudflareLogpushJob() *schema.Resource {
 			},
 			"ownership_challenge": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


cloudflare/resource_cloudflare_logpush_job.go
```diff
@@ -118,6 +119,12 @@ func resourceCloudflareLogpushJobCreate(d *schema.ResourceData, meta interface{}
 		return fmt.Errorf("error finding logpush job: %v", err)
 	}
 
+	var re = regexp.MustCompile(`^((datadog|splunk)://|s3://.+endpoint=)`)
```
jacobbednarz, line 122: I wonder if this would be better suited in `getJobFromResource` as the CRUD methods all use it?

taurelius: Sounds good! I avoided this initially just incase there was a scenario where this could prevent you deleting an existing resource, given we're checking this on create though I think you're right - it may as well be moved. 😄

jacobbednarz:
thanks @taurelius! this looks great and i appreciate your effort here 👏

taurelius:
No worries @jacobbednarz! If possible it'd be great to get the [API docs](https://api.cloudflare.com/#logpush-jobs-create-logpush-job) updated too. Happy to PR if there's a public repo for them, though haven't been able to find one 🤔

jacobbednarz:
afraid that's internal but i can pass it onto the team to rectify.

jacobbednarz:
just looping back on this one; the team was making doc changes at the same time so i lumped in this change. it is now reflecting the resource expectation.

PR #1053, opened 2021-05-07
Implement waiting room resource

bpaquet:
Need https://github.com/cloudflare/cloudflare-go/pull/627 to be merged and released before.


cloudflare/resource_cloudflare_waiting_room.go
```diff
@@ -0,0 +1,223 @@
...
+		TotalActiveUsers:      d.Get("total_active_users").(int),
+		NewUsersPerMinute:     d.Get("new_users_per_minute").(int),
+		CustomPageHtml:        d.Get("custom_page_html").(string),
+		SessionDuration:       d.Get("session_duration").(int),
+		QueueAll:              d.Get("queue_all").(bool),
+		DisableSessionRenewal: d.Get("disable_session_renewal").(bool),
+	}
+}
+
+func resourceCloudflareWaitingRoomReadFromWaitingRoomId(d *schema.ResourceData, meta interface{}, zoneID, waitingRoomID string) error {
```
jacobbednarz, line 117: why is this a seperate method? `d.Set` should really on be called in the `Read` method which in turn is called by the other CRUD methods.

it looks like you're calling this in the `Import` method as well which we shouldn't be doing but instead relying on the `Read` method to do it's thing.

cloudflare/resource_cloudflare_waiting_room.go
```diff
@@ -0,0 +1,223 @@
...
+	d.SetId(waitingRoom.ID)
+
+	return resourceCloudflareWaitingRoomRead(d, meta)
+}
+
+func resourceCloudflareWaitingRoomRead(d *schema.ResourceData, meta interface{}) error {
+	waitingRoomID := d.Id()
+	zoneID := d.Get("zone_id").(string)
+
+	return resourceCloudflareWaitingRoomReadFromWaitingRoomId(d, meta, zoneID, waitingRoomID)
```
jacobbednarz, line 166: this functionality should be pulled into the `Read` method

cloudflare/resource_cloudflare_waiting_room.go
```diff
@@ -0,0 +1,223 @@
...
+func resourceCloudflareWaitingRoomDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	waitingRoomID := d.Id()
+	zoneID := d.Get("zone_id").(string)
+
+	err := client.DeleteWaitingRoom(context.Background(), zoneID, waitingRoomID)
+
+	if err != nil {
+		name := d.Get("name").(string)
+		return fmt.Errorf("Error updating waiting room %q: %s", name, err)
```
jacobbednarz, line 195:
```suggestion
		return fmt.Errorf("Error deleting waiting room %q: %s", name, err)
```

cloudflare/resource_cloudflare_waiting_room.go
```diff
@@ -0,0 +1,223 @@
...
+	idAttr := strings.SplitN(d.Id(), "/", 2)
+	var zoneID string
+	var waitingRoomID string
+	if len(idAttr) == 2 {
+		zoneID = idAttr[0]
+		waitingRoomID = idAttr[1]
+	} else {
+		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"zoneID/waitingRoomID\" for import", d.Id())
+	}
+
```
jacobbednarz, line 211: checkout https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/resource_cloudflare_argo_tunnel.go#L84 for an example of using the `Import` method to call the `Read` instead of a custom method.

website/docs/r/waiting_room.html.markdown
```diff
@@ -0,0 +1,60 @@
...
+
+Provides a Cloudflare Waiting Room resource.
+
+## Example Usage
+
+```hcl
+resource "cloudflare_waiting_room" "example" {
+    zone_id              = "ae36f999674d196762efcc5abb06b345"
+    name                 = "foo"
+    host                 = "foo.bar.com"
```
jacobbednarz, line 19: let's use the documentation domains here from rfc5737 instead

```suggestion
    host                 = "foo.example.com"
```

cloudflare/resource_cloudflare_waiting_room.go
bpaquet: Done

cloudflare/resource_cloudflare_waiting_room.go
bpaquet: Done

cloudflare/resource_cloudflare_waiting_room.go
bpaquet: Done

jacobbednarz:
thanks for the updates here; it looks like we're almost there but still missing tests for this new functionality. are you able to add those to ensure this functionality works as intended and we can catch any regressions going forward?

bpaquet:
> thanks for the updates here; it looks like we're almost there but still missing tests for this new functionality. are you able to add those to ensure this functionality works as intended and we can catch any regressions going forward?

Do you mean unit test or acceptance test? I did not find a lot of unit test. And I'm not sure how acceptance test works. Do you have a doc around them?

jacobbednarz:
integration tests for the actual resources. see the README (https://github.com/cloudflare/terraform-provider-cloudflare#developing-the-provider) and Terraform's own documentation for acceptance testing https://www.terraform.io/docs/extend/testing/acceptance-tests/index.html

jacobbednarz:
@bpaquet are you still interested in completing this one?

bpaquet:
Yes, but I cannt write acceptance test, I need a dedicated Cloudflare
account for this.

Bertrand

jacobbednarz:
If you create a free zone, email me (or Cloudflare Support) the zone ID and I can take a look at adding the entitlements temporarily for you to add this.

jacobbednarz:
Nice, we're all green on the acceptance testing accounts too so this is good to merge.

```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareWaitingRoom_" -timeout 120m -parallel 1
=== RUN   TestAccCloudflareWaitingRoom_Create
=== PAUSE TestAccCloudflareWaitingRoom_Create
=== CONT  TestAccCloudflareWaitingRoom_Create
--- PASS: TestAccCloudflareWaitingRoom_Create (2.59s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/cloudflare	3.231s
```

Issue #1041, opened 2021-04-22
cloudflare_access_service_token should allow automatic rotation


uncycler:
### Current Terraform and Cloudflare provider version

```
Terraform v0.14.10
+ provider registry.terraform.io/cloudflare/cloudflare v2.20.0
```

### Description

When a service token expire, the cloudflare_access_service_token resource is not tainted which cause disruption on service that depends on it.

### Use cases

For example, a token that is used for healthcheck. In this use case, if the token expires, healthcheck will fail. The only way to fix the issue, is to issue "terraform taint cloudflare_access_service_token.healthcheck".

```hcl
resource "cloudflare_access_service_token" "healthcheck" {
  zone_id   = data.cloudflare_zones.myzone.id
  name      = "healthcheck"
}

resource "cloudflare_healthcheck" "zerotrust_healthcheck" {
  zone_id     = data.cloudflare_zones.myzone.id
  name        = "myhealthcheck"
  address     = "myservice.com"

  notification_email_addresses = ["alert@alert.com"]
  type            = "HTTPS"
  port            = "443"
  path            = "/healthz"
  expected_codes  = [200]

  follow_redirects = true
  header {
    header = "CF-Access-Client-Secret"
    values = [cloudflare_access_service_token.healthcheck.client_secret]
  }
  header {
    header = "CF-Access-Client-ID"
    values = [cloudflare_access_service_token.healthcheck.client_id]
  }
  consecutive_fails     = 3
  consecutive_successes = 2
}
```

### Potential Terraform configuration

We could leverage the "time_rotating" resource. But cloudflare_access_service_token resource doesnt have argument that force recreation. Maybe if there is an internal argument that trigger a recreation of the resource like this:

```hcl
resource "time_rotating" "service_token_healthcheck" {
  rotation_days = 180
}

resource "cloudflare_access_service_token" "healthcheck" {
  zone_id   = data.cloudflare_zones.myzone.id
  name      = "healthcheck"
  trigger   = time_rotating.service_token_healthcheck.unix
}
```

jacobbednarz:
Thanks for raising this one, it is definitely an interesting proposal. 

I think for starters, we don't need an additional resource. The service token itself provides an expires on value which we could store and on validation, force a new resource if the date has passed or is within a configured threshold. 

[...]

PR #1057, opened 2021-05-10
Cloudflare service token rotation

BojanZelic:
This PR closes https://github.com/cloudflare/terraform-provider-cloudflare/issues/1041

It adds support for cloudflare_access_service_token renewal if terraform is ran within the timeperiod defined in min_days_for_renewal.

See example usage:

```
# Generate a service token that will renew if terraform is ran within 30 days of expiration
resource "cloudflare_access_service_token" "my_app" {
  account_id = "d41d8cd98f00b204e9800998ecf8427e"
  name       = "CI/CD app renewed"
  min_days_for_renewal = 30

  lifecycle {
    create_before_destroy = true
  }
}
```


cloudflare/resource_cloudflare_access_service_tokens.go
```diff
@@ -19,6 +20,8 @@ func resourceCloudflareAccessServiceToken() *schema.Resource {
 			State: resourceCloudflareAccessServiceTokenImport,
 		},
 
+		CustomizeDiff: resourceCloudflareAccessServiceTokenCustomizeDiff,
```
jacobbednarz, line 23: is there a reason we're opting for a custom function here over something like [`customdiff.ForceNewIf`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/customdiff?utm_source=godoc#ForceNewIf)?

jacobbednarz:
I would also like to explore an integration test for this behaviour however I don't exactly have a good approach planned out yet. Thoughts?

cloudflare/resource_cloudflare_access_service_tokens.go
BojanZelic: Thanks for the suggestion - you're correct... I can make this work with `ForceNewIf` instead of the custom function to simplify abit. 

Initially I tried `ForceNewIfChange` as suggested in the linked issue but it required access to `schema.ResourceDiff` which it didn't provide.

BojanZelic:
> I would also like to explore an integration test for this behaviour however I don't exactly have a good approach planned out yet. Thoughts?

For an integration test I see two tests possibly:

1) set `min_days_for_renewal = 365`... run the apply 2x and make sure that the 2nd time it regenerates the token
2) set `min_days_for_renewal = 364`... run the apply 2x and make sure that it doesn't regenerate the token

BojanZelic:
@jacobbednarz I've added an acceptance test & updated a few things with the PR. Let me know if you have any further suggestions.

cloudflare/resource_cloudflare_access_service_tokens.go
```diff
@@ -100,6 +137,7 @@ func resourceCloudflareAccessServiceTokenCreate(d *schema.ResourceData, meta int
 	d.Set("name", serviceToken.Name)
 	d.Set("client_id", serviceToken.ClientID)
 	d.Set("client_secret", serviceToken.ClientSecret)
+	d.Set("expires_at", serviceToken.ExpiresAt.Format(time.RFC3339))
```
jacobbednarz, line 140: `cloudflare.AccessServiceTokenCreateResponse` doesn't have an `ExpiresAt` field - see https://github.com/cloudflare/cloudflare-go/blob/master/access_service_tokens.go#L37-L44. we'll need to get this into cloudflare-go first

jacobbednarz:
looks like a good approach! other than adding the field into cloudflare-go, this is looking ace. thanks!

jacobbednarz:
this is now ready to update from master to get this prepared for merge

BojanZelic:
@jacobbednarz I've rebased from master... let me know if you need anything else to get this merged.

jacobbednarz:
thanks, looks great and the integration tests are passing locally 🎉

PR #1058, opened 2021-05-11
Add teams list and device posture rule resources

Justin-Holmes:


jacobbednarz:
@Justin-Holmes are we able to drop the dependency upgrades here please? we generally bump them outside of the PRs and then just update the PRs to prevent conflicts (which it looks like we have here).

Justin-Holmes:
@jacobbednarz Are those linter errors a concern? I'm happy to fix them if so.

jacobbednarz:
Yeah, they are worth while listening to. They generally catch hard to debug issues (like type incompatibilites in the schema) which commonly silently fail.

Justin-Holmes:
@jacobbednarz Alright, should be fixed!

cloudflare/resource_cloudflare_device_posture_rule.go
```diff
@@ -0,0 +1,288 @@
...
+		Create: resourceCloudflareDevicePostureRuleCreate,
+		Read:   resourceCloudflareDevicePostureRuleRead,
+		Update: resourceCloudflareDevicePostureRuleUpdate,
+		Delete: resourceCloudflareDevicePostureRuleDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceCloudflareDevicePostureRuleImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"account_id": {
```
jacobbednarz, line 26: should this be required? it looks like we're accessing it directly further below for the API calls.

cloudflare/resource_cloudflare_teams_list.go
```diff
@@ -0,0 +1,211 @@
...
+		Create: resourceCloudflareTeamsListCreate,
+		Read:   resourceCloudflareTeamsListRead,
+		Update: resourceCloudflareTeamsListUpdate,
+		Delete: resourceCloudflareTeamsListDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceCloudflareTeamsListImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"account_id": {
```
jacobbednarz, line 25: similar to my other comment, i think we want this field to be required.

cloudflare/resource_cloudflare_teams_list.go
```diff
@@ -0,0 +1,211 @@
...
+	resourceCloudflareTeamsListRead(d, meta)
+
+	return nil
+}
+
+func resourceCloudflareTeamsListImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	attributes := strings.SplitN(d.Id(), "/", 2)
+
+	if len(attributes) != 2 {
+		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"accountID/accessApplicationID\"", d.Id())
```
jacobbednarz, line 170: should this be team list ID or similar?

jacobbednarz (commented):
couple of small nitpicks but we're nearly there!

cloudflare/resource_cloudflare_teams_list_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+)
+
+func TestAccCloudflareTeamsListBasic(t *testing.T) {
```
jacobbednarz, line 13: this test is currently failing

```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareTeams" -timeout 120m -parallel 1
=== RUN   TestAccCloudflareTeamsListBasic
    testing.go:705: Step 0 error: Check failed: Check 5/5 error: cloudflare_teams_list.ffyuqmxemk: Attribute 'items.0' expected "asdf-1234", got "asdf-5678"
--- FAIL: TestAccCloudflareTeamsListBasic (3.35s)
FAIL
FAIL	github.com/cloudflare/terraform-provider-cloudflare/cloudflare	3.910s
FAIL
make: *** [testacc] Error 1
```

suspect this may be a bug with constructing the list items.

cloudflare/resource_cloudflare_device_posture_rule.go
Justin-Holmes: Yep, good catch.

cloudflare/resource_cloudflare_teams_list_test.go
Justin-Holmes: Interestingly, the API returns the list items in reverse order.

jacobbednarz:
noice! 👏

PR #1089, opened 2021-06-02
fix(APITokens): change permission_groups type from list to set of strings

UrosSimovic:
**Problem**
A diff can be detected when creating API Tokens with multiple permission_groups in a single policy.

```hcl
resource "cloudflare_api_token" "test" {
  name = "test"

  policy {
    effect = "allow"
    permission_groups = [
      "3030687196b94b638145a3953da2b699",
      "da6d2d6f2ec8442eaadda60d13f42bca",
      "e6d2666161e84845a636613608cee8d5",
      "ed07f6c337da4195b4e72a1fb2c6bcae",
      "4755a26eedb94da69e1066d98aa820be",
      "1af1fa2adc104452b74a9a3364202f20",
      "43137f8d07884d3198dc0ee77ca6e79b",
    ]
    resources = { "com.cloudflare.api.account.*" = "*" }
  }
}
```

```terminal
$ terraform apply && terraform plan
...

  # cloudflare_api_token.test will be updated in-place
  ~ resource "cloudflare_api_token" "test" {
      ...
      - policy {
          - effect            = "allow" -> null
          - permission_groups = [
              - "3030687196b94b638145a3953da2b699",
              - "da6d2d6f2ec8442eaadda60d13f42bca",
              - "e6d2666161e84845a636613608cee8d5",
              - "ed07f6c337da4195b4e72a1fb2c6bcae",
              - "43137f8d07884d3198dc0ee77ca6e79b",
              - "4755a26eedb94da69e1066d98aa820be",
              - "1af1fa2adc104452b74a9a3364202f20",
            ] -> null
          - resources         = {
              - "com.cloudflare.api.account.*" = "*"
            } -> null
        }
      + policy {
          + effect            = "allow"
          + permission_groups = [
              + "3030687196b94b638145a3953da2b699",
              + "da6d2d6f2ec8442eaadda60d13f42bca",
              + "e6d2666161e84845a636613608cee8d5",
              + "ed07f6c337da4195b4e72a1fb2c6bcae",
              + "4755a26eedb94da69e1066d98aa820be",
              + "1af1fa2adc104452b74a9a3364202f20",
              + "43137f8d07884d3198dc0ee77ca6e79b",
            ]
          + resources         = {
              + "com.cloudflare.api.account.*" = "*"
            }
        }
    }

...
```

**Why?**
 CloudFlare API can return permission_groups in a different order then the posted/updated one.

**Solution**
Make permission_groups set of strings where order doesn't matter.

### DIFF
--- a/cloudflare/resource_cloudflare_api_token.go
+++ b/cloudflare/resource_cloudflare_api_token.go
@@ -22,9 +22,10 @@ func resourceCloudflareApiToken() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 			"permission_groups": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
 				Required: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
 			},
 			"effect": {
 				Type:         schema.TypeString,
@@ -165,7 +166,8 @@ func resourceDataToApiTokenPolices(d *schema.ResourceData) []cloudflare.APIToken
 	for _, p := range policies {
 		policy := p.(map[string]interface{})
 
-		permissionGroups := expandInterfaceToStringList(policy["permission_groups"])
+		permissionGroups := expandSetToStringList(policy["permission_groups"])
+
 		if len(permissionGroups) == 0 {
 			continue
 		}
--- a/cloudflare/resource_cloudflare_api_token_test.go
+++ b/cloudflare/resource_cloudflare_api_token_test.go
@@ -42,6 +42,29 @@ func TestAccAPIToken(t *testing.T) {
 	})
 }
 
+func TestAccCloudflareAPITokenMultiplePermissionsGroups(t *testing.T) {
+	// Temporarily unset CLOUDFLARE_API_TOKEN if it is set as the API token
+	// endpoint does not yet support the API tokens without an explicit scope.
+	if os.Getenv("CLOUDFLARE_API_TOKEN") != "" {
+		defer func(apiToken string) {
+			os.Setenv("CLOUDFLARE_API_TOKEN", apiToken)
+		}(os.Getenv("CLOUDFLARE_API_TOKEN"))
+		os.Setenv("CLOUDFLARE_API_TOKEN", "")
+	}
+
+	rnd := generateRandomResourceName()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCloudflareAPITokenMultiplePermissionsGroups(rnd),
+			},
+		},
+	})
+}
+
 func TestAccAPITokenAllowDeny(t *testing.T) {
 	// Temporarily unset CLOUDFLARE_API_TOKEN if it is set as the API token
 	// endpoint does not yet support the API tokens without an explicit scope.
@@ -250,4 +273,26 @@ func testAPITokenConfigAllowDeny(resourceID, name, permissionID, zoneID string, 
 		  %[4]s
 		}
 		`, resourceID, name, permissionID, add)
+}
+
+func testAccCloudflareAPITokenMultiplePermissionsGroups(rnd string) string {
+	return fmt.Sprintf(`
+	resource "cloudflare_api_token" "%[1]s" {
+		name = "%[1]s"
+
+		policy {
+			effect = "allow"
+			permission_groups = [
+				"3030687196b94b638145a3953da2b699",
+				"da6d2d6f2ec8442eaadda60d13f42bca",
+				"e6d2666161e84845a636613608cee8d5",
+				"ed07f6c337da4195b4e72a1fb2c6bcae",
+				"4755a26eedb94da69e1066d98aa820be",
+              	"1af1fa2adc104452b74a9a3364202f20",
+              	"43137f8d07884d3198dc0ee77ca6e79b",
+			]
+			resources = { "com.cloudflare.api.account.*" = "*" }
+		}
+	}
+`, rnd)
 }
--- a/cloudflare/utils.go
+++ b/cloudflare/utils.go
@@ -22,6 +22,16 @@ func expandInterfaceToStringList(list interface{}) []string {
 	return vs
 }
 
+func expandSetToStringList(setInterface interface{}) []string {
+	set := setInterface.(*schema.Set)
+	list := make([]string, set.Len())
+	for i, ele := range set.List() {
+		list[i] = ele.(string)
+	}
+
+	return list
+}
+
 func expandStringListToSet(list []string) *schema.Set {
 	values := schema.NewSet(schema.HashString, []interface{}{})
 	for _, h := range list {


cloudflare/resource_cloudflare_api_token.go
```diff
@@ -22,9 +22,10 @@ func resourceCloudflareApiToken() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 			"permission_groups": {
-				Type:     schema.TypeList,
```
jacobbednarz, line 25: as we are going from a list => set, we will need to 
1. bump the schema version 
2. provide a state migration (see https://www.terraform.io/docs/extend/resources/state-migration.html) as we will be changing the internal representation of the data

cloudflare/resource_cloudflare_api_token_test.go
```diff
@@ -42,6 +42,29 @@ func TestAccAPIToken(t *testing.T) {
...
+		}(os.Getenv("CLOUDFLARE_API_TOKEN"))
+		os.Setenv("CLOUDFLARE_API_TOKEN", "")
+	}
+
+	rnd := generateRandomResourceName()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
```
jacobbednarz, line 60: as we're now needing a schema migration we'll need to extend this a little.

the first is that we'll need to test the data migration from a list => set. https://www.terraform.io/docs/extend/resources/state-migration.html covers some examples of asserting that.

the second is that we'll want to update this test to be 2 steps. the first will create the resource with the permissions and the second will be the same permissions but in a different order and we'll want to assert there is no difference which will ensure we're ignoring the ordering.

jacobbednarz (changes requested):
great start! see comments inline for this changes needed for this one.

cloudflare/resource_cloudflare_api_token.go
UrosSimovic: Is this really necessary?

I tested this:
1. create resource with current provider version
2. upgrade to fixed provider
3. run `terraform apply`

No changes detected.

UrosSimovic: Because state presentation of a list or a set is the same - json array.

cloudflare/resource_cloudflare_api_token_test.go
UrosSimovic: I added a test which fails (1st commit) and then a fix for that (2nd commit).

But yeah, I get  your point. I can add another step with different order, but don't know how to assert there is no difference. Is there some example of something similar?

cloudflare/resource_cloudflare_api_token.go
jacobbednarz: `TypeList` and `TypeSet` definitely aren't the same state representation as [sets are stored indexed by a hash calculated based on the attributes in the set](https://www.terraform.io/docs/extend/schemas/schema-types.html#typeset) whereas [list items are stored as a zero indexed items](https://www.terraform.io/docs/extend/schemas/schema-types.html#typelist). I'm unsure why you're not seeing a difference here but we may need to dig into why as I've definitely been bit by this in the past.

cloudflare/resource_cloudflare_api_token.go
```diff
@@ -22,9 +22,10 @@ func resourceCloudflareApiToken() *schema.Resource {
...
+				Type:     schema.TypeSet,
 				Required: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
```
jacobbednarz, line 28: i don't think we need this as it's the default

cloudflare/resource_cloudflare_api_token.go
UrosSimovic: Sorry for late response. But I said that because I checked the state before and after the fix. Even at links you posted are examples, both of them with slices. I think terraform is doing hash keys on the fly and not storing them in the state file.

Just do the test please.

For the given resource
```hcl
resource "cloudflare_api_token" "test" {
  name = "test"

  policy {
    effect = "allow"
    permission_groups = [
      "3030687196b94b638145a3953da2b699",
      "da6d2d6f2ec8442eaadda60d13f42bca",
      "e6d2666161e84845a636613608cee8d5",
      "ed07f6c337da4195b4e72a1fb2c6bcae",
      "1af1fa2adc104452b74a9a3364202f20",
      "4755a26eedb94da69e1066d98aa820be",
    ]
    resources = { "com.cloudflare.api.account.*" = "*" }
  }
}
```


Current provider state
```json
{
  "version": 4,
  "terraform_version": "0.13.7",
  "serial": 1,
  "lineage": "f3059c1d-de72-cc99-752d-da6446426758",
  "outputs": {},
  "resources": [
    {
      "mode": "managed",
      "type": "cloudflare_api_token",
      "name": "test",
      "provider": "provider[\"registry.terraform.io/terraform-providers/cloudflare\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "condition": [],
            "id": "e8712ce735c7353116fe7586dcf5bacb",
            "issued_on": "2021-07-19T12:26:40Z",
            "modified_on": "2021-07-19T14:26:42.035223871+02:00",
            "name": "uros-test",
            "policy": [
              {
                "effect": "allow",
                "permission_groups": [
                  "3030687196b94b638145a3953da2b699",
                  "da6d2d6f2ec8442eaadda60d13f42bca",
                  "e6d2666161e84845a636613608cee8d5",
                  "ed07f6c337da4195b4e72a1fb2c6bcae",
                  "1af1fa2adc104452b74a9a3364202f20",
                  "4755a26eedb94da69e1066d98aa820be"
                ],
                "resources": {
                  "com.cloudflare.api.account.*": "*"
                }
              }
            ],
            "status": "active",
            "value": "REDACTED"
          },
          "private": "bnVsbA=="
        }
      ]
    }
  ]
}
```


Fixed provider state
```json
{
  "version": 4,
  "terraform_version": "0.13.7",
  "serial": 1,
  "lineage": "fb23297b-3d7b-2ce0-21fc-f1e9a7a75219",
  "outputs": {},
  "resources": [
    {
      "mode": "managed",
      "type": "cloudflare_api_token",
      "name": "test",
      "provider": "provider[\"registry.terraform.io/terraform-providers/cloudflare\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "condition": [],
            "id": "8a712ce0983ed8b47adc2710e0094ef6",
            "issued_on": "2021-07-19T12:28:08Z",
            "modified_on": "2021-07-19T14:28:10.23180693+02:00",
            "name": "uros-test",
            "policy": [
              {
                "effect": "allow",
                "permission_groups": [
                  "1af1fa2adc104452b74a9a3364202f20",
                  "3030687196b94b638145a3953da2b699",
                  "4755a26eedb94da69e1066d98aa820be",
                  "da6d2d6f2ec8442eaadda60d13f42bca",
                  "e6d2666161e84845a636613608cee8d5",
                  "ed07f6c337da4195b4e72a1fb2c6bcae"
                ],
                "resources": {
                  "com.cloudflare.api.account.*": "*"
                }
              }
            ],
            "status": "active",
            "value": "REDACTED"
          },
          "private": "bnVsbA=="
        }
      ]
    }
  ]
}
```

You can see that state representation of both is json array. And as said already said, I tested creating resource and then upgrading provider. It works like a charm.

I will do it proper way if you convince me that current approach is not working, but sorry, I will no do something that is unnecessary IMO.

LanceSandino:
bump.

UrosSimovic:
As said, MR was prepared, but not accepted for unproven reason. I don't want doing unnecessary things. Feel free to take it from here, @tete17.

tete17:
Hey @jacobbednarz I opened another MR with the commits of @UrosSimovic with some of my own addressing your comments. https://github.com/cloudflare/terraform-provider-cloudflare/pull/1425

I am still a bit confused with this whole migration thing given we seem to have been through a very similar scenario like this one and no migration was performed https://github.com/cloudflare/terraform-provider-cloudflare/commit/7704f43df41e26b1db7174ae2d37b11c8e41f9ae . 

This is the first time I deal with the go sdk so I hope you excuse me if it is 100% unrelated.

jacobbednarz:
moving to #1425

PR #1092, opened 2021-06-03
Cloudflare LoadBalancer support for Pool Latitude and Longitude

tpolich:
Support for Latitude/Longitude on pools

Removed duplicated float validation function replaced with standard
version from validation package.


tpolich:
didn't update first i'll make a new PR once i've fixed that

PR #1106, opened 2021-06-18
Update zone_lockdown.html.markdown

cf-vnkr:
Addidng explaination on how to create a rule with multiple IP addresses in it.


jacobbednarz (changes requested):
how about we add an example of multiple configuration blocks instead of this line? the example will show people instead of having them interpret it.

cf-vnkr:
That absolutely makes sense. I've added another example.

PR #1127, opened 2021-07-08
allow creating custom_hostname records without `ssl`

nickysemenza:
depends on https://github.com/cloudflare/cloudflare-go/pull/666, will need to be updated after cloudflare-go version is cut (temporarily pinning it)

### DIFF
--- a/cloudflare/resource_cloudflare_custom_hostname.go
+++ b/cloudflare/resource_cloudflare_custom_hostname.go
@@ -41,7 +41,7 @@ func resourceCloudflareCustomHostname() *schema.Resource {
 			},
 			"ssl": {
 				Type:     schema.TypeList,
-				Required: true,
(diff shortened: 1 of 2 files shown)


cloudflare/resource_cloudflare_custom_hostname.go
```diff
@@ -41,7 +41,7 @@ func resourceCloudflareCustomHostname() *schema.Resource {
 			},
 			"ssl": {
 				Type:     schema.TypeList,
-				Required: true,
```
nickysemenza, line 44: not too familiar with if this is a breaking change / needs to be modified differently? My feeling is that it's OK since it's going from required -> optional and not the other direction

jacobbednarz:
i've cherry picked this into #1131 to keep CI green while bumping dependencies.

PR #1150, opened 2021-08-05
Add support for notification webhooks resource

revathyrams:
Add support to manage webhooks destinations for notification policies.


jacobbednarz:
would you be able to remove the `vendor` directory on this PR? we use Go modules so don't need to check it in.

cloudflare/resource_cloudflare_notification_policy_webhooks.go
```diff
@@ -0,0 +1,175 @@
...
+	accountID := d.Get("account_id").(string)
+
+	notificationWebhooks := buildNotificationPolicyWebhooks(d)
+
+	webhooksDestination, err := client.CreateNotificationWebhooks(context.Background(), accountID, &notificationWebhooks)
+
+	if err != nil {
+		return fmt.Errorf("error connecting webhooks destination %s: %s", notificationWebhooks.Name, err)
+	}
+	formattedWebhookID, err := uuid.Parse(webhooksDestination.Result.ID)
```
jacobbednarz, line 72: i'm not sure we really need this. what is the intention over just using `webhooksDestination.Result.ID` as the `d.SetId()` value?

cloudflare/resource_cloudflare_notification_policy_webhooks.go
```diff
@@ -0,0 +1,175 @@
...
+
+	name := d.Get("name").(string)
+	if err != nil {
+		return fmt.Errorf("error retrieving notification webhooks %s: %s", name, err)
+	}
+	id, err := uuid.Parse(notificationWebhooks.Result.ID)
+	if err != nil {
+		return fmt.Errorf("error setting notification webhooks %s: %s", name, err)
+	}
+	d.Set("id", id.String())
```
jacobbednarz, line 97: the `Create ` should be suffice for setting the resource ID as we shouldn't need to change it after the fact.

side note: setting the resource ID should use `d.SetId()` as it is a special attribute.

cloudflare/resource_cloudflare_notification_policy_webhooks.go
```diff
@@ -0,0 +1,175 @@
...
+	webhooksDestinationID := d.Id()
+	accountID := d.Get("account_id").(string)
+
+	notificationWebhooks, err := client.GetNotificationWebhooks(context.Background(), accountID, webhooksDestinationID)
+
+	name := d.Get("name").(string)
+	if err != nil {
+		return fmt.Errorf("error retrieving notification webhooks %s: %s", name, err)
+	}
+	id, err := uuid.Parse(notificationWebhooks.Result.ID)
```
jacobbednarz, line 93: similar comment to https://github.com/cloudflare/terraform-provider-cloudflare/pull/1150/files#r683117132

PR #1151, opened 2021-08-05
Revathy/add webhooks resource

revathyrams:


cloudflare/resource_cloudflare_notification_policy_webhooks.go
```diff
@@ -0,0 +1,175 @@
...
+	accountID := d.Get("account_id").(string)
+
+	notificationWebhooks := buildNotificationPolicyWebhooks(d)
+
+	webhooksDestination, err := client.CreateNotificationWebhooks(context.Background(), accountID, &notificationWebhooks)
+
+	if err != nil {
+		return fmt.Errorf("error connecting webhooks destination %s: %s", notificationWebhooks.Name, err)
+	}
+	formattedWebhookID, err := uuid.Parse(webhooksDestination.Result.ID)
```
jacobbednarz, line 72: i'm not sure we really need this. what is the intention over just using `webhooksDestination.Result.ID` as the `d.SetId()` value?

cloudflare/resource_cloudflare_notification_policy_webhooks.go
```diff
@@ -0,0 +1,175 @@
...
+	webhooksDestinationID := d.Id()
+	accountID := d.Get("account_id").(string)
+
+	notificationWebhooks, err := client.GetNotificationWebhooks(context.Background(), accountID, webhooksDestinationID)
+
+	name := d.Get("name").(string)
+	if err != nil {
+		return fmt.Errorf("error retrieving notification webhooks %s: %s", name, err)
+	}
+	id, err := uuid.Parse(notificationWebhooks.Result.ID)
```
jacobbednarz, line 93: similar to https://github.com/cloudflare/terraform-provider-cloudflare/pull/1151/files#r683132017

cloudflare/resource_cloudflare_notification_policy_webhooks.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return fmt.Errorf("error setting notification webhooks %s: %s", name, err)
+	}
+	d.Set("id", id.String())
```
jacobbednarz, line 97: the `Create` should be suffice for setting the resource ID as we shouldn't need to change it after the fact. this value is often used as an "anchor" for lookups and should remain consistent.

side note: setting the resource ID should use `d.SetId()` as it is a special attribute.

cloudflare/resource_cloudflare_notification_policy_webhooks_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+)
+
+func TestAccCloudflareNotificationPolicyWebhooks(t *testing.T) {
```
jacobbednarz, line 11: acceptance tests are 🍏

```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareNotificationPolicyWebhooks" -timeout 120m -parallel 1
=== RUN   TestAccCloudflareNotificationPolicyWebhooks
--- PASS: TestAccCloudflareNotificationPolicyWebhooks (9.56s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/cloudflare	10.261s
```

jacobbednarz:
couple of questions but nothing major! such speed for your 2nd PR!

cloudflare/resource_cloudflare_notification_policy_webhooks.go
revathyrams: the notifications API checks if the webhook id received is a valid UUID or not. When we POST a webhook, the ID returned is not formatted with dashes `-` (although it's a valid UUID).  But when we update a webhook that's how it's expected.

revathyrams: When a webhook is created an ID is returned, which is a UUID. This may or may not have the dashes. But when we update a webhook we check if it's in the right format with dashes, so this is just an additional check to make sure they're formatted correctly.

jacobbednarz: Thanks for the info. We may want to address this service side in the future to return a consistent format as logic/safeguards like this in the Terraform provider will need to be duplicated into `cf-terraforming`. It's also not a great experience if we have potentially inconsistent identifier for resources. 

Let's proceed with this for now and discuss any improvements later on.

revathyrams: Agree with what you're saying. I'll create a ticket to fix it in our service and update here later on.

cloudflare/resource_cloudflare_notification_policy_webhooks.go
revathyrams: removed this now and verified that the tests are passing.

PR #1176, opened 2021-09-02
Add cname and UUID to tunnel

AndrewBurian:
It's exceptionally hard to use named tunnels in conjunction with either DNS or load balancers without returning the UUID or the CNAME record of the named tunnel.

This PR exposes that CNAME and uuid directly as a computed attribute of the schema instead of having to try and extract it from the ID.

I also moved the import logic into read so that calls to read the tunnel will actually verify it exists, as well as set the CNAME attribute on existing resources that aren't already set.

### DIFF
--- a/cloudflare/resource_cloudflare_argo_tunnel.go
+++ b/cloudflare/resource_cloudflare_argo_tunnel.go
@@ -34,6 +34,15 @@ func resourceCloudflareArgoTunnel() *schema.Resource {
 				Required:  true,
 				Sensitive: true,
 			},
+			"cname": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The generated CNAME for the named tunnel",
+			},
+			"uuid": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -55,10 +64,30 @@ func resourceCloudflareArgoTunnelCreate(d *schema.ResourceData, meta interface{}
 }
 
 func resourceCloudflareArgoTunnelRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	attributes := strings.Split(d.Id(), "/")
+
+	if len(attributes) != 2 {
+		return fmt.Errorf("invalid id (\"%s\") specified, should be in format \"accountID/argoTunnelUUID\"", d.Id())
+	}
+
+	accID, tunnelID := attributes[0], attributes[1]
+
+	tunnel, err := client.ArgoTunnel(context.Background(), accID, tunnelID)
+	if err != nil {
+		return fmt.Errorf("failed to fetch Argo Tunnel %s: %w", tunnelID, err)
+	}
+
+	d.Set("name", tunnel.Name)
+	d.Set("cname", fmt.Sprintf("%s.argotunnel.com", tunnelID))
+	d.Set("uuid", tunnelID)
+	d.SetId(tunnel.ID)
+
 	return nil
 }
 
-func resourceCloudflareArgoTunnelUpdate(d *schema.ResourceData, meta interface{}) error {
+func resourceCloudflareArgoTunnelUpdate(_ *schema.ResourceData, _ interface{}) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


cloudflare/resource_cloudflare_argo_tunnel.go
```diff
@@ -55,10 +64,30 @@ func resourceCloudflareArgoTunnelCreate(d *schema.ResourceData, meta interface{}
...
 func resourceCloudflareArgoTunnelRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	attributes := strings.Split(d.Id(), "/")
+
+	if len(attributes) != 2 {
+		return fmt.Errorf("invalid id (\"%s\") specified, should be in format \"accountID/argoTunnelUUID\"", d.Id())
+	}
+
+	accID, tunnelID := attributes[0], attributes[1]
```
jacobbednarz, line 75: this is `Import` specific behaviour, the `Read` doesn't need any of this. `Read` only syncs the remote API to the local state configuration.

cloudflare/resource_cloudflare_argo_tunnel.go
```diff
@@ -82,24 +111,8 @@ func resourceCloudflareArgoTunnelDelete(d *schema.ResourceData, meta interface{}
 }
 
 func resourceCloudflareArgoTunnelImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-	client := meta.(*cloudflare.API)
```
jacobbednarz, line 85: as mentioned at https://github.com/cloudflare/terraform-provider-cloudflare/pull/1176/files#r700663326, some of this functionality needs to stay as is as it's `Import` specific and `Read` doesn't use it.

jacobbednarz:
we'll also need to update the website documentation to include these new attributes too (https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/website/docs/r/argo_tunnel.html.markdown)

jacobbednarz:
thanks for the PR and i'm 👌 to ship the CNAME addition to help insulate any changes internally from being hardcoded on the customer end, but to your point that it's hard to use without these fields, we're not actually changing anything that dramatically so i'm not sure where the improvement is here. 

the `uuid` field you're proposing already exists as `id` on the resource and the CNAME can be concatenated into the resource usage using HCL.

```hcl
resource "cloudflare_argo_tunnel" "my_tunnel" {
  # .. snip
}

resource "cloudflare_record" "my_dns_record" {
  # .. snip
  value = "${cloudflare_argo_tunnel.my_tunnel.id}.argotunnel.com"
}
```

AndrewBurian:
@jacobbednarz thanks for the review! While I was testing it I pretty much proved your point that this achieves nothing. I got my wires crossed in how the import ID is represented as the union of account ID and tunnel ID, vs the actual ID just being the tunnel.

I'll refactor this down into a much smaller addition that just exposes the CNAME for convenience.

AndrewBurian:
Couple options for doing this, I could either use a state migration to set the cname record on all resources, since it's deterministic from the ID, or do like I've done here and implement read to fetch and check.

Terrraform plugin best practices are to read after a create/import anyways as an error check, so it's not wrong to do it this way, but maybe it's unnecessary considering it's an extra API call to check the state of an immutable resource. I could refactor this to use a state upgrade instead of the read, but I'm also starting to question if the whole thing is worth the convenience.

Also in testing it seems that terraform doesn't realize that all the fields are immutable so I added the `ForceNew` to all attributes so anything that would change them correctly calls out that it forces a replacement.

cloudflare/resource_cloudflare_argo_tunnel.go
```diff
@@ -97,6 +113,7 @@ func resourceCloudflareArgoTunnelImport(d *schema.ResourceData, meta interface{}
 	}
 
 	d.Set("name", tunnel.Name)
+	d.Set("cname", fmt.Sprintf("%s.argotunnel.com", tunnel.ID))
```
jacobbednarz, line 116: as this is happening in the `Read` and we're calling `resourceCloudflareArgoTunnelRead` at the end of this method, we don't really need this twice.

jacobbednarz:
The change to expose `cname` is 👌, thanks. I'm not sure I follow the following though

> Couple options for doing this, I could either use a state migration to set the cname record on all resources, since it's deterministic from the ID, or do like I've done here and implement read to fetch and check.

If you could elaborate, I'm happy to point you in the right direction.

jacobbednarz:
are you still interested in pursuing this? if so, we need to:
- get a test in place
- add the website documentation
- add a CHANGELOG entry (see [changelog process](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/docs/changelog-process.md) for details)

jacobbednarz:
superseded by #1259

PR #1181, opened 2021-09-08
GATE-2043: Uses teams rule actions enum defined in cloudflare-go

Michael9127:
Just pinning commit until new cloudflare-go version is released.

### DIFF
--- a/cloudflare/resource_cloudflare_teams_rules.go
+++ b/cloudflare/resource_cloudflare_teams_rules.go
@@ -11,21 +11,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 )
 
-var teamsRuleActions = []string{
-	"allow",
-	"block",
-	"safesearch",
-	"ytrestricted",
-	"on",
-	"off",
-	"scan",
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/cloudflare/terraform-provider-cloudflare
 go 1.15
 
 require (
-	github.com/cloudflare/cloudflare-go v0.21.0
+	github.com/cloudflare/cloudflare-go v0.21.1-0.20210905215638-8d67feea1cc0
```
jacobbednarz, line 6: can you please pull this one out? you/i can shim it in locally if we need to confirm it running. we have automated steps for bumping releases and these files will conflict with that.

Michael9127: theres no rush on this, so im fine with it not building for now until the next release. Just wanted to push it up while I was cleaning out the RM.

jacobbednarz:
Depends on cloudflare/cloudflare-go#702

go.mod
jacobbednarz: 🙇‍♂️

Issue #873, opened 2020-11-22
Add resources/configuration to support Tenant use


indygriffiths:
### Current Terraform version

0.13.5

### Description

The Tenant API has support for [creating and removing accounts](https://developers.cloudflare.com/tenant/tutorial/provisioning-resources) from Cloudflare, and the standard API also has support for [updating existing accounts](https://api.cloudflare.com/#accounts-update-account). It would be great if these could be added to support declaring new Cloudflare account and having them created.

As part of this work the `account_id` parameter [defined in the provider](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs#account_id) would most likely need to be added as a parameter to each resource type so that a provider isn't tied to a single account.

### Use-cases

* Cloudflare resellers declaring each Cloudflare account and their properties like zones within Terraform, rather than manually managing them manually or through a custom-built program using the Tenant API

### Potential Terraform configuration

```hcl
resource "cloudflare_account" "example_account" {
    name = "Example Account"
    type = "standard"
}

# Use the new account ID when creating a zone using a new account_id parameter
resource "cloudflare_zone" "example_zone" {
    zone       = "example.com"
    account_id = cloudflare_account.example_account.id
}
```

### Community note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
  to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull
  request, please leave a comment

jacobbednarz:
Thanks for raising this @indygriffiths! Looks like we'll need to add `cloudflare-go` support for this one first before we start using it in Terraform.

jacobbednarz:
While revisiting this, the functionality for Tenants managing Accounts already exist however the change to move the account ID definition into the resources which will be breaking changes. Will need to await for the next major release to address this.

jacobbednarz:
as of #1563 (landed in [v3.13.0](https://github.com/cloudflare/terraform-provider-cloudflare/releases/tag/v3.13.0), https://github.com/cloudflare/cloudflare-go/pull/850 upstream), this should be complete except for some Workers methods which are due for a revamp and won't be addressed due to the internal models and business logic attached.

if you find individual resources still impacted, please raise individual issues and we can work on getting them addressed.

jacobbednarz:
(deprecations in both systems will come with the next major version)

PR #1186, opened 2021-09-10
AUTH-3689: Adds support for access keys configuration

GreenStage:
- Documentation has already been updated. (https://api.cloudflare.com/#access-keys-configuration-properties)
- **Depends on** https://github.com/cloudflare/cloudflare-go/pull/704

### DIFF
--- a/cloudflare/resource_cloudflare_access_keys_configuration.go
+++ b/cloudflare/resource_cloudflare_access_keys_configuration.go
@@ -3,10 +3,10 @@ package cloudflare
 import (
 	"context"
 	"fmt"
+	"log"
+
 	cloudflare "github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
-	"log"
-	"strings"
 )
 
 func resourceCloudflareAccessKeysConfiguration() *schema.Resource {
@@ -39,12 +39,12 @@ func resourceCloudflareAccessKeysConfigurationRead(d *schema.ResourceData, meta 
 
 	keysConfig, err := client.AccessKeysConfig(context.Background(), accountID)
 	if err != nil {
-		if strings.Contains(err.Error(), "HTTP status 404") {
-			log.Printf("[INFO] Access Keys Configuration %s not enabled for account %s", d.Id(), accountID)
+		if err.(*cloudflare.APIRequestError).InternalErrorCodeIs(12109) {
+			log.Printf("[INFO] Access Keys Configuration not enabled for account %s", accountID)
 			d.SetId("")
 			return nil
 		}
-		return fmt.Errorf("error finding Access Keys Configuration %s: %s", d.Id(), err)
+		return fmt.Errorf("error finding Access Keys Configuration %s: %s", accountID, err)
 	}
 
 	d.SetId(accountID)
@@ -74,7 +74,7 @@ func resourceCloudflareAccessKeysConfigurationUpdate(d *schema.ResourceData, met
 
 	_, err := client.UpdateAccessKeysConfig(context.Background(), accountID, keysConfigUpdateReq)
 	if err != nil {
-		return fmt.Errorf("error updating Access Keys Configuration %s for account %s: %s", d.Id(), accountID, err)
+		return fmt.Errorf("error updating Access Keys Configuration for account %s: %s", accountID, err)
 	}
 
 	return resourceCloudflareAccessKeysConfigurationRead(d, meta)
--- a/cloudflare/resource_cloudflare_access_keys_configuration.go
+++ b/cloudflare/resource_cloudflare_access_keys_configuration.go
@@ -1,0 +1,97 @@
+package cloudflare
+
+import (
+	"context"
+	"fmt"
+	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"log"
+	"strings"
+)
+
+func resourceCloudflareAccessKeysConfiguration() *schema.Resource {
+	return &schema.Resource{
+		Read:   resourceCloudflareAccessKeysConfigurationRead,
+		Create: resourceCloudflareAccessKeysConfigurationCreate,
+		Update: resourceCloudflareAccessKeysConfigurationUpdate,
+		Delete: resourceCloudflareKeysConfigDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceCloudflareKeysConfigImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"account_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"key_rotation_interval_days": {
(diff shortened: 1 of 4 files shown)


cloudflare/resource_cloudflare_access_keys_configuration.go
```diff
@@ -0,0 +1,94 @@
...
+			},
+		},
+	}
+}
+
+func resourceCloudflareAccessKeysConfigurationRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	var keysConfig cloudflare.AccessKeysConfig
+	keysConfig, err := client.AccessKeysConfig(context.Background(), client.AccountID)
```
jacobbednarz, line 46: `account_id` should be configurable from the resource to support Tenant usage (#873)

cloudflare/resource_cloudflare_access_keys_configuration.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		if strings.Contains(err.Error(), "HTTP status 404") {
+			log.Printf("[INFO] Access Keys Configuration %s not enabled for account %s", d.Id(), client.AccountID)
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("error finding Access Keys Configuration %s: %s", d.Id(), err)
+	}
+
+	d.SetId(client.AccountID + "_keys_config")
```
jacobbednarz, line 56: `account_id` would be fine here if an end user can only set it at the account level

cloudflare/resource_cloudflare_access_keys_configuration.go
```diff
@@ -0,0 +1,94 @@
...
+		Delete: resourceCloudflareKeysConfigDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceCloudflareAccessMutualTLSCertificateImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"key_rotation_interval_days": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
```
jacobbednarz, line 28: `Computed` and `Optional` are mutually exclusive and you can't combine them. Need to decide on which one better suits the attribute here.

cloudflare/resource_cloudflare_access_keys_configuration.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+func resourceCloudflareAccessKeysConfiguration() *schema.Resource {
+	return &schema.Resource{
+		Read:   resourceCloudflareAccessKeysConfigurationRead,
+		Create: resourceCloudflareAccessKeysConfigurationCreate,
+		Update: resourceCloudflareAccessKeysConfigurationUpdate,
+		Delete: resourceCloudflareKeysConfigDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceCloudflareAccessMutualTLSCertificateImport,
```
jacobbednarz, line 21: Copy pasta? 😛

cloudflare/resource_cloudflare_access_keys_configuration.go
```diff
@@ -0,0 +1,94 @@
...
+			"key_rotation_interval_days": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+			},
+			"last_key_rotation_at": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"days_until_next_rotation": {
```
jacobbednarz, line 34: This is going to need a custom [`DiffSuppressFunc`](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#diffsuppressfunc) as otherwise applying this on multiple days will trigger a `Apply`/`Plan` difference. Otherwise we might need to think about the intended behaviour here.

cloudflare/resource_cloudflare_access_keys_configuration.go
```diff
@@ -0,0 +1,94 @@
...
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceCloudflareAccessKeysConfigurationRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	var keysConfig cloudflare.AccessKeysConfig
```
jacobbednarz, line 45:
```suggestion
```

we can do away with this as `client.AccessKeysConfig` returns a `cloudflare.AccessKeysConfig` type and you assign it to `keysConfig` anyway

cloudflare/resource_cloudflare_access_keys_configuration_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+)
+
+func TestAccessKeysConfiguration_WithKeyRotationIntervalDaysSet(t *testing.T) {
+	name := "cloudflare_access_keys_configuration.test"
```
jacobbednarz, line 11: let's swap this to use the `rnd := generateRandomResourceName()` pattern to avoid hardcoded dependency or resource collisions. example: https://github.com/cloudflare/terraform-provider-cloudflare/blob/a117eb773c303ebbaf657000eb59d8e736809141/cloudflare/resource_cloudflare_magic_firewall_ruleset_test.go#L17-L18

cloudflare/resource_cloudflare_access_keys_configuration_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+func testAccessKeysConfigurationWithKeyRotationIntervalDays(days int) string {
+	return fmt.Sprintf(`
+resource "cloudflare_access_keys_configuration" "test" {
+  key_rotation_interval_days = "%d"
+}`, days)
+}
+
+func TestAccessKeysConfiguration_WithoutKeyRotationIntervalDaysSet(t *testing.T) {
+	name := "cloudflare_access_keys_configuration.test"
```
jacobbednarz, line 37: same as https://github.com/cloudflare/terraform-provider-cloudflare/pull/1186/files#r706915217

cloudflare/resource_cloudflare_access_keys_configuration.go
GreenStage: Sure you dont mean `Computed` and `Required`? 

https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/resource_cloudflare_access_application.go#L30

cloudflare/resource_cloudflare_access_keys_configuration.go
GreenStage: 🙈

cloudflare/resource_cloudflare_access_keys_configuration.go
jacobbednarz: you're right, i had remembered point 2 at https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#optional in reverse

PR #1196, opened 2021-09-15
These should be outside the action_parameters block

thejml:
Local testing shows expression, enabled, and description should be located within the 'rules' block, not inside the 'action_parameters' block. This lines up with the other examples in this file.


jacobbednarz:
Thanks!

PR #1207, opened 2021-09-17
WDAPI-532 - adding support for warp split tunnel

kjhandy:
Adds terraform support for warp split tunnels. Depends on https://github.com/cloudflare/cloudflare-go/pull/709

### DIFF
--- a/cloudflare/provider.go
+++ b/cloudflare/provider.go
@@ -164,6 +164,8 @@ func Provider() terraform.ResourceProvider {
 			"cloudflare_zone_dnssec":                            resourceCloudflareZoneDNSSEC(),
 			"cloudflare_notification_policy":                    resourceCloudflareNotificationPolicy(),
 			"cloudflare_notification_policy_webhooks":           resourceCloudflareNotificationPolicyWebhooks(),
+			"cloudflare_split_tunnel_include":									 resourceCloudflareSplitTunnelInclude(),
+			"cloudflare_split_tunnel_exclude":									 resourceCloudflareSplitTunnelExclude(),
 		},
 	}
 
--- a/cloudflare/resource_cloudflare_split_tunnel_exclude.go
+++ b/cloudflare/resource_cloudflare_split_tunnel_exclude.go
@@ -1,0 +1,127 @@
+package cloudflare
+
+import (
+	"context"
+	"fmt"
+
+	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceCloudflareSplitTunnelExclude() *schema.Resource {
+	return &schema.Resource{
+		Read:   resourceCloudflareSplitTunnelsExcludeRead,
+		Update: resourceCloudflareSplitTunnelsExcludeUpdate,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"account_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"tunnels": {
+				Computed: true,
+				Type:     schema.TypeList,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"address": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Description: "The address for the tunnel.",
+						},
+						"host": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Description: "The domain name for the tunnel.",
+						},
+						"description": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Description: "A description for the tunnel.",
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func resourceCloudflareSplitTunnelsExcludeRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	accountID := d.Get("account_id").(string)
+
+  splitTunnel, err := client.SplitTunnelExclude(context.Background(), accountID)
+	if err != nil {
+		return fmt.Errorf("Error finding Include Split Tunnels %q", err)
+	}
+
+	tunnelList := make(map[string]interface{}, 0)
+	for _, t := range splitTunnel {
+		tunnelList = append(tunnelList, map[string]interface{}{
+			"address":		  t.Address,
+			"host":			    t.Host,
+			"description":	t.Description,
+		})
+	}
+
+	err = d.Set("tunnels", tunnelList)
+	if err != nil {
+		return fmt.Errorf("error setting Include Split Tunnels: %s", err)
+	}
+	return nil
+}
+
+func resourceCloudflareSplitTunnelsExcludeUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	accountID := d.Get("account_id").(string)
+
+	// get all of the existing split tunnels
+  existingSplitTunnel, err := client.SplitTunnelExclude(context.Background(), accountID)
+	if err != nil {
+		return fmt.Errorf("Error finding Include Split Tunnels %q", err)
+	}
+
+	tunnelList := make([]interface{}, 0)
+	for _, t := range existingSplitTunnel {
+		tun := cloudflare.SplitTunnel{}
+		if inputAddress, ok := t.GetOk("address"); ok {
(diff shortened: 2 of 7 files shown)


cloudflare/provider.go
```diff
@@ -164,6 +164,8 @@ func Provider() terraform.ResourceProvider {
 			"cloudflare_zone_dnssec":                            resourceCloudflareZoneDNSSEC(),
 			"cloudflare_notification_policy":                    resourceCloudflareNotificationPolicy(),
 			"cloudflare_notification_policy_webhooks":           resourceCloudflareNotificationPolicyWebhooks(),
+			"cloudflare_split_tunnel_include":									 resourceCloudflareSplitTunnelInclude(),
+			"cloudflare_split_tunnel_exclude":									 resourceCloudflareSplitTunnelExclude(),
```
jacobbednarz, line 168: what benefits do we get from splitting (pun only partially intended 😛 ) these two? would it make more sense to have this as an attribute on a single, shared resource? example:

```hcl
resource "cloudflare_split_tunnel" "..."
  account_id = "..."
  type = "include"
  tunnels {
    host = "example.com"
    description = "..."
  }
}
```

i'm thinking we can combine these into a single resource and just use the parameter to swap the calling method. what do you think?

cloudflare/resource_cloudflare_split_tunnel_exclude.go
```diff
@@ -0,0 +1,127 @@
...
+	"fmt"
+
+	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceCloudflareSplitTunnelExclude() *schema.Resource {
+	return &schema.Resource{
+		Read:   resourceCloudflareSplitTunnelsExcludeRead,
+		Update: resourceCloudflareSplitTunnelsExcludeUpdate,
```
jacobbednarz, line 14: we'll also need to include the `Create` here (even if it reuses the `Update` method) otherwise the resource won't be created in the API or state.

cloudflare/resource_cloudflare_split_tunnel_exclude.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+func resourceCloudflareSplitTunnelsExcludeRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	accountID := d.Get("account_id").(string)
+
+  splitTunnel, err := client.SplitTunnelExclude(context.Background(), accountID)
+	if err != nil {
+		return fmt.Errorf("Error finding Include Split Tunnels %q", err)
```
jacobbednarz, line 57:
```suggestion
		return fmt.Errorf("error finding Include Split Tunnels %q", err)
```

cloudflare/resource_cloudflare_split_tunnel_exclude.go
jacobbednarz: looking further down at the code, we may want to have a separate method 🤔 it looks like your update fetches all the existing values and sends them in the payload too which may not be desirable for a new resource given there will only be _new_ values to send.

cloudflare/resource_cloudflare_split_tunnel_exclude.go
```diff
@@ -0,0 +1,127 @@
...
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"account_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"tunnels": {
+				Computed: true,
```
jacobbednarz, line 25: do we really want this to be marked as `Computed`? i'd expect this would probably need to be an `Required` field otherwise we don't really have a resource to manage.

cloudflare/resource_cloudflare_split_tunnel_exclude.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func resourceCloudflareSplitTunnelsExcludeUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+	accountID := d.Get("account_id").(string)
+
+	// get all of the existing split tunnels
+  existingSplitTunnel, err := client.SplitTunnelExclude(context.Background(), accountID)
+	if err != nil {
+		return fmt.Errorf("Error finding Include Split Tunnels %q", err)
```
jacobbednarz, line 83:
```suggestion
		return fmt.Errorf("error finding Include Split Tunnels: %s", err)
```

cloudflare/resource_cloudflare_split_tunnel_exclude.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	err = d.Set("tunnels", tunnelList)
+	if err != nil {
+		return fmt.Errorf("error setting Include Split Tunnels: %s", err)
+	}
+	return nil
+}
+
+func resourceCloudflareSplitTunnelsExcludeUpdate(d *schema.ResourceData, meta interface{}) error {
```
jacobbednarz, line 76: somewhere (likely towards the end) in this method, we need to call `d.SetID(accountID)` to ensure that Terraform has an identifier in which to compare and validate the changes against. without this identifier, Terraform tests won't ever work as the "anchor" (this ID) is missing from the resource.

cloudflare/resource_cloudflare_split_tunnel_exclude.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+		if inputHost, ok := t.GetOk("host"); ok {
+			tun.Host = inputHost.(string)
+		}
+		if inputDescription, ok := t.GetOk("description"); ok {
+			tun.Description = inputDescription.(string)
+		}
+
+		tunnelList = append(tunnelList, tun)
+	}
```
jacobbednarz, line 100: i see what you're trying to do here but we shouldn't call the API to rebuild payloads for PUT requests. the assumption is that all your data _should_ be in a resource locally (in state) and then you inspect that (`d.Get("tunnels")`) to build the payload. if the local and remote differ, the `Read` method will get them back in sync.

cloudflare/resource_cloudflare_split_tunnel_exclude.go
```diff
@@ -0,0 +1,127 @@
...
+	tunnelList.append(tunnelList, newTunnel)
+
+	err = d.Set("tunnels", tunnelList)
+	if err != nil {
+		return fmt.Errorf("error setting Include Split Tunnels: %s", err)
+	}
+
+	_, updateErr := client.UpdateSplitTunnelExclude(context.Background(), accountID, tunnelList)
+	if updateErr != nil {
+		return fmt.Errorf("Error updating Include Split Tunnels %q", err)
```
jacobbednarz, line 123:
```suggestion
		return fmt.Errorf("error updating Include Split Tunnels: %s", err)
```

jacobbednarz:
i've taken a first pass at this one and can re-review once we determine how we will proceed with cloudflare/cloudflare-go#709.

this looks really solid for a first attempt, kudos! most of the stuff is Terraform nuances and me being nitpicky with error message formatting but hopefully we shouldn't be far off.

cloudflare/resource_cloudflare_split_tunnel_exclude.go
kjhandy: Reading through the docs, is this something that we need on every method? Even the `read` method in their example has `d.SetID()` at the end here: https://learn.hashicorp.com/tutorials/terraform/provider-setup?in=terraform/providers

cloudflare/resource_cloudflare_split_tunnel_exclude.go
kjhandy: Also these docs (https://learn.hashicorp.com/tutorials/terraform/provider-setup?in=terraform/providers) is what gave me the idea that we needed `Computed: true`. I thought because the API doesn't explicitly return a field called `tunnels` that it should have `computed` on it.

cloudflare/provider.go
kjhandy: I was mostly trying to future-proof it in case `include` and `exclude` ever behaved differently. Right now I can't imagine why that would be the case, but that's also often how these things go.

cloudflare/resource_cloudflare_split_tunnel_exclude.go
kjhandy: I was under the impression that `Create` matched up with a `POST` request and we don't have a `POST` request for split tunnels in the API. Am I wrong on that?

PR #1213, opened 2021-09-21
Add `cloudflare_zone` data resource (singular)

cmackenzie1:
This commit adds `cloudflare_zone` as a data source. While support for this technically exists by using `cloudflare_zones` and a `lookup`, I think we should explicitly support this approach as well for providing data about zones.

One concern I have is that the same set of fields should be returned when using `cloudflare_zone` or `cloudflare_zones`. Currently, `cloudflare_zones` only returns `id` and `name`.

### DIFF
--- a/cloudflare/data_source_zone.go
+++ b/cloudflare/data_source_zone.go
@@ -97,7 +97,14 @@ func dataSourceCloudflareZoneRead(d *schema.ResourceData, meta interface{}) erro
 	d.Set("status", zone.Status)
 	d.Set("paused", zone.Paused)
 	d.Set("plan", zone.Plan.Name)
-	d.Set("name_servers", zone.NameServers)
-	d.Set("vanity_name_servers", zone.VanityNS)
+
+	if err := d.Set("name_servers", zone.NameServers); err != nil {
+		return fmt.Errorf("failed to set name_servers attribute: %s", err)
+	}
+
+	if err := d.Set("vanity_name_servers", zone.VanityNS); err != nil {
+		return fmt.Errorf("failed to set vanity_name_servers attribute: %s", err)
+	}
+
 	return nil
 }
--- a/cloudflare/data_source_zone.go
+++ b/cloudflare/data_source_zone.go
@@ -1,0 +1,116 @@
+package cloudflare
+
+import (
+	"context"
+	"fmt"
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"log"
+)
+
+func dataSourceCloudflareZone() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceCloudflareZoneRead,
+
+		Schema: map[string]*schema.Schema{
+			"zone_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"name"},
+			},
+			"account_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"name": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"zone_id"},
+			},
+			"status": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"paused": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+			"plan": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"name_servers": {
+				Type:     schema.TypeList,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Computed: true,
+			},
+			"vanity_name_servers": {
+				Type:     schema.TypeList,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceCloudflareZoneRead(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG] Reading Zones")
+	client := meta.(*cloudflare.API)
+
+	zoneID, zoneIDExists := d.GetOk("zone_id")
+	zoneID = zoneID.(string)
+
+	name, nameExists := d.GetOk("name")
+	name = name.(string)
+
+	accountID := d.Get("account_id")
+	accountID = accountID.(string)
+
+	if nameExists && zoneIDExists {
+		return fmt.Errorf("zone_id and name arguments can't be used together")
+	}
+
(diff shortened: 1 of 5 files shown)


jacobbednarz:
Appreciate the effort here @cmackenzie1! 🙇 

I've been pretty reluctant to add a singular resource for zones largely because many people aren't aware that the zone name attribute isn't unique within Cloudflare. Meaning, if I have 5 accounts, it is possible to have 5 zones for "example.com" within each account, and it's totally acceptable (in fact some customers intentionally do this for migrations!). This makes a singular resource very difficult to use correctly and with any confidence. 

Using my example of 5 accounts with "example.com" and a singular resource, the configuration could look like this:

```hcl
data "cloudflare_zone" "my_zone" {
  name = "example.com"
}
```

Immediately, a bunch of questions come to mind. Which zone ID should be returned here? Do we introduce ordering to weight the zone status higher and just take that? How do we handle similar cases? Do we just error out and advise to be more specific with your lookup? There _are_solutions to some of these questions such as adding filtering to the resource and/or including the account ID in the resource but at that point, we're pretty much using the same functionality in the pluralised data source and the only difference is the 3-4 characters that are a part of the index access. 

At this point, I'm still struggling to see a valid use case for introducing and supporting this as a singular resource given the small improvements we make over the pluralised version. I'll sit on this for another few days and see what I can come up with and whether we introduce this.

cloudflare/data_source_zone.go
```diff
@@ -0,0 +1,109 @@
...
+				Optional: true,
+				Computed: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"status": {
+				Type:     schema.TypeBool,
```
jacobbednarz, line 32: `status` is a string in the [zone struct](https://github.com/cloudflare/cloudflare-go/blob/master/zone.go#L42)

```suggestion
				Type:     schema.TypeString,
```

cloudflare/data_source_zone.go
```diff
@@ -0,0 +1,109 @@
...
+			"account_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
```
jacobbednarz, line 30: looking at the use cases, i don't think it makes sense to ever apply these two together. if you want clarity, zone ID is unique and should be enough. what was your use case here?

```suggestion
			"zone_id": {
				Type:     schema.TypeString,
				Optional: true,
				Computed: true,
				ConflictsWith: []string{"name"},
			},
			"account_id": {
				Type:     schema.TypeString,
				Optional: true,
				Computed: true,
			},
			"name": {
				Type:     schema.TypeString,
				Optional: true,
				Computed: true,
				ConflictsWith: []string{"zone_id"},
			},
```

cloudflare/data_source_zone.go
```diff
@@ -0,0 +1,109 @@
...
+	client := meta.(*cloudflare.API)
+
+	zoneID, zoneIDExists := d.GetOk("zone_id")
+	zoneID = zoneID.(string)
+
+	name, nameExists := d.GetOk("name")
+	name = name.(string)
+
+	if nameExists && zoneIDExists {
+		return fmt.Errorf("zone_id and name arguments can't be used together")
```
jacobbednarz, line 68: heh, you're actually already preventing zone ID and name from being used together like i suggested further down. might just need to update the documentation to reflect this.

cloudflare/data_source_zone.go
```diff
@@ -0,0 +1,109 @@
...
+
+	var zone cloudflare.Zone
+	if nameExists && !zoneIDExists { // if only the name was provided
+		zones, err := client.ListZones(context.Background(), name.(string))
+		if err != nil {
+			return fmt.Errorf("error listing zones: %s", err)
+		}
+
+		if len(zones) >= 1 {
+			return fmt.Errorf("more than one zone was returned")
```
jacobbednarz, line 83:
```suggestion
			return fmt.Errorf("more than one zone was returned; you should consider using `cloudflare_zones` data source with filtering to target the zone more specifically")
```

cloudflare/data_source_zone.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	var zone cloudflare.Zone
+	if nameExists && !zoneIDExists { // if only the name was provided
+		zones, err := client.ListZones(context.Background(), name.(string))
+		if err != nil {
+			return fmt.Errorf("error listing zones: %s", err)
+		}
+
+		if len(zones) >= 1 {
```
jacobbednarz, line 82: the logic here is a little off; having a single record is 👌 

```suggestion
		if len(zones) > 1 {
```

cloudflare/data_source_zone.go
```diff
@@ -0,0 +1,109 @@
...
+		return fmt.Errorf("zone_id and name arguments can't be used together")
+	}
+
+	if !nameExists && !zoneIDExists {
+		return fmt.Errorf("either zone_id or name must be set")
+	}
+
+	var zone cloudflare.Zone
+	if nameExists && !zoneIDExists { // if only the name was provided
+		zones, err := client.ListZones(context.Background(), name.(string))
```
jacobbednarz, line 77: alternatively, you could use `ListZonesContext` here and pass in `WithZoneFilters` with the zone name and account ID if it is set. example: https://github.com/cloudflare/cloudflare-go/blob/b8935550e028d7c46a7adb20ab8c48f8e849bd06/cloudflare.go#L146

cloudflare/data_source_zone.go
cmackenzie1: Oh sweet! I didn't know there was a `ConflictsWith` option

cmackenzie1:
Thanks for the feedback and response @jacobbednarz 🙏! Until you pointed it out I thought zone's were globally unique on Cloudflare.

I incorporated some of the feedback you left here and added the ability to specify an `accountID` to be used when querying for zones. 

I'll hold off on writing tests and such until you decide if singular resource data definitions should be added or not.

cloudflare/data_source_zone.go
jacobbednarz: np! check out everything that is available at https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/schema.go#L196 if you're curious what else a schema can hold

jacobbednarz:
Sorry for the delay in coming back to this @cmackenzie1. I baked on this over the weekend and got caught up with some other reviews. I'm still not fully sold on this as a data source because of the reasoning I mentioned above but at the same time, we have an alternative if people hit those issues _and_ we've made it clear in the documentation there is a potential for getting the wrong data here. 

Let's land this and see how it goes. Thanks for getting the PR together!

PR #1228, opened 2021-09-30
Add data source load balancer pools

wayt:
This adds a Data Source for Load Balancer Pools resources.

i.e:
```
data "cloudflare_load_balancer_pools" "pools" {
}
```

Since there is no actual filter on the API client, the data source doesn't offer any either.

⚠️ this is my first time contributing here, let me know if I missed any guidelines.


cloudflare/data_source_load_balancer_pools_test.go
```diff
@@ -0,0 +1,34 @@
...
+	name := fmt.Sprintf("data.cloudflare_load_balancer_pools.%s", rnd)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCloudflareLoadBalancerPoolsConfig(rnd),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(name, "pools.#", "2"),
```
wayt, line 23: The value here was tested against my own account data, I could not find any information about what to use for upstream acceptance tests.

cloudflare/data_source_load_balancer_pools.go
```diff
@@ -0,0 +1,141 @@
...
+	return &schema.Resource{
+		Read: dataSourceCloudflareLoadBalancerPoolsRead,
+
+		Schema: map[string]*schema.Schema{
+			"pools": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
```
wayt, line 23: I hesitated to directly use `resourceCloudflareLoadBalancerPool().Schema` here, although it doesn't offer the `id` field.

Let me know if you think otherwise.

jacobbednarz: why not append the `id` if you need it?

```go
func dataSourceLoadBalancerPoolSchema() map[string]*schema.Schema {
	schema := resourceCloudflareLoadBalancerPool().Schema

	schema["id"] = *schema.Schema{
		Type:     schema.TypeString,
                Computed: true,
 
	}

	return schema
}

func dataSourceCloudflareLoadBalancerPools() *schema.Resource {
	return &schema.Resource{
		Read: dataSourceCloudflareLoadBalancerPoolsRead,
		Schema: dataSourceLoadBalancerPoolSchema(),
        }
}
```

to be honest though, i'm not sure why you'd need anything other than the `id` on it's own. what is your use case for having everything you've listed?

cloudflare/data_source_load_balancer_pools_test.go
jacobbednarz: you can create them all as a part of your test case and assert against that state. `data_source_zones_test` has a good example https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/data_source_zones_test.go#L228-L267

jacobbednarz:
thanks for solid start here @wayt, it's good! i've got some documentation i'm working on locally for PR minimums but for now, here is the tl;dr for this PR

- website documentation is required (see other data source documentation for examples)
- changelog note (see [changelog-process](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/docs/changelog-process.md))
- passing tests which you've almost got here

cloudflare/data_source_load_balancer_pools.go
jacobbednarz: actually, no, you don't want the `resourceCloudflareLoadBalancerPool` schema as the fields won't line up correctly as a data source 🤦‍♂️ it's geared towards being a managed resource as opposed to a data source.

wayt: > what is your use case for having everything you've listed?

We manage pools outside of Terraform (because of automation). But load balancers are in Terraform, so to avoid Terraform wiping them away we need that kind of data source.

We do some other filtering later within Terraform, unfortunately, that's not possible with the API.

XaF: tbh that would be weird having a data source that does not provide all the information about a pool; different usecases miight need different information from the pools, so might as well have the complete data source right away! (even if it's bound to evolve with lb pools evolutions)

cloudflare/data_source_load_balancer_pools.go
```diff
@@ -0,0 +1,141 @@
...
+		poolIds = append(poolIds, p.ID)
+	}
+
+	if err := d.Set("pools", pools); err != nil {
+		return fmt.Errorf("error setting load balaner pools: %w", err)
+	}
+
+	d.SetId(stringListChecksum(poolIds))
+	return nil
+}
```
XaF, line 141: Would be great to have filtering handled there too.
Even if not needed for the specific usecase here, I believe that it's kind of important to provide ways to filter the pools, either directly from the API request (if the API allows) or with basic pattern matching directly in the provider. Other data resources have similar approaches, waf groups as an example: 
- https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/data_source_waf_groups.go#L29-L46
- https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/data_source_waf_groups.go#L92-L96
- https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/data_source_waf_groups.go#L122-L128

cloudflare/data_source_load_balancer_pools.go
```diff
@@ -0,0 +1,141 @@
...
+			"notification_email": p.NotificationEmail,
+			"load_shedding":      flattenLoadBalancerLoadShedding(p.LoadShedding),
+			"created_on":         p.CreatedOn.Format(time.RFC3339Nano),
+			"modified_on":        p.ModifiedOn.Format(time.RFC3339Nano),
+		})
+		poolIds = append(poolIds, p.ID)
+	}
+
+	if err := d.Set("pools", pools); err != nil {
+		return fmt.Errorf("error setting load balaner pools: %w", err)
```
XaF, line 136:
```suggestion
		return fmt.Errorf("error setting load balancer pools: %w", err)
```

cloudflare/data_source_load_balancer_pools.go
wayt: I can add this if @jacobbednarz agrees here.

jacobbednarz:
@xens i've gone ahead and got this up to speed with the new conventions however, i'll have a play later this week to make sure it's working as expected. feel free to pull it locally and give it a whirl too.

PR #1251, opened 2021-10-14
add access groups data source

Lucretius:
- Access Groups data source, for either a particular zone or across an entire account
- Tests
- Documentation

I have tested this data source against my own Cloudflare account, for both zone_id and account_id.


jacobbednarz:
closing this one out since it has stagnated and both the provider and client library have advanced quite a bit which makes most of this not usable in the newer versions. if someone wants to repick this up and port it to the `terraform-plugin-framework` while taking into account the client improvements, i'm happy to review it and look at getting it into the provider.

PR #1308, opened 2021-11-11
Implicitly include cf.colo.id to the ruleset characteristics

Nmishin:
cf.colo.id implicitly included when using the dashboard, but
this is required as ratelimiting counting is processed at colocation level only.


jacobbednarz:
Thanks for this @Nmishin; appreciate it and apologies for the delay but I'm in and out for the next couple of months, so feedback will be spotty.

I like the intention here however I don't think we should implicitly include this by default as this is a service detail that Terraform _shouldn't_ care about. Could we instead update the documentation or example to mention this is currently a required characteristic?

Nmishin:
Sounds good to me, will check documentation and update if needed, thank you!

PR #1321, opened 2021-11-26
provider: separate schemas into dedicated files

jacobbednarz:
Updates the provider to separate the resource and schema. This unblocks
some possibilities with schema automation, schema reuse between
data sources and resources, and trims down some very large files.

To perform this, I largely relied on [comby](https://comby.dev/) and a couple of hand 
modified files for the outliers.

```
COMBY_M="$(cat <<"MATCH"
func :[methodname]() *schema.Resource {
	return &schema.Resource{
		:[rest]

			:[attrs]
		},
	}
}
MATCH
)"

COMBY_R="$(cat <<"REWRITE"
func :[methodname]() *schema.Resource {
	return &schema.Resource{
		Schema: :[methodname]Schema(),
		:[rest]	
	}
}

func :[methodname]Schema() map[string]*schema.Schema {
	return map[string]*schema.Schema{
			:[attrs]
	}
}
REWRITE
)"

comby "$COMBY_M" "$COMBY_R" .go -matcher .go
```

As this PR is largely just whitespace, I recommend viewing with `?w=1` 
enabled.

I've updated the documentation here too to mention why the file name conventions are 
the way they are and what they actually mean. See https://github.com/cloudflare/terraform-provider-cloudflare/pull/1321/files#diff-977f7a75a20da1290fb7ee0adb2dafa4d442433510400bcb01e2fd4ef95f5af5 for full details.


jacobbednarz:
acceptance tests are all still passing (like `master`).

Issue #1329, opened 2021-12-03
import of cloudflare_argo_tunnel resource fails with invalid API error


tjstansell:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```
# terraform -v
Terraform v1.0.11
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.4.0
+ provider registry.terraform.io/hashicorp/random v3.1.0
```

### Affected resource(s)

cloudflare_argo_tunnel

### Terraform configuration files

```hcl
resource "random_id" "tunnel_secret" {
  byte_length = 32
}

resource "cloudflare_argo_tunnel" "this" {
  account_id = var.account_id
  name       = var.name
  secret     = random_id.tunnel_secret.b64_std
}
```

### Debug output

initial part of import succeeds:
```
module.mytunnel.cloudflare_argo_tunnel.this: Importing from ID "<accountid>/<tunnelid>"...
2021-12-03T07:43:19.903Z [INFO]  provider.terraform-provider-cloudflare_v3.4.0: 2021/12/03 07:43:19 [DEBUG] Cloudflare API Request Details:
---[ REQUEST ]---------------------------------------
GET /client/v4/accounts/<accountid>/tunnels/<tunnelid> HTTP/1.1
Host: api.cloudflare.com
User-Agent: terraform/1.0.11 terraform-plugin-sdk/2.8.0 terraform-provider-cloudflare/3.4.0
Accept: application/json;version=1
Authorization: Bearer <token>
Content-Type: application/json
Accept-Encoding: gzip


-----------------------------------------------------: timestamp=2021-12-03T07:43:19.903Z
2021-12-03T07:43:20.129Z [INFO]  provider.terraform-provider-cloudflare_v3.4.0: 2021/12/03 07:43:20 [DEBUG] Cloudflare API Response Details:
---[ RESPONSE ]--------------------------------------
HTTP/2.0 200 OK
```
but then right after it uses an api call without the accountid:
```
-----------------------------------------------------: timestamp=2021-12-03T07:43:20.128Z
2021-12-03T07:43:20.154Z [INFO]  provider.terraform-provider-cloudflare_v3.4.0: 2021/12/03 07:43:20 [DEBUG] Cloudflare API Request Details:
---[ REQUEST ]---------------------------------------
[...]

PR #1330, opened 2021-12-03
fix cloudflare_argo_tunnel imports

tjstansell:
This should address #1329


jacobbednarz (approved):
looks good! I’ll give this some manual testing in the next day or so and we’ll get it merged. 

cheers!

PR #1359, opened 2021-12-27
Update go.mod

y0k41sec:
correcting the repo

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -79,7 +79,7 @@ require (
 	github.com/chavacava/garif v0.0.0-20210405164556-e8a0a408d6af // indirect
 	github.com/daixiang0/gci v0.2.9 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
-	github.com/denis-tingajkin/go-header v0.4.2 // indirect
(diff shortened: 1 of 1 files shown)


ganzcf (approved):
github user denis-tingajkin has renamed their account to denis-tingaikin

jacobbednarz:
going to hold on this for a week or so to confirm validity of the change.

patryk:
Looks to be valid, GitHub redirects:
```
$ curl https://github.com/denis-tingajkin/go-header -v 2>&1 | grep -i location:
< location: https://github.com/denis-tingaikin/go-header
```

jacobbednarz:
@y0k41sec this is missing the changes to the `go.sum` file. you'll need to add it for us to merge this PR.

y0k41sec:
Hi, it looks like this is an indirect dependency which implies that you are not using this dependency within the project, but there is some module that imports them. I find that go.sum creates a hash, so this is not a vulnerability, there is no risk of downloading a malicious dependency in this case. I'll close this pull request, thanks everyone for taking a look.

PR #1362, opened 2021-12-29
add support for account_member status

chopnico:
I would like to add support for account_member status.

### DIFF
--- a/cloudflare/resource_cloudflare_account_member.go
+++ b/cloudflare/resource_cloudflare_account_member.go
@@ -44,6 +44,7 @@ func resourceCloudflareAccountMemberRead(d *schema.ResourceData, meta interface{
 
 	d.Set("email_address", member.User.Email)
 	d.Set("role_ids", memberIDs)
+	d.Set("status", member.Status)
 	d.SetId(d.Id())
 
 	return nil
@@ -65,6 +66,7 @@ func resourceCloudflareAccountMemberDelete(d *schema.ResourceData, meta interfac
 func resourceCloudflareAccountMemberCreate(d *schema.ResourceData, meta interface{}) error {
 	memberEmailAddress := d.Get("email_address").(string)
 	requestedMemberRoles := d.Get("role_ids").(*schema.Set).List()
+	memberStatus := d.Get("status").(string)
 
 	client := meta.(*cloudflare.API)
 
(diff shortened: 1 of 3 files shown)


cloudflare/resource_cloudflare_account_member.go
```diff
@@ -73,7 +75,8 @@ func resourceCloudflareAccountMemberCreate(d *schema.ResourceData, meta interfac
 		accountMemberRoleIDs = append(accountMemberRoleIDs, roleID.(string))
 	}
 
-	r, err := client.CreateAccountMember(context.Background(), client.AccountID, memberEmailAddress, accountMemberRoleIDs)
+	r, err := client.CreateAccountMemberWithStatus(
```
jacobbednarz, line 78: this will need to be tested on different account types as most cannot set the status using the API (only Tenants from memory). this may need to be reverted.

chopnico: You can invite members from another account and it works just fine but it does bomb if a Cloudflare account doesn't exist. I do not have access to test on Tenants (I learned something new about Cloudflare...) so I'll revert this particular line back. 

Thanks!

jacobbednarz: actually, no need to test, I already did when this landed in cloudflare/cloudflare-go#715 and it’s behind entitlements.

cloudflare/resource_cloudflare_account_member.go
```diff
@@ -83,6 +86,8 @@ func resourceCloudflareAccountMemberCreate(d *schema.ResourceData, meta interfac
 		return fmt.Errorf("failed to find ID in create response; resource was empty")
 	}
 
+	d.Set("status", r.Status)
```
jacobbednarz, line 89: if an end user doesn’t set anything in the status, this is going to cause a permadiff as the result will be "pending" until they accept the invite. following that, once they accept, it will change to "accepted" resulting in a no-op apply to get the state in sync.

cloudflare/resource_cloudflare_account_member.go
chopnico: Solid. So is this a go or should I revert?

jacobbednarz (commented):
I’ve left some notes online as right now, this is going to cause state sync issues. can you elaborate on your use case for this? right now, it’s going to take some custom logic to use it without causing state issues and I’m not sure if it really makes sense to do.

cloudflare/resource_cloudflare_account_member.go
jacobbednarz: we can leave it if we proceed with the PR as Tenants could set this and everyone else will just be an omitted empty string.

PR #862, opened 2020-11-05
add resource 'api_token' and data source 'permission_groups'

UrosSimovic:

PR #1368, opened 2022-01-04
Fix cloudflare_api_token's modified_on field

danopia:
For some reason this field was always set to `time.Now()` when _reading_, which shows as drift in every Terraform plan:

```hcl
Terraform detected the following changes made outside of Terraform since the
last "terraform apply":

  # cloudflare_api_token.external-dns has changed
  ~ resource "cloudflare_api_token" "external-dns" {
        id          = "[redacted]"
      ~ modified_on = "2021-11-05T09:23:00.921389278Z" -> "2022-01-04T14:35:50.581109737Z"
        name        = "external dns"

        # (1 unchanged block hidden)
    }
```

It's always shown as the exact time of the plan, which is definitely not when every token was last modified!

It looks like the swap happened within the PR adding this resource: https://github.com/cloudflare/terraform-provider-cloudflare/pull/862/commits/02bb4db2805a2c3d740ed7e432f0717af463af0a so as an outsider I don't have an explanation for why `time.Now()` was inserted.

### DIFF
--- a/.changelog/1368.txt
+++ b/.changelog/1368.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_api_token: modified_on is now read correctly
+```
--- a/.golintci.yml
+++ b/.golintci.yml
@@ -1,28 +1,17 @@
 run:
-  # timeout for analysis, e.g. 30s, 5m, default is 1m
   timeout: 1m
-
-  # exit code when at least one issue was found, default is 1
   issues-exit-code: 1
-
-  # include test files or not, default is true
   tests: true
-
-  # default is true. Enables skipping of directories:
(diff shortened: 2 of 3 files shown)


jacobbednarz:
thanks for the PR! this needs a [CHANGELOG entry](https://github.com/cloudflare/terraform-provider-cloudflare/actions/workflows/lint.yml) added before we can merge this.

danopia:
I added a changelog, looks like they want to be one line otherwise I would've mentioned how it fixes `Objects have changed outside of Terraform` messages.

I also took the time to figure out how to use a custom provider build locally, and this patch does actually work & resolve the issue :)

jacobbednarz:
thank you! i'm really glad you took the time to go through those steps and got a better understanding of how this all works together 😄

Issue #1381, opened 2022-01-13
cloudflare_fallback_domain docs and usage are misleading and are mis-named IMO


tjstansell:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

latest

### Affected resource(s)

cloudflare_fallback_domain

### Terraform configuration files

```hcl
none
```

### Debug output

none

### Expected output

none

### Actual output

none

### Steps to reproduce

none

### Additional factoids

The new `cloudflare_fallback_domain` resource seems backwards in the use of plural attributes and such.

The documentation has: 
```hcl
# Adding example.com to the fallback domain list
resource "cloudflare_fallback_domain" "example_fallback_domain {
  account_id = "1d5fdc9e88c8a8c4518b068cd94331fe"
  domains {
    suffix = "example.com"
    description = "Example domain"
    dns_server = ["1.1.1.1", "2.2.2.2"]
  }
}
```

This is not accurate as there is a single API endpoint where you put the *entire list of fallback domains*.  This example is not adding (appending) just a "single" domain to the list.  It's the entire set of domains.  As such, the resource should really be called `cloudflare_fallback_domains` (since it's the entire set of domains) and the blocks inside should be called `domain {}` since it specifies the config for a single domain.  The example should show multiple:

```hcl
# adding example.com along with the defaults
resource "cloudflare_fallback_domains" "example" {
  account_id = "1d5fdc9e88c8a8c4518b068cd94331fe"
  dynamic "domain" {
    for_each = toset({"intranet", "internal", "private", "localdomain", "domain", "lan", "home", "host", "corp", "local", "localhost", "home.arpa", "invalid", "test"})
    content {
      suffix = domain.value
    }
  }
  domain {
    suffix = "example.com"
    description = "Example domain"
    dns_server = ["1.1.1.1", "2.2.2.2"]
  }
}

```

jacobbednarz:
I can see what you mean but this was the decision of the service team to model it the way they have. I’ll raise the discussion with them but closing as it was intentional and nothing to do from our side.

tjstansell:
How is this a "decision of the service team"?  Don't you maintain cloudflare-go and the terraform provider code?

jacobbednarz:
[...]

PR #1399, opened 2022-01-19
cloudflare_fallback_domain support for default entries

### LINKED ISSUE #1385: cloudflare_fallback_domain should support default entries so deleting resource does not remove all entries
### Confirmation

Terraform v1.1.3
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.7.0

### Affected resource(s)

cloudflare_fallback_domain

```hcl
resource "cloudflare_fallback_domain" "this" {
  account_id = local.account_id
  domains {
    suffix = "example.com"
    description = "Example domain"
    dns_server = ["1.1.1.1", "2.2.2.2"]
  }
}
```

### Debug output

n/a

### Panic output

_No response_

### Expected output

Per the documentation, I would expect this `example.com` domain to get added to my default set.  Instead, it replaces all domains including all of the default ones.

### Actual output

A plan shows that all other domains are removed:

```
  # cloudflare_fallback_domain.this will be updated in-place
  ~ resource "cloudflare_fallback_domain" "this" {
        id         = "<redacted>"

      ~ domains {
          + description = "Example domain"
          ~ dns_server  = [
              + "1.1.1.1",
              + "2.2.2.2",
            ]
          ~ suffix      = "intranet" -> "example.com"
        }
      - domains {
          - dns_server = [] -> null
          - suffix     = "internal" -> null
        }
      - domains {
          - dns_server = [] -> null
          - suffix     = "private" -> null
        }
<truncated>
```

### Steps to reproduce

1. create resource

Only the domains specified are included.

2. delete resource

The set of fallback domains is empty (with no way to reset to the default set of entries).

### Additional factoids

There are actually two problems here, but they're related.

1. The documentation is misleading (it implies a domain is added to existing settings), but also there's no handling of the "default entries" as they're called in the UI.  There should be an additional parameter `include_default_entries` that defaults to `true` and appends the set of domains in the resource to the default entries so that this resource behaves more how folks would assume it behaves (and how the documention implies).
2. When the resource is deleted, the set of domains is emptied instead of reverting back to the default set.  You should be able to control this behavior by enabling/disabling a setting.

### References

_No response_

tjstansell:
This should close #1385 and adds two new options (enabled by default) to support the default entries provided by teams initially and to support restoring those entries on resource deletion.  This updates the behavior to align with the docs and the intended behavior most folks expect, while providing control to customize the behavior.


cloudflare/resource_cloudflare_fallback_domain.go
```diff
@@ -3,11 +3,20 @@ package cloudflare
...
 	"fmt"
+	"log"
 
 	cloudflare "github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+// use this since we cannot use const for lists.
+func default_domains() []string {
+	return []string{
```
jacobbednarz, line 14: @kjhandy i can see internally we have this stored; is it something we plan on exposing via the public API to save storing this in Terraform?

cloudflare/resource_cloudflare_fallback_domain.go
```diff
@@ -3,11 +3,20 @@ package cloudflare
...
 	"context"
 	"fmt"
+	"log"
 
 	cloudflare "github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+// use this since we cannot use const for lists.
+func default_domains() []string {
```
jacobbednarz, line 13: not sure why the linter didn't pick this up but we should use camel casing in Go where the method isn't exported.

```suggestion
func fallbackDefaultDomains() []string {
```

cloudflare/schema_cloudflare_fallback_domain.go
```diff
@@ -10,14 +10,29 @@ func resourceCloudflareFallbackDomainSchema() map[string]*schema.Schema {
 			Type:     schema.TypeString,
 			Required: true,
 		},
+		"include_default_domains": {
+			Optional: true,
+			Type:     schema.TypeBool,
+			Default:  true,
+		},
+		"restore_default_domains_on_delete": {
```
jacobbednarz, line 18: i'm not sure if this attribute makes sense to exist. from the service perspective, the resource is always present we are just modifying it (even before it is configured by the end user). why would we not just update the `Delete` method to send only the default domains instead?

cloudflare/resource_cloudflare_fallback_domain.go
jacobbednarz: alternatively, a `var` will also work here.

cloudflare/resource_cloudflare_fallback_domain.go
```diff
@@ -18,6 +27,7 @@ func resourceCloudflareFallbackDomain() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: resourceCloudflareFallbackDomainImport,
 		},
+		CustomizeDiff: resourceCloudflareFallbackDomainDiff,
```
jacobbednarz, line 30: we should avoid `CustomizeDiff` where we can as it is pretty riddled with edge cases for certain types like `TypeMap` and behaviours on computed fields. it's not really something we want more of.'

if we're using the `CustomizeDiff` here for including the default fallback domains in addition to custom values and avoid permadiffs, we could instead make end users explicitly provide them which feels like expected behaviour given we are customising the already existing resource with their own domains.

jacobbednarz: a note in the documentation mentioning this for the end user seems like it would address the surprise without any additional work in the provider itself.

cloudflare/resource_cloudflare_fallback_domain.go
tjstansell: I saw the teams api endpoint (api.teams.cloudflare.com) allowed `&reset_defaults=true` during a `DELETE` call, but the public API doesn't allow this.  So, I opted to just duplicate it here.

jacobbednarz:
thanks for the PR! i've gone through and added comments inline but this PR does raise some provider design questions which i don't think we've quite ironed out in the current state.

the two big things:

- `restore_default_domains_on_delete` as an attribute doesn't really make a whole heap of sense if it is the default state, and we return to that state upon no longer managing the resource. like zone settings, we can just set those values in the `Delete` operation back to the original state and leave it at that.
- the `CustomizeDiff` behaviour seems to be trying to do too much implicitly for the end user at the risk of introducing bugs through the types weirdness and incompatibilities. i would recommend a documentation update over this approach mentioning the values need to be explicitly provided should end users wish to include any of the default domains (favouring explicitness in managing the resource).

cloudflare/resource_cloudflare_fallback_domain.go
jacobbednarz: i'll chat to some folks and see what the intention is there. that would make more sense to get publicly supported so the service is the source of truth for this list instead of Terraform.

cloudflare/resource_cloudflare_fallback_domain.go
tjstansell: Yeah. I started this by just updating Delete to reset the entries back to the default set, and then figured, well, if we already have to store the list internally and the existing docs all imply you're _adding_ domains, might as well support the currently documented behavior.  I also tried to point out this documentation problem in #1381, but you immediately closed that issue.

tjstansell: And without `CustomizeDiff`, we'd be able to change the type to `TypeSet` instead, which is actually more appropriate, since the order doesn't matter.

cloudflare/resource_cloudflare_fallback_domain.go
jacobbednarz: cool, confirmed with the team this is 👌 if we add it to cloudflare-go, we can do away with this list entirely.

cloudflare/resource_cloudflare_fallback_domain.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+var default_domains = []string{
```
jacobbednarz, line 11: we can do away with this now because i added support for this via cloudflare/cloudflare-go#784

cloudflare/resource_cloudflare_fallback_domain.go
```diff
@@ -61,7 +66,7 @@ func resourceCloudflareFallbackDomainDelete(d *schema.ResourceData, meta interfa
 	client := meta.(*cloudflare.API)
 	accountID := d.Get("account_id").(string)
 
-	_, err := client.UpdateFallbackDomain(context.Background(), accountID, nil)
+	_, err := client.UpdateFallbackDomain(context.Background(), accountID, getDefaultDomains())
```
jacobbednarz, line 69:
```suggestion
	err := client.RestoreFallbackDomainDefaults(context.Background(), accountID)
```

jacobbednarz:
pending cloudflare/cloudflare-go#784

cloudflare/resource_cloudflare_fallback_domain.go
tjstansell: great.  any idea when that will be functional in the public api?  doesn't seem to work in my testing.

jacobbednarz: public documentation is incoming however it already works. it is a DELETE request with the query parameters, and an empty JSON array just as the browser does.

anything that works via the UI has to work with the API as they are serviced from the same infrastructure.

tjstansell: well, crap. yeah, i was using `&` instead of `?` to separate my query parameters in the curl test ... and the fact it requires a valid json body seemed odd too... but whatever. it does work.

PR #1420, opened 2022-02-01
cloudflare_access_identity_provider import did not include config block

### LINKED ISSUE #1419: `cloudflare_access_identity_provider` does not import config block settings
### Confirmation

# terraform -v
Terraform v1.1.4
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.8.0

### Affected resource(s)

* cloudflare_access_identity_provider

```hcl
resource "cloudflare_access_identity_provider" "okta" {
  account_id = local.account_id
  name       = "Okta SAML"
  type       = "saml"
  config {
    attributes           = ["email", "groups"]
    issuer_url           = var.issuer_url
    sso_target_url       = var.sso_target_url
    email_attribute_name = ""
    sign_request         = false
    idp_public_cert      = local.okta_sso_cert
  }
}
```

### Debug output

Notice how the result from the API call includes a `config {}` block.  I would expect this to get imported.
```
-----------------------------------------------------: timestamp=2022-01-31T23:57:51.029Z
2022-01-31T23:57:51.262Z [INFO]  provider.terraform-provider-cloudflare_v3.8.0: 2022/01/31 23:57:51 [DEBUG] Cloudflare API Response Details:
---[ RESPONSE ]--------------------------------------
HTTP/2.0 200 OK
Cf-Cache-Status: DYNAMIC
Cf-Ray: 6d66e27b3c4e7118-SJC
Content-Security-Policy: frame-ancestors 'none'; default-src https: 'unsafe-inline'
Content-Type: application/json; charset=UTF-8
Date: Mon, 31 Jan 2022 23:57:51 GMT
Expect-Ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
Server: cloudflare
Strict-Transport-Security: max-age=31536000; includeSubDomains
Vary: Accept-Encoding
X-Content-Type-Options: nosniff
X-Envoy-Upstream-Service-Time: 4
X-Frame-Options: DENY
X-Xss-Protection: 1; mode=block

{
  "result": {
    "id": "1241a7e1-92ad-4c9d-9e97-22024662201f",
    "type": "saml",
    "uid": "1241a7e1-92ad-4c9d-9e97-22024662201f",
    "name": "Okta SAML",
    "config": {
      "acs_url": "<REDACTED>",
      "attributes": [
        "email",
"groups"
      ],
      "email_attribute_name": "",
      "header_attributes": null,
      "idp_public_cert": "**********************************",
      "issuer_url": "<REDACTED>",
      "sign_request": false,
      "sso_target_url": "<REDACTED>"
    },
    "version": "a261190c440acc4d945728355612ab11"
  },
  "success": true,
  "errors": [],
  "messages": []
}

-----------------------------------------------------: timestamp=2022-01-31T23:57:51.262Z
cloudflare_access_identity_provider.okta: Import prepared!
  Prepared cloudflare_access_identity_provider for import
cloudflare_access_identity_provider.okta: Refreshing state... [id=1241a7e1-92ad-4c9d-9e97-22024662201f]
```

### Panic output

_No response_

### Expected output

I expected the existing saml `config {}` block to be imported into the state file.

### Actual output

The `config {}` block of settings was not imported.
```
# terraform import cloudflare_access_identity_provider.okta <accountid>/1241a7e1-92ad-4c9d-9e97-22024662201f
cloudflare_access_identity_provider.okta: Importing from ID "<REDACTED>/1241a7e1-92ad-4c9d-9e97-22024662201f"...
cloudflare_access_identity_provider.okta: Import prepared!
  Prepared cloudflare_access_identity_provider for import
cloudflare_access_identity_provider.okta: Refreshing state... [id=1241a7e1-92ad-4c9d-9e97-22024662201f]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.

Releasing state lock. This may take a few moments...
# terraform state show cloudflare_access_identity_provider.okta
# cloudflare_access_identity_provider.okta:
resource "cloudflare_access_identity_provider" "okta" {
    account_id = "<REDACTED>"
    id         = "1241a7e1-92ad-4c9d-9e97-22024662201f"
    name       = "Okta SAML"
    type       = "saml"
}
```

### Steps to reproduce

import any identity providers with a config block.

### Additional factoids

_No response_

### References

_No response_

tjstansell:
This should close #1419.  When importing, we need a stub config block so that Read will look for an existing one.

### DIFF
--- a/.changelog/1420.txt
+++ b/.changelog/1420.txt
@@ -1,4 +1,4 @@
 ```release-note:bug
-resource/cloudflare_access_identity_provider: import config block settings
+resource/cloudflare_access_identity_provider: import now includes config block settings
 ```
 
--- a/.changelog/1420.txt
+++ b/.changelog/1420.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+resource/cloudflare_access_identity_provider: import config block settings
+```
+
--- a/cloudflare/resource_cloudflare_access_identity_provider.go
+++ b/cloudflare/resource_cloudflare_access_identity_provider.go
@@ -167,6 +167,11 @@ func resourceCloudflareAccessIdentityProviderImport(d *schema.ResourceData, meta
 
 	log.Printf("[DEBUG] Importing Cloudflare Access Identity Provider: accountID=%s accessIdentityProviderID=%s", accountID, accessIdentityProviderID)
 
+	// create a non-empty config here so Read will import any existing configs
+	m := map[string]interface{}{"support_groups": true}
+	config := []interface{}{m}
+	d.Set("config", config)
+
 	d.Set("account_id", accountID)
 	d.SetId(accessIdentityProviderID)
 


jacobbednarz (changes requested):
I'm afraid I can't merge this; the "fix" here is just covering the actual bug which lies in the [`convertStructToSchema` method](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/resource_cloudflare_access_identity_provider.go#L215-L217)

The `convertStructToSchema` shouldn't be looking at the existing schema, instead it should be pulling from the `cloudflare.AccessIdentityProviderConfiguration` that is being passed in.

jacobbednarz:
Let's also add a regression test here via `ImportStateVerify` to prevent this being missed in any future changes. It will also give us a way to validate the fix actually works.

tjstansell:
Since it seems like there are several instances like this that are looking at the `ResourceData` to decide what to return, I assumed that it was basically "ignoring" the config on the resource side if you didn't configure it locally.  If that's not the case and it should always return the full resource config, we can change how this PR works.  I figured this was the least impact just to get imports to work properly.

jacobbednarz:
definitely not intentionally ignoring those values. there is documentation on how the `flatten*` and `expand*` style methods (which we are using here, just with a different naming convention) should be used https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/docs/data-handling-and-processing.md

PR #1421, opened 2022-02-01
cloudflare_access_application does not import cors_headers block

tjstansell:
This is similar to #1420 where optional blocks are not getting imported.


jacobbednarz (changes requested):
same issue as #1420; this is just masking the underlying problem of reading the data from the wrong place.

PR #1425, opened 2022-02-03
Fix api tokens issue with premission_groups

tete17:
This MR is a superseed of https://github.com/cloudflare/terraform-provider-cloudflare/pull/1089

I went ahead and added the state migration that @jacobbednarz requested. I am still a bit unsure whether I did the right thing.

The sdk documentation doesn't make it very clear that the rawState argument of the migration function is indeed a *schema.Set. I went with that assumption please correct me if wrong.

I also had to do some tinkering in order to unit test the migration function since the structure `*schema.Set` contains a non nil function and `DeepEqual` can't cope with that. It is my first MR regarding terraform code so I am all open to suggestions and feedback.

### DIFF
--- a/.changelog/1425.txt
+++ b/.changelog/1425.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_api_token: Avoid recreating the resources when multiple `permission_groups` exist.
+```
--- a/.changelog/1425.txt
+++ b/.changelog/1425.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-resource/cloudflare_api_token: Avoid recreating the resources when multiple `permission_groups` exist.
+resource/cloudflare_api_token: ignore ordering of `permission_group` IDs
 ```
--- a/cloudflare/resource_cloudflare_api_token_test.go
+++ b/cloudflare/resource_cloudflare_api_token_test.go
@@ -42,6 +42,29 @@ func TestAccAPIToken(t *testing.T) {
 	})
 }
 
+func TestAccCloudflareAPITokenMultiplePermissionsGroups(t *testing.T) {
+	// Temporarily unset CLOUDFLARE_API_TOKEN if it is set as the API token
+	// endpoint does not yet support the API tokens without an explicit scope.
+	if os.Getenv("CLOUDFLARE_API_TOKEN") != "" {
+		defer func(apiToken string) {
+			os.Setenv("CLOUDFLARE_API_TOKEN", apiToken)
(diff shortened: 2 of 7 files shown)


.changelog/1425.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_api_token: Avoid recreating the resources when multiple `permission_groups` exist.
```
jacobbednarz, line 2: i think this is slightly confusing as we're not fixing multiple cases of `permission_groups` but instead, ignoring the ordering change in the permission IDs that are returned from the API.

```suggestion
resource/cloudflare_api_token: ignore ordering of `permission_group` IDs
```

cloudflare/resource_cloudflare_api_token_test.go
```diff
@@ -42,6 +42,29 @@ func TestAccAPIToken(t *testing.T) {
...
+
+	rnd := generateRandomResourceName()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCloudflareAPITokenMultiplePermissionsGroups(rnd),
+			},
```
jacobbednarz, line 63: for this to check the behaviour we're changing, we need another step to run a `diff` on the existing configuration too. we should just be able to duplicate the existing config and swap the values around. (no need to assert `ExpectEmptyPlan: true` as that is the default)

```suggestion
			{
				Config: testAccCloudflareAPITokenMultiplePermissionsGroups(rnd),
			},
			{
				Config: testAccCloudflareAPITokenMultiplePermissionsGroupsReordered(rnd),
			},
```

jacobbednarz: fixed in c851cf3f84384bc0fe381235e23360828ffe15fd

jacobbednarz:
thanks for getting this one over the line @tete17! i made a few updates to the test suite but other than that, this looks great! thanks again.

tete17:
I think this wasn't ready to be merged 
image

jean-mercier-hivebrite:
Same error on my side

jacobbednarz:
both test suites were passing and this looked complete; hence the merge. by the looks of things we’re not exercise enough of the code path in the unit test to catch this. I’ll revert with a hot fix and you can resubmit the PR with the full fix and an updated test.

jacobbednarz:
released in v3.9.1 - feel free to submit a PR addressing this one again.

PR #1444, opened 2022-02-11
Add a `cloudflare_zone_cache_variants` resource

maleblond:
We'd like to be able to manage [cache variants](https://api.cloudflare.com/#zone-cache-settings-get-variants-setting) through terraform so I added a.`cloudflare_zone_cache_variants` resource.

I added some unit tests + I did some manual testing. A `cloudflare_zone_cache_variants` resource would look like this:

```
resource "cloudflare_zone_cache_variants" "variants" {
  zone_id = <zone_id>
  gif     = ["image/avif", "image/webp", "image/gif"]
  webp    = ["image/whatever"]
  ...
}
```

It's my first time updating a terraform provider, so there may be some very ugly things in there not following the best practices 😅

### DIFF
--- a/cloudflare/resource_cloudflare_zone_cache_variants.go
+++ b/cloudflare/resource_cloudflare_zone_cache_variants.go
@@ -13,7 +13,7 @@ import (
 func resourceCloudflareZoneCacheVariants() *schema.Resource {
 	return &schema.Resource{
 		Schema: resourceCloudflareZoneCacheVariantsSchema(),
-		Create: resourceCloudflareZoneCacheVariantsCreate,
+		Create: resourceCloudflareZoneCacheVariantsUpdate,
 		Read:   resourceCloudflareZoneCacheVariantsRead,
 		Update: resourceCloudflareZoneCacheVariantsUpdate,
 		Delete: resourceCloudflareZoneCacheVariantsDelete,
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/aws/aws-sdk-go v1.40.35 // indirect
 	github.com/bflad/tfproviderlint v0.28.1
 	github.com/client9/misspell v0.3.4
-	github.com/cloudflare/cloudflare-go v0.31.0
```
jacobbednarz, line 13: please remove these; we don’t bump version dependencies in the same PR. 

this is also already in master.

cloudflare/schema_cloudflare_zone_cache_variants.go
```diff
@@ -0,0 +1,42 @@
...
+func resourceCloudflareZoneCacheVariantsSchema() map[string]*schema.Schema {
+	variantsSchema := map[string]*schema.Schema{
+		"zone_id": {
+			Type:     schema.TypeString,
+			Required: true,
+			ForceNew: true,
+		},
+	}
+
+	for _, ext := range supportedCacheVariantsExtensions {
```
jacobbednarz, line 30: not really sure how i feel about the dynamic nature of creating schema items due to making it difficult for discovery. i'll have a think about this.

cloudflare/resource_cloudflare_zone_cache_variants.go
```diff
@@ -0,0 +1,151 @@
...
+	"strings"
+
+	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceCloudflareZoneCacheVariants() *schema.Resource {
+	return &schema.Resource{
+		Schema: resourceCloudflareZoneCacheVariantsSchema(),
+		Create: resourceCloudflareZoneCacheVariantsCreate,
```
jacobbednarz, line 16: given create and update are essentially the same funcitonality, you could point `Create` and `Update` here to the same method.

maleblond: :+1: I merged both function

cloudflare/schema_cloudflare_zone_cache_variants.go
maleblond: :+1: My main goal was not to repeat the whole schema over and over again. An alternative would be to do something like:

```
"avif": extensionSchema(),
"gif": extensionSchema(),
```

I think an upside with this alternative is that the schema is a little bit easier to read. I'll do that, LMK what you think

Issue #1128, opened 2021-07-12
resource/cloudflare_record: `allow_overwrite` fails to find root domain when `@` supplied as name variable


MrLemur:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

terraform v1.0.2
+ provider registry.terraform.io/cloudflare/cloudflare v2.23.0 (pulled latest version as of today)

### Affected resource(s)

cloudflare_record

### Terraform configuration files

```hcl
resource "cloudflare_record" "example_com" {
  zone_id         = module.zone.id
  name            = "@"
  value           = "google.com"
  type            = "CNAME"
  allow_overwrite = true
}
```

### Debug output

2021-07-12T17:39:45.308+0100 [DEBUG] provider.terraform-provider-cloudflare_99.0.0: -----------------------------------------------------
2021-07-12T17:39:45.308+0100 [DEBUG] provider.terraform-provider-cloudflare_99.0.0: 2021/07/12 17:39:45 [DEBUG] Cloudflare API Request Details:
2021-07-12T17:39:45.308+0100 [DEBUG] provider.terraform-provider-cloudflare_99.0.0: ---[ REQUEST ]---------------------------------------
2021-07-12T17:39:45.308+0100 [DEBUG] provider.terraform-provider-cloudflare_99.0.0: GET /client/v4/zones/xxxxxxxxxxxx/dns_records?name=%40.example.com&page=1&per_page=100&type=CNAME HTTP/1.1
2021-07-12T17:39:45.308+0100 [DEBUG] provider.terraform-provider-cloudflare_99.0.0: Host: api.cloudflare.com
2021-07-12T17:39:45.308+0100 [DEBUG] provider.terraform-provider-cloudflare_99.0.0: User-Agent: HashiCorp Terraform/1.0.1 (+https://www.terraform.io) Terraform Plugin SDK/1.17.2 terraform-provider-cloudflare/dev
2021-07-12T17:39:45.308+0100 [DEBUG] provider.terraform-provider-cloudflare_99.0.0: Authorization: Bearer xxxxxxxxxxxx
2021-07-12T17:39:45.308+0100 [DEBUG] provider.terraform-provider-cloudflare_99.0.0: Content-Type: application/json
[...]

PR #1468, opened 2022-02-17
fix: allow_override when full domain name is used

debendraoli:
This will fix the issue where `allow_overwrite` will fail to insert the record when full domain name is passed.

ref:
https://github.com/cloudflare/terraform-provider-cloudflare/issues/1128


debendraoli:
> 

updated

debendraoli:
> Oops! It looks like no changelog entry is attached to this PR. Please include a release note as described in https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/docs/changelog-process.md.
> 
> Example:
> 
> ```
> ```release-note:TYPE
> Release note
> ```
> ```
> 
> If you do not require a release note to be included, please add the `workflow/skip-changelog-entry` label.

Updated.

jacobbednarz:
can you please elaborate on the bug you're seeing here? this is currently working with full domains so we need to be specific about what we're fixing.

debendraoli:
`allow_overwrite` will fail when full domain `subdomain.domain.tld` is used.

Example:

Overwrite fails:

name:  `subdomain.example.com`

Suceceeds:

name: `subdomain`

PR #1405, opened 2022-01-24
resource/cloudflare_ruleset: Yet Another tristate bool fix

jacobbednarz:
Updates the check for whether or not to populate the `enabled` flag on
category/rule based overrides to use the `GetOkExists` method (not
recommended due to undocumented edge cases) to check the existence
before converting the pointer to a value.

Closes #1403

PR #1492, opened 2022-03-04
resource/cloudflare_ruleset: fix ruleset category overrides in DDoS L7 phase by allowing the provider to omit the `enabled` flag, as required by Cloudflare's API

Quentin-M:
In the DDoS L7 phase, category overrides do not support the `enabled` flag which must therefore be omitted. This is pretty closely related to https://github.com/cloudflare/terraform-provider-cloudflare/pull/1405. 

### Terraform example

```
resource "cloudflare_ruleset" "example" {
  zone_id     = cloudflare_zone.example.id
  name        = "zone_managed_ddos"
  description = "Zone-level Cloudflare DDoS Managed Rulesets"
  kind        = "zone"
  phase       = "ddos_l7"

  // DDoS L7 ruleset
  rules {
    action      = "execute"
    description = "Override DDoS L7 Ruleset"
    expression  = "true"
    enabled     = true

    action_parameters {
      id = "4d21379b4f9f4bb088e0729962c8b3cf"

      overrides {
        // Enable all unusual request rules
        categories {
          category = "unusual-requests"
          action   = "challenge"
        }
```

### API Request (sample) & error before this PR

```
                        "categories": [
                          {
                            "action": "challenge",
                            "category": "unusual-requests",
                            "enabled": false
                          },
                          {
                            "action": "challenge",
                            "category": "botnets",
                            "enabled": false
                          }
                        ],
```

Leading to the following `400 Bad Request` error returned by the Cloudflare API (doubled, because two categories):

```
    DDoS L7 overrides cannot use the enabled field, DDoS L7 overrides cannot use the enabled field
    DDoS L7 overrides cannot use the enabled field, DDoS L7 overrides cannot use the enabled field
```

### API Request (sample) after this PR

```
     "categories": [
      {
       "category": "unusual-requests",
       "action": "challenge"
      },
      {
       "category": "botnets",
       "action": "challenge"
      }
     ],
```
 
### Notes

- This PR requires the following cloudflare-go schema: https://github.com/cloudflare/cloudflare-go/pull/824
- This PR also includes a second commit to fix a panic encountered in resourceCloudflareRulesetStateUpgradeV0ToV1 during testing

### DIFF
--- a/.changelog/1492.txt
+++ b/.changelog/1492.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+resource/cloudflare_ruleset: fix category overrides for ddos_l7 phase
+```
+
--- a/cloudflare/resource_cloudflare_ruleset.go
+++ b/cloudflare/resource_cloudflare_ruleset.go
@@ -460,12 +460,19 @@ func buildRulesetRulesFromResource(d *schema.ResourceData) ([]cloudflare.Ruleset
 
 							// Category based overrides
 							if val, ok := overrideParamValue.(map[string]interface{})["categories"]; ok {
-								for _, category := range val.([]interface{}) {
+								for categoryOverrideCounter, category := range val.([]interface{}) {
 									cData := category.(map[string]interface{})
+
+									var enabled *bool
+									//nolint:staticcheck
+									if value, ok := d.GetOkExists(fmt.Sprintf("rules.%d.action_parameters.0.overrides.%d.categories.%d.enabled", rulesCounter, overrideCounter, categoryOverrideCounter)); ok {
+										enabled = &[]bool{value.(bool)}[0]
+									}
+
 									categories = append(categories, cloudflare.RulesetRuleActionParametersCategories{
 										Category: cData["category"].(string),
 										Action:   cData["action"].(string),
-										Enabled:  cData["enabled"].(bool),
+										Enabled:  enabled,
 									})
 								}
 							}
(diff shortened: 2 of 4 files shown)


Quentin-M:
Resource creation now works appropriately, however, it appears that `"enabled": false` still gets written to the Terraform state despite being skipped in `buildRulesetRulesFromResource`, and never returned by the Cloudflare API - which makes updates break as the boolean ends up getting into the request during updates. I am not 100% sure how it gets written to state at this point.

jacobbednarz:
FWIW, this isn’t the first time we’ve tangled with tristate booleans…I have a SDK bug open to address exactly this (https://github.com/hashicorp/terraform-plugin-sdk/issues/817)

Quentin-M:
Yeah, I've been reading that thread while trying to fix up the provider, looks fundamentally broken. I am not 100% sure how the SDK/cty ends up writing a `*Bool{nil}` as `bool{false}` but it did not seem like a great use of time to dive deep into it, I just assumed TypeBool is just a two-state bool and therefore we're losing information on state save/read.. making it basically impossible to determine later on whether we need to include it in our API requests or not, and thus breaking `ddos_l7` updates. From my past experience facing Terraform core issues, it is most likely going to be significantly faster and probably much more robust to fix the problem within terraform-provider-cloudflare by patching up a TypeString<->*bool conversion between the state and the cloudflare-go API.

jacobbednarz:
looks like you'll need to update from `master` to resolve the conflicts. additionally, i recently landed some helpers (https://github.com/cloudflare/cloudflare-go/pull/825) to swap between types easily. you'll probably want this for `BoolPtr` like we've done in other places recently.

Issue #1276, opened 2021-10-25
cloudflare_origin_ca_certificate requested_validity causes new certificates to be created on every apply


mattmichal:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform 1.0.4
cloudflare 3.3.0

### Affected resource(s)

cloudflare_origin_ca_certificate

### Terraform configuration files

```hcl
resource "tls_private_key" "app_origin" {
  algorithm = "RSA"
}

resource "tls_cert_request" "app_origin" {
  key_algorithm   = tls_private_key.app_origin.algorithm
  private_key_pem = tls_private_key.app_origin.private_key_pem

  subject {
    common_name    = "CloudFlare Origin Certificate"
    street_address = []
  }
}

resource "cloudflare_origin_ca_certificate" "app_origin" {
  csr                = tls_cert_request.app_origin.cert_request_pem
  hostnames          = [cloudflare_record.cname.hostname, "www.${cloudflare_record.cname.hostname}"]
  request_type       = "origin-rsa"
  requested_validity = 15 * 365
}
```

### Debug output

-

### Panic output

-

### Expected output

No change to resource.

### Actual output

```
   # module.app-xyz.cloudflare_origin_ca_certificate.app_origin will be updated in-place
  ~ resource "cloudflare_origin_ca_certificate" "app_origin" {
        id                 = "xxx"
      ~ requested_validity = 5472 -> 5475
        # (5 unchanged attributes hidden)
    }
```
3 days after the last apply and the resource is showing a diff. A new certificate is generated causing any dependent resources to be updated as well (Azure App Service Certificate for example).

### Steps to reproduce

1. Use cloudflare provider 3.3.0 to create a `cloudflare_origin_ca_certificate` resource with `requested_validity`.
2. Run `terraform apply`.
3. Wait at least one day.
4. Run `terraform plan` and observe that the `cloudflare_origin_ca_certificate` resource will be updated.

### Additional factoids

This was previously discussed in #1031 for version 2.x series. However, when I look at the current code I don't see the same logic as was implemented with PR #1078. The `ignore_changes` fix is still possible but it would be good to have a native fix. Is it possible to update the docs in the interim?

The API is still not returning `requested_validity`; is there any issue or documentation that can be referenced or voted on?

### References

#1031 
#1078

[...]

Issue #1031, opened 2021-04-15
cloudflare_origin_ca_certificate diff lead to new certificate creation


azhurbilo:
### Confirmation

My issue isn't already found on the issue tracker.
I have replicated my issue using the latest version of the provider and it is still present.

### Terraform version

Terraform v0.12.24

### Affected resource(s)

cloudflare_origin_ca_certificate

### Terraform configuration files

```hcl
resource "cloudflare_origin_ca_certificate" "ca" {
  csr                = tls_cert_request.cloudflare_origin_ca.cert_request_pem
  hostnames          = ["XXXX"]
  request_type       = "origin-rsa"
  requested_validity = 5475
}

resource "aws_iam_server_certificate" "cloudflare_origin_ca" {
  name_prefix      = "cloudflare_origin_ca"
  certificate_body = cloudflare_origin_ca_certificate.ca.certificate
  private_key      = tls_private_key.cloudflare_origin_ca.private_key_pem

  lifecycle {
    create_before_destroy = true
  }
}
```

### Expected output

Maybe it's possible to show that new "cloudflare_origin_ca_certificate" resource will be created 

### Actual output

3 step diff, but as result new certificate was generated

```
# cloudflare_origin_ca_certificate.ca will be updated in-place
  ~ resource "cloudflare_origin_ca_certificate" "ca" {
        ....
        id                 = "XXXXXX"
        request_type       = "origin-rsa"
      ~ requested_validity = 5059 -> 5475
    }
Plan: 0 to add, 2 to change, 0 to destroy.
```

and next terraform run we got new diff related to other resources
4 step diff >>

```
 # aws_iam_server_certificate.cloudflare_origin_ca must be replaced
+/- resource "aws_iam_server_certificate" "cloudflare_origin_ca" {
      ~ arn              = "xxxxx" -> (known after apply)
      ~ certificate_body = <<~EOT # forces replacement
      ...
      ~ expiration       = "xxxx" -> (known after apply)
      ~ id               = "xxxxx" -> (known after apply)
      ~ name             = "xxxx" -> (known after apply)
        name_prefix      = "xxxx"
        path             = "/"
      ~ private_key      = (sensitive value)
      - tags             = {} -> null
      ~ upload_date      = "xxxx" -> (known after apply)
}
```

### Steps to reproduce

1. use cloudflare provider 2.7.0 with "requested_validity" bug (in TF state requested_validity value = 5059 but valid value is 5475)
[...]

PR #1502, opened 2022-03-10
Use requested_validity from API response

boekkooi-lengoo:
Instead of trying and failing to calculate the requested validity of the certificate we can use the returned validity.
It looks like `requested_validity` was being used in 2 ways.
The first one was to request the certificate for a time period and the second to indicate the amount of day the certificate is still valid.

The calculation is no longer done and could be seen as a BC break but since the current setup is causing issue like #1448, #1276 and #1031. I expect that this is acceptable.

Related API docs: https://api.cloudflare.com/#origin-ca-get-certificate

### DIFF
--- a/.changelog/1502.txt
+++ b/.changelog/1502.txt
@@ -1,0 +1,3 @@
+```release-note:breaking-change
+resource/cloudflare_origin_ca_certificate: `requested_validity` is no longer calculated (to be the amount of days until the certificate expires) but is now the amount of days the certificate was requested for.
+```
--- a/.changelog/1502.txt
+++ b/.changelog/1502.txt
@@ -1,3 +1,3 @@
-```release-note:breaking-change
-resource/cloudflare_origin_ca_certificate: `requested_validity` is no longer calculated (to be the amount of days until the certificate expires) but is now the amount of days the certificate was requested for.
+```release-note:note
+resource/cloudflare_origin_ca_certificate: `requested_validity` no longer decrements until the `expires_on` value but is now the amount of days the certificate was requested for.
 ```
--- a/cloudflare/resource_cloudflare_origin_ca_certificate.go
+++ b/cloudflare/resource_cloudflare_origin_ca_certificate.go
@@ -6,7 +6,6 @@ import (
 	"encoding/pem"
 	"fmt"
 	"log"
-	"math"
 	"strings"
 	"time"
 
@@ -92,19 +91,7 @@ func resourceCloudflareOriginCACertificateRead(d *schema.ResourceData, meta inte
 	d.Set("expires_on", cert.ExpiresOn.Format(time.RFC3339))
 	d.Set("hostnames", hostnames)
 	d.Set("request_type", cert.RequestType)
-
-	// lazy approach to extracting the date from a known timestamp in order to
-	// `time.Parse` it correctly. Here we are getting the certificate expiry and
-	// calculating the validity as the API doesn't return it yet it is present in
-	// the schema.
-	date := strings.Split(cert.ExpiresOn.Format(time.RFC3339), "T")
-	certDate, _ := time.Parse("2006-01-02", date[0])
-	now := time.Now()
-	duration := certDate.Sub(now)
-	var validityDays int
-	validityDays = int(math.Ceil(duration.Hours() / 24))
-
-	d.Set("requested_validity", validityDays)
+	d.Set("requested_validity", cert.RequestValidity)
 
 	return nil
 }
--- a/cloudflare/resource_cloudflare_origin_ca_certificate.go
+++ b/cloudflare/resource_cloudflare_origin_ca_certificate.go
@@ -6,6 +6,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"log"
+	"math"
 	"strings"
 	"time"
 
@@ -91,7 +92,12 @@ func resourceCloudflareOriginCACertificateRead(d *schema.ResourceData, meta inte
 	d.Set("expires_on", cert.ExpiresOn.Format(time.RFC3339))
 	d.Set("hostnames", hostnames)
 	d.Set("request_type", cert.RequestType)
-	d.Set("requested_validity", cert.RequestValidity)
+
+	x509Cert, err := x509.ParseCertificate([]byte(cert.Certificate))
+	if err != nil {
+		return fmt.Errorf("error parsing OriginCACertificate %q: %s", certID, err)
+	}
+	d.Set("requested_validity", calculateRequestedValidityFromCertificate(x509Cert))
 
 	return nil
 }
@@ -124,4 +130,30 @@ func validateCSR(v interface{}, k string) (ws []string, errors []error) {
 		errors = append(errors, fmt.Errorf("%q: %s", k, err.Error()))
 	}
 	return
+}
+
+func calculateRequestedValidityFromCertificate(cert *x509.Certificate) int {
+	diff := cert.NotAfter.UTC().Sub(cert.NotBefore.UTC())
+	days := math.Round(diff.Hours() / 24)
+
+	validateDays := []float64{7, 30, 90, 365, 730, 1095, 5475}
+
+	// Find the closest matching requested validity (in days) to avoid possible leap second issue.
+	i := 0
+	d := math.Abs(days - validateDays[i])
+	distanceIdx := i
+	distance := d
+	for i < len(validateDays) {
+		d := math.Abs(validateDays[i] - days)
+		if d == 0 {
+			return int(days)
+		}
+
+		if d < distance {
+			distanceIdx = i
+			distance = d
+		}
+		i++
+	}
+	return int(validateDays[distanceIdx])
 }
--- a/cloudflare/resource_cloudflare_origin_ca_certificate.go
+++ b/cloudflare/resource_cloudflare_origin_ca_certificate.go
@@ -93,7 +93,12 @@ func resourceCloudflareOriginCACertificateRead(d *schema.ResourceData, meta inte
 	d.Set("hostnames", hostnames)
 	d.Set("request_type", cert.RequestType)
 
-	x509Cert, err := x509.ParseCertificate([]byte(cert.Certificate))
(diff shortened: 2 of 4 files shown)


jacobbednarz (changes requested):
while i appreciate the effort here, `requested_validity` is an auto decrementing value in the API (hence the logic to handle that). taking the API value as is will cause a permadiff for users of this resource.

boekkooi-lengoo:
Hey @jacobbednarz, 
Thank you for the information. 

Would a possible solution be to [parse the certificate](https://pkg.go.dev/crypto/x509#ParseCertificate) and use the [Certificate](https://pkg.go.dev/crypto/x509#Certificate) `NotBefore` and `NotAfter` to calculate the correct value?

jacobbednarz:
can you elaborate on how this change improves the situation? as far as i can tell, this is still going to have the same issues as the current and your previous attempt as the value is auto decrementing (either from the API or the internal calculation). the only difference is that we're using the certificate instead of the API response itself.

personally, i think we need to take a step back and dig into why the issue is popping up. between these two things, it _should_ be working as expected so we need to better understand that before we start trying to change more functionality.

boekkooi-lengoo:
Hey @jacobbednarz,

Sure let me try and explain my reasoning for the last commit (e4246ab29df6f75444be48529dcef1fc2285cad6) that is using the [Certificate](https://pkg.go.dev/crypto/x509#Certificate) `NotBefore` and `NotAfter` to resolve the `requested_validity`.

To my understanding `requested_validity` is at the moment used for the following 2 purposes that seem to be conflicting and are causing issue.
1. `requested_validity` indicates the amount of days that a certificate should be valid for on creation. (via user input)
2. `requested_validity` indicates the amount of days that the certificate is still valid for. (via API/computation)

What I'm doing in commit e4246ab29df6f75444be48529dcef1fc2285cad6 is to remove the second purpose of the `requested_validity` by ensuring that `requested_validity` is used to indicate the amount of days that a certificate should be valid for on creation.  
This should resolve the issue where `requested_validity` keeps changing into [invalid values](https://github.com/cloudflare/terraform-provider-cloudflare/blob/6b1450760209cdab1be0408fdec93e005d13f91f/cloudflare/schema_cloudflare_origin_ca_certificate.go#L41).  
It also allows the removal of the [`DiffSuppressFunc`](https://www.terraform.io/plugin/sdkv2/schemas/schema-behaviors#diffsuppressfunc) which currently suppresses any change to the `requested_validity` property.

In regards to commit e4246ab29df6f75444be48529dcef1fc2285cad6 auto decrementing `requested_validity` this is no longer true because the certificate is created once with a `notBefore` and `notAfter` which should not change (see https://datatracker.ietf.org/doc/html/rfc5280#section-4.1).  
Because of this we can calculate the `requested_validity` without it decreasing in amount.

I hope this explains my reasoning well enough. 

Please let me know if I'm way of the mark here and thanks again for taking the time to look at this.

jacobbednarz:
gotcha, I understand your thought process now. if we want to take that route, there is also the option of just using `d.Get("certificate_validity")` and make it a required parameter. 

additionally, we already have `expires_on` so that functionality doesn’t need replacing making the job here pretty straight forward.

boekkooi-lengoo:
I personally feel that exposing the amount of days the certificate will still be valid for before it expires is not that useful for the resource also because there is already expires on. Then again I'm happy to add another argument or attribute if that is what you want :smile: 

Feel free to review and/or edit this PR or open another if you feel this one is incorrect :smile:

jacobbednarz:
we already have `expires_on` which i think is fine for most to use if they need the expiry.

let's go ahead and swap https://github.com/cloudflare/terraform-provider-cloudflare/pull/1502/files#diff-42699afda6f492fcdb4d92095ae6d2af27c66eee17a034f36f02f68d0234c236R100 to be `d.Set("certificate_validity", d.Get("certificate_validity").(string))` and make it required. that will really trim down this PR and we'll just issue a note in the changelog that the value no longer decrements.

boekkooi-lengoo:
I'm not sure but would not calculating `requested_validity` cause an issue with `terraform import` since `requested_validity` may differ from the specified value in Terraform?

jacobbednarz:
it _shouldn’t_ be different but it totally can be if someone botches the calculation themselves. 

my concern doing it this way is that theoretically, someone can create a certificate for a date range outside of the schema and that value gets saved during import…🤔 I suppose we can just chalk that up to user error and leave the happy path as is. 

I’ll give this a run with the acceptance test suite in the morning and all going to plan, we can land this as is. 

thanks for the discussion and feedback!

jacobbednarz:
running the acceptance tests locally is showing this erroring

```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareOriginCACertificate_Basic" -count 1 -parallel 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareOriginCACertificate_Basic
    resource_cloudflare_origin_ca_certificate_test.go:34: Step 1/2 error: Error running apply: exit status 1

        Error: error parsing OriginCACertificate "105532533569008664101985694143040053618752100187": x509: malformed certificate

          with cloudflare_origin_ca_certificate.qsrzolpbtz,
          on terraform_plugin_test.tf line 2, in resource "cloudflare_origin_ca_certificate" "qsrzolpbtz":
           2: resource "cloudflare_origin_ca_certificate" "qsrzolpbtz" {

--- FAIL: TestAccCloudflareOriginCACertificate_Basic (8.60s)
FAIL
FAIL	github.com/cloudflare/terraform-provider-cloudflare/cloudflare	8.848s
FAIL
make: *** [testacc] Error 1
```

here is the HTTP payload if you need something to test against

```
{
  "success": true,
  "result": {
    "id": "105532533569008664101985694143040053618752100187",
    "certificate": "-----BEGIN CERTIFICATE-----\nMIIEsjCCA5qgAwIBAgIUEnw+EpiIfufXbLQ1Vi2Kn91MF1swDQYJKoZIhvcNAQEL\nBQAwgYsxCzAJBgNVBAYTAlVTMRkwFwYDVQQKExBDbG91ZEZsYXJlLCBJbmMuMTQw\nMgYDVQQLEytDbG91ZEZsYXJlIE9yaWdpbiBTU0wgQ2VydGlmaWNhdGUgQXV0aG9y\naXR5MRYwFAYDVQQHEw1TYW4gRnJhbmNpc2NvMRMwEQYDVQQIEwpDYWxpZm9ybmlh\nMB4XDTIyMDMxNjAyMDkwMFoXDTIyMDMyMzAyMDkwMFowYjEZMBcGA1UEChMQQ2xv\ndWRGbGFyZSwgSW5jLjEdMBsGA1UECxMUQ2xvdWRGbGFyZSBPcmlnaW4gQ0ExJjAk\nBgNVBAMTHUNsb3VkRmxhcmUgT3JpZ2luIENlcnRpZmljYXRlMIIBIjANBgkqhkiG\n9w0BAQEFAAOCAQ8AMIIBCgKCAQEAxuhEL5/ibLTcZoaEroYm+zd24vZTVkkNusp4\niVwu0JrSruXlzfuhgoXHYTyu/SeN3QX4IA9TmeU3A0KmMMJRmT6ArXyw52Emc7U6\nYTvCESeSFWnMcUtfUJpCVVuxetelVkMhGAdZzx4lu5kPzlg2lxMiBDxeQesfricb\nVdWiBx4QbkIXjMd3v2MIY2nmAT84dy4HFpney9pU/91JglCUzlPcdzlTfFHr8iDw\n7hQ0eEvZGItM5TpleBc0wUu3+5q+ofIBz4bCNIgRvbysTXhLMI+JrnrJ1/X830Eh\nQ3bNS03aSs2FbyTbUHMpLiSvwjvtfHoUuRkzxjnLYsU7muaHsQIDAQABo4IBNDCC\nATAwDgYDVR0PAQH/BAQDAgWgMB0GA1UdJQQWMBQGCCsGAQUFBwMCBggrBgEFBQcD\nATAMBgNVHRMBAf8EAjAAMB0GA1UdDgQWBBS/bOEmKmPldegtxeFS6bNn1RYSFzAf\nBgNVHSMEGDAWgBQk6FNXXXw0QIep65TbuuEWePwppDBABggrBgEFBQcBAQQ0MDIw\nMAYIKwYBBQUHMAGGJGh0dHA6Ly9vY3NwLmNsb3VkZmxhcmUuY29tL29yaWdpbl9j\nYTA1BgNVHREELjAsghUqLnRlcnJhZm9ybS5jZmFwaS5uZXSCE3RlcnJhZm9ybS5j\nZmFwaS5uZXQwOAYDVR0fBDEwLzAtoCugKYYnaHR0cDovL2NybC5jbG91ZGZsYXJl\nLmNvbS9vcmlnaW5fY2EuY3JsMA0GCSqGSIb3DQEBCwUAA4IBAQC/lAOE8Z5ddgTr\ntsq1ZEgns437GDp3Jac1Lzte7eeR98RFqTdvY3tppQ5sU6g33biHRdBQziTcuH/+\nEh5R5b5DXPwvnNwQyE19dxR/jlXhz7b00va8Heo6GZxxehBk/ZwLKneKEbMVhX94\nEpKkQq9WDW4oJRyAN00t+Vh7oPVu55afcmhZtvreIBM5w665xUVZNy6atYIWaEQB\nUqiE4AKTRAk5+8M5Gzg4qmcenM0DLg54U59IZSWwCtYKzQKtlbRH4lrIKAoGdVol\neJQzDboeUdjl042hJOo6UYB7NlkmdMIVu8ONRoqZu0OHOxWKz1OY5dLrUk4C/PAy\nPl7w7LJA\n-----END CERTIFICATE-----\n",
    "expires_on": "2022-03-23 02:09:00 +0000 UTC",
    "request_type": "origin-rsa",
    "hostnames": [
      "*.terraform.cfapi.net",
      "terraform.cfapi.net"
    ]
  },
  "errors": [],
  "messages": []
}
```

boekkooi-lengoo:
Sorry about that @jacobbednarz based on a quick test it seems I forgot to decode the cert before parsing it. This issue should be resolved by ad26425050565f7913739de8a18eaae83c25c14e.

Issue #1486, opened 2022-02-28
Failure to create DNSSEC resource with "error unmarshalling the JSON response"


Malcolm:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```powershell
PS C:\CloudFlare-Issue> terraform -v
Terraform v1.1.6
on windows_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.9.1
```

### Affected resource(s)

cloudflare_zone_dnssec

### Terraform configuration files

```hcl
terraform {
  required_version = "~> 1.1.0"

  required_providers {
    cloudflare = {
      source  = "cloudflare/cloudflare"
      version = "~> 3.9.0"
    }
  }
}

variable "api_token" {
  type = string
}

provider "cloudflare" {
  api_token = var.api_token
}

resource "cloudflare_zone" "this" {
  zone = "mydomain.com"
}

resource "cloudflare_zone_dnssec" "this" {
  zone_id = cloudflare_zone.this.id
}

output "dnssec" {
  value = cloudflare_zone_dnssec.this
}
```

### Debug output

I've included the following files
- [01_run.log](https://github.com/cloudflare/terraform-provider-cloudflare/files/8154283/01_run.log): This is the initial run where the DNS zone is created (successfully) and DNSSEC fails
- [02_run.log](https://github.com/cloudflare/terraform-provider-cloudflare/files/8154284/02_run.log): Attempting to re-run terraform apply, the zone is seen as successfully created, attempts to create the DNSSEC resource but fails with error `HTTP status 400: DNSSEC is already enabled (1003)`
- [03_run.log](https://github.com/cloudflare/terraform-provider-cloudflare/files/8154285/03_run.log): In the Cloudflare dashboard, for this zone, I remove the DNSSEC config (DNSSEC: Cancel Setup) then re-run terraform apply, DNSSEC is successfully enabled in the zone
- [04_run.log](https://github.com/cloudflare/terraform-provider-cloudflare/files/8154281/04_run.log): Terraform destroy, both resources are successfully removed
[...]

PR #1530, opened 2022-03-27
fix: detect zone dnssec as enabled when in pending status

xabinapal:
Related to #1486

This PR does not fully fix that issue because it addresses two different errors and I think one of them is a bug either in the Cloudflare API or the Go client: https://github.com/cloudflare/terraform-provider-cloudflare/issues/1486#issuecomment-1079902813.

With this change, zones with DNSSEC in `pending` state (when the DS record has not been added or detected yet) won't throw a `DNSSEC is already enabled` error. The code expected the `state` value of the DNSSEC API request was only one of `active` or `disabled`, but the `pending` state also represents an enabled DNSSEC, only that is not fully configured yet.

### DIFF
--- a/cloudflare/resource_cloudflare_zone_dnssec.go
+++ b/cloudflare/resource_cloudflare_zone_dnssec.go
@@ -14,6 +14,7 @@ import (
 // The supported status for a Zone DNSSEC setting
 const (
 	DNSSECStatusActive   = "active"
+	DNSSECStatusPending  = "pending"
 	DNSSECStatusDisabled = "disabled"
 )
 
@@ -41,7 +42,7 @@ func resourceCloudflareZoneDNSSECCreate(d *schema.ResourceData, meta interface{}
 	if err != nil {
 		return fmt.Errorf("error finding Zone DNSSEC %q: %s", zoneID, err)
 	}
-	if currentDNSSEC.Status != DNSSECStatusActive {
+	if currentDNSSEC.Status != DNSSECStatusActive && currentDNSSEC.Status != DNSSECStatusPending {
 		_, err := client.UpdateZoneDNSSEC(context.Background(), zoneID, cloudflare.ZoneDNSSECUpdateOptions{Status: DNSSECStatusActive})
 
 		if err != nil {
--- a/.changelog/1530.txt
+++ b/.changelog/1530.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_zone_dnssec: don't try to enable DNSSEC when state is "pending"
+```


cloudflare/resource_cloudflare_zone_dnssec.go
```diff
@@ -41,7 +42,7 @@ func resourceCloudflareZoneDNSSECCreate(d *schema.ResourceData, meta interface{}
 	if err != nil {
 		return fmt.Errorf("error finding Zone DNSSEC %q: %s", zoneID, err)
 	}
-	if currentDNSSEC.Status != DNSSECStatusActive {
+	if currentDNSSEC.Status != DNSSECStatusActive && currentDNSSEC.Status != DNSSECStatusPending {
```
jacobbednarz, line 45: it seems like we should invert this check and just check against `DNSSECStatusDisabled`, no?

xabinapal: That was my first thought, but then I checked the API docs (https://api.cloudflare.com/#dnssec-properties) and there are two extra possible values for this field: `pending-disabled` and `error`. If I'm not wrong, neither of them represent a enabled status.

jacobbednarz: that's fair - my assumption was we only wanted to attempt to update this value if it was explicitly disabled however, perhaps we also want it on `error` too. we don't want to attempt the `Create` if there is an operation in progress (pending, pending-disabled) or already active. in which case, i think we also need to extend this to include the `pending-disabled` status too.

could i also ask that you add these two other conditions to the constants so we have them all?

xabinapal: I've been trying some tests. I don't know how to trigger an `error` state, so I'm not sure what it does mean or if we should try or not to create the resource.

`pending-disable` is an intermediate state before showing DNSSEC as fully disabled; the UI shows the following message:

> Your DNSSEC setup will be disabled as soon as we detect that the DS record has been removed from your registrar.

This state is effectively disabled and DNSSEC can be enabled again with no issues. I don't think we should skip `Create` in this state.

I can add two new constants, `DNSSECStatusPendingDisabled` and `DNSSECStatusError` but I don't think we will be using them (at least the first one, not sure about the `Error` one).

jacobbednarz: I'm struggling to come up with scenarios where we wouldn't want to prevent attempting to recreate when the status is pending disabled but a handful o log scenarios where you wouldn't want successive calls to attempt to create the resource while it is pending cleanup. The only things I can think of are if you disabled it by accident?

xabinapal: As I see it, `pending-disable` and `disable` are the same status, with the existence of the DS record (or rather, Cloudflare not having already detected its absence) on the domain as the only difference, but this is something "out of scope" of Cloudflare. For example, if we don't allow enabling DNSSEC in this state, `terraform taint cloudflare_zone_dnssec.example && terraform apply` would have different behaviour depending on if the DNS changes in the domain have propagated or not.

I always tend to assume that Terraform providers should behave exactly like direct API requests. If I can re-enable DNSSEC through the web UI or the API even in a `pending-disable`, why would Terraform block me to do it? It seems a little counter-intuitive.

Of course, almost 100% of the use cases are either accidents or testing, in normal execution I also don't see why anyone would do it. For example, my current use case is that we are developing our DNS related IaC code and at the same time, we are migrating our DNS to Cloudflare. I would not expect to wait until DS changes from our registrar are propagated everytime a `terraform destroy` is executed after a test so that the status is detected as `disabled` instead of `pending-disabled`.

If we don't allow enabling DNSSEC when in `pending-disabled` status, the same would apply to disabling it when in `pending` state. And this would be quite odd: do I need to add a DS record to a domain and wait for it to propagate before Terraform allows me to disable DNSSEC?

jacobbednarz: I agree with you on the API compatibility. The difference though is that Terraform has lifecycle management which means in some places, it has to differ. I'm not strongly opposites enough here though so let's try it and see what happens.

Issue #1314, opened 2021-11-18
Can't add Warp and Gateway device_posture to policy


maksymivash:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```
Terraform v1.0.6
on darwin_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.4.0
+ provider registry.terraform.io/hashicorp/archive v2.2.0
+ provider registry.terraform.io/hashicorp/null v3.1.0
```

### Affected resource(s)

cloudflare_access_group

### Terraform configuration files

```hcl
resource "cloudflare_device_posture_rule" "warp" {
  account_id = local.account_id
  name       = "Warp"
  type       = "warp"
}

resource "cloudflare_device_posture_rule" "gateway" {
  account_id = local.account_id
  name       = "Gateway"
  type       = "gateway"
}

resource "cloudflare_access_group" "whole_team" {
  account_id = local.account_id
  name       = "Whole Team"

  include {
    email          = []
    device_posture = []
    email_domain  = ["example.com"]
    geo           = []
    group         = []
    ip            = []
    login_method  = []
    service_token = []
  }

  require {
    email = []
    device_posture = [cloudflare_device_posture_rule.warp.id, cloudflare_device_posture_rule.gateway.id]
    email_domain   = []
    geo            = []
    group          = []
    ip             = []
    login_method   = []
    service_token  = []
  }
}
```

### Debug output
```
---[ REQUEST ]---------------------------------------
PUT /client/v4/accounts/ACCOUNT_ID_REDACTED/access/groups/GROUP_ID_REDACTED HTTP/1.1
Host: api.cloudflare.com
User-Agent: terraform/1.0.6 terraform-plugin-sdk/2.8.0 terraform-provider-cloudflare/3.4.0
Content-Length: 337
Authorization: Bearer AUTH_REDACTED
Content-Type: application/json
Accept-Encoding: gzip

{
 "id": "GROUP_ID_REDACTED",
 "created_at": null,
 "updated_at": null,
 "name": "Whole Team",
 "include": [
  {
   "email_domain": {
    "domain": "example.com"
   }
  }
 ],
 "exclude": null,
 "require": [
  {
   "device_posture": {
    "integration_uid": "WARP_DEVICE_POSTURE_ID_REDACTED"
   }
  },
  {
[...]

PR #1536, opened 2022-03-29
add access group gateway

fwieffering-blockfi:
if this gets merged https://github.com/cloudflare/cloudflare-go/pull/846/files I can update to using the struct defined there

https://github.com/cloudflare/terraform-provider-cloudflare/issues/1314

### DIFF
--- a/.changelog/1536.txt
+++ b/.changelog/1536.txt
@@ -1,3 +1,3 @@
 ```release-note:enhancement
-add support for gateway attribute for access_group
+resource/cloudflare_access_group: add support for gateway attribute
 ```
--- a/.changelog/1536.txt
+++ b/.changelog/1536.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+add support for gateway attribute for access_group
+```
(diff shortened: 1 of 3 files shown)


Justin-Holmes:
I left a comment on the related cloudflare-go PR here: https://github.com/cloudflare/cloudflare-go/pull/846#issuecomment-1101442483. To summarize, gateway/warp rules can be created as a `device_posture` rule.  Creating using the `gateway` or `warp` rule type is not publicly supported and have no guarantees of support in the future. Here is an example on how to create a require gateway rule:
```terraform
resource "cloudflare_device_posture_rule" "my_gateway_dpr" {
  account_id  = var.account_id
  type        = "gateway"
  name        = "Gateway"
}

resource "cloudflare_access_policy" "my_policy" {
  application_id = cloudflare_access_application.my_app.id
  account_id     = var.account_id
  name           = "my asdf test"
  precedence     = "0"
  decision       = "allow"

  require {
    device_posture = [cloudflare_device_posture_rule.gateway_dpr.id]
  }
}
```

PR #1548, opened 2022-04-04
formatting and simplification of test params

jacobbednarz:


jacobbednarz:
there is a leaking test somewhere in this that i need to track down.

PR #1550, opened 2022-04-06
Adding IP type support for cloudflare teams list

sanicheev:
This PR adds IP type support to cloudflare teams list.

I checked the cloudflare-go repo and could not find any validation performed on a type of a list.

Even though API docs does not mention IP type:
Screenshot from 2022-04-06 12-54-00

By looking at the response from GET '/lists' i was able verify that IP is actually supported: 
Screenshot from 2022-04-06 17-08-14

I also tried to test it locally using terraform.
So far so good.

Please let me know if there is anything i am missing.
As our team would like to have that juicy feature available via terraform

### DIFF
--- a/.changelog/1550.txt
+++ b/.changelog/1550.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/cloudflare_teams_list: Add support for IP type
+```
--- a/cloudflare/schema_cloudflare_teams_list.go
+++ b/cloudflare/schema_cloudflare_teams_list.go
@@ -18,7 +18,7 @@ func resourceCloudflareTeamsListSchema() map[string]*schema.Schema {
 		"type": {
 			Type:         schema.TypeString,
 			Required:     true,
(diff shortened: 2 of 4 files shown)


jacobbednarz:
I'm afraid we cannot accept this until we have the public documentation updated to reflect it the value being expected for public consumption. Have you got a support ticket requesting the documentation be updated yet?

sanicheev:
> I'm afraid we cannot accept this until we have the public documentation updated to reflect it the value being expected for public consumption. Have you got a support ticket requesting the documentation be updated yet?

No i havn't. Whats the best way to proceed?

jacobbednarz:
I can make a ticket for the team however, there are no guarantees when it will be actioned. If you need it, I'd recommend opening a customer support ticket that will end up in the hands of the right team.

sanicheev:
Thanks @jacobbednarz will do. Should we leave this PR opened until API will be updated?

jacobbednarz:
yep, that's fine. once we have the documentation we merge this in.

jacobbednarz:
the public docs will be released next week but have been merged, thanks for the contributions!

PR #1565, opened 2022-04-15
Allow 0 value timeout in cf rate_limit

manatarms:
The API seems to be returning timeouts of 0. So I'm assuming this is  a valid value.


jacobbednarz:
are you seeing the 0 in the API response or the cf-terraforming output? if it's not from the API, this might be a terraform zero value that is being output.

manatarms:
I'm seeing it from the API, so I figured it's allowed
image

manatarms:
The API seems to be sending these back but not accepting them. Closing this out, to fix on the API side.

Issue #1559, opened 2022-04-11
cloudflare_certificate_pack - optionally block until status is active


ct-dh:
### Current Terraform and Cloudflare provider version

All versions

### Description

The `cloudflare_certificate_pack` resource currently does not provide a way to wait for an active status. This means, even if using `create_before_destroy` you can't be very confident that an outage wouldn't be caused by "updating" a certificate pack. The documentation in the resource says to do a 2 phase change if you want to be sure the new cert pack is active before deleting the old one.

Adding a new optional boolean optional such as `wait_for_active` (open to suggestions on a better name) which defaults to `false` would potentially help with this situation.

I am willing to work on a PR to add this to test it out, but didn't want to spend any time on it without checking if this was seen as a valid approach.

### Use cases

I have a use case where I will be ordering many advanced certificate packs with 100s of hostnames and the plan would be to use the terraform `chunklist()` function to batch into blocks of 50 hostnames and use create_before_destroy. I need to plug this terraform configuration into an existing deployment workflow so doing a 2 phase change won't be easy to implement.

### Potential Terraform configuration

```hcl
resource "cloudflare_certificate_pack" "advanced_example_for_digicert" {
  zone_id               = "1d5fdc9e88c8a8c4518b068cd94331fe"
  type                  = "advanced"
  hosts                 = ["example.com", "sub.example.com"]
  validation_method     = "txt"
  validity_days         = 30
  certificate_authority = "digicert"
  cloudflare_branding   = false
  wait_for_active       = true
}
```

ct-dh:
If it helps, I have a local branch where I have tested adding this in and it seems to work pretty well.

ct-dh:
I added a PR to share the change that seemed to work in my limited testing - I know it doesn't meet all the contribution guidelines but I didn't want to just throw it away if this is seen as a feature that should be implemented.

PR #1567, opened 2022-04-20
Add ability to wait for cert packs to be active

ct-dh:
This is probably not ready for merge - I made this little patch to test the feature request in #1559 and won't be able to follow up in all likelihood by the time there is a response due to a change in employer (so I will no longer personally be working with Cloudflare - but the use case for the employer is still valid/relevant), but I didn't want to just delete the local code.

Please feel free to use/modify/drop this as needed.

It adds a new option to the ﻿﻿`cloudflare_certificate_pack` resource to block/wait on creation until all certificates in the pack are marked as active in the API - with the aim of this making use of `create_before_destroy` less likely to cause service disruption on updates.


cloudflare/resource_cloudflare_certificate_pack.go
```diff
@@ -62,6 +63,32 @@ func resourceCloudflareCertificatePackCreate(d *schema.ResourceData, meta interf
...
+			if len(certificatePack.Certificates) == 0 {
+				return resource.RetryableError(fmt.Errorf("certificate list in response is empty"))
+			}
+			for _, certificate := range certificatePack.Certificates {
+				if certificate.Status != "active" {
+					return resource.RetryableError(fmt.Errorf("expected all certificates in certificate pack to be active state but certificate %s was in state %s", certificate.ID, certificate.Status))
+				}
+			}
+
+			// if we get here, the new cert pack is active
```
jacobbednarz, line 81: is there a reason we're duplicating the `d.SetId` and `Read` operations here? we should be able to reuse what we already have a few lines further down.

jacobbednarz:
before we consider merging this, we'll also need some tests and to clean up the imports for these files.

cloudflare/resource_cloudflare_certificate_pack.go
ct-dh: No good reason, I had rushed this out so just hadn't noticed the duplication. Should be resolved in latest commits.

ct-dh: No good reason, I had rushed this out so just hadn't noticed the duplication. Should be resolved in latest commits.

jacobbednarz:
> I can add a `resource.TestCheckResourceAttr` check for the new attribute to check the default

yep, just a check that it ends up in the state is fine. we can also add another test with the `wait_for_active_status` flag set to on so that we exercise that code path and there aren't any errors too.

ct-dh:
Added the tests and rebased from master if you can take another look please @jacobbednarz.

I don't have a test Cloudflare account that would be safe to run the test suite in (from a brief look at the way they work, the sweepers will destroy unrelated resources) so I don't know if the tests actually pass or not.

jacobbednarz:
acceptance tests look good

```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCertificatePack_" -count 1 -parallel 1 -timeout 120m -parallel 1
=== RUN   TestAccCertificatePack_AdvancedDigicert
--- PASS: TestAccCertificatePack_AdvancedDigicert (9.51s)
=== RUN   TestAccCertificatePack_AdvancedLetsEncrypt
--- PASS: TestAccCertificatePack_AdvancedLetsEncrypt (9.86s)
=== RUN   TestAccCertificatePack_DedicatedCustom
    resource_cloudflare_certificate_pack_test.go:133: Pending investigation into ACM entitlements
--- SKIP: TestAccCertificatePack_DedicatedCustom (0.00s)
=== RUN   TestAccCertificatePack_WaitForActive
--- PASS: TestAccCertificatePack_WaitForActive (14.06s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/cloudflare	34.009s
```

PR #1572, opened 2022-04-21
adds cloudflare_teams_route and acceptance tests

terinjokes:


cloudflare/resource_cloudflare_teams_route.go
```diff
@@ -0,0 +1,144 @@
...
+
+func resourceCloudflareTeamsRouteRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	tunnelRoute, err := client.GetTunnelRouteForIP(context.Background(), cloudflare.TunnelRoutesForIPParams{
+		AccountID: d.Get("account_id").(string),
+		Network:   d.Get("network").(string),
+	})
+	if err != nil {
+		// TODO: implement logic for missing route
```
jacobbednarz, line 51: `d.SetId("")` will remove it from the state file if the response indicates the resource no longer exists. might need to check the response status or exception though. something like https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/resource_cloudflare_access_rule.go#L102-L106 will work.

cloudflare/resource_cloudflare_teams_route.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return nil
+}
+
+func resourceCloudflareTeamsRouteCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	resource := cloudflare.TunnelRoutesCreateParams{
+		AccountID: d.Get("account_id").(string),
+		TunnelID:  d.Get("tunnel_id").(string),
```
jacobbednarz, line 71: given the `TunnelID` is read only, this is probably better suited to be used as the `SetID` call.

```suggestion
```

cloudflare/resource_cloudflare_teams_route.go
```diff
@@ -0,0 +1,144 @@
...
+	if comment, ok := d.Get("comment").(string); ok {
+		resource.Comment = comment
+	}
+
+	newTunnelRoute, err := client.CreateTunnelRoute(context.Background(), resource)
+	if err != nil {
+		return fmt.Errorf("error creating Tunnel Route for Network %q: %w", d.Get("network").(string), err)
+	}
+
+	d.SetId(newTunnelRoute.Network)
```
jacobbednarz, line 84:
```suggestion
	d.SetId(newTunnelRoute.TunnelID)
```

cloudflare/resource_cloudflare_teams_route.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return resourceCloudflareTeamsRouteRead(d, meta)
+}
+
+func resourceCloudflareTeamsRouteUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	resource := cloudflare.TunnelRoutesUpdateParams{
+		AccountID: d.Get("account_id").(string),
+		TunnelID:  d.Get("tunnel_id").(string),
```
jacobbednarz, line 94: as the parameter is read only, we don't need to send it here.

```suggestion
```

cloudflare/resource_cloudflare_teams_route.go
```diff
@@ -0,0 +1,134 @@
...
+		return fmt.Errorf("error deleting Tunnel Route for Network %q: %w", d.Get("network").(string), err)
+	}
+
+	return nil
+}
+
+func resourceCloudflareTeamsRouteImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	attributes := strings.SplitN(d.Id(), "/", 3)
+
+	if len(attributes) != 2 {
```
tjstansell, line 117: This import doesn't seem like it'll work as it splits into 3 parts but looks for only 2 here.  Similarly, since the read function only uses the `account_id` and `network` to find the route, shouldn't import just require those two fields?  `tunnel_id` here isn't actually used, as the read call overrides it from the api response.

terinjokes: Good catch, I originally implemented it with a different API call where `tunnel_id` was required. I'll send a follow-up PR in a few moments.

cloudflare/resource_cloudflare_teams_route.go
```diff
@@ -0,0 +1,134 @@
...
+		Importer: &schema.ResourceImporter{
+			State: resourceCloudflareTeamsRouteImport,
+		},
+	}
+}
+
+func resourceCloudflareTeamsRouteRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	tunnelRoute, err := client.GetTunnelRouteForIP(context.Background(), cloudflare.TunnelRoutesForIPParams{
```
tjstansell, line 29: The use of this to read the resource seems fragile.  If you are managing two routes, say `10.0.0.0/8` and `10.10.0.0/24` via terraform and `10.10.0.0/24` gets manually removed on accident ... terraform will scan resources and actually find the route for `10.0.0.0/8` for the `10.10.0.0/24` resource.  Was it not feasible to use the List Tunnel Routes API passing in the tunnel_id parameter to specify an exact tunnel?

tjstansell: Alternatively, perhaps this should verify the network returned is an exact match for what was requested to prevent this?

tjstansell: i guess more than one network could route to the same tunnel, but you could pass in `network_subnet=10.10.0.0/24` and `nework_superset=10.10.0.0/24` to get the exact route.  And probably include `is_deleted=false` too ...

cloudflare/resource_cloudflare_teams_route.go
```diff
@@ -0,0 +1,134 @@
...
+func resourceCloudflareTeamsRouteUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*cloudflare.API)
+
+	resource := cloudflare.TunnelRoutesUpdateParams{
+		AccountID: d.Get("account_id").(string),
+		TunnelID:  d.Get("tunnel_id").(string),
+		Network:   d.Get("network").(string),
+	}
+
+	if comment, ok := d.Get("comment").(string); ok {
```
tjstansell, line 88: If you have a comment in a resource and then simply delete the line with the intent to remove the comment, will it remove it?  Perhaps the comment parameter should have a default of `""` so it's always set to something since the API seems to always return an empty string as the comment field.

cloudflare/resource_cloudflare_teams_route.go
terinjokes: I'll take a look at doing that instead. It's a lot of workaround for an endpoint that doesn't follow the normal pattern of the  V4 API.

PR #1581, opened 2022-04-25
cloudflare_tunnel_route: read by filtering list

terinjokes:
The `GetTunnelRouteForIP` SDK function will return the first route that
either exactly matches the specified IP or larger CIDR. If routes exist
for both `10.0.0.0/8` and `10.0.0.0/24`, and the latter is accidentally
deleted, the SDK function will start returning the former route.

This changeset resolves this issue by providing filters to the
`ListTunnelRoutes` such that the network subset and superset are both
the specific CIDR being managed. The API returns 1 result if there is an
exact match, and 0 results otherwise.

Bug: K8S-4828

### DIFF
--- a/cloudflare/resource_cloudflare_tunnel_route.go
+++ b/cloudflare/resource_cloudflare_tunnel_route.go
@@ -2,8 +2,8 @@ package cloudflare
 
 import (
 	"context"
-	"errors"
 	"fmt"
+	"log"
 	"strings"
 
 	"github.com/cloudflare/cloudflare-go"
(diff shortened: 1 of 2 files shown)


cloudflare/resource_cloudflare_tunnel_route.go
```diff
@@ -25,27 +25,31 @@ func resourceCloudflareTunnelRoute() *schema.Resource {
...
-	tunnelRoute, err := client.GetTunnelRouteForIP(context.Background(), cloudflare.TunnelRoutesForIPParams{
-		AccountID: d.Get("account_id").(string),
-		Network:   d.Get("network").(string),
+	accountID := d.Get("account_id").(string)
+	network := d.Get("network").(string)
+
+	isDeleted := false
+	tunnelRoutes, err := client.ListTunnelRoutes(context.Background(), cloudflare.TunnelRoutesListParams{
+		AccountID:       accountID,
+		IsDeleted:       &isDeleted,
```
jacobbednarz, line 34: there is also [`cloudflare.BoolPtr`](https://github.com/cloudflare/cloudflare-go/blob/master/convert_types.go#L75-L86) if you'd like to do this in a single line.

tjstansell:
Any chance we'll get a new cloudflare-go and terraform provider release soon to pick up the upstream changes?  I tried to use the new `cloudflare_tunnel_route` resource and it keeps matching the wrong ones (either a delete one or a larger one).

PR #1587, opened 2022-04-28
cloudflare_teams_list: add IP list type

terinjokes:
The dashboard allows creating Gateway lists of the "IP" type, but the
schema here rejects attempts to define resources using it.

Bug: K8S-4829


dhaynespls (approved):

jacobbednarz:
looks like we've already got a PR for this in #1550 and we're just waiting on the public documentation. let's go with that one since you have sorted the public docs.

PR #1590, opened 2022-04-30
cloudflare_argo_tunnel: add tunnel token to state

### LINKED ISSUE #1439: `cloudflare_argo_tunnel` import does not include secrets
### Confirmation

Terraform v1.1.5
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.8.0
+ provider registry.terraform.io/hashicorp/random v3.1.0

### Affected resource(s)

* cloudflare_argo_tunnel

```hcl
resource "cloudflare_argo_tunnel" "this" {
  account_id = var.account_id
  name       = var.name
  secret     = random_id.tunnel_secret.b64_std
}
```

### Debug output

The plan output after an import of a tunnel:
```
  # cloudflare_argo_tunnel.this must be replaced
-/+ resource "cloudflare_argo_tunnel" "this" {
      ~ cname      = "<REDACTED>.cfargotunnel.com" -> (known after apply)
      ~ id         = "<REDACTED>" -> (known after apply)
        name       = "my-tunnel.example.com"
      + secret     = (sensitive value) # forces replacement
    }
```

### Panic output

_No response_

### Expected output

When importing a `cloudflare_argo_tunnel` resource, the secret is not available by directly querying the apis.  This means that while the basic tunnel resource is imported, the secret is not and so any plan will see a change in the secret and will then want to replace the tunnel.  The import should be updated to allow specifying the secret to import as well.

### Actual output

after import, a plan shows the secret will change and it will want to replace the tunnel.

### Steps to reproduce

1. import a `cloudflare_argo_tunnel` resource
2. plan will show the secret needs updating and want to replace the tunnel

### Additional factoids

_No response_

### References

_No response_

bradleyglover:
The `cloudflared` service supports passing in a token to allow a connector to authenticate and run a tunnel. This token is a base64 encoded string made up of the account ID, tunnel ID, and secret ID. Since these items are already stored in state, this PR represents a nice quality of life improvement by constructing, encoding, and storing the token in state too.

Closes #1439


jacobbednarz:
i'm going to put this one on the back burner until we land the changes from cloudflare/cloudflare-go#866 and update the methods internally. as is, this won't work because the tunnel secret is only available on the `Create` response; not the individual fetch. however, we did recently add a method to directly get the token via https://github.com/cloudflare/cloudflare-go/blob/master/tunnel.go#L266 which we can use instead.

jacobbednarz:
the upstream changes landed this morning if you'd like to incorporate them into this PR and we can get it merged.

bradleyglover:
Nice, I'll take a stab at refactoring this tomorrow or the next day.

Looking at https://github.com/cloudflare/terraform-provider-cloudflare/issues/1439#issuecomment-1115039598, it seems we will expect errors setting the token state for older tunnels. Is there any example I can lean on for how the provider would handle that? aka leave the token field empty, return a message, etc?

jacobbednarz:
for older tunnels that return the 404, let's leave it empty. we can issue a `diag.Warning` from the `Read` operation to notify the end user it won't be set and perhaps recreate it if they want the token to work in Terraform. since we have a way to fetch the accurate value, we can't populate it with dummy or wrong data as people may rely on it.

bradleyglover:
Updated to use the upstream `TunnelToken` method, but this will not be functional until the changes from https://github.com/cloudflare/cloudflare-go/pull/881 are released

.changelog/1590.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+resource/cloudflare_argo_tunnel: Add computed `tunnel_token` to schema
```
jacobbednarz, line 2:
```suggestion
resource/cloudflare_argo_tunnel: add `tunnel_token` support
```

internal/provider/schema_cloudflare_argo_tunnel.go
```diff
@@ -24,5 +24,9 @@ func resourceCloudflareArgoTunnelSchema() map[string]*schema.Schema {
 			Type:     schema.TypeString,
 			Computed: true,
 		},
+		"tunnel_token": {
```
jacobbednarz, line 27: isn't this the same as `secret` that we already have the schema?

jacobbednarz:
@bradleyglover this is ready for you to pick back up and test now.

internal/provider/schema_cloudflare_argo_tunnel.go
bradleyglover: no, this is the token used by hosts to authenticate and run tunnels. this does technically include the `secret` but also the account id and tunnel id, all being base64 encoded

bradleyglover:
@jacobbednarz thanks! all tests passed for me locally (with the exception of some existing lint failures)

internal/provider/resource_cloudflare_argo_tunnel.go
```diff
@@ -50,7 +50,18 @@ func resourceCloudflareArgoTunnelRead(ctx context.Context, d *schema.ResourceDat
 		return diag.FromErr(fmt.Errorf("failed to fetch Argo Tunnel: %w", err))
 	}
 
+	token, err := client.TunnelToken(context.Background(), cloudflare.TunnelTokenParams{
```
jacobbednarz, line 53:
```suggestion
	token, err := client.TunnelToken(ctx, cloudflare.TunnelTokenParams{
```

internal/provider/resource_cloudflare_argo_tunnel.go
```diff
@@ -50,7 +50,18 @@ func resourceCloudflareArgoTunnelRead(ctx context.Context, d *schema.ResourceDat
...
+		AccountID: accID,
+		ID:        tunnel.ID,
+	})
+	if err != nil {
+		fmt.Sprintf("[WARN] Unable to set the tunnel_token in state because it's not found in API")
```
jacobbednarz, line 58:
```suggestion
		tflog.Warn("unable to set the tunnel_token in state because it's not found in API")
```

PR #1593, opened 2022-05-02
Update default instructions to configure the provider

pablogamboa:
`api_token` is preferred over `email` + `api_key`

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -33,8 +33,7 @@ terraform {
 }
 
 provider "cloudflare" {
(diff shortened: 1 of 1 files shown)


jacobbednarz (changes requested):
thanks for the contribution! while API tokens are preferred, not all services support them yet so this cannot be the default for some people. we either need to stick with the API key/email combo or explain that either is acceptable.

pablogamboa:
ok @jacobbednarz I see, thought this was an easier win. Closing the PR.

Issue #1398, opened 2022-01-18
import cloudflare_workers_kv documentation is wrong


antigenius0910:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

v3.6.0

### Affected resource(s)

cloudflare_workers_kv

### Terraform configuration files

```hcl
I test what you have on terraform registry as example
https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/workers_kv


resource "cloudflare_workers_kv_namespace" "example_ns" {
  title = "test-namespace"
}

resource "cloudflare_workers_kv" "example" {
  namespace_id = cloudflare_workers_kv_namespace.example_ns.id
  key = "test-key"
  value = "test value"
}
```

### Debug output

```terraform
ᐅ TF_LOG=DEBUG terraform import cloudflare_workers_kv_namespace.example 895ac02c008443aab205e8939e471bb9_test-key
2022-01-18T10:47:51.958-0600 [DEBUG] Adding temp file log sink: /var/folders/xb/77gkpnrj527cqm82yf4x8_l40000gp/T/terraform-log788947437
2022-01-18T10:47:51.958-0600 [INFO]  Terraform version: 1.0.9
2022-01-18T10:47:51.958-0600 [INFO]  Go runtime version: go1.16.4
2022-01-18T10:47:51.958-0600 [INFO]  CLI args: []string{"/usr/local/bin/terraform", "import", "cloudflare_workers_kv_namespace.example", "895ac02c008443aab205e8939e471bb9_test-key"}
2022-01-18T10:47:51.958-0600 [DEBUG] Attempting to open CLI config file: /Users/yen/.terraformrc
2022-01-18T10:47:51.958-0600 [DEBUG] File doesn't exist, but doesn't need to. Ignoring.
2022-01-18T10:47:51.959-0600 [DEBUG] ignoring non-existing provider search directory terraform.d/plugins
2022-01-18T10:47:51.959-0600 [DEBUG] ignoring non-existing provider search directory /Users/yen/.terraform.d/plugins
2022-01-18T10:47:51.959-0600 [DEBUG] ignoring non-existing provider search directory /Users/yen/Library/Application Support/io.terraform/plugins
[...]

PR #1595, opened 2022-05-04
Fixed documentation error for workers kv import

azy141:
Fixes documentation for workers kv import. Issue https://github.com/cloudflare/terraform-provider-cloudflare/issues/1398

### DIFF
--- a/website/docs/r/workers_kv.html.markdown
+++ b/website/docs/r/workers_kv.html.markdown
@@ -38,7 +38,7 @@ The following arguments are supported:
 Workers KV Namespace settings can be imported using it's ID.  **Note:** Because the same key can exist in multiple namespaces, the ID is a composite key of the format <namespace_id>/<key>.
 
 ```
-$ terraform import cloudflare_workers_kv.example beaeb6716c9443eaa4deef11763ccca6/"test-key"
+$ terraform import cloudflare_workers_kv.example beaeb6716c9443eaa4deef11763ccca6/test-key
 ```
 
 where:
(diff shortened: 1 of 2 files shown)


website/docs/r/workers_kv.html.markdown
```diff
@@ -35,10 +35,10 @@ The following arguments are supported:
 
 ## Import
 
-Workers KV Namespace settings can be imported using it's ID.  **Note:** Because the same key can exist in multiple namespaces, the ID is a composite key of the format <namespace_id>_<key>.
+Workers KV Namespace settings can be imported using it's ID.  **Note:** Because the same key can exist in multiple namespaces, the ID is a composite key of the format <namespace_id>/<key>.
 
 ```
-$ terraform import cloudflare_workers_kv_namespace.example beaeb6716c9443eaa4deef11763ccca6_test-key
+$ terraform import cloudflare_workers_kv.example beaeb6716c9443eaa4deef11763ccca6/"test-key"
```
jacobbednarz, line 41: we don't need the last bit quoted

```suggestion
$ terraform import cloudflare_workers_kv.example beaeb6716c9443eaa4deef11763ccca6/test-key
```

jacobbednarz, line 41: the resource is `cloudflare_workers_kv_namespace`. the filename should probably be updated to reflect that.

jacobbednarz, line 41: we don't need the quotes around the last bit here.

jacobbednarz: That's likely a shell issue. The importer splits on `/` and spaces are fine. I'm not in front of a computer at the moment but I think `terraform import resource.foo "abc/key here"` will also work as expected.

jacobbednarz:
LGTM, thanks!

Issue #1360, opened 2021-12-27
cloudflare_access_rule does not update value after v3 release


yurymuski:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```sh
Terraform v1.1.0
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.6.0
```

### Affected resource(s)

cloudflare_access_rule

### Terraform configuration files

```hcl
resource "cloudflare_access_rule" "fw_whitelist" {
  mode  = "whitelist"
  configuration {
    target = "ip"
    value  = "1.2.3.4"
  }
}
```

### Debug output

```sh
Terraform will perform the following actions:

  # cloudflare_access_rule.fw_whitelist will be updated in-place
  ~ resource "cloudflare_access_rule" "fw_whitelist" {
        id   = "08ed90275a3246019129200a5a69b1f2"
        # (1 unchanged attribute hidden)

      ~ configuration {
          ~ value  = "1.2.3.4" -> "5.6.7.8"
            # (1 unchanged attribute hidden)
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.

terraform apply "geo.plan"
cloudflare_access_rule.fw_whitelist: Modifying... [id=08ed90275a3246019129200a5a69b1f2]
cloudflare_access_rule.fw_whitelist: Modifications complete after 6s [id=08ed90275a3246019129200a5a69b1f2]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.


```

### Expected output

Initially IP address adds correctly to cloudflare_access_rule

Update of the resource shows successful status but does not apply changes

1.2.3.4 => 5.6.7.8

firewall/tools =>
![image](https://user-images.githubusercontent.com/20399073/147478690-67a805c0-5386-4238-8c18-38ac8cbaa357.png)

### Actual output

-

### Steps to reproduce

1. Create and apply cloudflare_access_rule
```sh
resource "cloudflare_access_rule" "fw_whitelist" {
  mode  = "whitelist"
  configuration {
    target = "ip"
    value  = "1.2.3.4"
  }
}
```
2. Make changes to the IP and apply
```sh
resource "cloudflare_access_rule" "fw_whitelist" {
  mode  = "whitelist"
  configuration {
    target = "ip"
    value  = "5.6.7.8"
  }
}
```
[...]

PR #1601, opened 2022-05-07
fix(#1360): Fix lifecycle of access_rule update

Nmishin:
According to documentation, to modify an existing Access Rule,
need to create a new Access Rule and delete the existing one.


jacobbednarz:
perfecto! thank you!

PR #1613, opened 2022-05-10
update hc-install from v0.3.1 to v0.3.2

jafowler:
Due to a recent change (https://github.com/hashicorp/hc-install/issues/56) hc-install now throws errors on content-type when attempting to install terraform.


jacobbednarz:
closing in favour of dependabot.

PR #1628, opened 2022-05-17
Add uptycs, intune and crwodstrike

suhrit-cf:


jacobbednarz:
i've made some changes to the underlying structure of this repository and attempted to update the PR here however, it looks like you don't allow maintainers to update your fork. you'll need to manually fetch and merge the current master branch here to resolve these conflicts (or allow maintainers to edit your PR and i'll fix it for you).

suhrit-cf:
Weird `Allow edits by maintainers` is checked

internal/provider/resource_cloudflare_device_posture_integration.go
```diff
@@ -168,6 +173,42 @@ func setDevicePostureIntegrationConfig(integration *cloudflare.DevicePostureInte
 				return fmt.Errorf("api_url is a string")
 			}
 			integration.Config = config
+		case crowdstrike:
+			if config.ClientID, ok = d.Get("config.0.client_id").(string); !ok {
```
jacobbednarz, line 177: these error conditions don't really make sense here (looking above, it looks like this was copy pasted but we should still fix it). i think what you're looking to achieve is erroring if the value isn't provided? but then, the schema doesn't enforce it so perhaps we need to be more strict on the checks in here instead?

jacobbednarz: sorry, i think i wasn't clear enough. `return fmt.Errorf("FOO is a string")` isn't a clear error message as to what is happening. let's update these to be specific to the error we're aiming to handle.

jacobbednarz: @suhrit-cf do you need anything else from me here? i'd like to get this out with the next release if we can as this has sat for a bit now.

jacobbednarz:
you'll also need to rebase this against `master` as the documentation has moved. the file you are after now lives at `templates/resources/cloudflare_device_posture_integration.md`

internal/provider/resource_cloudflare_device_posture_integration.go
suhrit-cf: Hi @jacobbednarz made error message more descriptive.

suhrit-cf:
Also rebased and is now without conflicts.

jacobbednarz (approved):

jacobbednarz:
thank you! this is great.

Issue #1646, opened 2022-05-25
breaking changes for 4.x


jacobbednarz:
Tracking for all the planned cleanup in 4.x

## Internal cleanup 

- [x] Remove all uses of `UsingAccount` in favour of explicit account ID parameters. (#2139)
- [x] Remove reliance on `userBaseUrl`.
- [x] Remove all references to `client.AccountID`. (#2154)

## Schema updates

- [x] `cloudflare_byo_ip_prefix` `prefix_id` should be internal `id`, not a custom attribute.

### Deletions/modifications

- [x] `cloudflare_spectrum_application` needs to support a list of config options (not just IP list) in `edge_ips` (#2219)
- [x] `cloudflare_zone` to swap `account_id` to `Required: true` (from optional) (#2153)
- [x] `cloudflare_load_balancer_pool` to swap `account_id` to `Required: true` (from optional) (#2153)
- [x] `cloudflare_load_balancer_monitor` to swap `account_id` to `Required: true` (from optional) (#2153)
- [x] `cloudflare_account_member` to swap `account_id` to `Required: true` (from optional) (#2153)
- [x] `cloudflare_workers_kv_namespace` to swap `account_id` to `Required: true` (from optional) (#2153)
- [x] `cloudflare_workers_kv` to swap `account_id` to `Required: true` (from optional) (#2153)
- [x] `cloudflare_workers_script` to swap `account_id` to `Required: true` (from optional) (#2153)
- [x] remove ability to create user level resources
  - [x] `cloudflare_access_rule` (#2157)
  - [x] `cloudflare_load_balancer` (#2157)
  - [x] `cloudflare_load_balancer_monitor` (#2157)
  - [x] `cloudflare_load_balancer_pool` (#2157)

### Additions

- [x] `cloudflare_zone` needs to support `account_id` (#1767)
- [x] `cloudflare_account_member` needs to support `account_id` (#1767)
- [x] `cloudflare_access_rule` needs to support `account_id` (#1790)
- [x] `cloudflare_load_balancer_pool` needs to support `account_id` (#1986)
- [x] `cloudflare_load_balancer_monitor` needs to support `account_id` (#1986)
- [x] `cloudflare_workers_kv` needs to support `account_id` (#2049)
- [x] `cloudflare_workers_kv_namespace` needs to support `account_id` (#2049)
- [x] `cloudflare_worker_script` needs to support `account_id` (#2102)
[...]

PR #1647, opened 2022-05-26
AUTH-4232 add saas app remove bookmark specific app resource

AkemiDavisson:


jacobbednarz:
thanks for this @AkemiDavisson! as this is removing a resource causing a breaking change, we need to leave the cleanup until the next major release and support this until we get there. in the meantime, we can deprecate the resource by updating the method calls to emit a `diag.Warn` about the deprecation and include a callout in the documentation to make sure people don't keep using this resource.

i've added this to track via #1646.

PR #1648, opened 2022-05-26
Waitingroom event bugfix

broswen:
- Set ID on waiting room event expansion. 
cloudflare-go uses the id in the struct to create the path for the API request so it must be set. (https://github.com/cloudflare/cloudflare-go/blob/master/waiting_room.go#L342) 
- Propagating changes to use time pointer for PrequeueStartTime. (https://github.com/cloudflare/cloudflare-go/pull/878)

### DIFF
--- a/internal/provider/resource_cloudflare_waiting_room_event.go
+++ b/internal/provider/resource_cloudflare_waiting_room_event.go
@@ -56,7 +56,7 @@ func expandWaitingRoomEvent(d *schema.ResourceData) (cloudflare.WaitingRoomEvent
 		Name:                  d.Get("name").(string),
 		EventStartTime:        eventStartTime,
 		EventEndTime:          eventEndTime,
-		PrequeueStartTime:     &prequeueStartTime,
+		PrequeueStartTime:     prequeueStartTime,
 		Description:           d.Get("description").(string),
 		QueueingMethod:        d.Get("queueing_method").(string),
 		ShuffleAtEventStart:   d.Get("shuffle_at_event_start").(bool),
--- a/internal/provider/resource_cloudflare_waiting_room_event.go
+++ b/internal/provider/resource_cloudflare_waiting_room_event.go
@@ -45,14 +45,13 @@ func expandWaitingRoomEvent(d *schema.ResourceData) (cloudflare.WaitingRoomEvent
 	var prequeueStartTime *time.Time
 	if t, ok := d.GetOk("prequeue_start_time"); ok {
 		prequeueStartTimeValue, err := time.Parse(time.RFC3339, t.(string))
-		prequeueStartTime = &prequeueStartTimeValue
+		prequeueStartTime = cloudflare.TimePtr(prequeueStartTimeValue)
 		if err != nil {
 			return cloudflare.WaitingRoomEvent{}, err
 		}
 	}
 
 	return cloudflare.WaitingRoomEvent{
-		ID:                    d.Id(),
 		Name:                  d.Get("name").(string),
 		EventStartTime:        eventStartTime,
 		EventEndTime:          eventEndTime,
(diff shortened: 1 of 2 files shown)


internal/provider/resource_cloudflare_waiting_room_event.go
```diff
@@ -42,19 +42,21 @@ func expandWaitingRoomEvent(d *schema.ResourceData) (cloudflare.WaitingRoomEvent
 		return cloudflare.WaitingRoomEvent{}, err
 	}
 
-	prequeueStartTime := time.Time{}
+	var prequeueStartTime *time.Time
 	if t, ok := d.GetOk("prequeue_start_time"); ok {
-		prequeueStartTime, err = time.Parse(time.RFC3339, t.(string))
+		prequeueStartTimeValue, err := time.Parse(time.RFC3339, t.(string))
+		prequeueStartTime = &prequeueStartTimeValue
```
jacobbednarz, line 48: we also have `Bool` and `BoolPtr` helper methods if you'd like to use them instead? 

https://github.com/cloudflare/cloudflare-go/blob/9af8b925a93059adb1840ab07c5dbc79c1a428fb/convert_types.go#L75-L86

internal/provider/resource_cloudflare_waiting_room_event.go
```diff
@@ -42,19 +42,21 @@ func expandWaitingRoomEvent(d *schema.ResourceData) (cloudflare.WaitingRoomEvent
...
 		if err != nil {
 			return cloudflare.WaitingRoomEvent{}, err
 		}
 	}
 
 	return cloudflare.WaitingRoomEvent{
+		ID:                    d.Id(),
```
jacobbednarz, line 55: this is going to error in the `Create` where it is referenced as `d.SetId()` isn't called until after the API calls are made and this value doesn't exist.

internal/provider/resource_cloudflare_waiting_room_event.go
```diff
@@ -42,19 +42,21 @@ func expandWaitingRoomEvent(d *schema.ResourceData) (cloudflare.WaitingRoomEvent
 		return cloudflare.WaitingRoomEvent{}, err
 	}
 
-	prequeueStartTime := time.Time{}
+	var prequeueStartTime *time.Time
 	if t, ok := d.GetOk("prequeue_start_time"); ok {
-		prequeueStartTime, err = time.Parse(time.RFC3339, t.(string))
+		prequeueStartTimeValue, err := time.Parse(time.RFC3339, t.(string))
```
jacobbednarz, line 47: we have `Time` and `TimePtr` if you'd like to use those here instead to manually go back and forth.

https://github.com/cloudflare/cloudflare-go/blob/master/convert_types.go#L67-L69

https://github.com/cloudflare/cloudflare-go/blob/master/convert_types.go#L916-L923

internal/provider/resource_cloudflare_waiting_room_event.go
broswen: Makes sense, I'll just extract from the ResourceData and set it manually.

jacobbednarz:
is there a test case we should be adding for this as well?

broswen:
> is there a test case we should be adding for this as well?

Before this change, we would see `prequeue_start_time` set to the default time.Time value (epoch start). Now it should not be set (null).
Is there a terraform test function we can use to compare a state drift? Before the change we would see that prequeue_start_time would be set even though the terraform config doesn't have it set.

broswen:
The latest commit should get that TestCheck working. Since it wasn't using a pointer before it would not set the value if it was a zero value (epoch start).

jacobbednarz:
acceptance tests are passing

```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareWaitingRoomEvent_" -count 1 -parallel 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareWaitingRoomEvent_Create
=== PAUSE TestAccCloudflareWaitingRoomEvent_Create
=== CONT  TestAccCloudflareWaitingRoomEvent_Create
--- PASS: TestAccCloudflareWaitingRoomEvent_Create (16.79s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/provider	17.360s
```

PR #1659, opened 2022-05-31
Fix Go module version

vavsab:
The following commands are failing:
`go get github.com/cloudflare/terraform-provider-cloudflare@v3.15.0`
`go get github.com/cloudflare/terraform-provider-cloudflare/v3@v3.15.0`

For both you get:
`go: github.com/cloudflare/terraform-provider-cloudflare@v3.15.0: invalid version: module contains a go.mod file, so module path must match major version ("github.com/cloudflare/terraform-provider-cloudflare/v3")`

[The manual says](https://therootcompany.com/blog/bump-go-package-to-v2/) that you must change a package name to make it work.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/cloudflare/terraform-provider-cloudflare
+module github.com/cloudflare/terraform-provider-cloudflare/v3
 
 go 1.17
 
--- a/main.go
+++ b/main.go
@@ -4,7 +4,7 @@ import (
 	"flag"
 	"log"
 
-	"github.com/cloudflare/terraform-provider-cloudflare/internal/provider"
+	"github.com/cloudflare/terraform-provider-cloudflare/v3/internal/provider"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/plugin"
 )
 


jacobbednarz:
terraform providers are not intended to be `go get`-able. they need to be run as single binaries via terraform itself. 

as such, this isn't something we'll be accepting a change for.

vavsab:
It's a pity since pulumi is using this go module as a package to wrap the provider. Not following go module conventions requires us to use commit hash instead of readable tags

vavsab:
@jacobbednarz You may see [here](https://github.com/pulumi/pulumi-cloudflare/blob/0bcafa403394fd60a4e9324f4bfb19c1908fb644/provider/go.mod#L6) how it looks. It's actually v3.15.0 but it's totally not obvious.

jacobbednarz:
i knew i recalled seeing something from Hashicorp on this; see https://www.terraform.io/plugin/sdkv2/best-practices/depending-on-providers#do-not-import-providers-as-go-modules for alternatives.

vavsab:
Thanks for the explanation. Yeah. Looks like pulumi wrappers do not follow best practices 🤷‍♂️.
Will just continue using hashes.

PR #1667, opened 2022-06-02
Noan/add pkce idps

Noanan:
Proof Key for Code Exchange (PKCE) has been recently added to the API and UI. This PR increases terraform functionality to match the API/UI. 

As there was no unit tests relevant to the IDPs that support PKCE (azure, fb, google, google apps, okta, oidc), none were added.

However, creation and modification of these relevant IPDs with and without PKCE was tested with this updated provider against production.


.changelog/1652.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
```
jacobbednarz, line 1: this file should be `1667.txt` (match the PR number)

internal/provider/schema_cloudflare_access_identity_provider.go
```diff
@@ -130,6 +130,10 @@ func resourceCloudflareAccessIdentityProviderSchema() map[string]*schema.Schema
 						Type:     schema.TypeString,
 						Optional: true,
 					},
+					"pkce_enabled": {
```
jacobbednarz, line 133: should we default this to `false`?

Noanan: Don't think we need to. This var has the `omitempty` tag in the go struct so the API just won't populate it.

jacobbednarz:
looks good, just pending the upstream changes.

you'll need to rebase this against `master` as the documentation has moved. the file you're editing is now at `docs/resources/cloudflare_access_identity_provider`. alternatively, you can follow in the footsteps of #1677 and swap this resource to be automatically generated resources :)

jacobbednarz:
looks like we have some CI failures here that will need some attention before we merge.

internal/provider/resource_cloudflare_access_identity_provider.go
```diff
@@ -207,6 +207,7 @@ func convertSchemaToStruct(d *schema.ResourceData) (cloudflare.AccessIdentityPro
 		IDPConfig.SsoTargetURL = d.Get("config.0.sso_target_url").(string)
 		IDPConfig.SupportGroups = d.Get("config.0.support_groups").(bool)
 		IDPConfig.TokenURL = d.Get("config.0.token_url").(string)
+		IDPConfig.PKCEEnabled = d.Get("config.0.pkce_enabled").(bool)
```
jacobbednarz, line 210:
```suggestion
		IDPConfig.PKCEEnabled = cloudflare.BoolPtr(d.Get("config.0.pkce_enabled").(bool))
```

.changelog/1652.txt
jacobbednarz: this is still the old PR number

docs/resources/cloudflare_access_identity_provider.md
```diff
@@ -72,7 +72,8 @@ The following arguments are supported:
 - `type` - (Required) The provider type to use. Must be one of: `"centrify"`,
   `"facebook"`, `"google-apps"`, `"oidc"`, `"github"`, `"google"`, `"saml"`,
   `"linkedin"`, `"azureAD"`, `"okta"`, `"onetimepin"`, `"onelogin"`, `"yandex"`.
-- `config` - (Optional) Provider configuration from the [developer documentation][access_identity_provider_guide].
+* `config` - (Optional) Provider configuration from the [developer documentation][access_identity_provider_guide].
+* `pkce_enabled` - (Optional) Enable Proof Key for Code Exchange (PKCE). Only use this if your identity provider supports PKCE for confidential clients.
```
jacobbednarz, line 76: you need to make the changes to the `templates` directory and then run `make docs` to have this updated. `docs` is the generated output so this will be overwritten on next generation.

PR #1668, opened 2022-06-02
Add vnet param to tunnel routes

### LINKED ISSUE #1605: Improvement: add the `virtual_network_id` reference to `cloudflare_tunnel_route` resource

Terraform v1.1.4
provider registry.terraform.io/cloudflare/cloudflare v3.14.0

### Description

Add the `virtual_network_id` (or potentially `virtual_network_name`?) property to `cloudflare_tunnel_route` resource in order to assign it to specific [Tunnel Virtual Network](https://api.cloudflare.com/#tunnel-virtual-network-properties).

It would be an equivalent of existing `cloudflared` CLI command:
`cloudflared tunnel route ip add --vnet vnet-frankfurt 10.1.0.0/16 tunnel-fra`

More info:
https://blog.cloudflare.com/building-many-private-virtual-networks-through-cloudflare-zero-trust/

Related with:
#1602 

### Use cases

It would be great to have an option to manage Tunnel Virtual Networks and assignment of Tunnels' Routes through Terraform instead of CLI/API.

Example 1:
```hcl
resource "cloudflare_argo_tunnel" "tunnel" {
  account_id = "c4a7362d577a6c3019a474fd6f485821"
  name       = "my_tunnel"
  secret     = "AQIDBAUGBwgBAgMEBQYHCAECAwQFBgcIAQIDBAUGBwg="
}

resource "cloudflare_tunnel_route" "example" {
  account_id = "c4a7362d577a6c3019a474fd6f485821"
  tunnel_id  = cloudflare_argo_tunnel.tunnel.id
  network    = "192.0.2.24/32"
  comment    = "New tunnel route for documentation"
  # New reference property
  virtual_network_id = "1d5fdc9e88c8a8c4518b068cd94331fe"
}
```

Example 2 related with #1602:
```hcl

resource "cloudflare_argo_tunnel" "tunnel" {
  account_id = "c4a7362d577a6c3019a474fd6f485821"
  name       = "my_tunnel"
  secret     = "AQIDBAUGBwgBAgMEBQYHCAECAwQFBgcIAQIDBAUGBwg="
}

# New Virtual Network resource from #1602
resource "cloudflare_tunnel_virtual_network" "example" {
  name = "us-east-1-vpc"
  is_default_network = true
  comment = Staging VPC for data science
}

resource "cloudflare_tunnel_route" "example" {
  account_id = "c4a7362d577a6c3019a474fd6f485821"
  tunnel_id  = cloudflare_argo_tunnel.tunnel.id
  network    = "192.0.2.24/32"
  comment    = "New tunnel route for documentation"
  # New reference property
  virtual_network_id = cloudflare_tunnel_virtual_network.example.id
}
```

### References

_No response_

vavsab:
Add missing vnet param
Closes #1605


internal/provider/resource_cloudflare_tunnel_route.go
```diff
@@ -76,7 +89,12 @@ func resourceCloudflareTunnelRouteCreate(ctx context.Context, d *schema.Resource
 		return diag.FromErr(fmt.Errorf("error creating Tunnel Route for Network %q: %w", d.Get("network").(string), err))
 	}
 
-	d.SetId(newTunnelRoute.Network)
+	if virtualNetworkID != "" {
+		// It's possible to create several routes with the same network but different virtual network ids.
+		d.SetId(fmt.Sprintf("%s/%s", newTunnelRoute.Network, virtualNetworkID))
```
jacobbednarz, line 94: if you are doing this, you'll want to make `virtual_network_id` `ForceNew` otherwise, the underlying ID will change everytime the `virtual_network_id` updates and you'll have issues in `Update`/`Read` operations.

we should also use [`stringChecksum`](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/provider/utils.go#L63-L69) to keep the ID value looking consistent with others.

internal/provider/resource_cloudflare_tunnel_route.go
```diff
@@ -105,28 +124,40 @@ func resourceCloudflareTunnelRouteUpdate(ctx context.Context, d *schema.Resource
...
+	err := client.DeleteTunnelRoute(ctx, resource)
 	if err != nil {
-		return diag.FromErr(fmt.Errorf("error deleting Tunnel Route for Network %q: %w", d.Get("network").(string), err))
+		return diag.FromErr(fmt.Errorf("error deleting Tunnel Route for Network %q: %w", network, err))
 	}
 
 	return nil
 }
 
 func resourceCloudflareTunnelRouteImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
jacobbednarz, line 143: this seems unnecessarily complex; let's just force it to include the `virtual_network_id` for explicitness.

jacobbednarz:
as you've updated the `templates` directory, we'll also need to run `make docs` to generate the `docs` equivalent. or if you really want, swap this resource over to the autogenerating documentation in this PR 😄

vavsab:
No problem. Will do 🙂

internal/provider/resource_cloudflare_tunnel_route.go
vavsab: As a `cloudflared` user I know that vnet is totally optional param there.
Making it required during import will create additional headache for developers that are always working inside of the "default" vnet since it's not easy at all to get GUID of the default vnet (you cannot do it through UI. Only by using API or `cloudflared`).

But anyway for me it does not make any difference. So let's make it required.

internal/provider/resource_cloudflare_tunnel_route.go
vavsab: image
Have not found a way to update vnet. Changed it for `ForceNew` and used `stringChecksum` (btw, do you know why we need to use it (except following project code style)?)

vavsab: We always had slashes in ID even before (because CIDR always contains slash) but were not using `stringChecksum`.. so what is the real purpose?

jacobbednarz:
awesome 🍭 thanks!

PR #1674, opened 2022-06-03
Require block vs argument

swoodford-clear:
Fixes error in example:
An argument named "require" is not expected here. Did you mean to define a block of type "require"?


swoodford-clear:
note: trying to add label `workflow/skip-changelog-entry` I'm unable to add any labels

PR #1676, opened 2022-06-06
Autogenerate registry documentation

### LINKED ISSUE #923: Swap documentation CI generation to use hashicorp/terraform-plugin-docs
Hashicorp have deprecated the `make website-test` target in favour of https://github.com/hashicorp/terraform-plugin-docs. To fix CI and get it green again, we need to make the swap and perform the migration.

The steps have been removed in #925 but will need to be reinstated once migrated.

jacobbednarz:
Updates the provider to use terraform-plugin-docs to achieve documentation generation instead of maintaining it by hand.

This lays the foundations for all resources to be updated (individually, to prevent a whole lotta merge conflicts).

Closes #923


internal/provider/provider.go
```diff
@@ -29,86 +55,87 @@ func New(version string) func() *schema.Provider {
...
+					Description: "Whether to print logs from the API client (using the default log library logger). Alternatively, can be configured using the `CLOUDFLARE_API_CLIENT_LOGGING` environment variable.",
 				},
 
 				"account_id": {
 					Type:        schema.TypeString,
 					Optional:    true,
 					DefaultFunc: schema.EnvDefaultFunc("CLOUDFLARE_ACCOUNT_ID", nil),
-					Description: "Configure API client to always use that account",
+					Description: "Configure API client to always use a specific account. Alternatively, can be configured using the `CLOUDFLARE_ACCOUNT_ID` environment variable.",
+					Deprecated:  "Use resource specific `account_id` attributes instead.",
```
noah-d, line 124: Hi, 

Some of the resources don't accept `zone_id` or `account_id` (i.e. [worker_script](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/provider/schema_cloudflare_worker_script.go#L58)). This appears to mean we need to keep the `account_id` defined as input at the cloudflare provider level, even though it warns as Deprecated. Am I understanding this correctly? Thanks!

Issue #1397, opened 2022-01-18
[Ruleset] Single rule override in a managed ruleset disables all other rules


nickbabkin:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

terraform -v
Terraform v1.1.3
on darwin_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.7.0
+ provider registry.terraform.io/hashicorp/aws v3.72.0
+ provider registry.terraform.io/hashicorp/http v2.1.0
+ provider registry.terraform.io/tchupp/env v0.0.2

### Affected resource(s)

cloudflare_ruleset

### Terraform configuration files

```hcl
resource "cloudflare_ruleset" "terraform_managed_resource_8313e85ab7164e90adfaee5254de6322" {
  description = "Cloudflare Managed WAF ruleset"
  kind        = "zone"
  name        = "Managed WAF"
  phase       = "http_request_firewall_managed"
  zone_id     = "94691220711a494300a5b3021d0a8443"
  rules {
    action      = "execute"
    description = "Execute Cloudflare Managed Ruleset on my zone-level phase entry point ruleset"
    enabled     = true
    expression  = "true"
    # id          = "db6a5a81d2a14d92be019b27852c227e"
    # ref         = "db6a5a81d2a14d92be019b27852c227e"
    # version     = "1"
    action_parameters {
      overrides {
        rules {
          action  = "log"
          enabled = true
          id      = "0f2da91cec674eb58006929e824b817c"
        }
        rules {
          action  = "log"
          enabled = true
          id      = "81718f38edde45a58298189e113e4f59"
        }
      }
      id      = "efb7b8c949ac4650a09736fc376e9aee"
      version = "latest"
    }
  }
}
```

### Debug output

PLAN:
```
-----------------------------------------------------: timestamp=2022-01-18T16:08:37.257Z
[...]

PR #1684, opened 2022-06-08
add warning about #1397

### LINKED ISSUE #1683: Update docs for #1397 until a fix is in place

latest

### Description

Due to the potential impact of #1397 to disable WAF we should add a line in the docs to let customers know that issues may arise if they don't set 'enabled=true' correctly.

### Use cases

Help customers not make mistakes that could have significant impact on configuration while bugs exist.

```hcl
...
```

### References

_No response_

garrettgalow:
closes #1683


docs/resources/cloudflare_ruleset.md
```diff
@@ -33,6 +33,8 @@ and clean up the resources before attempting to configure them with
 Terraform. This is because Terraform will fail to apply if configuration
 already exists to prevent blindly overwriting changes.
 
+~> **Note:** Until [issue #1397](https://github.com/cloudflare/terraform-provider-cloudflare/issues/1397) is fixed, when configuring a ruleset with overrides, you will need to set `enabled = true` under `action_parameters` to ensure rules are not unintentially disabled.
```
jacobbednarz, line 36: we need to move this to `templates/resources/cloudflare_ruleset.md`. `docs/` is the generated artifacts whereas `templates` is the source.

note: this is also the case for all `enabled` fields in the resource.

Issue #1579, opened 2022-04-25
Page rule validation error message not being returned


brandonstrohmeyer:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.1.9
on darwin_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.13.0

### Affected resource(s)

cloudflare_page_rule

### Terraform configuration files

```hcl
resource "cloudflare_page_rule" "this" {
  zone_id  = module.cloudflare_stro_debug.website.zone.id
  target   = "${module.cloudflare_stro_debug.website.zone.zone}/test"
  priority = 1

  actions {

    always_use_https = true

    cache_key_fields {
      header {}
      query_string {}
      user {}
      cookie {}
      host {}
    }
  }
}
```

### Debug output

```
---[ RESPONSE ]--------------------------------------
HTTP/2.0 400 Bad Request
Connection: close
Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0
Cf-Cache-Status: DYNAMIC
Cf-Ray: 7017f3c91ba9e3ca-ATL
Content-Type: application/json
Date: Mon, 25 Apr 2022 15:01:09 GMT
Expect-Ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
Expires: Sun, 25 Jan 1981 05:00:00 GMT
Pragma: no-cache
Server: cloudflare
Set-Cookie: __cflb=0H28vgHxwvgAQtjUGU4vq74ZFe3sNVUZaMRnT3FnvWR; SameSite=Lax; path=/; expires=Mon, 25-Apr-22 17:31:10 GMT; HttpOnly
Set-Cookie: __cfruid=fb89178db7f78f805459602d86f6c6a8d414fd04-1650898869; path=/; domain=.api.cloudflare.com; HttpOnly; Secure; SameSite=None
Strict-Transport-Security: max-age=31536000
Vary: Accept-Encoding
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN

{
 "success": false,
 "errors": [
  {
   "code": 1004,
   "message": "Page Rule validation failed: See messages for details."
  }
 ],
 "messages": [
  {
   "code": 1,
[...]

PR #1693, opened 2022-06-11
Update misleading page rule docs

etienne-napoleone:
Apply **will** fail if `ignore` is set as `true` and any of `exclude` or `include` is non-empty. It is not ignored.

It was a pain to debug due to https://github.com/cloudflare/terraform-provider-cloudflare/issues/1579


jacobbednarz:
this change needs to be made to the file in `templates` directory and then run `make docs` to generate this file. without it, your changes will be lost on next generation.

etienne-napoleone:
ah bummers, i will let you take care of it

PR #1702, opened 2022-06-15
Add documentation to cloudflare_notification_policy for workers alerts

jwstanly:
Hi, I was able to create and destroy `workers_alert` and `weekly_account_overview` notifications with the Terraform template below thanks to these alert types being supported in the underlying Cloudflare API.

I've added documentation to reflect this.

```
resource "cloudflare_notification_policy" "my_workers_alert" {
  account_id = "account-id-here"
  name = "Policy for Worker Alerts"
  description = "Notification policy to alert when my Worker has unusual CPU usage"
  enabled     =  true
  alert_type  = "workers_alert"

  email_integration {
    id   =  "my@email.com"
  }
}

resource "cloudflare_notification_policy" "my_workers_weekly_overview" {
  account_id = "account-id-here"
  name = "Policy for Worker Weekly Oberview"
  description = "Notification policy to alert when my Worker Weekly Account Overview"
  enabled     =  true
  alert_type  = "weekly_account_overview"

  email_integration {
    id   =  "my@email.com"
  }
}
```


jacobbednarz (changes requested):
the documentation structure here is unfortunately confusing and incorrect; `alert_type` is a string that accepts these not a nested block so this doesn't reflect the reality here.

additionally, `docs` is the generated output from `make docs`. the source lives in `templates` or the resource schema itself so we'll need to adjust those to clean this up.

jacobbednarz:
let me see if i can quickly swap this over so the documentation reflects what is happening here better.

jacobbednarz:
i've swapped this resource over to the correct format and automatically generating the documentation so this is now good. however, we have lost the more verbose descriptions of the individual alert types so i've directed those to the developer docs which have a better write up of each. one drawback to this is that the developer docs don't have the API values we should be expecting here. 

@jwstanly @broswen WYDT? should we get the API documentation updated with those expected values?

broswen:
> i've swapped this resource over to the correct format and automatically generating the documentation so this is now good. however, we have lost the more verbose descriptions of the individual alert types so i've directed those to the developer docs which have a better write up of each. one drawback to this is that the developer docs don't have the API values we should be expecting here.
> 
> @jwstanly @broswen WYDT? should we get the API documentation updated with those expected values?

I had trouble finding the `alert_type` values when I was working on this resource previously. I resorted to digging in the dash and inspecting the API calls to find them. Ideally I think it would be best to have the raw API value of the `alert_type` in the dev docs along with the existing descriptions of the alert type. And maybe even a raw list of values in the API docs (without descriptions).

There is also[ this endpoint that lists the available alert types](https://api.cloudflare.com/#notification-alert-types-get-alert-types). It could be useful to mention it more prominently. The api documentation for notification policy [mentions an endpoint](https://api.cloudflare.com/#notification-policies-create-a-notification-policy) to use but doesn't link to it.

jacobbednarz:
sounds like a plan to me. i'll create a ticket for notifications/PCX to have a look at this and in the meantime, we can land this.

thanks @jwstanly for raising this one!

Issue #1422, opened 2022-02-01
`cloudflare_teams_list` with more than 50 items does not work


tjstansell:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```
Terraform v1.1.4
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.8.0
+ provider registry.terraform.io/hashicorp/external v2.2.0
```

### Affected resource(s)

* cloudflare_teams_list

### Terraform configuration files

```hcl
resource "cloudflare_teams_list" "bad_domains" {
  account_id = local.account_id
  name       = "bad domains"
  type       = "DOMAIN"
  items      = local.bad_domains
}
```

### Debug output

n/a

### Expected output

The list is correctly created, but subsequent plans always return diffs if there are more than 50 items in the list.   If you try to apply the diff, it regularly thinks it's removing a bunch of items but also adding others, which end up being conflicting entries.

This is actually an API limitation in that it doesn't properly do paging and, by default, only returns 50 items.  I discovered after tracing the teams UI that it does, however, support an undocumented `?limit=` parameter.  I manually tested the API with `?limit=100000` and it returned the 1240 items in my list just fine.  It doesn't seem there's a limit to the size of the limit parameter.

I don't know if the provider here needs to be updated or if the `cloudflare-go` library should just always use a large limit parameter to get the entire list (since there's no paging or way to know if your list has been truncated by the API).

### Actual output

n/a

### Steps to reproduce

1. create a `cloudflare_teams_list` with more than 50 items
2. apply (works)
3. plan (always shows diffs)

tjstansell:
FWIW, the [docs on teams lists](https://developers.cloudflare.com/cloudflare-one/policies/lists) says:
> Your lists can include up to 5,000 entries for Enterprise subscriptions and 1,000 for Standard subscriptions.

tjstansell:
FYI, I opened support ticket 2431495 regarding this and they claim there's already a backend gateway engineering ticket to handle this, so there's likely nothing to do here on the terraform side other than wait for the API to actually get fixed.  If others who run into this could also push on this to help accelerate it, that'd be great.  I've discussed this issue with the engineering side directly multiple times over the past few months.  It's unfortunate this bug hasn't been addressed yet.

tjstansell:
[...]

PR #1706, opened 2022-06-16
Adds recently supported per_page/pagination teams_list

AkemiDavisson:
Relates to: https://github.com/cloudflare/terraform-provider-cloudflare/issues/1422
cf-go PR: https://github.com/cloudflare/cloudflare-go/pull/897


internal/provider/resource_cloudflare_teams_list.go
```diff
@@ -73,6 +73,7 @@ func resourceCloudflareTeamsListRead(ctx context.Context, d *schema.ResourceData
 	listItems, _, err := client.TeamsListItems(ctx, cloudflare.TeamsListItemsParams{
 		AccountID: accountID,
 		ListID:    d.Id(),
+		PaginationOptions: cloudflare.PaginationOptions{ PerPage: 1000 },
```
jacobbednarz, line 76: this is a step in the right direction but doesn't account for when more than 1000 items are available. ideally, we need to paginate all available pages instead.

jacobbednarz: i've added the iteration via e1a3384 so we should be right with large list items now.

jacobbednarz:
@tjstansell you've mentioned in a couple of issues you've got near 5k team list items. are you able to take this PR for a spin and confirm you get them all back and managed correctly?

jacobbednarz:
Oops! It looks like no changelog entry is attached to this PR. Please include a release note as described in https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/changelog-process.md.

Example: 

~~~
```release-note:TYPE
Release note
```
~~~

If you do not require a release note to be included, please add the `workflow/skip-changelog-entry` label.

jacobbednarz:
Oops! It looks like no changelog entry is attached to this PR. Please include a release note as described in https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/changelog-process.md.

Example: 

~~~
```release-note:TYPE
Release note
```
~~~

If you do not require a release note to be included, please add the `workflow/skip-changelog-entry` label.

jacobbednarz:
Oops! It looks like no changelog entry is attached to this PR. Please include a release note as described in https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/changelog-process.md.

Example: 

~~~
```release-note:TYPE
Release note
```
~~~

If you do not require a release note to be included, please add the `workflow/skip-changelog-entry` label.

jacobbednarz:
Oops! It looks like no changelog entry is attached to this PR. Please include a release note as described in https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/changelog-process.md.

Example: 

~~~
```release-note:TYPE
Release note
```
~~~

If you do not require a release note to be included, please add the `workflow/skip-changelog-entry` label.

jacobbednarz:
Oops! It looks like no changelog entry is attached to this PR. Please include a release note as described in https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/changelog-process.md.

Example: 

~~~
```release-note:TYPE
Release note
```
~~~

If you do not require a release note to be included, please add the `workflow/skip-changelog-entry` label.

jacobbednarz:
Oops! It looks like no changelog entry is attached to this PR. Please include a release note as described in https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/changelog-process.md.

Example: 

~~~
```release-note:TYPE
Release note
```
~~~

If you do not require a release note to be included, please add the `workflow/skip-changelog-entry` label.

jacobbednarz:
Oops! It looks like no changelog entry is attached to this PR. Please include a release note as described in https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/changelog-process.md.

Example: 

~~~
```release-note:TYPE
Release note
```
~~~

If you do not require a release note to be included, please add the `workflow/skip-changelog-entry` label.

jacobbednarz:
Oops! It looks like no changelog entry is attached to this PR. Please include a release note as described in https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/changelog-process.md.

Example: 

~~~
```release-note:TYPE
Release note
```
~~~

If you do not require a release note to be included, please add the `workflow/skip-changelog-entry` label.

internal/provider/resource_cloudflare_teams_list.go
```diff
@@ -70,14 +70,32 @@ func resourceCloudflareTeamsListRead(ctx context.Context, d *schema.ResourceData
...
+		PaginationOptions: cloudflare.PaginationOptions{PerPage: 1000},
 	})
 	if err != nil {
 		return diag.FromErr(fmt.Errorf("error finding Teams List %q: %w", d.Id(), err))
 	}
-	d.Set("items", convertListItemsToSchema(listItems))
+
+	allListItems = append(allListItems, listItems...)
+
+	for i := 2; i < responseInfo.TotalPages; i++ {
```
tjstansell, line 86: Shouldn't this be `<=`?

internal/provider/resource_cloudflare_teams_list.go
AkemiDavisson: Thanks Jacob this makes a lot more sense than the hardcoded default. I was unable to get to this Friday and appreciate the assist! 💯

internal/provider/resource_cloudflare_teams_list.go
tjstansell: Actually, in my list that has 1240 items, if I query the API for page 2 with a page size of 10 I get the following response:
```
  "result_info": {
    "page": 2,
    "per_page": 10,
    "count": 10,
    "total_count": 1240
  }
```
There is no TotalPages in there.   Is that something the cloudflare-go library is determining?

tjstansell: If i query for page 2 with a page size of 1000, I get:
```
  "result_info": {
    "page": 2,
    "per_page": 1000,
    "count": 240,
    "total_count": 1240
  }
```
So maybe it should be checking if the `count` returned is less than the `page_size` requested to see if there's more?  Or, I guess do math to figure out if it got the last item on the list with something like `(($page-1) * $per_page) + $count) < $total_count`.  The API response is kinda terrible.  Normally there's a clear indicator if there are additional items and a `next` token or page number to key off of.

jacobbednarz: `TotalPages` is a part of the v4 envelope that teams should be emitting for paginated results. let me confirm why it isn't present here before we swap to using the total count instead.

fwwieffering:
@jacobbednarz is this PR still waiting on anything?

jacobbednarz:
@fwwieffering see two comments up from yours 😄

geota:
@jacobbednarz API now returnst his: https://api.cloudflare.com/#zero-trust-lists-zero-trust-list-items 

```
  "result_info": {
    "page": 1,
    "per_page": 20,
    "count": 1,
    "total_count": 2000
  }
  ```
  
  
  Can you confirm this is not sufficient and you need total_pages as well? E.g. total_count / per_page? That seems redundant, but we can add it.

**Comment by jacobbednarz** [comment]
there is a little bit of nuance in the v4 envelope that can be confusing.

- `page` is the current page of results
- `per_page` is the number of results expected in the current response
- `count` is the number of *filtered* results found (say if you provide `param1=blah` to the request)
- `total_count` is the number of all results without filters
- `total_pages` is to return the number of all results without filters

i don't agree with this pattern (including filtered + unfiltered options in the response, as really, we don't actually care) however, that is the intended use of those fields which we're lacking some of here. some teams internally don't follow this and that's fine but we're trying to make sure all the fields at least exist for a consistent payload.

we are also in the process of standardising pagination in cloudflare-go so in the near future, consumers can do something like `paginateResults(...)` and this is will fetch all the results - see https://github.com/cloudflare/cloudflare-go/blob/dc6c002de52f15566cbcedb5be9a5e758858dabb/pagination.go. this relies on the `TotalPages` field being present.

**Comment by jacobbednarz** [comment]
i've updated the internals of cloudflare-go to handle the pagination automatically via cloudflare/cloudflare-go#1114 so we no longer need it here. once that is released, this PR will be good to merge.

acceptance tests are all green

```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareTeamsList_" -count 1 -parallel 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareTeamsList_Basic
--- PASS: TestAccCloudflareTeamsList_Basic (10.73s)
=== RUN   TestAccCloudflareTeamsList_LottaListItems
--- PASS: TestAccCloudflareTeamsList_LottaListItems (31.84s)
=== RUN   TestAccCloudflareTeamsList_Reordered
--- PASS: TestAccCloudflareTeamsList_Reordered (21.29s)
PASS
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/provider   64.198s
```

PR #1712, opened 2022-06-20
LogpushJob.Filter is optional

sbfaulkner:
> **Note** depends on the new release of cloudflare-go

this PR changes the usage of LogpushJob.Filter after the fix in cloudflare/cloudflare-go#937

* set pointer instead of copying struct to create LogpushJob from resource
* check nil pointer when creating resource from LogpushJob


sbfaulkner:
rebased to not include changes from our fork

Issue #154, opened 2018-11-15
cloudflare_record used with aws_acm_certificate always thinks it has to update


SpamapS:
### Terraform Version
0.11.10

### Affected Resource(s)
Please list the resources as a list, for example:
- cloudflare_record
- aws_acm_certificate

### Terraform Configuration Files
```hcl
variable "domain" {}
variable "access_key" {}
variable "secret_key" {}
variable "cluster_region" {}
variable "cf_email" {}
variable "cf_token" {}
variable "subject" {}

provider "aws" {
  region     = "${var.cluster_region}"
  access_key = "${var.access_key}"
  secret_key = "${var.secret_key}"
}

provider "cloudflare" {
  email = "${var.cf_email}"
  token = "${var.cf_token}"
}

resource "aws_acm_certificate" "cert" {
  domain_name       = "${var.subject}"
  validation_method = "DNS"

  lifecycle {
    create_before_destroy = true
  }
}

resource "cloudflare_record" "validation" {
  domain = "${var.domain}"
  name   = "${aws_acm_certificate.cert.domain_validation_options.0.resource_record_name}"
  type   = "${aws_acm_certificate.cert.domain_validation_options.0.resource_record_type}"
  value  = "${aws_acm_certificate.cert.domain_validation_options.0.resource_record_value}"
}

resource "aws_acm_certificate_validation" "cert" {
  certificate_arn         = "${aws_acm_certificate.cert.arn}"
  validation_record_fqdns = ["${cloudflare_record.validation.hostname}"]
}

output "cert_arn" {
  value = "${aws_acm_certificate.cert.arn}"
}
```

### Debug Output
This has secret data in it, and is part of a much bigger configuration. If developers cannot reproduce then I can provide but redaction may take a long time.

### Panic Output
N/A

### Expected Behavior
Resource is already up to date, so I expect there to be no changes.

### Actual Behavior
No matter what, terraform completes successfully, but the next run always wants to add a single "." to the value, which is entirely unnecessary and is ignored by the API as equivalent.

```
  ~ module.zuul-cert.cloudflare_record.validation
      value: "_XXXXXXXXXX.yyyyyyyyyy.acm-validations.aws" => "_XXXXXXXXXX.yyyyyyyyyy.acm-validations.aws."
```

### Steps to Reproduce

1. `terraform apply`

### Important Factoids
N/A

### References
N/A

jacobbednarz:
[...]

PR #1713, opened 2022-06-20
resource_cloudflare_record: Add DiffSuppressFunc to value param

stack72:
Fixes: #154

When a trailing period is passed to the cloudflare_record resource,
we get a perpetual diff that looks as follows:

```
~ value: "_dbdxxx05384ca.hnyhpvdqhv.acm-validations.aws" => "_dbdxxxf205384ca.hnyhpvdqhv.acm-validations.aws."
```

We should ensure that we are adhering to the following RFCs where it
states that an FQDN can have a trailing `.`

Therefore, we can now use a DiffSuppressFunc to check that the old
value with the final `.` removed and the new value with the `.`
removed are the same and if they are then we don't need to cause a
diff

### DIFF
--- a/.changelog/1713.txt
+++ b/.changelog/1713.txt
@@ -1,0 +1,3 @@
+``release-note:enhancement
+resource/cloudflare_record: ensure trailing `.` in `value` don't cause surious diffs
+```
--- a/internal/provider/schema_cloudflare_record.go
+++ b/internal/provider/schema_cloudflare_record.go
@@ -42,6 +42,9 @@ func resourceCloudflareRecordSchema() map[string]*schema.Schema {
 			Optional:      true,
 			Computed:      true,
 			ConflictsWith: []string{"data"},
+			DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+				return strings.TrimSuffix(old, ".") == strings.TrimSuffix(new, ".")
+			},
 		},
 
 		"data": {


jacobbednarz:
thanks for this @stack72, appreciate it! as I mentioned in #154 (and a couple of follow up issues), I'm still not sure if we want to be covering this up as they are potentially different things depending on that character. as this tool is used by people migrating from services that rely on this to Cloudflare that makes it optional, I'm worried we may be covering up a config issue in doing this. let me check with the DNS folks whether we have better safeguards in place now to support this. 

you mentioned adhering to RFCs; could you provide which part we are missing as to the best of my knowledge, the mention of trailing dots is optional based on the implementation.

stack72:
Hi @jacobbednarz 

Thanks for the response, the following [RFC](https://datatracker.ietf.org/doc/html/rfc1034) - section 3.1 - states this:

```
When a user needs to type a domain name, the length of each label is
omitted and the labels are separated by dots (".").  Since a complete
domain name ends with the root label, this leads to a printed form which
ends in a dot.  We use this property to distinguish between:

   - a character string which represents a complete domain name
     (often called "absolute").  For example, "poneria.ISI.EDU."

   - a character string that represents the starting labels of a
     domain name which is incomplete, and should be completed by
     local software using knowledge of the local domain (often
     called "relative").  For example, "poneria" used in the
     ISI.EDU domain.
```

Right now the provider only handles a user based domain name and not the printed form (including the trailing dot)

jacobbednarz:
thanks 🙏 the API definitely allows both however, (this is where my memory gets fuzzy) the response can be normalised with the trailing period as not everyone had the entitlements to use it in that. iirc, it _was_ being phased out but I will need to confirm.

stack72:
I understand - thanks for confirmation on this. Just to add that AWS ACM does standardise on the printed form now so it would be really good to be able to support this easily and without having to make users cut their strings :)

I'm happy to pull this PR into a Pulumi fork of the provider if you'd rather I did that so that we can take advantage in pulumi-cloudflare

Paul

jacobbednarz:
the other thing we have to consider here as well is that given a zone of `example.com`, a value of `foo` results in the API response being `foo.example.com`. in it's current state, i think that would result in a permadiff but something to keep in mind.

jacobbednarz:
i've been chatting internally with the DNS team and i now have a list of the available inputs, their expected outputs and some normalisation that we modified recently. 

within this API, we _do_ take into account the trailing comma to denote whether the value is canonical or not however, that is no longer considered in the response itself since the input value is restricted and must now contain the zone name itself for that to take effect. 

| input          | zone name     | output                |
| -------------- | ------------- | --------------------- |
| `foo`          | `example.com` | `foo.example.com`     |
| `foo.`         | `example.com` | `foo.example.com`     |
| `foo.bar`      | `example.com` | `foo.bar.example.com` |
| `@`            | `example.com` | `example.com`         |
| ` `/`""`         | `example.com` | `example.com`       |
| `example.com.` | `example.com` | `example.com`         |
| `example.com`  | `example.com` | `example.com`         |


the other scenarios (such as `.` and `foo.`) won't be an issue here without resulting in a permadiff since the API expands those completely.

i think we're safe with the normalisation as is now to merge this.

PR #1717, opened 2022-06-21
validate logpush job name

sbfaulkner:
when terraforming a `cloudflare_logpush_job` an invalid `name` attribute will successfully plan, but fail to apply

this PR adds validation using the appropriate regex so that it can be caught during plan instead

### DIFF
--- a/.changelog/1717.txt
+++ b/.changelog/1717.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/cloudflare_logpush_job: validate name attribute
+```
--- a/internal/provider/schema_cloudflare_logpush_job.go
+++ b/internal/provider/schema_cloudflare_logpush_job.go
@@ -2,6 +2,7 @@ package provider
 
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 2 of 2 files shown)


ChristophShyper:
@jacobbednarz @sbfaulkner  Is this correct https://github.com/cloudflare/terraform-provider-cloudflare/pull/1717/files#diff-edc89067cd072336937f133d5ec6e811e3557e0670723f455231935a12cf0e0bR33 ?
It's `^[a-zA-Z0-9.-]+$`, but shoduln't it be `^[a-zA-Z0-9.-]*$`?
Because the new version is failing with:
```
│ Error: invalid value for name (must contain only alphanumeric characters, hyphens, and periods)
│ 
│   with module.logpush.cloudflare_logpush_job.this,
│   on ../submodule-cloudflare-logpush/main.tf line 25, in resource "cloudflare_logpush_job" "this":
│   25:   name                = ""
│ 
╵
```

It's another time this month when new release is breaking integration...

jacobbednarz:
yep, regex should swap out those characters. if you send over a PR to address, I'll review in the morning.

sbfaulkner:
sorry about that - I didn't realize that name could be blank

ChristophShyper:
When you create a logpush job from console page it's created without a name.
Or at least it's behaving like that when we imported it to Terraform.

PR #1721, opened 2022-06-22
Fix typo in ruleset documentation

bszuldyner-ledger:


jacobbednarz:
the documentation in `docs` is automatically generated so this change will have no effect on next generation. 

as this is to do with the headers, which are automatically added, this is something you'll need to address in terraform-plugin-docs if this is a bug.

PR #1745, opened 2022-06-30
fix logpush job name validation regex

ChristophShyper:
Should be `^[a-zA-Z0-9.-]*$` instead of `^[a-zA-Z0-9.-]+$`.


ChristophShyper:
wrong

PR #1689, opened 2022-06-10
resource/cloudflare_ruleset: deprecate `enabled` (immediately) in overrides in favour of `status`

jacobbednarz:
Within Rulesets, an overridden property can have three `enabled` states:

- `true`: The value is forced on.
- `false`: The value is forced off.
- missing: The value reverts to the default value provided by the
  service.

Using `cloudflare-go` directly, all the configuration variations work as
expected thanks to the boolean pointers. However, Terraform doesn't
support the missing state since `TypeBool`s default is `false` (not
omitted). This leads to some rough inconsistencies where operations that
rely on the defaults and then attempt another run will result in it
being disabled without the operators knowledge.

To address the underlying issue, we're immediately deprecating `enabled`
(TypeBool) inside of overridden properties in favour of `status` with
the three following states:

- `enabled`: Value is forced on.
- `disabled`: Value is forced off.
- `""`: Value is reset to the service default.

We've added the translation layer in Terraform as that is where the bug
is instead of updating the service and the API responses.

Closes #1397

PR #1746, opened 2022-06-30
Fix default values for redirect list updates.

orium:


internal/provider/resource_cloudflare_list.go
```diff
@@ -209,8 +209,8 @@ func buildListItemsCreateRequest(resource *schema.ResourceData, items []interfac
 			var preservePathSuffix *bool = nil
 
 			hasField := func(field string) bool {
-				_, has := resource.GetOkExists(fmt.Sprintf("item.%d.value.0.redirect.0.%s", i, field))
-				return has
+				v := getRawValue(fmt.Sprintf("item.%d.value.0.redirect.0.%s", i, field), resource.GetRawConfig())
```
jacobbednarz, line 212: we really shouldn't be using `getRawValue` as it calls Terraform internal methods without going via the abstraction layers. it was initially planned to help with tristate booleans in Terraform configs but we instead took another approach due to the stability and on going issues with it. it really should have been removed.

what is the issue we're hitting here that requires it?

jacobbednarz: this is the exact issue type we ended up solving in #1689 with a breaking change. as i suspect it is the same underlying cause of tristate boolean values, here was the testing outcome i did then with `d.GetRawConfig`.

|Initial state| Updated state|Outcome|
|---|---|---|
|enabled: false|enabled: true| ✅ |
| enabled: false | enabled: nil | ❌ |
| enabled: true | enabled: false | ✅ |
| enabled: true | enabled: nil | ✅ |
| enabled: nil | enabled: true | ✅ |
| enabled: nil | enabled: false | ❌ |

as you can see, this didn't fully solve the issue despite getting us most of the way there. another drawback is that this approach is using internal methods to Terraform and isn't recommended to be relied upon.

i'll grab some time with you to discuss what we can do going forward here.

orium: :( It really shouldn't be this hard to get optional values to work.  In any case, I've updated the PR and avoided the issue by negating the field name, so the default is now `false`.  This is, of course, a breaking change, but the field was misbehaving any way, and was just introduced in the last release.

jacobbednarz: agree; i've flicked you an internal ticket that covers the context and our potential next steps. as this is an issue in Terraform core (not the provider) we are pretty limited in the options to take. 

eventually, we'll move to terraform-plugin-framework which doesn't have this issue with uninitialised boolean values.

jacobbednarz:
i'm having a look through this PR and would supplying `false` as the default solve the issue here? also, i don't think the latest commit actually solves this as it is still a boolean - for the workaround to work (heh), we'll likely need a string that allows "enabled"/"disabled"/"" as the values.

internal/provider/schema_cloudflare_list.go
```diff
@@ -69,29 +69,33 @@ var listItemElem = &schema.Resource{
...
+									Description:  "Whether the redirect target url should keep the query string of the request's url.",
+									Type:         schema.TypeString,
+									Optional:     true,
+									ValidateFunc: validation.StringInSlice([]string{"disabled", "enabled"}, false),
 								},
 								"preserve_path_suffix": {
-									Description: "Whether to preserve the path suffix when doing subpath matching.",
-									Type:        schema.TypeBool,
-									Optional:    true,
+									Description:  "Whether to preserve the path suffix when doing subpath matching.",
```
jacobbednarz, line 95:
```suggestion
									Description:  fmt.Sprintf("Whether to preserve the path suffix when doing subpath matching. %s", renderAvailableDocumentationValuesStringSlice([]string{"disabled", "enabled"})),
```

jacobbednarz:

```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareList_" -count 1 -parallel 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareList_Exists
--- PASS: TestAccCloudflareList_Exists (10.63s)
=== RUN   TestAccCloudflareList_UpdateDescription
--- PASS: TestAccCloudflareList_UpdateDescription (18.67s)
=== RUN   TestAccCloudflareList_Update
--- PASS: TestAccCloudflareList_Update (49.49s)
PASS
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/provider   79.346s
```

PR #1762, opened 2022-07-08
Add SaaS Application support

### LINKED ISSUE #1226: Support for SaaS Access Application

```
Terraform v1.0.7
on darwin_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.1.0
```

### Description

We'd like to manage our SaaS application through Terraform.

However, the `cloudflare_access_application` resource currently [doesn't support the `saas` type](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/cloudflare/resource_cloudflare_access_application.go#L55).
The API seems to [support it](https://api.cloudflare.com/#access-applications-create-access-application).

I tried to patch the provider to accept this type but it's more complicated as we need to be able to configure the `saas_app` block and the `domain` required field would be the value returned after creating the app (eg: `<organization>.cloudflareaccess.com/cdn-cgi/access/sso/saml/<uid>`).
This doesn't seem to be documented in the API so I'm not sure if it's actually supported.

### Use cases

Support for SaaS applications.

```hcl
resource "cloudflare_access_application" "saas_app" {
  account_id = var.cloudflare_account_id
  name       = "My Great Saas App"
  type       = "saas"
  domain     = "???"

  saas_app {
      consumer_service_url = "https://saas-app.example/sso/saml/consume",
      sp_entity_id  = "saas-app.example"
      name_id_format =  "email"
  }

  ...
}  
```

### References

_No response_

ouranos:
Closes #1226

Added in cloudflare-go: https://github.com/cloudflare/cloudflare-go/pull/900

I've tested in our environment which had a custom fork of the SDK previously and the plan is not showing any difference when using the new SDK. Opening a draft PR to get some feedback and I'll do more testing in the meantime.

I need a bit of help handling the `domain` attribute. From my research (before the doc was available), it seems that it is not required for SaaS applications and will be computed for them (eg: `https://<organisation>.cloudflareaccess.com/cdn-cgi/access/sso/saml/<uuid>`.
The updated documentation seems to confirm it: https://api.cloudflare.com/#access-applications-properties

### DIFF
--- a/internal/provider/resource_cloudflare_access_application.go
+++ b/internal/provider/resource_cloudflare_access_application.go
@@ -64,6 +64,10 @@ func resourceCloudflareAccessApplicationCreate(ctx context.Context, d *schema.Re
 		newAccessApplication.CorsHeaders = CORSConfig
 	}
 
+	if _, ok := d.GetOk("saas_app"); ok {
+		newAccessApplication.SaasApplication = convertSaasSchemaToStruct(d)
+	}
+
 	tflog.Debug(ctx, fmt.Sprintf("Creating Cloudflare Access Application from struct: %+v", newAccessApplication))
 
 	identifier, err := initIdentifier(d)
@@ -133,6 +137,11 @@ func resourceCloudflareAccessApplicationRead(ctx context.Context, d *schema.Reso
 		return diag.FromErr(fmt.Errorf("error setting Access Application CORS header configuration: %w", corsConfigErr))
 	}
 
+	saasConfig := convertSaasStructToSchema(d, accessApplication.SaasApplication)
+	if saasConfigErr := d.Set("saas_app", saasConfig); saasConfigErr != nil {
+		return diag.FromErr(fmt.Errorf("error setting Access Application SaaS app configuration: %w", saasConfigErr))
+	}
+
 	return nil
 }
 
@@ -160,6 +169,10 @@ func resourceCloudflareAccessApplicationUpdate(ctx context.Context, d *schema.Re
 		ServiceAuth401Redirect:  d.Get("service_auth_401_redirect").(bool),
 	}
 
+	if appType != "saas" {
+		updatedAccessApplication.Domain = d.Get("domain").(string)
+	}
+
 	if len(allowedIDPList) > 0 {
 		updatedAccessApplication.AllowedIdps = allowedIDPList
 	}
@@ -170,6 +183,11 @@ func resourceCloudflareAccessApplicationUpdate(ctx context.Context, d *schema.Re
 			return diag.FromErr(err)
 		}
 		updatedAccessApplication.CorsHeaders = CORSConfig
+	}
+
+	if _, ok := d.GetOk("saas_app"); ok {
+		saasConfig := convertSaasSchemaToStruct(d)
+		updatedAccessApplication.SaasApplication = saasConfig
 	}
 
 	tflog.Debug(ctx, fmt.Sprintf("Updating Cloudflare Access Application from struct: %+v", updatedAccessApplication))
--- a/internal/provider/resource_cloudflare_access_application_test.go
+++ b/internal/provider/resource_cloudflare_access_application_test.go
@@ -39,6 +39,7 @@ func TestAccCloudflareAccessApplication_BasicZone(t *testing.T) {
 					resource.TestCheckResourceAttr(name, "type", "self_hosted"),
 					resource.TestCheckResourceAttr(name, "session_duration", "24h"),
 					resource.TestCheckResourceAttr(name, "cors_headers.#", "0"),
+					resource.TestCheckResourceAttr(name, "saas_app.#", "0"),
 					resource.TestCheckResourceAttr(name, "auto_redirect_to_identity", "false"),
 				),
 			},
@@ -68,6 +69,7 @@ func TestAccCloudflareAccessApplication_BasicAccount(t *testing.T) {
 					resource.TestCheckResourceAttr(name, "type", "self_hosted"),
 					resource.TestCheckResourceAttr(name, "session_duration", "24h"),
 					resource.TestCheckResourceAttr(name, "cors_headers.#", "0"),
+					resource.TestCheckResourceAttr(name, "sass_app.#", "0"),
 					resource.TestCheckResourceAttr(name, "auto_redirect_to_identity", "false"),
 				),
 			},
@@ -100,6 +102,36 @@ func TestAccCloudflareAccessApplication_WithCORS(t *testing.T) {
 					resource.TestCheckResourceAttr(name, "cors_headers.0.allowed_origins.#", "1"),
 					resource.TestCheckResourceAttr(name, "cors_headers.0.max_age", "10"),
 					resource.TestCheckResourceAttr(name, "auto_redirect_to_identity", "false"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccCloudflareAccessApplication_WithSaas(t *testing.T) {
+	rnd := generateRandomResourceName()
+	name := fmt.Sprintf("cloudflare_access_application.%s", rnd)
+	accountID := os.Getenv("CLOUDFLARE_ACCOUNT_ID")
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccessAccPreCheck(t)
+		},
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckCloudflareAccessApplicationDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCloudflareAccessApplicationConfigWithSaas(rnd, accountID),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(name, "account_id", accountID),
+					resource.TestCheckResourceAttr(name, "name", rnd),
+					resource.TestCheckResourceAttr(name, "type", "saas"),
+					resource.TestCheckResourceAttr(name, "session_duration", "24h"),
+					resource.TestCheckResourceAttr(name, "saas_app.#", "1"),
+					resource.TestCheckResourceAttr(name, "saas_app.0.sp_entity_id", "saas-app.example"),
+					resource.TestCheckResourceAttr(name, "saas_app.0.consumer_service_url", "https://saas-app.example/sso/saml/consume"),
+					resource.TestCheckResourceAttr(name, "saas_app.0.name_id_format", "email"),
 				),
 			},
 		},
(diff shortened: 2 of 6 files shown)


jacobbednarz:
@Justin-Holmes can i get your eyes on this too please?

internal/provider/resource_cloudflare_access_application_test.go
```diff
@@ -103,6 +105,34 @@ func TestAccCloudflareAccessApplication_WithCORS(t *testing.T) {
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccessAccPreCheck(t)
+		},
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckCloudflareAccessApplicationDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCloudflareAccessApplicationConfigWithSaas(rnd, accountID),
```
jacobbednarz, line 120: accountID is undefined here. you'll need `accountID := os.Getenv("CLOUDFLARE_ACCOUNT_ID")` to use it.

internal/provider/schema_cloudflare_access_application.go
```diff
@@ -121,6 +123,28 @@ func resourceCloudflareAccessApplicationSchema() map[string]*schema.Schema {
 				},
 			},
 		},
+		"saas_app": {
```
jacobbednarz, line 126: all these fields need to be documented with `Description` fields to generate the documentation

internal/provider/resource_cloudflare_access_application_test.go
ouranos: I copied the syntax from other tests in this file (eg: `TestAccCloudflareAccessApplication_BasicAccount` & `TestAccCloudflareAccessApplication_WithADefinedIdps`) and they don't seem to define it either. Is that an issue too?

internal/provider/schema_cloudflare_access_application.go
ouranos: Done

internal/provider/resource_cloudflare_access_application.go
```diff
@@ -64,6 +64,11 @@ func resourceCloudflareAccessApplicationCreate(ctx context.Context, d *schema.Re
 		newAccessApplication.CorsHeaders = CORSConfig
 	}
 
+	if _, ok := d.GetOk("saas_app"); ok {
+		saasApp := convertSaasSchemaToStruct(d)
+		newAccessApplication.SaasApplication = saasApp
```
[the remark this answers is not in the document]
ouranos: Done

Justin-Holmes:
Looks great! And you're correct @ouranos, the `domain` is not required for `saas` applications and is generated for you on creation. 

Also, the `custom_attributes` field is not used often so I'm cool omitting that for now.

ouranos:
Thanks, 

> the `domain` is not required for `saas` applications and is generated for you on creation.

So this is the correct way then?
```diff
"domain": {
    Type:     schema.TypeString,
-     Required: true,
+     Optional: true,
+     Computed: true, 
    ...
}
```

Let me know if you are happy with it or if it needs more change and I'll rebase/clean up/squash.

jacobbednarz:
> So this is the correct way then?

`Computed: true` is fine. `Optional` implies the end user is also able to set it.

internal/provider/resource_cloudflare_access_application_test.go
jacobbednarz: looks like `resource_cloudflare_access_group_test.go` is defining `accountID` globally and the others are relying on it. definitely don't rely on this and instead define it at the test level.

internal/provider/schema_cloudflare_access_application.go
```diff
@@ -37,15 +37,17 @@ func resourceCloudflareAccessApplicationSchema() map[string]*schema.Schema {
 			Description: "Friendly name of the Access Application.",
 		},
 		"domain": {
-			Type:        schema.TypeString,
-			Required:    true,
+			Type:     schema.TypeString,
+			Optional: true,
+			Computed: true, // REVIEW
+			// Required:    true,
```
jacobbednarz, line 43:
```suggestion
			Computed: true,
```

internal/provider/schema_cloudflare_access_application.go
```diff
@@ -121,6 +123,32 @@ func resourceCloudflareAccessApplicationSchema() map[string]*schema.Schema {
 				},
 			},
 		},
+		"saas_app": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			Description: "SaaS configuration for the Access Application. See below for reference structure.",
```
jacobbednarz, line 129:
```suggestion
			Type:        schema.TypeList,
			Optional:    true,
			MaxItems:    1,
			Description: "SaaS configuration for the Access Application.",
```

internal/provider/resource_cloudflare_access_application_test.go
```diff
@@ -103,6 +105,34 @@ func TestAccCloudflareAccessApplication_WithCORS(t *testing.T) {
 	})
 }
 
+func TestAccCloudflareAccessApplication_WithSass(t *testing.T) {
+	rnd := generateRandomResourceName()
+	name := fmt.Sprintf("cloudflare_access_application.%s", rnd)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccessAccPreCheck(t)
```
jacobbednarz, line 114: you need both of these here (sets up the mock client)

```suggestion
            testAccPreCheck(t)
			testAccessAccPreCheck(t)
```

ouranos:
> > So this is the correct way then?
> 
> `Computed: true` is fine. `Optional` implies the end user is also able to set it.

It's computed for SaaS apps (and returns a Cloudflare generated domain) but it's required for the other types (eg: self-hosted)

internal/provider/schema_cloudflare_access_application.go
ouranos: We still need to be able to set it for `self_hosted`, wouldn't that break it?

internal/provider/resource_cloudflare_access_application_test.go
ouranos: Understood, done.

ouranos:
> don't forget to run `make docs` to get the documentation updated here. you'll need to fix the tests as well now that the field is computed.

- Squashed all previous commits together
- Addressed the new review comments
- Generated the docs

So I think the only remaining points are:
- `domain`: it's `required` for `self_hosted` but computed for `saas`
- acceptance tests: I only ran `make test` so far and I'm now trying to set up an environment to run `make testacc`.  We're only using Zero-Trust with Saas apps so we don't have any domains/zones set up. Looks like I can only add root domains on the free plan (we do have a paid plan for ZeroTrust). I'm going to try to set up one of our unused test domains.

internal/provider/schema_cloudflare_access_application.go
```diff
@@ -37,15 +37,17 @@ func resourceCloudflareAccessApplicationSchema() map[string]*schema.Schema {
...
+			// Required:    true,
 			Description: "The complete URL of the asset you wish to put Cloudflare Access in front of. Can include subdomains or paths. Or both.",
 		},
 		"type": {
 			Type:         schema.TypeString,
 			Optional:     true,
 			Default:      "self_hosted",
-			ValidateFunc: validation.StringInSlice([]string{"self_hosted", "ssh", "vnc", "file"}, false),
+			ValidateFunc: validation.StringInSlice([]string{"self_hosted", "saas", "ssh", "vnc", "file"}, false),
 			Description:  fmt.Sprintf("The application type. %s", renderAvailableDocumentationValuesStringSlice([]string{"self_hosted", "ssh", "vnc", "file"})),
```
jacobbednarz, line 51: need to add `saas` here too
```suggestion
			Description:  fmt.Sprintf("The application type. %s", renderAvailableDocumentationValuesStringSlice([]string{"self_hosted", "saas", "ssh", "vnc", "file"})),
```

internal/provider/schema_cloudflare_access_application.go
```diff
@@ -121,6 +123,33 @@ func resourceCloudflareAccessApplicationSchema() map[string]*schema.Schema {
 				},
 			},
 		},
+		"saas_app": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			MaxItems:    1,
+			Description: "SaaS configuration for the Access Application. See below for reference structure.",
```
jacobbednarz, line 130:
```suggestion
			Description: "SaaS configuration for the Access Application.",
```

internal/provider/schema_cloudflare_access_application.go
jacobbednarz: that would; if we need to set it, this is fine.

internal/provider/resource_cloudflare_access_application_test.go
```diff
@@ -103,6 +105,36 @@ func TestAccCloudflareAccessApplication_WithCORS(t *testing.T) {
 	})
 }
 
+func TestAccCloudflareAccessApplication_WithSass(t *testing.T) {
```
jacobbednarz, line 108:
```suggestion
func TestAccCloudflareAccessApplication_WithSaas(t *testing.T) {
```

internal/provider/resource_cloudflare_access_application_test.go
```diff
@@ -103,6 +105,36 @@ func TestAccCloudflareAccessApplication_WithCORS(t *testing.T) {
...
+				Config: testAccCloudflareAccessApplicationConfigWithSaas(rnd, accountID),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(name, "account_id", accountID),
+					resource.TestCheckResourceAttr(name, "name", rnd),
+					resource.TestCheckResourceAttr(name, "type", "saas"),
+					resource.TestCheckResourceAttr(name, "session_duration", "24h"),
+					resource.TestCheckResourceAttr(name, "saas_app.#", "1"),
+					resource.TestCheckResourceAttr(name, "saas_app.0.sp_entity_id", "1"),
+					resource.TestCheckResourceAttr(name, "saas_app.0.consumer_service_url", "1"),
+					resource.TestCheckResourceAttr(name, "saas_app.0.name_id_format", "1"),
```
jacobbednarz, line 131: need to fix these assertions

```suggestion
					resource.TestCheckResourceAttr(name, "saas_app.0.sp_entity_id", "saas-app.example"),
					resource.TestCheckResourceAttr(name, "saas_app.0.consumer_service_url", "https://saas-app.example/sso/saml/consume"),
					resource.TestCheckResourceAttr(name, "saas_app.0.name_id_format", "email"),
```

jacobbednarz: if you fix these, the acceptance tests all pass

ouranos:
> if you fix these, the acceptance tests all pass

Great thanks 🙏 

I've:
- applied all suggested changes
- cleaned up the comments regarding `domain`
- rebased from latest `master` & squashed

Let me know if I need to do anything else

ouranos:
> did you re-run `make docs`? that should be the only thing.

Nope 🤦‍♂️  
Done now. It fixes doc for #1792 too

jacobbednarz:
cool! I'll run the acceptance tests again in the morning and we should be good, thanks!

p.s for future PRs if you check the box to allow maintainers to make changes, we can limit the back and forth as I can provide a patch within the branch for you.

PR #1766, opened 2022-07-13
update waitingroom parameter defaults

broswen:
Update Waiting Room default values for some parameters.
This prevents constant state drift when a parameter isn't set.


.changelog/1766.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_waiting_room: Fix default waiting room session duration and path
```
jacobbednarz, line 2:
```suggestion
resource/cloudflare_waiting_room: fix default waiting room `session_duration` and `path` values
```

Issue #1776, opened 2022-07-15
Crash while generating error message while updating zone lockdown


schans:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform: v1.2.2
Cloudflare provider: v3.19.0

### Affected resource(s)

resource_cloudflare_zone_lockdown

### Terraform configuration files

```hcl
Nothing special..
```

### Debug output

Can't easily get this from our build server. I think the error is obvious enough that this is not needed. If so, please let me know.

### Panic output

Stack trace from the terraform-provider-cloudflare_v3.19.0 plugin:

panic: interface conversion: interface {} is nil, not string

goroutine 73 [running]:
github.com/cloudflare/terraform-provider-cloudflare/internal/provider.resourceCloudflareZoneLockdownUpdate({0xf95540, 0xc0003204e0}, 0xc000820c80, {0xe2b4e0?, 0xc0003ac600})
	github.com/cloudflare/terraform-provider-cloudflare/internal/provider/resource_cloudflare_zone_lockdown.go:150 +0x765
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).update(0xc000522380, {0xf95578, 0xc000305020}, 0xd?, {0xe2b4e0, 0xc0003ac600})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.18.0/helper/schema/resource.go:741 +0x12e
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).Apply(0xc000522380, {0xf95578, 0xc000305020}, 0xc00045a8f0, 0xc0001ce880, {0xe2b4e0, 0xc0003ac600})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.18.0/helper/schema/resource.go:847 +0x82c
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).ApplyResourceChange(0xc000395230, {0xf95578?, 0xc000304f00?}, 0xc0009ee4b0)
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.18.0/helper/schema/grpc_provider.go:1021 +0xe3c
[...]

PR #1777, opened 2022-07-15
Log zoneID as d.Get(zone) might return nil

schans:
This should fix https://github.com/cloudflare/terraform-provider-cloudflare/issues/1776 where the provider crashes when an error is encountered.

### DIFF
--- a/.changelog/1777.txt
+++ b/.changelog/1777.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_zone_lockdown: Fix crash when logging upstream error message
+```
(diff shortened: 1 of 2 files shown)


jacobbednarz:
how does this address the issue? the variable you are using here is identical to the value at the start of the method.

jacobbednarz:
also, given `zone_id` is a required attribute, I'm unsure how it is possible for the value to be nil.

schans:
> also, given `zone_id` is a required attribute, I'm unsure how it is possible for the value to be nil.

Hi! Thanks for looking into this. I had the exact same question initially ;-)  Note that the original line referenced `zone` and not `zone_id`. I think that was the culprit.

jacobbednarz:
ah! good catch. that makes more sense :)

PR #1778, opened 2022-07-15
update misc SSL attributes

nickysemenza:
```release-note:enhancement
* remove references to long-deprecated dedicated certs (replaced by `advanced`)
* update the list of allowed certificate authorities
* fix some of the custom hostname docs copy
```


jacobbednarz:
i've swapped these two resources over to [autogenerated documentation](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md#migration) but in doing so, uncovered that we don't have all the fields documented. up to you if you'd like to go through and add these before we merge. i'm fine either way.

something else that is problematic here is that we're using a `TypeMap` for the `ownership_verification` and `ownership_verification_http` nested objects. using a `TypeMap` pretty much means we don't have control over the keys and as such, can't document them automatically (among other internal issues). we may want to introduce a [breaking change for 4.x]https://github.com/cloudflare/terraform-provider-cloudflare/issues/1646) to add this to the schema as a `TypeList` instead.

jacobbednarz (approved):

nickysemenza:
Oh cool I was curious about the 2 different docs. I can file a followup to improve the documentation more, I just fixed this since it was noticeably wrong :)

PR #1785, opened 2022-07-18
Flip cache flag semantics in Cache Rules

WowVeryLogin:
This product is still under unstable entitlement and no actual customers are using it, so it is safe to rename parameters without migration


jacobbednarz:
cherry picked into #1793 due to the chicken/egg issue we have here

PR #1805, opened 2022-07-29
Fetch/modify `origin_max_http_version` as a single setting

reedloden:
Unfortunately, https://api.cloudflare.com/#zone-settings-get-all-zone-settings does not include `origin_max_http_version`, so need to add it to `fetchAsSingleSetting` so that it is handled as a one-off.

### DIFF
--- a/.changelog/1805.txt
+++ b/.changelog/1805.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_zone_settings_override: Fetch/modify `origin_max_http_version` as a single setting.
+```
--- a/internal/provider/resource_cloudflare_zone_settings_override.go
+++ b/internal/provider/resource_cloudflare_zone_settings_override.go
@@ -31,6 +31,7 @@ var fetchAsSingleSetting = []string{
 	"h2_prioritization",
 	"image_resizing",
 	"early_hints",
+	"origin_max_http_version",
 }
 
 func resourceCloudflareZoneSettingsOverrideCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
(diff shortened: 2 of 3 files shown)


jacobbednarz:
i'll have to dig into this one internally as it looks like the test account we have doesn't have this visible and the acceptance tests fail. PR looks fine though.

reedloden:
@jacobbednarz Any luck following up on this internally?

jacobbednarz:
I did and it turns out the URL doesn't follow the zone settings conventions so we need to either modify the provider or fix the URL. I'm still working out next steps with the team.

reedloden:
> I did and it turns out the URL doesn't follow the zone settings conventions so we need to either modify the provider or fix the URL. I'm still working out next steps with the team.

Checking in again on this one. Would love to be able to set this setting via terraform.

jacobbednarz:
nothing new, still pending the service team to decide what they would like to do here.

jacobbednarz:
we've deployed workaround for this endpoint and now acceptance tests are all passing.

```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareZoneSettingsOverride_" -count 1 -parallel 1 -timeout 120m -parallel 1
--- PASS: TestAccCloudflareZoneSettingsOverride_Full (41.35s)
=== RUN   TestAccCloudflareZoneSettingsOverride_RemoveAttributes
--- PASS: TestAccCloudflareZoneSettingsOverride_RemoveAttributes (41.15s)
PASS
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/provider   83.323s
```

thanks for the patience here @reedloden! 🏆

Issue #1802, opened 2022-07-27
`cloudflare_fallback_domain` shows perpetual changes to apply


tjstansell:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```
Terraform version v1.2.5
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.19.0
+ provider registry.terraform.io/hashicorp/aws v4.23.0
```

### Affected resource(s)

* `cloudflare_fallback_domain`

### Terraform configuration files

```hcl
locals {
  default_domains  = ["intranet", "internal", "private", "localdomain", "domain", "lan", "home", "host", "corp", "local", "localhost", "home.arpa", "invalid", "test"]
  internal_domains = ["10.in-addr.arpa", "mydomain.net", "mydomain.com"]
  dns_servers      = ["10.1.1.1", "10.1.1.2"]
}

resource "cloudflare_fallback_domain" "this" {
  account_id = local.account_id

  dynamic "domains" {
    for_each = local.default_domains
    content {
      suffix = domains.value
    }
  }
  dynamic "domains" {
    for_each = local.internal_domains
    content {
      suffix     = domains.value
      dns_server = local.dns_servers
    }
  }
}
```

### Debug output

I'm not including the full debug output here because most of it is irrelevant.
```
  # cloudflare_fallback_domain.this will be updated in-place
  ~ resource "cloudflare_fallback_domain" "this" {
        id         = "<redacted>"
        # (1 unchanged attribute hidden)

      ~ domains {
          ~ dns_server = [
              - "10.1.1.1",
              - "10.1.1.2",
            ]
          ~ suffix     = "10.in-addr.arpa" -> "intranet"
        }
      ~ domains {
          ~ dns_server = [
              - "10.1.1.1",
              - "10.1.1.2",
            ]
          ~ suffix     = "mydomain.com" -> "internal"
        }
      ~ domains {
          ~ suffix     = "home.arpa" -> "private"
            # (1 unchanged attribute hidden)
        }
      ~ domains {
          ~ dns_server = [
              - "10.1.1.1",
              - "10.1.1.2",
            ]
          ~ suffix     = "mydomain.net" -> "localdomain"
        }
      ~ domains {
          ~ suffix     = "intranet" -> "domain"
            # (1 unchanged attribute hidden)
        }
[...]

PR #1828, opened 2022-08-09
Use set instead of to avoid infinite changes

vavsab:
Closes https://github.com/cloudflare/terraform-provider-cloudflare/issues/1802

### DIFF
--- a/.changelog/1828.txt
+++ b/.changelog/1828.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_fallback_domain: fix perpetual changes due to ordering
+```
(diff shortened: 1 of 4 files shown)


vavsab:
> thanks for the PR! you'll need to update the acceptance tests as well since right now they are based on `TypeList` - https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/provider/resource_cloudflare_fallback_domain_test.go

The test is green now.
image

jacobbednarz:
this is odd that the test is passing with the following assertion

    resource.TestCheckResourceAttr(name, "domains.0.description", "example domain"),

for TypeSet, the index is a computed hash of the inner elements (such as `12345678`). i'll see if i can find an explanation as to why this is still working after it should have broke.

jacobbednarz:
i had a look into this and it seems like it somehow works with nested schemas. i can't find specifically where the difference in nested vs simple schemas is (that shows the behaviour difference) but it works 🤷

PR #1839, opened 2022-08-16
fix: doc: loadbalancing region URL

xens:
Fixing the documentation URL for the `check_regions` inside the `cloudflare_load_balancer_pool` resource.


xens:
workflow/skip-changelog-entry

PR #1849, opened 2022-08-22
fix Access test assertions

jacobbednarz:
Updates the "not found" matchers in `Read` operations to correctly identify missing resources.

### DIFF
--- a/internal/provider/resource_cloudflare_teams_location.go
+++ b/internal/provider/resource_cloudflare_teams_location.go
@@ -30,7 +30,7 @@ func resourceCloudflareTeamsLocationRead(ctx context.Context, d *schema.Resource
 
 	location, err := client.TeamsLocation(ctx, accountID, d.Id())
 	if err != nil {
-		if strings.Contains(err.Error(), "HTTP status 400") {
+		if strings.Contains(err.Error(), "Location ID is invalid") {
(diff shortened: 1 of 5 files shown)


internal/provider/resource_cloudflare_teams_location.go
```diff
@@ -30,7 +30,7 @@ func resourceCloudflareTeamsLocationRead(ctx context.Context, d *schema.Resource
 
 	location, err := client.TeamsLocation(ctx, accountID, d.Id())
 	if err != nil {
-		if strings.Contains(err.Error(), "HTTP status 400") {
+		if strings.Contains(err.Error(), "Location ID is invalid") {
```
jacobbednarz, line 33: this should really be a HTTP 404

internal/provider/resource_cloudflare_teams_proxy_endpoints.go
```diff
@@ -30,7 +30,7 @@ func resourceCloudflareTeamsProxyEndpointRead(ctx context.Context, d *schema.Res
 
 	endpoint, err := client.TeamsProxyEndpoint(ctx, accountID, d.Id())
 	if err != nil {
-		if strings.Contains(err.Error(), "HTTP status 400") {
+		if strings.Contains(err.Error(), "Proxy Endpoint ID is invalid") {
```
jacobbednarz, line 33: this should really be a HTTP 404

PR #1892, opened 2022-09-07
add support for devcontainer

jacobbednarz:


Cyb3r-Jak3:
Don't know if it is worth adding, but I use the Terraform extension in my devcontainer

jacobbednarz:
sure thing @Cyb3r-Jak3; added it since it makes sense here.

PR #1894, opened 2022-09-07
Add cloudflare_user_agent_blocking_rule resource

tamas-jozsa:
Marking as draft, as I need to update cf-go to use common method signatures: https://github.com/cloudflare/cloudflare-go/blob/master/user_agent.go 

Adding new User Agent Blocking rules resource: https://api.cloudflare.com/#user-agent-blocking-rules-properties


jacobbednarz:
if the methods are already there (and not requiring zone and account level access), it's fine to leave the methods as is for now. we can update them in another pass. 

(This is older functionality fwiw so it may end up deprecated for ERE)

tamas-jozsa:
I followed up with ERE and confirmed, that this will be deprecated. I agree, that we should go with these methods for now and do the update after official deprecation.

Marking PR for ready for review.

jacobbednarz:

```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareUserAgentBlockingRule" -count 1 -parallel 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareUserAgentBlockingRule
--- PASS: TestAccCloudflareUserAgentBlockingRule (8.40s)
PASS
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/provider   9.187s
```

thanks!

PR #1902, opened 2022-09-09
Add support for account resource

mlhnono68:


internal/provider/resource_cloudflare_account.go
```diff
@@ -0,0 +1,150 @@
...
+		CreateContext: resourceCloudflareAccountCreate,
+		ReadContext:   resourceCloudflareAccountRead,
+		UpdateContext: resourceCloudflareAccountUpdate,
+		DeleteContext: resourceCloudflareAccountDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Description: heredoc.Doc(`
+			Provides a Cloudflare Account resource. Account is the basic resource for
+			working with Cloudflare zones, teams and users. Requires the tenant entitlement
```
jacobbednarz, line 33:
```suggestion
			working with Cloudflare zones, teams and users. 
```

internal/provider/resource_cloudflare_account.go
```diff
@@ -0,0 +1,150 @@
...
+		`),
+	}
+}
+
+func resourceCloudflareAccountCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*cloudflare.API)
+	accountName := d.Get("name").(string)
+	var accountType string
+	if d.Get("type").(string) == "" {
```
jacobbednarz, line 42: we can do this logic with a combination of `Default: "standard"` and `validation.StringInSlice` (example: https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/provider/schema_cloudflare_list.go#L23) for confirming only allowed values.

internal/provider/resource_cloudflare_account.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("error creating account %q: %w", accountName, err))
+	}
+
+	d.SetId(acc.ID)
+
+	return resourceCloudflareAccountRead(ctx, d, meta)
+}
+
+func getCloudflareAccontFromId(accountID string, client *cloudflare.API, ctx context.Context) (cloudflare.Account, error) {
```
jacobbednarz, line 67: can you elaborate on why we need this method? we already have a library method to fetch by ID - https://github.com/cloudflare/cloudflare-go/blob/aaf504612ca6040b262b8867d21bd0f9c7ddc818/accounts.go#L75-L90

internal/provider/resource_cloudflare_account.go
```diff
@@ -0,0 +1,150 @@
...
+		return cloudflare.Account{}, fmt.Errorf("Account %s does not exist", accountID)
+	}
+}
+
+func resourceCloudflareAccountRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*cloudflare.API)
+	accountID := d.Id()
+
+	foundAcc, err := getCloudflareAccontFromId(accountID, client, ctx)
+	tflog.Debug(ctx, fmt.Sprintf("AccountDetails error: %#v", err))
```
jacobbednarz, line 87: this shouldn't be a debug; if we hit an error, we need to handle it and propagate that to the end user.

internal/provider/resource_cloudflare_account.go
```diff
@@ -0,0 +1,150 @@
...
+
+	return nil
+}
+
+func resourceCloudflareAccountUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*cloudflare.API)
+	accountID := d.Id()
+	foundAcc, err := getCloudflareAccontFromId(accountID, client, ctx)
+
+	tflog.Debug(ctx, fmt.Sprintf("AccountDetails error: %#v", err))
```
jacobbednarz, line 113: similar to https://github.com/cloudflare/terraform-provider-cloudflare/pull/1902/files#r967906187. this should be full error handling.

internal/provider/resource_cloudflare_account.go
```diff
@@ -0,0 +1,150 @@
...
+
+	log.Printf("[INFO] Updating Cloudflare Account: id %s", accountID)
+
+	if accountName, ok := d.GetOkExists("name"); ok && d.HasChange("name") {
```
jacobbednarz, line 117: you shouldn't use `GetOkExists` unless you really have to. a `GetOk` should be suffice here.

internal/provider/resource_cloudflare_account.go
```diff
@@ -0,0 +1,150 @@
...
+	if enforce_twofactor, ok := d.GetOkExists("enforce_twofactor"); ok && d.HasChange("enforce_twofactor") {
+		foundAcc.Settings.EnforceTwoFactor = enforce_twofactor.(bool)
+	}
+
+	_, err = client.UpdateAccount(ctx, accountID, foundAcc)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("error updating Account %q: %w", d.Id(), err))
+	} else {
+		return resourceCloudflareAccountRead(ctx, d, meta)
+	}
```
jacobbednarz, line 134: this doesn't need to be an `if/else`, a simpler `if` will do as the error handler will return if it is encountered

```suggestion
	if err != nil {
		return diag.FromErr(fmt.Errorf("error updating Account %q: %w", d.Id(), err))
	} 
	
	return resourceCloudflareAccountRead(ctx, d, meta)
```

internal/provider/resource_cloudflare_account.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func resourceCloudflareAccountUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*cloudflare.API)
+	accountID := d.Id()
+	foundAcc, err := getCloudflareAccontFromId(accountID, client, ctx)
+
+	tflog.Debug(ctx, fmt.Sprintf("AccountDetails error: %#v", err))
+
+	log.Printf("[INFO] Updating Cloudflare Account: id %s", accountID)
```
jacobbednarz, line 115: you shouldn't use `log.Printf`. use `tflog` instead.

internal/provider/resource_cloudflare_account.go
```diff
@@ -0,0 +1,150 @@
...
+	} else {
+		return resourceCloudflareAccountRead(ctx, d, meta)
+	}
+}
+
+func resourceCloudflareAccountDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*cloudflare.API)
+	accountID := d.Id()
+
+	log.Printf("[INFO] Deleting Cloudflare Account: id %s", accountID)
```
jacobbednarz, line 141: you shouldn't use `log.Printf`. use `tflog` instead.

internal/provider/schema_cloudflare_account.go
```diff
@@ -0,0 +1,28 @@
+package provider
```
jacobbednarz, line 1: you'll want to add the documentation examples and then run `make docs` to have the schema generate the documentation for you.

internal/provider/resource_cloudflare_account.go
mlhnono68: Agreed. It was already checked at schema level this way so I removed this useless check here.

internal/provider/resource_cloudflare_account.go
mlhnono68: no need to elaborate. Simply not used enough to your repo and missed this one. Just removed my redundant function.

internal/provider/resource_cloudflare_account.go
mlhnono68: just cleaned this useless remaining debug and added the error management for it.

internal/provider/resource_cloudflare_account.go
mlhnono68: Changed

internal/provider/resource_cloudflare_account.go
mlhnono68: Changed

internal/provider/resource_cloudflare_account.go
mlhnono68: switched to tflog

internal/provider/resource_cloudflare_account.go
mlhnono68: ok. statement has been simplified

internal/provider/schema_cloudflare_account.go
mlhnono68: examples added

mlhnono68:
Hope you feel ok with it now @jacobbednarz . Thanks a lot for the detailed review !

internal/provider/resource_cloudflare_account.go
```diff
@@ -0,0 +1,130 @@
...
+	if enforce_twofactor, ok := d.GetOk("enforce_twofactor"); ok && d.HasChange("enforce_twofactor") {
+		foundAcc.Settings.EnforceTwoFactor = enforce_twofactor.(bool)
+	}
+
+	_, err = client.UpdateAccount(ctx, accountID, foundAcc)
+	if err != nil {
+		var cfError *cloudflare.RequestError
+		if errors.As(err, &cfError) {
+			errCodes := cfError.ErrorCodes()
+			if len(errCodes) == 1 && errCodes[0] == 1001 {
```
jacobbednarz, line 105: can you elaborate on this issue? with the correct entitlements, this should work.

mlhnono68: This is kind of a workaround to maybe something that should be fixed in cloudflare-go library. For example, running such basic code will actually make the update to work but will raise a 1001 error on `err3`:
```
	acc, _, err2 := api.Account(ctx, someAccountID)
	if err2 != nil {
		log.Fatal(err2)
	}

	fmt.Println(acc.ID)
	acc.Name = acc.Name + "x"
	_, err3 := api.UpdateAccount(ctx, acc.ID, acc)
	if err3 != nil {
		log.Fatal(err3)
	}
```

However, the update of the name will actually succeed. I think a change would be needed in https://github.com/cloudflare/cloudflare-go/blob/master/accounts.go#L95 so that accountType is not sent to backend while updating

mlhnono68: I will make this handling more clean by setting Type to empty as the JSON will omit it if empty on update

mlhnono68: https://github.com/cloudflare/terraform-provider-cloudflare/pull/1902#:~:text=prevent%201001%20error-,bdaf421,-Merge%20state will make sure the error does not pop up by omitting Type in the update.

mlhnono68:
Do you feel it is ok now @jacobbednarz ?

jacobbednarz:
appreciate the work to get this one sorted! thanks @mlhnono68 👏

PR #1906, opened 2022-09-12
Add DNS record data source

### LINKED ISSUE #1904: Add cloudflare_record data source.

Terraform v1.1.7
on darwin_arm64

Your version of Terraform is out of date! The latest version
is 1.2.9. You can update by downloading from https://www.terraform.io/downloads.html

### Description

Add a new data source (may called cloudflare_record) to be able to retrieve a record details in case that record not created in the same terraform configuration files. 
Being able to pass the record ID or FQDN in the data source block and get all possible details for that record like record type, name, and value.

### Use cases

I need to create a new AWS ACM certificate (let's call it cert A) and associate it with a the same domain that's already associated with another AWS ACM certificate (let's call it cert B) created in another AWS region. So when trying to validate the new certificate (cert A) against that domain, I got an error stating that those validation records already exist so I can't add new ones.

I need to use "cert A" with AWS cloudfront distribution while "cert B" is already verified and used with AWS APIGW.

Why I need to create 2 different AWS ACM certificates for the same domain? Because the application is deployed in a "X" region, but the AWS Cloudfront distribution needs to be in the global AWS region

```hcl
...
```
data "cloudflare_record" "default" {
      id = "abcd123"
}

resource "aws_acm_certificate_validation" "us_east_1" {
  provider                            = aws.global_region
  certificate_arn                 = aws_acm_certificate.global_reion.arn
  validation_record_fqdns = [data.cloudflare_record.default.hostname]
}

### References

_No response_

Cyb3r-Jak3:
Adds a data source for an existing DNS record

Resolves #1904


docs/data-sources/record.md
```diff
@@ -0,0 +1,33 @@
+---
+page_title: "cloudflare_record Data Source - Cloudflare"
+subcategory: ""
+description: |-
+  
+---
+
+# cloudflare_record (Data Source)
+
+
```
jacobbednarz, line 10: need to add an `example` here

internal/provider/data_source_record.go
```diff
@@ -0,0 +1,80 @@
...
+	"fmt"
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceCloudflareRecord() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceCloudflareRecordRead,
```
jacobbednarz, line 14: need to add `Description` to this so that it will automatically fill out the docs.

internal/provider/data_source_record.go
```diff
@@ -0,0 +1,80 @@
...
+		Schema: map[string]*schema.Schema{
+			"zone_id": {
+				Description: "The zone identifier to target for the resource.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"record_id": {
```
jacobbednarz, line 21: we can drop this in favour of `id` that we get for free with `d.SetId`

internal/provider/data_source_record.go
```diff
@@ -0,0 +1,80 @@
...
+			},
+			"zone_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceCloudflareRecordRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
jacobbednarz, line 62: i don't quite think this is usable. if someone already has the ID, wouldn't they use it? usually, data sources are used for when you have a fuzzy idea of what you're after but can't pin point it. i.e. a DNS record for "@" in a particular zone.

.changelog/1906.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:new-data-source
+cloudflare_record: Add data source for DNS record.
```
jacobbednarz, line 2:
```suggestion
cloudflare_record
```

Cyb3r-Jak3:
For the search, what about using the list DNS record API call (https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records) but applying a filter of hostname and a type of "A" unless set?

jacobbednarz:
> For the search, what about using the list DNS record API call ([api.cloudflare.com/#dns-records-for-a-zone-list-dns-records](https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records)) but applying a filter of hostname and a type of "A" unless set? Could even make the page response only be 1 record on 1 page.

yep, where possible, offload the actual searching the Cloudflare API and don't do it in the data source itself. check out [`cloudflare_zones`](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/provider/data_source_zones.go) for an example of this.

PR #1920, opened 2022-09-21
Permit setting account member status directly in Terraform schema

### LINKED ISSUE #1654: cloudflare_account_member: support status option

Terraform v1.0.6
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v3.9.1

### Description

`cloudflare_account_member` resource doesn't support the `status` attribute. [This setting supports](https://api.cloudflare.com/#account-members-add-member) automatically accepting that invite,
[cloudflare-go supports this field](https://github.com/cloudflare/cloudflare-go/blob/master/account_members.go#L95)

### Use cases

When adding a user through Terraform, they have to accept that invitation via email. That email often gets missed and we need to follow up with them. The invites also expire.

```hcl
resource "cloudflare_account_member" "jane_doe" {
  email_address = "jane.doe@example.com"
  send_email_invite = false
  role_ids = [
    "deadbeef",
  ]
  lifecycle {
    create_before_destroy = true
  }
}
```

Alternatively we could use `status = "accepted"` but since that field only receives two possible values a Boolean makes more sense to me 🤷 

### References

_No response_

markblackman:
Currently, the status always defaults to "" which translates to "pending" by default, but this change allows setting a status attribute in the Terraform schema.  

Closes #1654


internal/provider/resource_cloudflare_account_member.go
```diff
@@ -101,7 +101,8 @@ func resourceCloudflareAccountMemberCreate(ctx context.Context, d *schema.Resour
 		accountID = client.AccountID
 	}
 
-	r, err := client.CreateAccountMember(ctx, accountID, memberEmailAddress, accountMemberRoleIDs)
+	status := d.Get("status").(string)
```
jacobbednarz, line 104: you'll need to add this field to the schema - https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/provider/schema_cloudflare_account_member.go

i'd also recommend defaulting to `""` and having a conditional in here that if the status is not empty (the default), use the new method with the status.

markblackman: Ok, thanks, I'll add the schema field. doesn't `d.Get("status").(string)` default to "" for me?

jacobbednarz: > The function CreateAccountMember in cloudflare-go simply calls CreateAccountMemberwithStatus using "" as the status. [cloudflare/cloudflare-go@master/account_members.go#L114](https://github.com/cloudflare/cloudflare-go/blob/master/account_members.go?rgh-link-date=2022-09-22T08%3A39%3A58Z#L114) , so as long as d.Get("status").(string) results in "", then I don't think a conditional is required here. What do you think?

yep, that works for me! thanks for digging into it.

internal/provider/schema_cloudflare_account_member.go
```diff
@@ -22,5 +22,10 @@ func resourceCloudflareAccountMemberSchema() map[string]*schema.Schema {
 			Elem:        &schema.Schema{Type: schema.TypeString},
 			Description: "List of account role IDs that you want to assign to a member.",
 		},
+		"status": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "A member's status in the account. default value in API: pending; valid values: accepted, pending",
```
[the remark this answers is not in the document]
jacobbednarz: disregard, i ran this locally for you :)

jacobbednarz:
thanks for the PR! 🙌

PR #1925, opened 2022-09-22
Fix array manipulation bug with the zone settings resource.

turbomaze:
This PR addresses a potential bug with the `cloudflare_zone_settings_override` resource. I was reviewing the implementation before provisioning it, and I encountered a bit of array manipulation code that might contain a bug. I haven't observed a problem in practice, but out of caution, I don't want to apply this resource until I'm sure it won't cause any problems for my zone.

The potential bug relates to how zone settings that must be fetched as single zone settings are removed from the zone settings list. I think there's an off-by-one type of error happening as the zone settings array is mutated.

NOTE: the code in this PR has never been tested or even ran

### DIFF
--- a/internal/provider/resource_cloudflare_zone_settings_override.go
+++ b/internal/provider/resource_cloudflare_zone_settings_override.go
@@ -229,11 +229,9 @@ func updateSingleZoneSettings(ctx context.Context, zoneSettings []cloudflare.Zon
 		}
 	}
 
-	// precondition: indexesToCut is in ascending order
 	offset := 0
 	for _, indexToCut := range indexesToCut {
 		adjustedIndexToCut := indexToCut - offset
-		// loop postcondition: len(zoneSettings) decreases by 1 each iteration
 		zoneSettings = append(zoneSettings[:adjustedIndexToCut], zoneSettings[adjustedIndexToCut+1:]...)
 		offset += 1
 	}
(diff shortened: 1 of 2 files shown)


jacobbednarz:
@turbomaze are you able to add a test case that demonstrates the issue and the fix here? i'm not quite sure if this is actually changing the behaviour.

internal/provider/resource_cloudflare_zone_settings_override.go
```diff
@@ -229,8 +229,13 @@ func updateSingleZoneSettings(ctx context.Context, zoneSettings []cloudflare.Zon
 		}
 	}
 
+	// precondition: indexesToCut is in ascending order
```
jacobbednarz, line 232:
```suggestion
```

internal/provider/resource_cloudflare_zone_settings_override.go
```diff
@@ -229,8 +229,13 @@ func updateSingleZoneSettings(ctx context.Context, zoneSettings []cloudflare.Zon
...
+	offset := 0
 	for _, indexToCut := range indexesToCut {
-		zoneSettings = append(zoneSettings[:indexToCut], zoneSettings[indexToCut+1:]...)
+		adjustedIndexToCut := indexToCut - offset
+		// loop postcondition: len(zoneSettings) decreases by 1 each iteration
```
jacobbednarz, line 236:
```suggestion
```

jacobbednarz:
@turbomaze after you add the CHANGELOG entry (https://github.com/cloudflare/terraform-provider-cloudflare/pull/1925#issuecomment-1255585508) we can get this merged.

turbomaze:
> @turbomaze after you add the CHANGELOG entry ([#1925 (comment)](https://github.com/cloudflare/terraform-provider-cloudflare/pull/1925#issuecomment-1255585508)) we can get this merged.

Should be good now, thanks

PR #1929, opened 2022-09-26
Fix typo in load balancer docs

AdamFrey:
"A disabled rule will **be** be executed." -> "A disabled rule will **not** be executed."


jacobbednarz:
good spot! thank you 🍰

PR #1931, opened 2022-09-26
Fix backwards `Sprintf` parameters for error message

mthadley:
While attempting to import a `cloudflare_rate_limit` resource, I received the following error:

```sh
$ terraform import cloudflare_rate_limit.graphql_production [zone_id]/[rate_limit_id]
╷
│ Error: Error reading rate limit resource from API for resource [zone_id] in zone [rate_limit_id]: Authentication error (10000)
│
│
╵
```

Based on the error message, I initially interpreted it as meaning I had the Zone ID and the Rate Limit ID backwards. However, after reviewing the code, I the correct order is `[zone_id]/[rate_limit_id]`, and that the ID's are being interpolated into the message in the wrong order.

I still have authentication problems, but I don't think it's related to any bugs in this Terraform provider.

Anyways, I hope this change helps, and thanks for supporting Terraform!

### DIFF
--- a/internal/provider/resource_cloudflare_rate_limit.go
+++ b/internal/provider/resource_cloudflare_rate_limit.go
@@ -271,7 +271,7 @@ func resourceCloudflareRateLimitRead(ctx context.Context, d *schema.ResourceData
 			return nil
 		} else {
 			return diag.FromErr(errors.Wrap(err,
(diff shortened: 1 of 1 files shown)


jacobbednarz:
thanks!

Issue #377, opened 2019-06-12
Resource cloudflare_zone_settings_override not importable


pjmpsu:
### Terraform Version
Terraform v0.12.1
+ provider.cloudflare v1.15.0

### Affected Resource(s)
cloudflare_zone_settings_override

### Debug Output
```
terraform import -allow-missing-config cloudflare_zone_settings_override.example abc123
cloudflare_zone_settings_override.example: Importing from ID "abc123"...

Error: resource cloudflare_zone_settings_override doesn't support import
```

### Expected Behavior
It should import successfully.

jacobbednarz:
I'm on the fence about this as the `cloudflare_zone_settings_override` isn't intended (as is) to manage all the zone settings available in a subscription; instead, it is meant to be for overriding the existing default configurations with your changes. This does, however, clash with the intent of managing all infrastructure as code via Terraform.

Something for the longer term, but perhaps the resource name and purpose needs to be changed to reflect being able to manage all settings and not just overridding defaults. I would entertain that discussion for 3.x milestone.

kawikao:
This is a fatal error because when trying to delete the resource, it tries to update settings at Cloudflare marked as read-only which causes the delete to fail. So, I can't update a read-only setting with the existing value or null and I can't delete the resource. It's essentially a zombie resource that will never update or go away.

billinghamj:
@jacobbednarz so what is the intended way to manage all the zone settings?

jacobbednarz:
Closing this off as we won't be making `cloudflare_zone_settings_override` importable and any changes to manage all settings will likely be in a new resource.

cofyc:
> Closing this off as we won't be making `cloudflare_zone_settings_override` importable and any changes to manage all settings will likely be in a new resource.

Out of curiosity, when will this new resource be available?

gerbyzation:
@jacobbednarz 

> Something for the longer term, but perhaps the resource name and purpose needs to be changed to reflect being able to manage all settings and not just overriding defaults. I would entertain that discussion for 3.x milestone.

The provider is on version 4 now, but this doesn't seem to have changed. What are the prospects of having better support for these settings?

nwalters512:
The first thing I did when I was setting up Cloudflare+Terraform just now was try to figure out how to disable email obfuscation. It seems like the only option for that is via the `cloudflare_zone_settings_override` resource, which to this day still isn't importable. Are there any improvements for this on the horizon?

PR #1940, opened 2022-09-30
Adding an ability to import cloudflare_zone_settings_override resource

sanicheev:
This PR adds support for cloudflare_zone_settings_override resource import.
Currently importer is not assigned in a constructor.

Please let me know if there is anything i am missing.

### DIFF
--- a/.changelog/1940.txt
+++ b/.changelog/1940.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/cloudflare_zone_settings_override: Add importer
+```
--- a/internal/provider/resource_cloudflare_zone_settings_override.go
+++ b/internal/provider/resource_cloudflare_zone_settings_override.go
@@ -23,6 +23,9 @@ func resourceCloudflareZoneSettingsOverride() *schema.Resource {
 		ReadContext:   resourceCloudflareZoneSettingsOverrideRead,
 		UpdateContext: resourceCloudflareZoneSettingsOverrideUpdate,
 		DeleteContext: resourceCloudflareZoneSettingsOverrideDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceCloudflareZoneSettingsOverrideImport,
+		},
 	}
 }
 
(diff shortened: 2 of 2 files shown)


jacobbednarz:
thanks for the PR however, the zone settings override resource is intentionally not importable since it is only intended as a way to override the defaults. there are some previous discussions in #998 and #377.

sanicheev:
@jacobbednarz the biggest problem with this - list of changes are not visible. What i mean by that, imagine you have an existing zone with custom settings(you changed lots of defaults) and suddenly decided to manage this as terraform code. You define override resource and try to check terraform plan. List of changes will not be available to you and you might miss some important setting which will be flipped back to its original value and whole zone will be affected. I suggest reconsider looking at this issue, i am happy to adjust my PR to use a different naming probably if override part is causing confusion or do any other adjustments.

jacobbednarz:
i understand the issues with this however, the resource is in it's current state is not designed to be used like this. i mentioned this in the linked issues and until we replace this resource with the next iteration of it, we won't be supporting the import due to the issues it will create.

you can use [cf-terraforming](https://github.com/cloudflare/cf-terraforming) to generate all the available settings and remove those which you aren't overriding in the zone. if you keep them all (outside of the ones you've customised), you will hit issues sooner or later.

sanicheev:
@jacobbednarz i can try and create separate resource: cloudflare_zone_settings in the context of this PR as the next iteration. Do you think it will work? Or you don't have any strong opinion on how it should look like yet?

jacobbednarz:
no, i wouldn't recommend you spending effort on designing it. we are still working out (internally) how zone settings will be handled going forward and how we want to manage them externally. this is on the list for 4.x improvements (#1646) and will be addressed when we come around to that.

sanicheev:
@jacobbednarz thanks for the information :+1:

florianmutter:
Seems like it did not make it into 4.x version. Are there new plans when to revisit the settings resource?

Regarding cf-terraforming: How do I know if a setting is the default or not? Is there any way to find out?

markbaird:
It's been a year since the [last update](https://github.com/cloudflare/terraform-provider-cloudflare/pull/1940#issuecomment-1263093028) on this issue, and I don't see any open tickets addressing this issue. The massive amount of settings in the `cloudflare_zone_settings_override` resource, and the lack of ability to import any of those settings, makes this Terraform provider unable to manage a big chunk of settings for any pre-existing Cloudflare account. Can we please get an update on this?

markbaird:
@jacobbednarz Two years ago [you said](https://github.com/cloudflare/terraform-provider-cloudflare/pull/1940#issuecomment-1263093028) you would be handling this in version 4.x. Version 4 has been out for a while, but this issue was not fixed in version 4. No communication with users in 2 years about a fundamental issue with this Terraform provider, and closing out any issues opened here regarding this issue is unacceptable.

Issue #1937, opened 2022-09-29
Cannot specify type = drop for cloudflare_email_routing_catch_all


yan12125:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```
$ ./terraform -v
Terraform v1.3.1
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.24.0
```

### Affected resource(s)

cloudflare_email_routing_catch_all

### Terraform configuration files

```hcl
terraform {
    required_providers {
        cloudflare = {
            source = "cloudflare/cloudflare"
        }
    }
}

provider "cloudflare" {
    api_token = "0123456789012345678901234567890123456789"
}

resource "cloudflare_email_routing_catch_all" "foobar" {
    zone_id = "YYY"
    name = "Catch-all rule"
    enabled = true
    matcher {
        type = "all"
    }
    action {
        type = "drop"
        value = []
    }
}
```

### Link to debug output

https://gist.github.com/yan12125/f683cbbd61c09b8e2544c30cd6acad37

### Expected output

Catch-all rule for email routing is enabled and configured to drop all mails.

### Actual output

The provider does not like type = drop for `cloudflare_email_routing_catch_all`.

### Steps to reproduce

1. Run `terraform plan`

### Additional factoids

Per https://api.cloudflare.com/#email-routing-routing-rules-update-catch-all-rule, the catch-all rule accepts type = drop, forward or worker.

jacobbednarz:
moving to enhancement as this is a new option. 

feel free to send over a PR if you'd it addressed immediately.

yan12125:
Thanks! It's not urgent. I may have a look recently.

PR #1947, opened 2022-10-03
resource/cloudflare_email_routing_catch_all: switch to a dedicated scheme

yan12125:
Compared to a normal email routing rule, the catch-all rule:

* Does not have a priority
* For matchers, it accepts only type = "all", and there are no "field" and "value" properties
* For actions, it accepts type = "drop" besides "forward" and "worker"

Ref: https://api.cloudflare.com/#email-routing-routing-rules-update-routing-rule
Ref: https://api.cloudflare.com/#email-routing-routing-rules-update-catch-all-rule

Fixes https://github.com/cloudflare/terraform-provider-cloudflare/issues/1937

### DIFF
--- a/internal/provider/schema_cloudflare_email_routing_catch_all.go
+++ b/internal/provider/schema_cloudflare_email_routing_catch_all.go
@@ -1,0 +1,70 @@
+package provider
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+func resourceCloudflareEmailRoutingCatchAllSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"zone_id": {
+			Description: "The zone identifier to target for the resource.",
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"tag": {
+			Description: "Routing rule identifier.",
+			Type:        schema.TypeString,
+			Computed:    true,
+		},
+		"name": {
+			Description: "Routing rule name.",
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"enabled": {
+			Description: "Routing rule status.",
+			Type:        schema.TypeBool,
+			Optional:    true,
+		},
+		"matcher": {
+			Description: "Matching patterns to forward to your actions.",
+			Type:        schema.TypeSet,
+			Required:    true,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"type": {
+						Description:  "Type of matcher.",
+						Type:         schema.TypeString,
+						Required:     true,
+						ValidateFunc: validation.StringInSlice([]string{"all"}, true),
+					},
+				},
+			},
+		},
+		"action": {
+			Description: "List actions patterns.",
+			Type:        schema.TypeSet,
+			Required:    true,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"type": {
+						Description:  "Type of supported action.",
+						Type:         schema.TypeString,
+						Required:     true,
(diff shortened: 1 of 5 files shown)


yan12125:
As a side note, an API response for the catch-all rule contains `priority`. For example,
```json
{
 "result": {
  "tag": "xxx",
  "name": "Catch-all rule",
  "matchers": [
   {
    "type": "all"
   }
  ],
  "actions": [
   {
    "type": "drop"
   }
  ],
  "enabled": true,
  "priority": 2147483647
 },
 "success": true,
 "errors": [],
 "messages": []
}
```
I'm not sure if `priority` should be added as a computed field in the new scheme or not.

jacobbednarz:
> I'm not sure if `priority` should be added as a computed field in the new scheme or not.

if it's not in the public API, we'll need to wait until it's declared stable (and documented) before we implement it here.

are you able to run `make docs` for this locally? the automatic generation of documentation will need to be kicked since we've updated the fields.

internal/provider/schema_cloudflare_email_routing_catch_all.go
```diff
@@ -0,0 +1,70 @@
...
+			Optional:    true,
+		},
+		"matcher": {
+			Description: "Matching patterns to forward to your actions.",
+			Type:        schema.TypeSet,
+			Required:    true,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"type": {
+						Description:  "Type of matcher.",
```
jacobbednarz, line 37:
```suggestion
						Description:  fmt.Sprintf("Type of matcher. %s", renderAvailableDocumentationValuesStringSlice([]string{"all"})),
```

(don't forget to update your imports for `fmt`)

internal/provider/schema_cloudflare_email_routing_catch_all.go
```diff
@@ -0,0 +1,70 @@
...
+			},
+		},
+		"action": {
+			Description: "List actions patterns.",
+			Type:        schema.TypeSet,
+			Required:    true,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"type": {
+						Description:  "Type of supported action.",
```
jacobbednarz, line 52:
```suggestion
						Description:  fmt.Sprintf("Type of supported action. %s", renderAvailableDocumentationValuesStringSlice([]string{"drop", "forward", "worker"})),
```

internal/provider/schema_cloudflare_email_routing_catch_all.go
```diff
@@ -0,0 +1,70 @@
...
+						Type:         schema.TypeString,
+						Required:     true,
+						ValidateFunc: validation.StringInSlice([]string{"drop", "forward", "worker"}, true),
+					},
+					"value": {
+						Description: "An array with items in the following form.",
```
jacobbednarz, line 58:
```suggestion
						Description: "A list with items in the following form.",
```

yan12125:
Thanks! Descriptions are improved.

> if it's not in the public API, we'll need to wait until it's declared stable (and documented) before we implement it here.

Got it! Then I will omit the priority in codes for now.

> are you able to run `make docs` for this locally? the automatic generation of documentation will need to be kicked since we've updated the fields.

`docs/resources/email_routing_catch_all.md` is updated and committed. However, during `make docs`, `docs/data-sources/devices.md` is also modified, and changes in https://github.com/cloudflare/terraform-provider-cloudflare/commit/924f1e8101c6134fdb467ae8df8eac43369ac8a8#diff-410ef65dd8c8160d34c379f6a0d303020f23b6631bdd91955b2d27962e7025e9 are reverted. I assume that's unrelaed to this PR, so I didn't commit it.

jacobbednarz:
thanks @yan12125 💯 you rock!

yan12125:
Cool, thanks!

PR #1950, opened 2022-10-05
Add chromeos and unique_client_id to device posture rules

kjhandy:
Adds `chromeos` an a valid value for `platform` and adds `unique_client_id` as a valid value for `type` in device posture rules.


jacobbednarz:
we've got the changelog correct but the filename doesn't match the PR number so the check here is failing. you'll want to rename `.changelog/22346.txt` => `.changelog/1950.txt`

you'll also want to run `make docs` + commit locally to update the automatically generated documentation. you can see more about the process at https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md

kjhandy:
Thanks @jacobbednarz. For some reason when I read the guidelines, I thought the filename was supposed to match up with the Jira ticket number. I'm not able to run `make docs` on my forked version (but for whatever reason, it works perfectly fine on my cloned repo). I get `Error executing command: unable to generate website: configuration is invalid` every time. Have you seen this before?

jacobbednarz:
> I get `Error executing command: unable to generate website: configuration is invalid` every time.

not since we've swapped, i'd have to have a look and see if i can replicate.

jacobbednarz:
thanks for this one! 🍰

PR #1951, opened 2022-10-05
Set the ID for the byoip prefix during import.

### LINKED ISSUE #1890: Unable to import BYOIP prefix
### Confirmation

```
Terraform v1.0.11
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v3.22.0
```

### Affected resource(s)

* `cloudflare_byo_ip_prefix`

```hcl
resource "cloudflare_byo_ip_prefix" "foo" {
  account_id    = var.cloudflare_account_id
  prefix_id     = "947fe4f12003420fbfac95429b58865"
  description   = ""
  advertisement = "off"
}
```

### Link to debug output

https://gist.github.com/oefd/e82fb16a665f38825619739f443d6bc5

### Panic output

_No response_

### Expected output

The import of the resource via a web request to the API endpoint including the account's ID.

### Actual output

The import fails.

### Steps to reproduce

* create a prefix in the UI
* attempt to import it to terraform

### Additional factoids

The provider attempts to get information on the resource from the wrong endpoint. This is part of the error text terraform outputs:

```
Could not route to /accounts/addressing/prefixes/xyz, perhaps your object identifier is invalid? (7003), No route for that URI (7000)
```

Note that error text is directly from the API, which you can confirm with:

```
$ curl -H "Authorization: Bearer xyz" https://api.cloudflare.com/client/v4/accounts/addressing/prefixes/947fe4f12003420fbfac95429b58865
```

This is happening because that's the incorrect endpoint, the account id should appear after `accounts/` as per [the api docs](https://api.cloudflare.com/#ip-address-management-prefixes-prefix-details). I can confirm the account id and token combo I used to attempt the import functions correctly when the account ID is inserted in at the appropriate place.

### References

_No response_

wolfmd:
Support for splitting the account_id and prefix_id out of the user-supplied ID had been added in f9b3d1e, but when the object is being read, the prefix id supplied to the API includes both the account id and prefix id.

This sets the ID of the schema object to the prefix id in a similar manner as resourceCloudflareBYOIPPrefixCreate

Closes #1890


jacobbednarz:
solid work @wolfmd 👏

wolfmd:
Yeah, real monumental change here 😅
 
Thanks for the review and merge!

PR #1954, opened 2022-10-07
[Backwards compatible] Migrate cloudflare_load_balancer to autogen docs

tc80:
### Migrate cloudflare_load_balancer docs to use the autogen process rather than templating, as documented [here](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md#migration).

While migrating, I learned that the object `session_affinity_attributes` was being represented with a `TypeMap`, which [does not have support for defining specific keys](https://www.terraform.io/plugin/sdkv2/schemas/schema-types#typemap). As a result, descriptions of any of its keys were discarded when running `make docs`. ~~In other to document this object properly, I refactored it to a `TypeSet`, which is a **breaking change**. I used `TypeSet` rather than a `TypeList` (of length 1) since the order of keys does not matter, but happy to refactor it if `TypeList` is more appropriate.~~

To preview the documentation changes, copy the file `docs/resources/load_balancer.md` and paste it into the [Terraform doc preview](https://registry.terraform.io/tools/doc-preview). To see the difference with the current docs, [the current docs can be found here](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/load_balancer).

**Other changes include:**
- Improved the written descriptions to align more with the API docs (and fixed typos).
- Improved the formatting of the descriptions (formatting property `values` and linking to other property Markdown anchors).
- Added a terraform import example.
- Moved all schema-related structs into `schema_cloudflare_load_balancer.go`.
- Avoided reusing common `schema.Elem` definitions between LB rule overrides and main definitions. They were made separate as there are minor differences between some of the LB rule overrides and the main definitions, such as `session_affinity_attributes` not supporting the property `drain_duration` when it is a rule override. In addition, validation may be slightly different for rule overrides. The descriptions for all rule overrides are in the form `See (<main definition property name>)`, such as `See session_affinity_attributes`.

**Open questions**
1.  I noticed that the main load balancer schema definitions include `fallback_pool_id` and `default_pool_ids` properties, which correspond to the [API definitions](https://api.cloudflare.com/#load-balancers-create-load-balancer) `fallback_pool` and `default_pools`. The LB rule overrides for these are named `fallback_pool` and `default_pools`, which is a bit inconsistent. Since we are already introducing a breaking change, would it make sense to rename both occurrences to the same name? Are there any other breaking change improvements we can think of that might be nice to make now?
2.  We have properties `pop_pools`/`country_pools`/`region_pools`, all which share a common sub-property `pool_ids`. As a result, when trying to create a Markdown anchor link to `pools_ids`, like so — `[pool_ids](#pool_ids)` — it will link to the first `pool_ids` definition, every time. Do you know if it is possible to tell Terraform to create an anchor so that we can link to a particular definition? (Of course, we can just change the autogenerated Markdown, but this is not a great solution).

**Future work / Thoughts**
- I noticed that some of the properties have `Computed: true`, which aren't _really_ computed and caused some confusion for me. Also, we can't set default values when `Computed: true`. For instance, `ttl` has this set, which prevents us from setting `Default: 30`. `steering_policy` also has it set and I believe this is because the API will return either `"off"` or `"geo"` when given an empty string. I still don't fully understand why `Computed` is needed in some of these places.
- I tried to maintain some level of consistency between the main schema definitions and rule override definitions, but there were a few places I was afraid to touch, thinking there might be a possible corner case. For instance, the rule overrides `fallback_pool` and `default_pools` do not have any validation.

<details>
  <summary>Acceptance tests pass</summary>
 
  ```
 TESTARGS='-run "^TestAccCloudflareLoadBalancer_*" -count 1 -parallel 1' make testacc
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareLoadBalancer_*" -count 1 -parallel 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareLoadBalancerMonitor_Import
=== PAUSE TestAccCloudflareLoadBalancerMonitor_Import
=== RUN   TestAccCloudflareLoadBalancerPool_Import
=== PAUSE TestAccCloudflareLoadBalancerPool_Import
=== RUN   TestAccCloudflareLoadBalancer_Import
=== PAUSE TestAccCloudflareLoadBalancer_Import
=== RUN   TestAccCloudflareLoadBalancerMonitor_Basic
--- PASS: TestAccCloudflareLoadBalancerMonitor_Basic (3.92s)
=== RUN   TestAccCloudflareLoadBalancerMonitor_FullySpecified
--- PASS: TestAccCloudflareLoadBalancerMonitor_FullySpecified (10.49s)
=== RUN   TestAccCloudflareLoadBalancerMonitor_EmptyExpectedBody
--- PASS: TestAccCloudflareLoadBalancerMonitor_EmptyExpectedBody (4.10s)
=== RUN   TestAccCloudflareLoadBalancerMonitor_TcpFullySpecified
--- PASS: TestAccCloudflareLoadBalancerMonitor_TcpFullySpecified (5.07s)
=== RUN   TestAccCloudflareLoadBalancerMonitor_PremiumTypes
--- PASS: TestAccCloudflareLoadBalancerMonitor_PremiumTypes (7.36s)
=== RUN   TestAccCloudflareLoadBalancerMonitor_NoRequired
--- PASS: TestAccCloudflareLoadBalancerMonitor_NoRequired (2.28s)
=== RUN   TestAccCloudflareLoadBalancerMonitor_Update
--- PASS: TestAccCloudflareLoadBalancerMonitor_Update (5.50s)
=== RUN   TestAccCloudflareLoadBalancerMonitor_CreateAfterManualDestroy
--- PASS: TestAccCloudflareLoadBalancerMonitor_CreateAfterManualDestroy (5.72s)
=== RUN   TestAccCloudflareLoadBalancerMonitor_ChangingHeadersCauseReplacement
--- PASS: TestAccCloudflareLoadBalancerMonitor_ChangingHeadersCauseReplacement (5.26s)
=== RUN   TestAccCloudflareLoadBalancerPool_Basic
=== PAUSE TestAccCloudflareLoadBalancerPool_Basic
=== RUN   TestAccCloudflareLoadBalancerPool_FullySpecified
=== PAUSE TestAccCloudflareLoadBalancerPool_FullySpecified
=== RUN   TestAccCloudflareLoadBalancerPool_CreateAfterManualDestroy
=== PAUSE TestAccCloudflareLoadBalancerPool_CreateAfterManualDestroy
=== RUN   TestAccCloudflareLoadBalancer_Basic
=== PAUSE TestAccCloudflareLoadBalancer_Basic
=== RUN   TestAccCloudflareLoadBalancer_SessionAffinity
=== PAUSE TestAccCloudflareLoadBalancer_SessionAffinity
=== RUN   TestAccCloudflareLoadBalancer_AdaptiveRouting
=== PAUSE TestAccCloudflareLoadBalancer_AdaptiveRouting
=== RUN   TestAccCloudflareLoadBalancer_LocationStrategy
=== PAUSE TestAccCloudflareLoadBalancer_LocationStrategy
=== RUN   TestAccCloudflareLoadBalancer_RandomSteering
=== PAUSE TestAccCloudflareLoadBalancer_RandomSteering
=== RUN   TestAccCloudflareLoadBalancer_GeoBalanced
=== PAUSE TestAccCloudflareLoadBalancer_GeoBalanced
=== RUN   TestAccCloudflareLoadBalancer_ProximityBalanced
=== PAUSE TestAccCloudflareLoadBalancer_ProximityBalanced
=== RUN   TestAccCloudflareLoadBalancer_Rules
=== PAUSE TestAccCloudflareLoadBalancer_Rules
=== RUN   TestAccCloudflareLoadBalancer_DuplicatePool
=== PAUSE TestAccCloudflareLoadBalancer_DuplicatePool
=== RUN   TestAccCloudflareLoadBalancer_Update
=== PAUSE TestAccCloudflareLoadBalancer_Update
=== RUN   TestAccCloudflareLoadBalancer_CreateAfterManualDestroy
=== PAUSE TestAccCloudflareLoadBalancer_CreateAfterManualDestroy
=== CONT  TestAccCloudflareLoadBalancerMonitor_Import
--- PASS: TestAccCloudflareLoadBalancerMonitor_Import (3.72s)
=== CONT  TestAccCloudflareLoadBalancer_LocationStrategy
--- PASS: TestAccCloudflareLoadBalancer_LocationStrategy (5.62s)
=== CONT  TestAccCloudflareLoadBalancer_CreateAfterManualDestroy
--- PASS: TestAccCloudflareLoadBalancer_CreateAfterManualDestroy (8.08s)
=== CONT  TestAccCloudflareLoadBalancer_Update
--- PASS: TestAccCloudflareLoadBalancer_Update (7.38s)
=== CONT  TestAccCloudflareLoadBalancer_DuplicatePool
--- PASS: TestAccCloudflareLoadBalancer_DuplicatePool (3.06s)
=== CONT  TestAccCloudflareLoadBalancer_Rules
--- PASS: TestAccCloudflareLoadBalancer_Rules (5.21s)
=== CONT  TestAccCloudflareLoadBalancer_ProximityBalanced
--- PASS: TestAccCloudflareLoadBalancer_ProximityBalanced (5.36s)
=== CONT  TestAccCloudflareLoadBalancer_GeoBalanced
--- PASS: TestAccCloudflareLoadBalancer_GeoBalanced (5.76s)
=== CONT  TestAccCloudflareLoadBalancer_RandomSteering
--- PASS: TestAccCloudflareLoadBalancer_RandomSteering (6.18s)
=== CONT  TestAccCloudflareLoadBalancerPool_CreateAfterManualDestroy
--- PASS: TestAccCloudflareLoadBalancerPool_CreateAfterManualDestroy (5.83s)
=== CONT  TestAccCloudflareLoadBalancer_AdaptiveRouting
--- PASS: TestAccCloudflareLoadBalancer_AdaptiveRouting (5.33s)
=== CONT  TestAccCloudflareLoadBalancer_SessionAffinity
--- PASS: TestAccCloudflareLoadBalancer_SessionAffinity (12.39s)
=== CONT  TestAccCloudflareLoadBalancer_Basic
--- PASS: TestAccCloudflareLoadBalancer_Basic (5.23s)
=== CONT  TestAccCloudflareLoadBalancerPool_Basic
--- PASS: TestAccCloudflareLoadBalancerPool_Basic (4.01s)
=== CONT  TestAccCloudflareLoadBalancerPool_FullySpecified
--- PASS: TestAccCloudflareLoadBalancerPool_FullySpecified (4.02s)
=== CONT  TestAccCloudflareLoadBalancer_Import
--- PASS: TestAccCloudflareLoadBalancer_Import (5.72s)
=== CONT  TestAccCloudflareLoadBalancerPool_Import
--- PASS: TestAccCloudflareLoadBalancerPool_Import (4.26s)
PASS
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/provider   147.055s
  ```
</details>

This schema still needs lots of more love, but hopefully these changes put it back on the right track!


jacobbednarz:
thanks for this massive effort @tc80 👏 i'm very much in favour of getting this to a more consistent state however, we can't do big breaking changes like this outside of next major version bump (tracking 4.x in #1646). what i would suggest is for now, to put two PRs together. one that has all the backwards compatible changes in it to merge immediately and a second with all the breaking changes that we can track for 4.x. i'll have to schedule some time soon to start on putting the 4.x release together and iron out a bunch of the changes mentioned in #1646. the breaking changes will also need to provide a schema migration to try and automatically help out operators as much as possible with their migration to 4.x when it comes (`TypeMap` => `TypeSet`, field renames, etc).

> I noticed that the main load balancer schema definitions include fallback_pool_id and default_pool_ids properties, which correspond to the [API definitions](https://api.cloudflare.com/#load-balancers-create-load-balancer) fallback_pool and default_pools. The LB rule overrides for these are named fallback_pool and default_pools, which is a bit inconsistent. Since we are already introducing a breaking change, would it make sense to rename both occurrences to the same name? Are there any other breaking change improvements we can think of that might be nice to make now?

none that come to mind (other than adding `account_id` to other specific LB resources) but this is a good call out.

> We have properties pop_pools/country_pools/region_pools, all which share a common sub-property pool_ids. As a result, when trying to create a Markdown anchor link to pools_ids, like so — [pool_ids](#pool_ids) — it will link to the first pool_ids definition, every time. Do you know if it is possible to tell Terraform to create an anchor so that we can link to a particular definition? (Of course, we can just change the autogenerated Markdown, but this is not a great solution).

none to my knowledge. the automatic documentation uses the nested level as part of the anchor however, that doesn't seem to extend to the properties of the nested schema. this might be something we can suggest/PR to the [upstream documentation tool, terraform-plugin-docs](https://github.com/hashicorp/terraform-plugin-docs). i couldn't find an open issue for this one.

> I still don't fully understand why Computed is needed in some of these places.

from the [docs](https://www.terraform.io/plugin/sdkv2/schemas/schema-behaviors#computed): 

> `Computed` is often used to represent values that are not user configurable or can not be known at time of `terraform plan` or `apply`, such as date of creation or a service specific UUID. `Computed` can be combined with other attributes to achieve specific behaviors, and can be used as output for interpolation into other resources

i think it was likely that because we defined the default in the API instead, we opted to have these listed as `Computed` in the schema to allow the API to provide the default when one wasn't provided. it's entirely possible, this is wrong too 😄

tc80:
Thanks for all the feedback @jacobbednarz! Really appreciate it 🙏  
You are also super prompt at responding!! 🥇

I amended the commit so that it is backwards compatible (and naturally reverted all changes to the test file). I will open a new PR once this merges with the breaking change to refactor `TypeMap => TypeSet`.

Since `session_affinity_attributes` is a `TypeMap`, I just shoved all the definitions of its sub-properties into it, like so:
session_affinity_attributes

jacobbednarz:
thanks for understanding. unfortunately, `TypeMap` is still a thorn in our side and we're slowly moving everything off it since 2.x due to issues like this in the internals.

that's a good workaround for the `session_affinity_attributes`. i was going to say, if you don't have any better approaches, we could use a custom template (in `templates/resources/load_balancer.md.tmpl) that looks like this and append those descriptions to the bottom to link to. 

```
---
page_title: "{{.Name}} {{.Type}} - {{.RenderedProviderName}}"
subcategory: ""
description: |-
{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
---

# {{.Name}} ({{.Type}})

{{ .Description | trimspace }}

## Example Usage

{{ tffile (printf "%s%s%s" "examples/resources/" .Name "/resource.tf") }}

{{ .SchemaMarkdown | trimspace }}

### Nested schema for `session_affinity_attributes` 
<a href="#nested-schema-for-session_affinity_attributes"></a>

Optional: 

- `secure` (String) Configures the SameSite attribute on session affinity cookie. Value Auto will be translated to Lax or None depending if Always Use HTTPS is enabled. Note: when using value None, then you can not set [secure="Never"](https://registry.terraform.io/tools/doc-preview#secure). Available values: `Auto`, `Lax`, `None`, `Strict`. 
- ...

## Import

Import is supported using the following syntax:

{{ codefile "shell" (printf "%s%s%s" "examples/resources/" .Name "/import.sh") }}
```

then in your `session_affinity_attributes` link to `#nested-schema-for-session_affinity_attributes`. then once this breaking change happens, remove the custom template and you'll be 100% autogen then.

i'll take a final pass over this tomorrow but it's looking good! thanks again.

tc80:
Thanks for the idea! I didn't realize custom templates could be used like so.

To be honest, I would prefer avoiding the custom template if we can and keeping the documentation in the same place. However, if you feel strongly I have no problem making that change 😄

tc80:
Never mind go for it my dude 😆

jacobbednarz:
i just whipped in the custom template for you to take a look and let me know what you think. it keeps the same feel, etc so it blends in a bit better than all in a single attribute but let me know your thoughts. it's just as easy to revert!

tc80:
Sweet, let's keep it. Thanks for pushing that up!

When we merge this I will open another PR to do the breaking change and revert this custom template 😄

tc80:
Got to be careful with the Available Values! 

Screen Shot 2022-10-11 at 1 02 42 AM

templates/resources/load_balancer.md.tmpl
```diff
@@ -0,0 +1,32 @@
...
+{{ tffile (printf "%s%s%s" "examples/resources/" .Name "/resource.tf") }}
+
+{{ .SchemaMarkdown | trimspace }}
+
+### Nested schema for `session_affinity_attributes`
+<a href="#nested-schema-for-session_affinity_attributes"></a>
+
+Optional:
+
+- `samesite` (String) Configures the SameSite attribute on session affinity cookie. Value `Auto` will be translated to `Lax` or `None` depending if Always Use HTTPS is enabled. Note: when using value `None`, then you can not set `secure="Never". Available values: `Auto`, `Lax`, `None`, `Strict`.
```
tc80, line 23:
```suggestion
- `samesite` (String) Configures the SameSite attribute on session affinity cookie. Value `Auto` will be translated to `Lax` or `None` depending if Always Use HTTPS is enabled. Note: when using value `None`, then you can not set `secure="Never"`. Available values: `Auto`, `Lax`, `None`, `Strict`.
```

tc80:
Pushed one more commit to update the documentation (`make docs`).

tc80:
Thanks for catching that! 🙏

jacobbednarz:
thanks a bunch @tc80! this is awesome!

PR #1960, opened 2022-10-12
ACCT-4178: Add raw permissions struct to provider

### LINKED ISSUE #1004: when creating cloudflare_api_token with permission for logs Write on the zone level, permission is assigned on the account level
### Confirmation

### Terraform version

0.14.8

### Affected resource(s)

cloudflare_api_token

resource "cloudflare_api_token" "zone_tokens" {
  for_each                               = local.zonemap
  name                                   = "${each.key}-key"
  policy {
    permission_groups = [
      data.cloudflare_api_token_permission_groups.all.permissions["Logs Write"],
    ]
    resources = {
        "com.cloudflare.api.account.${var.account_id}" = "*"
    }
  }
  policy {
    permission_groups = [
      data.cloudflare_api_token_permission_groups.all.permissions["DNS Write"],
      data.cloudflare_api_token_permission_groups.all.permissions["Access: Apps and Policies Write"],
      data.cloudflare_api_token_permission_groups.all.permissions["SSL and Certificates Write"],
      data.cloudflare_api_token_permission_groups.all.permissions["Zone Settings Write"],
      data.cloudflare_api_token_permission_groups.all.permissions["Zone Write"],
      data.cloudflare_api_token_permission_groups.all.permissions["Firewall Services Write"],
      data.cloudflare_api_token_permission_groups.all.permissions["Page Rules Write"],
      data.cloudflare_api_token_permission_groups.all.permissions["Workers Routes Write"],
      data.cloudflare_api_token_permission_groups.all.permissions["Logs Write"],
    ]
    resources = {
        "com.cloudflare.api.account.zone.${each.value}" = "*"
    }
  }
}

### Debug output

_No response_

### Panic output

_No response_

### Expected output

token created with permissions for Logs Write on the zone level

### Actual output

token created with permissions for Logs Write on the account  level

### Steps to reproduce

execute the code above

### Additional factoids

in the api docs for permissions : https://developers.cloudflare.com/api/tokens/create/permissions
logs write, like many others exists on both account and zone level, all work fine with the exception of the logs write so far.

### References

_No response_

cloudflare-tf

ianmobbs:
At the moment, there is an issue with API Token permission groups. When fetching permission groups, there are 5 or 6 groups that can be either account-scoped or zone-scoped. Each of these permission groups results in two objects in the permission group response with the same name but different scopes - for example (pseudocode):
```
[
    {
        "name": "Permission Group 1",
        "scope": "com.cloudflare.api.account"
    },
    {
        "name": "Permission Group 1",
        "scope": "com.cloudflare.api.zone"
    }
]
```

As we're indexing by name only when calling `d.Set`, this can cause the ID for `Permission Group 1` (account scope) to be overridden with the ID for `Permission Group 1` (zone scope).

The resolution here is to index by name AND scope. To achieve this, we are adding a new `zone`, `account` and `user` attributes that contain *only* those for the resource level as opposed to all together. This still allows the ability of building any permission policy you need, but with explicit resource levels as part of the reference.

Closes #1004


internal/provider/data_source_api_token_permission_groups.go
```diff
@@ -32,6 +32,11 @@ func dataSourceCloudflareApiTokenPermissionGroupsRead(ctx context.Context, d *sc
 		return diag.FromErr(fmt.Errorf("error listing API Token Permission Groups: %w", err))
 	}
 
+	err = d.Set("raw_permissions", permissions)
```
jacobbednarz, line 35: this field doesn't exist on the schema so it won't have any effect.

what i would recommend we do here is add two new fields to the schema, `account` and `zones` each with their respective permissions that sit under it. that way, if someone wants to target `Foo Permission` on their zone, it's pretty explicit as `data.cloudflare_api_token_permissions.zone["Foo Permission"]`

PR #1973, opened 2022-10-15
Add defaults and domain importer

### LINKED ISSUE #1924: Pages project should have deployments enabled by default

Terraform v1.2.4
on darwin_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.24.0
+ provider registry.terraform.io/hashicorp/vault v3.7.0

### Description

Currently the default value for `deployments_enabled` in the `pages_project` resource is `false`. This should instead be `true`.

### Use cases

Setting `deployments_enabled` to `true` by default or explicitly documenting that it's `false` by default would prevent unnecessary confusion surrounding a project being created but seeing no deployments.

```hcl
resource "cloudflare_pages_project" "example" {
  account_id        = var.account_id
  name              = "example"
  production_branch = "master"

  build_config {
    build_command   = "yarn build"
    destination_dir = "build"
  }

  source {
    type = "github"

    config {
      owner               = "cloudflare"
      repo_name           = "1.1.1.1"
      production_branch   = "master"
      pr_comments_enabled = false
    }
  }
}
```

### References

_No response_

### LINKED ISSUE #1958: Add import feature to cloudflare_pages_domain

0.13.5 and 3.25.0

### Description

I'm getting similar errors as #1946 and was hoping I could temporarily fix it by importing the domain. I don't see a way to import in the documentation or in the code when comparing to something like cloudflare_pages_project.

ID is not easily found without opening the dev console in a browser. Perhaps the name of the domain would work best in combination with account ID? Perhaps project too?

Also, this seems way too small to file a bug report, but https://github.com/cloudflare/terraform-provider-cloudflare/blob/1cd58f49a1b7454d0c42ccf75e08c8f7bb53620f/internal/provider/resource_cloudflare_pages_domain.go#L39 is incorrect and says it will output a project but outputs the account ID.

### Use cases

See above. Importing is a common feature for resources.

n/a

### References

_No response_

WalshyDev:
Adds defaults to Pages deployment config - closes #1924

Adds Pages domain importer - Closes #1958


.changelog/1973.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+resource/cloudflare_pages_domain: add Pages project domain importer. Closes #1958
```
jacobbednarz, line 2:
```suggestion
resource/cloudflare_pages_domain: add Pages project domain importer
```

internal/provider/resource_cloudflare_pages_domain.go
```diff
@@ -78,3 +80,39 @@ func resourceCloudflarePagesDomainDelete(ctx context.Context, d *schema.Resource
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to find domains for project %q: %w", d.Id(), err)
+	}
+
+	tflog.Info(ctx, fmt.Sprintf("Found domains: %+v", domains))
+
+	resourceData := []*schema.ResourceData{}
+
+	for _, domain := range domains {
+		d.SetId(domain.ID)
```
jacobbednarz, line 108: i don't think this is going to work the way we expect it to here. `Import` is only called expected to be called once per resource and only save that state (once for `example.com` and once for `example.net`, for example). here, it looks like we are expecting multiple domains and overwriting `ID` each time meaning only the last one will actually save. the resource schema and docs implies this is singular as well which is confusing here.

jacobbednarz: i think you want [`GetPagesDomain`](https://github.com/cloudflare/cloudflare-go/blob/e2c82ef2466084055e6706b1104a7875038cf2a6/pages_domain.go#L86) here and to do it for each resource individually.

WalshyDev: That makes sense, got it :)

jacobbednarz:
nice one! thanks @WalshyDev !

Issue #1974, opened 2022-10-15
cloudflare_access_application missing validation for type="private_ip"


csautter:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Tested with:
- Terraform v1.1.8
- Terraform v1.3.2
- cloudflare v3.25.0

### Affected resource(s)

- cloudflare_access_application 

### Terraform configuration files

```hcl
resource "cloudflare_access_application" "test" {
  account_id       = local.cloudflare_account_id
  name             = "test"
  type             = "private_ip"
  session_duration = "24h"
}
```

### Link to debug output

https://gist.github.com/csautter/2f2bd62cd48ae5f11d51d913ba946f77

### Panic output

````
╷
│ Error: expected type to be one of [self_hosted saas ssh vnc bookmark], got private_ip
│
│   with cloudflare_access_application.test,
│   on cloudflare_access_application_prox_backup_1.tf line 11, in resource "cloudflare_access_application" "test":
│   11:   type             = "private_ip"
│
╵
````

### Expected output

Ressource **cloudflare_access_application** accepts `type  = "private_ip"`

### Actual output

Error: expected type to be one of [self_hosted saas ssh vnc bookmark], got private_ip

### Steps to reproduce

1. set `type  = "private_ip"`

### Additional factoids

It's possible to import a ressource with `type="private_ip"` but actually it isn't possible to set this value in terraform code.
````txt
$ terraform import "cloudflare_access_application"."test" .../..
$ terraform state show "cloudflare_access_application"."test"
# cloudflare_access_application.test:
resource "cloudflare_access_application" "test" {
    account_id                 = "<removed>"
    allowed_idps               = []
    app_launcher_visible       = true
    aud                        = "<removed>"
    auto_redirect_to_identity  = false
    enable_binding_cookie      = false
    http_only_cookie_attribute = false
    id                         = "<removed>"
    name                       = "test"
    service_auth_401_redirect  = false
    skip_interstitial          = false
    type                       = "private_ip"
}

````
Looking at the API Docs it seems that `type="private_ip"`  is undocumented.
https://api.cloudflare.com/#access-applications-add-the-access-app-launcher

### References

[...]

PR #1975, opened 2022-10-15
resource/cloudflare_access_application: add type=private_ip

csautter:
Quick Fix for Issue

Closes https://github.com/cloudflare/terraform-provider-cloudflare/issues/1974


jacobbednarz:
@csautter to get this over the line, you'll need to:

- address the inline comments
- add a CHANGELOG entry (see comment above)
- run `make docs` locally to update the generated registry documentation (see https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md)

jacobbednarz:
@Justin-Holmes do we have this in the backlog to be documented publicly?

Justin-Holmes:
> @Justin-Holmes do we have this in the backlog to be documented publicly?

Unfortunately, we don't have plans on adding private apps to the API docs because they're considered unstable at the moment. We'll be doing work on private apps in the coming months and we'll publicly document after that work is complete!

jacobbednarz:
thanks! closing this out in the meantime while unstable.

F21:
Sorry for bumping up a closed issue, but is the private apps api considered stable now? They aren't labeled as beta in the zero trust UI and seem to be well documented here: https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/private-net/connect-private-networks/

itsmechlark:
@jacobbednarz , any news? Is this possible to get release so we can provision access apps for private ips.

kurochan:
I am looking forward to it as well.

bipbipbipbipbip:
@jacobbednarz it has been a more than a year since this PR was opened. As far as we can tell the API is considered stable, so when could we expect an update on these resources?

derence-boxed:
Is it possible to add this type to the resource cloudflare_access_application? face the same situation here as well

nlkaisert:
@jacobbednarz @Justin-Holmes is the API considered stable now? The type seems to still be missing from the cloudflare_access_application, but when using the cf-terraforming tool, then it gets generated with the type "private_ip". This is highly confusing. Is this scenario supported or not?

PR #2006, opened 2022-10-31
Default value for root_dir in pages_project.md

pedrocabido:
The simple description of the default value makes it much more clearer.
I've tested with `root_dir = "/"` and I got his value on Cloudflare console: `//`.

E.g.:
image

### DIFF
--- a/docs/resources/pages_project.md
+++ b/docs/resources/pages_project.md
@@ -135,7 +135,7 @@ Optional:
 
 - `build_command` (String) Command used to build project.
 - `destination_dir` (String) Output directory of the build.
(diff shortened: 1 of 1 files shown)


jacobbednarz (changes requested):
our documentation is automatically generated so you'll need to review https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md as to propose this change and have it not be overwritten next run.

`Defaults to ...` is based off the schema as well (which we currently don't have) so for this to be true, we'll want to add that to the schema.

Issue #2011, opened 2022-11-07
An argument named "origin" is not expected here. Did you mean to define a block of type "origin"?


puzzledvacuum:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```bash
$ terraform -v
Terraform v1.3.4
on darwin_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.27.0
+ provider registry.terraform.io/datadog/datadog v2.26.1
+ provider registry.terraform.io/hashicorp/aws v3.75.2
+ provider registry.terraform.io/hashicorp/external v2.2.2
+ provider registry.terraform.io/hashicorp/helm v2.7.1
+ provider registry.terraform.io/hashicorp/http v2.2.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.15.0
+ provider registry.terraform.io/hashicorp/null v3.2.0
+ provider registry.terraform.io/hashicorp/random v3.4.3
+ provider registry.terraform.io/integrations/github v4.31.0
+ provider registry.terraform.io/mongodb/mongodbatlas v1.4.6
```

### Affected resource(s)

Only `cloudflare_ruleset`.

### Terraform configuration files

```hcl
resource "cloudflare_ruleset" "http_origin_example" {
  zone_id     = "0da42c8d2132a9ddaf714f9e7c920711"
  name        = "Change to some origin"
  description = "Change origin for a route"
  kind        = "zone"
  phase       = "http_request_origin"

  rules {
    action = "route"
    action_parameters {
      host_header = "some.host"
      origin = {
        host = "some.host"
        port = 80
      }
    }
    expression  = "(http.request.uri.path matches \"^/api/\")"
    description = "change origin to some.host"
    enabled     = true
  }
}
# should be =>
resource "cloudflare_ruleset" "http_origin_example" {
  zone_id     = "0da42c8d2132a9ddaf714f9e7c920711"
  name        = "Change to some origin"
  description = "Change origin for a route"
  kind        = "zone"
  phase       = "http_request_origin"

  rules {
    action = "route"
    action_parameters {
      host_header = "some.host"
      origin {
        host = "some.host"
        port = 80
      }
    }
    expression  = "(http.request.uri.path matches \"^/api/\")"
    description = "change origin to some.host"
    enabled     = true
  }
}
```

### Link to debug output

[...]

PR #2020, opened 2022-11-10
Update to fix documentation issue 2011

tenaciousdlg:
Closes https://github.com/cloudflare/terraform-provider-cloudflare/issues/2011


templates/resources/ruleset.md.tmpl
```diff
@@ -1,31 +0,0 @@
----
```
jacobbednarz, line 1: this file needs to hang around

jacobbednarz:
once you add that file back, run `make docs` locally and you should be good to go.

PR #2021, opened 2022-11-10
Allowing usage of separate providers with api_user_service_key and normal credentials

mlhnono68:
As of now, the terraform provider for Cloudflare only setting either `api_user_service_key`, `api_token` or `api_key` which might make sense at first glance. However, there are use cases where some terraform resources need to use both specific `api_user_service_key` credentials as well as normal tokens credentials. But with current codebase, setting both environment variables `CLOUDFLARE_API_USER_SERVICE_KEY` and one of `CLOUDFLARE_API_TOKEN` or `CLOUDFLARE_API_KEY` leads to an error because of `ExactlyOneOf` constraints in the provider schema.

With this proposed change, it becomes possible to use different providers and override the environment variables the way we need like hereunder :
```
provider "cloudflare" {
  # This provider will assume the normal API credentials
  api_user_service_key = ""
}

provider "cloudflare" {
  # This provider alias will be used specifically for calls in need for api_user_service_key
  alias = "origin"
  api_token = ""
}
```

I don't think this needs a release note

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -92,8 +92,9 @@ func New(version string) func() *schema.Provider {
 					Optional:     true,
 					DefaultFunc:  schema.EnvDefaultFunc("CLOUDFLARE_API_KEY", nil),
 					Description:  "The API key for operations. Alternatively, can be configured using the `CLOUDFLARE_API_KEY` environment variable. API keys are [now considered legacy by Cloudflare](https://developers.cloudflare.com/api/keys/#limitations), API tokens should be used instead.",
-					ExactlyOneOf: []string{"api_key", "api_token", "api_user_service_key"},
-					ValidateFunc: validation.StringMatch(regexp.MustCompile("[0-9a-f]{37}"), "API key must be 37 characters long and only contain characters 0-9 and a-f (all lowercased)"),
+					AtLeastOneOf: []string{"api_key", "api_token", "api_user_service_key"},
+					ExactlyOneOf: []string{"api_key", "api_token"},
+					ValidateFunc: validation.StringMatch(regexp.MustCompile("[0-9a-f]{37}|"), "API key must be 37 characters long and only contain characters 0-9 and a-f (all lowercased)"),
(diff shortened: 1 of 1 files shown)


jacobbednarz:
thanks but i'm afraid i'm going to have to decline this. when we send authentication credentials, we don't ever want there to be a case where you can send more than one of these at a time. it makes debugging incredibly difficult in both the provider and internal systems.

> However, there are use cases where some terraform resources need to use both specific api_user_service_key credentials as well as normal tokens credentials. 

i'm not sure what resources you are referring to as this should never be the case. the only time you need the API service key is for origin CA certificates and for now, you should either alias the provider or set it into a different Terraform directory and reference the resource as needed. i'm hoping in the near future, this auth mechanism will go away in favour of API keys or tokens.

mlhnono68:
I was actually referring to the origin CA certificates. I wanted with this merge request to address the problem that it still creates issues when using provider aliases. The problem is that in case you leverage environment variables to provide values for `api_user_service_key` and `api_token` both provider aliases will complain about the `ExactlyOneOf` constraint because both aliases will find a value for `api_user_service_key` and `api_token` from the environment variables. 

With my proposed change, the idea was at least to be able to override the `DefaultFunc` with an empty string to enable using 2 provider alises with environment variables.

Your concerns about debugging difficulty should be addressed by the fact I kept `ExactlyOneOf: []string{"api_key", "api_token"},`

I hope you might reconsider the merge closure :) @jacobbednarz

jacobbednarz:
> Your concerns about debugging difficulty should be addressed by the fact I kept ExactlyOneOf: []string{"api_key", "api_token"},

the issues are with all three auth types, not just those two which is why we enforce being able to only set a single one and avoid the headache. 

i'm afraid at this point, there isn't a good way to allow a combination of the keys to be set and have a clear path for which auth type is to be used internally. as the impact here is relatively low for those that use origin CAs, that is the trade off we're making for now until we have a better solution for this single endpoint.

PR #2034, opened 2022-11-17
resource/cloudflare_account: initialize AccountSettings

### LINKED ISSUE #2031: v3.28.0 crash on cloudflare_account update
### Confirmation

Terraform v0.13.7

I know this is pretty old but the environement I'm working with cannot be upgraded at the moment

### Affected resource(s)

`cloudflare_account`

```hcl
terraform {
  required_providers {
    cloudflare = {
      source = "cloudflare/cloudflare"
      version = "3.28.0"
    }
  }
}

provider "cloudflare" {
  account_id = "REDACTED"
  api_token = "REDACTED"
}

resource "cloudflare_account" "test" {
  name = "test-new"
  enforce_twofactor = true
  type = "standard"
}
```

### Link to debug output

https://gist.github.com/jfchevrette/a6fe4852b3e58a248716469891b4d8ac

### Panic output

See panic at bottom of https://gist.github.com/jfchevrette/a6fe4852b3e58a248716469891b4d8ac

### Expected output

```
# cloudflare_account.test will be updated in-place
  ~ resource "cloudflare_account" "test" {
      ~ enforce_twofactor = false -> true
        id                = "2a98f16f04b782a94f252de9a6277653"
      ~ name              = "test" -> "test-new"
        type              = "standard"
    }
```

### Actual output

See https://gist.github.com/jfchevrette/a6fe4852b3e58a248716469891b4d8ac

### Steps to reproduce

See above config and expected output

### Additional factoids

_No response_

### References

_No response_

jfchevrette:
Attempt to fix a panic issue with uninitialized `account.Settings` in `resourceCloudflareAccountUpdate`

Fixes #2031

### DIFF
--- a/.changelog/2034.txt
+++ b/.changelog/2034.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_account: Fix uninitialized cloudflare.Account.Settings
+```
--- a/internal/provider/resource_cloudflare_account.go
+++ b/internal/provider/resource_cloudflare_account.go
@@ -86,7 +86,9 @@ func resourceCloudflareAccountUpdate(ctx context.Context, d *schema.ResourceData
 
 	tflog.Debug(ctx, fmt.Sprintf("Updating Cloudflare Account: id %s", accountID))
 
-	updatedAcc := cloudflare.Account{}
+	updatedAcc := cloudflare.Account{
+		Settings: &cloudflare.AccountSettings{},
+	}
 	if accountName, ok := d.GetOk("name"); ok {
 		updatedAcc.Name = accountName.(string)
 	}
--- a/internal/provider/resource_cloudflare_account.go
+++ b/internal/provider/resource_cloudflare_account.go
@@ -86,15 +86,15 @@ func resourceCloudflareAccountUpdate(ctx context.Context, d *schema.ResourceData
 
 	tflog.Debug(ctx, fmt.Sprintf("Updating Cloudflare Account: id %s", accountID))
 
-	updatedAcc := cloudflare.Account{
-		Settings: &cloudflare.AccountSettings{},
-	}
+	updatedAcc := cloudflare.Account{}
 	if accountName, ok := d.GetOk("name"); ok {
 		updatedAcc.Name = accountName.(string)
 	}
 
 	if enforce_twofactor, ok := d.GetOk("enforce_twofactor"); ok {
-		updatedAcc.Settings.EnforceTwoFactor = enforce_twofactor.(bool)
+		updatedAcc.Settings = &cloudflare.AccountSettings{
+			EnforceTwoFactor: enforce_twofactor.(bool),
+		}
 	}
 
 	_, err := client.UpdateAccount(ctx, accountID, updatedAcc)
--- a/internal/provider/resource_cloudflare_account_test.go
+++ b/internal/provider/resource_cloudflare_account_test.go
@@ -22,17 +22,13 @@ func TestAccCloudflareAccount_Basic(t *testing.T) {
 			{
 				Config: testAccCheckCloudflareAccountName(rnd, fmt.Sprintf("%s_old", rnd)),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr(
-						name, "name", fmt.Sprintf("%s_old", rnd),
-					),
+					resource.TestCheckResourceAttr(name, "name", fmt.Sprintf("%s_old", rnd)),
 				),
 			},
 			{
 				Config: testAccCheckCloudflareAccountName(rnd, fmt.Sprintf("%s_new", rnd)),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr(
-						name, "name", fmt.Sprintf("%s_new", rnd),
-					),
+					resource.TestCheckResourceAttr(name, "name", fmt.Sprintf("%s_new", rnd)),
 				),
 			},
 		},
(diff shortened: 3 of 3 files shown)


jacobbednarz:
thanks for this @jfchevrette. appreciate it!

instead of unconditionally initialising the `Settings` i opted to only initialise it when 2FA is present. this avoids unnecessary allocations only to be ignored by the struct. added a test as well to confirm the functionality works as intended.

```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareAccount_" -count 1 -parallel 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareAccount_Basic
=== PAUSE TestAccCloudflareAccount_Basic
=== RUN   TestAccCloudflareAccount_2FAEnforced
=== PAUSE TestAccCloudflareAccount_2FAEnforced
=== CONT  TestAccCloudflareAccount_Basic
--- PASS: TestAccCloudflareAccount_Basic (18.17s)
=== CONT  TestAccCloudflareAccount_2FAEnforced
--- PASS: TestAccCloudflareAccount_2FAEnforced (18.58s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/provider	37.356s
```

kwilczynski:
@jacobbednarz, thank you for accepting this change! Much appreciated.

However, with the changes you added atop of the original changeset, you won't ever be able to disable the 2-factor requirements via the `enforce_twofactor` attribute (the test added wouldn't catch this). This is because of how `schema.ResourceData.GetOK()` and Go zero-values work.

Have a look at the following. I added `GetOk()`, `GetOkExists()` and printed the variable that hold the `Account` type, as per:

- Enabling `enforce_twofactor`:

```go
2022-11-18T21:49:01.484+0900 [WARN]  unexpected data: registry.terraform.io/cloudflare/cloudflare:stdout="(string) (len=7"
2022-11-18T21:49:01.484+0900 [WARN]  unexpected data: registry.terraform.io/cloudflare/cloudflare:stdout=") "GetOk()""
2022-11-18T21:49:01.484+0900 [WARN]  unexpected data: registry.terraform.io/cloudflare/cloudflare:stdout="(bool) true
(bool) true
(string) (len=13) "GetOkExists()"
(bool) true
(bool) true
(string) (len=10) "updatedAcc"
(cloudflare.Account) {
 ID: (string) "",
 Name: (string) (len=11) "kwilczynski",
 Type: (string) "",
 CreatedOn: (time.Time) 0001-01-01 00:00:00 +0000 UTC,
 Settings: (*cloudflare.AccountSettings)(0xc000044c20)({
  EnforceTwoFactor: (bool) true
 })
}"
```

- Disabling `enforce_twofactor`:

```go
2022-11-18T23:08:03.778+0900 [WARN]  unexpected data: registry.terraform.io/cloudflare/cloudflare:stdout=""GetOk()"
(bool) false
(bool) false
(string) (len=13) "GetOkExists()"
(bool) false
(bool) true
(string) (len=10) "updatedAcc"
(cloudflare.Account) {
 ID: (string) "",
 Name: (string) (len=11) "kwilczynski",
 Type: (string) "",
 CreatedOn: (time.Time)"
2022-11-18T23:08:03.778+0900 [WARN]  unexpected data: registry.terraform.io/cloudflare/cloudflare:stdout="0001-01-01 00:00:00 +0000 UTC,"
2022-11-18T23:08:03.778+0900 [WARN]  unexpected data: registry.terraform.io/cloudflare/cloudflare:stdout=Settings:
2022-11-18T23:08:03.778+0900 [WARN]  unexpected data: registry.terraform.io/cloudflare/cloudflare:stdout=(*cloudflare.AccountSettings)(<nil>)
2022-11-18T23:08:03.778+0900 [WARN]  unexpected data: registry.terraform.io/cloudflare/cloudflare:stdout=}
```

Even though the logging in Terraform can be a bit unwieldy, can you see where the problem manifests itself?

jacobbednarz:
the zero value issue will need to be solved by updating the struct in cloudflare-go to be a boolean pointer. regardless of where the settings struct would have been initialised, we still would have had this issue due to zero values in Terraform, not only Go (this is something I've raised with the core team a couple of times).

jacobbednarz:
actually disregard; this is slightly different to the underlying issue at https://github.com/hashicorp/terraform-plugin-sdk/issues/817.

as the schema has `false` as the default, the update shouldn't be conditionally setting this at all. the payload should always include the value regardless. 

you're welcome to send a PR removing the conditional or I can do it next week.

kwilczynski:
Hi @jacobbednarz,

> actually disregard; this is slightly different to the underlying issue at [hashicorp/terraform-plugin-sdk#817](https://github.com/hashicorp/terraform-plugin-sdk/issues/817).

I am glad you raised this with them. I was worried that it would be a hill I, and many other people, would die on alone, so to speak. This has been an issue in Terraform for almost half a decade now.

> as the schema has `false` as the default, the update shouldn't be conditionally setting this at all. the payload should always include the value regardless.
>
> you're welcome to send a PR removing the conditional or I can do it next week.

You can do it at your convenience as long as it makes it into the next release. No rush otherwise. Also, thank you in advance for your help! Much appreciated.

jacobbednarz:
i've raised #2036 which covers the removal use case however, in testing this, i also found [a bug](https://github.com/cloudflare/terraform-provider-cloudflare/pull/2036/files#diff-cf7f35502d88a193002a158317c3770facbb87e3f277422efa20d3201b7d217eR66-R68) that prevents creating a new account with the settings from the start. the only way to apply 2FA is with a second run to update the resource. thanks for helping uncover these two issues!

kwilczynski:
Hi @jacobbednarz,

> i've raised #2036 which covers the removal use case however, in testing this, i also found [a bug](https://github.com/cloudflare/terraform-provider-cloudflare/pull/2036/files#diff-cf7f35502d88a193002a158317c3770facbb87e3f277422efa20d3201b7d217eR66-R68) that prevents creating a new account with the settings from the start. the only way to apply 2FA is with a second run to update the resource.

This amazing. Thank you so much!

>  thanks for helping uncover these two issues!

My pleasure! This is some amazing work you and your team are doing with this provider, so I am always happy to help as its user. :)

PR #2048, opened 2022-11-22
Add logic to renew the origin_ca_certificate when expires

### LINKED ISSUE #2043: Implement the logic to auto renew origin_ca certificate

Terraform v1.3.5
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v3.27.0

### Description

Right now the cloudflare_origin_ca_certificate resource is not handling the renewal, I would like to have a control similar to eary renewal hours from hashicorpt tls (https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/self_signed_cert#early_renewal_hours) or early_renewal_duration from Amazon ACM (https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/acm_certificate#early_renewal_duration)

### Use cases

Per internal security policy I am not allowed to generate an Origin CA cert longer than 1 year, so I must renew it. With the current capabilities I must do a manual change in the certificate to force the replacement (like changing the requested validity)

```hcl
resource "cloudflare_origin_ca_certificate" "example" {
  # use this only if we are using CloudFlare
  provider = cloudflare.user_service_key
  csr = tls_cert_request.example.cert_request_pem
  hostnames = "my-hostname.domain.tld"
  request_type = "origin-rsa"
  requested_validity = 90
  min_validity = 15
}
...
```

### References

_No response_

peruzzof:
This is an attempt to implement #2043.
I based the logic in the hashicorp self signed and AWS providers, I have no previous experience with Go so any suggestion of improvement is welcome.

Closes #2043


peruzzof:
I changed the attribute UpdateContext from Create to Read as the previous was generating uneeded certificates. I could not find any issue with this change as most of the attributes will trigger a new certificate anyway.

peruzzof:
I fixed the merge conflicts and set the Update as Read, as we have the "remaining_days" property that should be set locally, but there is no change to be done upstream. Any property change in a certificate should trigger a renew.
Let me know if I misunderstood anything.

jacobbednarz:
thanks @peruzzof. i updated the schema definition and called out specifically that a Terraform operation has to be run as this isn't a service side option.

jacobbednarz:
i've also just updated the attribute to be `min_days_for_renewal` to align with another resource that we do this with - https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/access_service_token

PR #2051, opened 2022-11-23
DS-12892: Support Workers Analytics Engine bindings

jimhawkridge:
Closes #2058

PR #2059, opened 2022-11-26
Add support for durable objects binding

### LINKED ISSUE #1320: Durable objects?

Whatever runs on app.terraform.io

### Description

I'd like to be able to register durable object classes with terraform, mimicing the `[durable_objects]` from `wrangler.toml`.

### Use cases

Not having to use wrangler and use terraform for DO lifecycle :)

```hcl
... no idea
```

### References

_No response_

afranche:
This PR addresses #1320 and follows upon the work done on https://github.com/cloudflare/cloudflare-go/pull/949! 🙂 

Closes #1320


jacobbednarz (changes requested):
thanks for starting this PR @afranche! unfortunately, we still have a bit of work to do here to get this over the line.

- we need to add the `durable_object_binding` to the worker script schema binding. right now, there is no way to configure/use the new functionality.
- we'll need to fix the tests. right now, they don't pass as i suspect the durable object setup is going to be more involved (looking at https://developers.cloudflare.com/workers/learning/using-durable-objects/)

i have a feeling there may be more API calls involved here too as just testing this out on a dev account i have doesn't have DO working as smoothly as with wrangler.

afranche:
@jacobbednarz Thanks a lot for the review! 🙇🏼 

> we need to add the durable_object_binding to the worker script schema binding. right now, there is no way to configure/use the new functionality.

I'm unsure as to what the worker script schema binding is in this context. My thought process was given that the schema methods for [Create](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/provider/resource_cloudflare_workers_script.go#L117), [Read](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/provider/resource_cloudflare_workers_script.go#L158) and [Update](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/provider/resource_cloudflare_workers_script.go#L272) all have a call to `parseWorkerBindings`, what had to be done to support DOs had to be done from there. If I understand correctly there is more to be done then ?

> we'll need to fix the tests. right now, they don't pass as i suspect the durable object setup is going to be more involved (looking at https://developers.cloudflare.com/workers/learning/using-durable-objects/)

I see for the tests. Based on the link you sent I read that if we're using ES Modules, the main module has to be provided to the DO when creating it hence I'm wondering how well is it working if we're providing the DO without providing the main module with it, that might also be an issue! 😄

> i have a feeling there may be more API calls involved here too as just testing this out on a dev account i have doesn't have DO working as smoothly as with wrangler.

I'd be happy knowing more details about how it worked if you may, and also what routine should be tested regarding how which cases we can stumble upon. 🙂

jacobbednarz:
for the missing schema, check out #2051. it is adding another binding type that should point you in the right direction. 

as for what setup is involved for actual use and tests, I'm chasing down the internal team for clarification. if I can't get an answer in the next couple of days, I'll have to dissect wrangler and see what it is doing under the hood and hope the entire API is documented (or I'm just missing something completely)

davidfernandezm:
Hi all, what happened to this PR? I'm looking at deploying a worker with a Durable Object binding and this is not currently supported... is there any other way to use terraform to bind the Durable Object to the worker?

afranche:
@davidfernandezm Sorry for the late answer. I gave up on Terraform on the worker I was implementing the following for.

> as for what setup is involved for actual use and tests, I'm chasing down the internal team for clarification. if I can't get an answer in the next couple of days, I'll have to dissect wrangler and see what it is doing under the hood and hope the entire API is documented (or I'm just missing something completely)

@jacobbednarz I don't know if you have found an answer regarding this, I'm willing to take back the work on this if you have some answers for me 😄

skyf0l:
We can just give up on durable objects and use wrangler as system D

PR #2093, opened 2022-12-16
fix h1 title

ThomasIAm:
I ran across this page and saw that the H1 title in the doc was wrong. This is my fix.


jacobbednarz (changes requested):
thanks but instead of manually updating this, we should swap it to the autogenerated documentation. check out https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md for how to do this.

PR #2120, opened 2023-01-04
docs: log_to_cloudflare usage notice

satanTime:


docs/resources/zone_settings_override.md
```diff
@@ -72,7 +72,7 @@ These can be specified as "on" or "off" string. Similar to boolean values, but h
 - `http3` (default: `off`)
 - `ip_geolocation` (default: `on`)
 - `ipv6` (default: `off`)
-- `log_to_cloudflare` (default: `off`)
+- `log_to_cloudflare` (default: `off`). Note that it should be provided only if China Network has been enabled for the zone
```
jacobbednarz, line 75: instead of this note, i'd rather we explain what the purpose of the setting is and allow consumers to make their own decisions about this (since `log_to_cloudflare` also isn't required if you're using the China Network).

PR #2121, opened 2023-01-04
docs: visitor_ip usage notice

satanTime:


jacobbednarz:
thanks for the PR, however, this is true for multiple attributes and we call it out broadly at https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/zone_settings_override#plan-dependent-settings. I wouldn't want to add this to all plan dependant settings.

this resource is also quite confusing as it is interpreted as a way to manage all settings however, that isn't the intention. it should only be used for setting values that are not the default. we plan to address this later in 4.x by rearchitecting zone settings.

satanTime:
Hi @jacobbednarz, thanks for the info.

In the case of ENT plan, there are only 2 settings which aren't available out of the box, and for them I've created 2 PRs.

Unfortunately, it's hard to catch [#plan-dependent-settings](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/zone_settings_override#plan-dependent-settings) section, I think that's why customers still add all the settings to their tf, and with this PR, it will be clear that they should skip `visitor_ip` and `log_to_cloudflare`.

PR #2131, opened 2023-01-10
docs: Update cloudflare_list example to use underscore in name

wolfmd:
The Cloudflare API rejects lists that have whitespace characters in their name. This updates the example resources we have to use a name with an underscore instead.

The error I had gotten previously was
```
╷
│ Error: invalid value for name (List name must only contain lowercase letters, numbers and underscores)
│ 
│   with cloudflare_list.example,
│   on cfdata-staging.tf line 3, in resource "cloudflare_list" "example":
│    3:   name        = "example list"
│ 
╵
```


jacobbednarz:
🙇 @wolfmd

PR #2134, opened 2023-01-11
Add support for Cloudflare Queues

jbw1991:
Starting a WIP PR to discuss adding Cloudflare Queues to the terraform provider.

This PR includes:
* The Queue Resource. The provider can now create/update/delete queues.
* Queue Bindings for a workers script: A script can bind to a queue by name

Upcoming work:
* Queue consumer config for a workers script. This is blocked on a "list consumers for script" API endpoint


jbw1991:
Whoops, forgot that this depends on introducing the "Queue Binding" type to cloudflare-go: https://github.com/cloudflare/cloudflare-go/pull/1176

internal/provider/resource_cloudflare_workers_queue.go
```diff
@@ -0,0 +1,131 @@
...
+	err := client.DeleteQueue(ctx, cloudflare.AccountIdentifier(accountID), d.Get("name").(string))
+	if err != nil {
+		return diag.FromErr(errors.Wrap(err, "error deleting workers queue"))
+	}
+
+	d.SetId("")
+	return nil
+}
+
+// TODO(now) is this needed?
```
jbw1991, line 128: @jacobbednarz Hey Jacob, I think I have everything in mostly working order.

I'm a bit unclear on this part (the "import"). I'm not very familiar with terraform, could you point me in the right direction on how to implement this?

jbw1991: I gave this a shot and can seemingly import a Queue resource (provided that I set an account_id on my top level cloudflare provider config).

.changelog/2134.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:new-resource
+cloudflare_workers_queue
```
jacobbednarz, line 2: shouldn't this be just `cloudflare_queue`?

internal/sdkv2provider/resource_cloudflare_workers_queue.go
```diff
@@ -0,0 +1,131 @@
...
+		Description: "Provides the ability to manage Cloudflare Workers Queue features.",
+	}
+}
+
+func resourceCloudflareWorkersQueueCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*cloudflare.API)
+
+	accountID := d.Get("account_id").(string)
+	if accountID == "" {
+		accountID = client.AccountID
```
jacobbednarz, line 33: `client.AccountID` is deprecated. instead, we should enforce `account_id` and rely on that value explicitly.

internal/sdkv2provider/resource_cloudflare_workers_queue.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	queueName := d.Get("name").(string)
+	req := cloudflare.CreateQueueParams{
+		Name: queueName,
+	}
+
+	tflog.Debug(ctx, fmt.Sprintf("[Info] Creating Cloudflare Workers Queue from struct: %+v", req))
```
jacobbednarz, line 41: no need for `[LOG LEVEL]` prefixes here. `tflog` handles those for us.

```suggestion
	tflog.Debug(ctx, fmt.Sprintf("Creating Cloudflare Workers Queue from struct: %+v", req))
```

internal/sdkv2provider/resource_cloudflare_workers_queue.go
```diff
@@ -0,0 +1,131 @@
...
+	return resourceCloudflareWorkersQueueRead(ctx, d, meta)
+}
+
+func resourceCloudflareWorkersQueueRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*cloudflare.API)
+	queueID := d.Id()
+
+	accountID := d.Get("account_id").(string)
+	if accountID == "" {
+		accountID = client.AccountID
```
jacobbednarz, line 65: similar to above, this should be removed in favour of the explicit value.

internal/sdkv2provider/resource_cloudflare_workers_queue.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	resp, _, err := client.ListQueues(ctx, cloudflare.AccountIdentifier(accountID), cloudflare.ListQueuesParams{})
```
jacobbednarz, line 68: any reason why we don't have a `GetQueue` style method for use here?

internal/sdkv2provider/schema_cloudflare_workers_queue.go
```diff
@@ -0,0 +1,19 @@
+package sdkv2provider
+
+import "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+func resourceCloudflareWorkersQueueSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"account_id": {
+			Description: "The account identifier to target for the resource.",
+			Type:        schema.TypeString,
+			Optional:    true,
```
jacobbednarz, line 10:
```suggestion
			Required:    true,
```

internal/sdkv2provider/resource_cloudflare_workers_queue.go
```diff
@@ -0,0 +1,131 @@
...
+	err := client.DeleteQueue(ctx, cloudflare.AccountIdentifier(accountID), d.Get("name").(string))
+	if err != nil {
+		return diag.FromErr(errors.Wrap(err, "error deleting workers queue"))
+	}
+
+	d.SetId("")
+	return nil
+}
+
+func resourceCloudflareWorkersQueueImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
jacobbednarz, line 128: you'll want to pass in the account ID here to ensure you are targeting the correct account when importing the resource.

internal/sdkv2provider/resource_cloudflare_workers_queue.go
```diff
@@ -0,0 +1,131 @@
...
+func resourceCloudflareWorkersQueueUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*cloudflare.API)
+
+	accountID := d.Get("account_id").(string)
+	if accountID == "" {
+		accountID = client.AccountID
+	}
+
+	// TODO(soon) fix the cloudflare-go UpdateQueue implementation: updating a queue should accept the existing name, as well as the new name. Other properties are read-only.
+	_, err := client.UpdateQueue(ctx, cloudflare.AccountIdentifier(accountID), cloudflare.UpdateQueueParams{
```
jacobbednarz, line 100: this API feels kinda weird. if it's only using the `name` as the identifier, i would expect that changing it would be creating a new resource. in terraform, we can do this with `ForceNew: true` however, on a broader sense, any reason why we don't have a md5 style checksum for IDs?

internal/sdkv2provider/resource_cloudflare_workers_queue.go
jacobbednarz: check out the `access_application` approach for setting these values.

.changelog/2134.txt
jbw1991: A bit of a tough call. The product is indeed named "Cloudflare Queues". But "Queues" lives under the "Workers" tab in Dash and is under the "workers" section of the Cloudflare V4 API (accounts/<id>/workers/queues/<name>).

Your suggestion is probably for the best though, I'll go ahead and switch it to drop the workers prefix.

internal/sdkv2provider/resource_cloudflare_workers_queue.go
jbw1991: Unfortunately, the `GetQueue` API endpoint gets a queue given its *name*. But AFAICT, we only have a Queue's ID here.
(Queues do have a UUID associated with them, but there is currently no API endpoint to fetch by UUID).

internal/sdkv2provider/resource_cloudflare_workers_queue.go
jbw1991: Queues are associated with a UUID, but the current V4 API endpoints all use the queue's name instead...

Currently, the "Update Queue" V4 endpoint (https://api.cloudflare.com/#queue-update-queue) is only capable of changing the queue's name. The existing name is in the URL (/queues/<existing-name>) while the desired new name is in the request body. The Queue's UUID remains unchanged though...

The cloudflare-go `UpdateQueue` function seems to be wrong currently as there is no way to pass the parameters correctly (existing name and queue name).

.changelog/2134.txt
jbw1991: Done

internal/sdkv2provider/resource_cloudflare_workers_queue.go
jbw1991: ✅

jbw1991:
Now blocked on https://github.com/cloudflare/cloudflare-go/pull/1188

jacobbednarz:
this all looks fine now but i'm hitting a DSR issue with the Worker bindings acceptance tests. i'll work through this internally but i think we're ok here.

jacobbednarz:
acceptance tests all good!

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^(TestAccCloudflareWorkerScript_|TestAccCloudflareQueue_)" -count 1 -parallel 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareWorkerScript_Import
--- PASS: TestAccCloudflareWorkerScript_Import (15.20s)
=== RUN   TestAccCloudflareQueue_Basic
=== PAUSE TestAccCloudflareQueue_Basic
=== RUN   TestAccCloudflareWorkerScript_ModuleUpload
=== PAUSE TestAccCloudflareWorkerScript_ModuleUpload
=== CONT  TestAccCloudflareQueue_Basic
--- PASS: TestAccCloudflareQueue_Basic (19.33s)
=== CONT  TestAccCloudflareWorkerScript_ModuleUpload
--- PASS: TestAccCloudflareWorkerScript_ModuleUpload (10.07s)
--- PASS: TestAccCloudflareWorkerScript_MultiScriptEnt (40.30s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	85.218s
```

PR #2135, opened 2023-01-12
resource/cloudflare_argo_tunnel: rename to `cloudflare_tunnel`

jacobbednarz:


jacobbednarz:
```
TF_ACC=1 go test $(go list ./...) -v -run "^TestAccCloudflareTunnelCreate_" -count 1 -parallel 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareTunnelCreate_Basic
--- PASS: TestAccCloudflareTunnelCreate_Basic (7.60s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/provider	8.160s
``

**Comment by AllexVeldman** [comment]
@jacobbednarz Renaming the resource fails in terraform with
```
│ Warning: Missing resource schema from provider
│ 
│ No resource schema found for cloudflare_argo_tunnel when decoding prior
│ state
╵
╷
│ Error: no schema available for module.cf-argo-tunnel.cloudflare_argo_tunnel.argo-tunnel while reading state; this is a bug in Terraform and should be reported
│ 
│ 
```

As there is no version with both `cloudflare_argo_tunnel` and `cloudflare_tunnel` the only upgrade path we have is remove existing tunnel on v3, update, re-add tunnel in v4.

See https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-rename.

afreyermuth98:
@jacobbednarz @AllexVeldman @carl-armis 
Maybe we'll have more visibility if we re-open a new issue on this subject ? Because talking on a merged PR doesn't seem to be efficient :/

AllexVeldman:
I luckily am in the position where I could destroy and recreate the tunnel and dependant resources.
But I agree [an Issue](https://github.com/cloudflare/terraform-provider-cloudflare/issues/2598) makes this more visible.

PR #2158, opened 2023-01-13
Added note regarding "partial" setup DNS

wolruf:
It's not obvious for customers that "partial" setup type will fail unless the account team or customer success added a specific enablement to the zone.


jacobbednarz (changes requested):
thanks for the PR however the docs directory is a generated artifact. check out https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md for how to make changes to the documentation. 

the second part of this is that I don't think this sort of information belongs in the Terraform repository. we aim to keep the source of product truth in the developer docs (such as https://developers.cloudflare.com/dns/zone-setups/partial-setup/setup/). in this repository, we generally try to document the syntax and how features are used in Terraform.

PR #2174, opened 2023-01-23
Changing associated_hostnames from list to set for cloudflare_access_mutual_tls_certificate

mlhnono68:
The `associated_hostnames` attribute of the `cloudflare_access_mutual_tls_certificate` resource is currently ordered whereas it should not be in my opinion. There is no consistency in the ordering of the hostnames that is returned by APIs. Therefore, shouldn't we replace the list by a set ?

### DIFF
--- a/internal/sdkv2provider/schema_cloudflare_access_mutual_tls_certificate.go
+++ b/internal/sdkv2provider/schema_cloudflare_access_mutual_tls_certificate.go
@@ -32,7 +32,7 @@ func resourceCloudflareAccessMutualTLSCertificateSchema() map[string]*schema.Sch
 			Description: "The Root CA for your certificates.",
 		},
 		"associated_hostnames": {
-			Type:     schema.TypeList,
+			Type:     schema.TypeSet,
 			Optional: true,
 			Elem: &schema.Schema{
 				Type: schema.TypeString,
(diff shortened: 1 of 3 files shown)


internal/sdkv2provider/schema_cloudflare_access_mutual_tls_certificate.go
```diff
@@ -32,7 +32,7 @@ func resourceCloudflareAccessMutualTLSCertificateSchema() map[string]*schema.Sch
 			Description: "The Root CA for your certificates.",
 		},
 		"associated_hostnames": {
-			Type:     schema.TypeList,
```
jacobbednarz, line 35: this doesn't "just work". you have to update the usage references in the CRUD methods otherwise you get a type error as the iterations are accessed differently.

jacobbednarz (changes requested):
> There is no consistency in the ordering of the hostnames that is returned by APIs. Therefore, shouldn't we replace the list by a set ?

it's less about the assumptions and more about why it was originally done that way. for example, in another case, we had multiple CAs issuing certificates and Cloudflare was the middle operator, these needed to be a particular way per CA (which varies between customers usage). I don't know if any such thing exists here, however, we need to confirm before we blindly accept the changes. 

can you elaborate on the issue you're hitting here? or is this purely a "it looks wrong" scenario and not causing any actual problem?

PR #2178, opened 2023-01-24
feat: Logpush for worker script

bjorand:


bjorand:
@jacobbednarz do we have a clean way of reading the logpush setting for a worker? I can't find anything in the API or in the SDK.
I see the logpush status in `/accounts/<account_id>/workers/services/<script_name>` but I don't know if it is the right way to retrieve it.

docs/resources/worker_script.md
```diff
@@ -83,6 +83,8 @@ resource "cloudflare_worker_script" "my_script" {
 - `secret_text_binding` (Block Set) (see [below for nested schema](#nestedblock--secret_text_binding))
 - `service_binding` (Block Set) (see [below for nested schema](#nestedblock--service_binding))
 - `webassembly_binding` (Block Set) (see [below for nested schema](#nestedblock--webassembly_binding))
+- `logpush` (Boolean) Whether to enable [Logpush](https://developers.cloudflare.com/workers/platform/logpush/)
```
jacobbednarz, line 86: this is a generated artifact. check out https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md for how to update these files.

jacobbednarz:
> @jacobbednarz do we have a clean way of reading the logpush setting for a worker? I can't find anything in the API or in the SDK.
I see the logpush status in /accounts/<account_id>/workers/services/<script_name> but I don't know if it is the right way to retrieve it.

not to my knowledge. if it doesn't exist in the SDK/API docs, we won't add support for it here as the stability isn't yet ready for public use.

kamilkisiela:
Any chance to get this in?

PR #2210, opened 2023-02-01
Add allowed_match_count in DLP Profiles

pierreca:
Corresponding cloudflare-go PR: https://github.com/cloudflare/cloudflare-go/pull/1193
released with v0.60.0


internal/sdkv2provider/resource_cloudflare_dlp_profile.go
```diff
@@ -161,6 +163,8 @@ func resourceCloudflareDLPProfileUpdate(ctx context.Context, d *schema.ResourceD
 		Type: d.Get("type").(string),
 	}
 	updatedDLPProfile.Description, _ = d.Get("description").(string)
+	updatedDLPProfile.AllowedMatchCount, _ = d.Get("allowed_match_count").(uint)
```
jacobbednarz, line 166: similar to https://github.com/cloudflare/terraform-provider-cloudflare/pull/2210/files#r1093789250

internal/sdkv2provider/resource_cloudflare_dlp_profile_test.go
```diff
@@ -71,12 +73,13 @@ func TestAccCloudflareDLPProfile_Custom_MultipleEntries(t *testing.T) {
...
 }
 
-func testAccCloudflareDLPProfileConfigCustom(accountID, rnd, description string) string {
+func testAccCloudflareDLPProfileConfigCustom(accountID, rnd, description string, allowedMatchCount uint) string {
 	return fmt.Sprintf(`
 resource "cloudflare_dlp_profile" "%[1]s" {
   account_id                = "%[3]s"
   name                      = "%[1]s"
   description               = "%[2]s"
+  allowed_match_count       = %[4]d
```
jacobbednarz, line 82: i don't recommend allowing 0 here if you can avoid it. we're working on a larger body of work for `cloudflare_ruleset` related to zero values in terraform and this will invoke the same issues.

pierreca: 0 is a valid value, and means that the rule will immediately be triggered (rather than after a certain number of hits). What's wrong with 0 values in terraform? should it be "omitted if 0, >=1 if specified" ?

jacobbednarz: you can read the full details in the core issue i've had open - https://github.com/hashicorp/terraform-plugin-sdk/issues/817. while the issue title is about booleans, this impacts all zero values.

pierreca: thanks for the context! it looks like this is an issue is when an uninitialized value is a "valid" configuration choice and should not be sent in the payload? in our case, uninitialized would be treated as 0, so uninitialized and zero value are the same. is that ok or am I missing something?

jacobbednarz: > so uninitialized and zero value are the same

that is the crux of the issue. terraform cannot differentiate between the two and will often send and save them interchangeably which can cause issues with the service.

pierreca: right. I know nothing about what methods are used by terraform to differentiate "create" and "update" operations  but I could see a problematic case if we have a value that is set, and gets overwritten when it's omitted because it's assumed to be 0 rather than left untouched.

here's the desired behavior:
- if creating a new profile:
  - If the value is absent, it'll be set to 0 server-side
  - if set to 0 or any other value between 0 and 1000, it should be set to that value (the upper bound may change in the future, the lower bound will always be 0)
- if updating an existing profile
  - if the value is absent, preserve the existing one
  - if the value is set, overwrite the existing one

with my current understanding of things, it seems as though the update operation could be the tricky one, and if that's not the default behavior, maybe we can document somehow to always do a read before a write and always specify the value?

from the service perspective, 0 is a valid value, I don't think that's something that can be redesigned and I don't see how we'd do it. 

from an code cleanliness perspective, i'm starting to regret the choice of `uint` in cloudflare-go because that just makes the terraform code more complicated for no good reason. but that's unrelated

jacobbednarz: if you always send the value (aka, a required field) and expect it this should be 👌 the problems only start coming about if you need to conditionally send/set it.

pierreca: given it's a new field in an existing structure, wouldn't it be a breaking change to make it required?

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.19
 require (
 	github.com/agext/levenshtein v1.2.3 // indirect
 	github.com/apparentlymart/go-cidr v1.1.0 // indirect
-	github.com/cloudflare/cloudflare-go v0.60.0
```
jacobbednarz, line 8: you can remove this version bump. we have automation that does this and once the new version lands, this PR can just be synced.

.changelog/2210.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+dlp_profile: Add new allowed_match_count field to profiles
```
jacobbednarz, line 2:
```suggestion
resource/cloudflare_dlp_profile: Add new `allowed_match_count` field to profiles
```

jacobbednarz:
acceptance tests all passing

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareDLPProfile_" -count 1 -parallel 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareDLPProfile_Custom
--- PASS: TestAccCloudflareDLPProfile_Custom (10.72s)
=== RUN   TestAccCloudflareDLPProfile_Custom_MultipleEntries
--- PASS: TestAccCloudflareDLPProfile_Custom_MultipleEntries (9.35s)
=== RUN   TestAccCloudflareDLPProfile_CustomWithAllowedMatchCount
--- PASS: TestAccCloudflareDLPProfile_CustomWithAllowedMatchCount (9.69s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	30.330s
```

Issue #1917, opened 2022-09-19
notification policy import: only last email gets saved in tfstate


Done203:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.2.9
on windows_386
provider registry.terraform.io/cloudflare/cloudflare v3.23.0

### Affected resource(s)

terraform import cloudflare_notification_policy.<ressource_name> <account_id>/<policy_id>
cloudflare_notification_policy

### Terraform configuration files

```hcl
resource "cloudflare_notification_policy" "example_name" {
  account_id = "<account_id>"
  alert_type = "health_check_status_notification"
  enabled    = true
  name       = "example_name"

  description = ""
  email_integration {
    id = "email_1@company.com"
  }
  email_integration {
    id = "email_2@company.com"
  }
  filters {
    health_check_id = [<healthcheck_id>]
    status          = ["Healthy", "Unhealthy"]
  }
}
```

### Link to debug output

https://gist.github.com/Done203/14d59eba0dd291c7936ad611d107b9cc

### Expected output

taken from terraform.tfstate:

{
  "version": 4,
  "terraform_version": "1.2.9",
  "serial": 1,
  "lineage": "7171fcc1-5e00-ae83-140e-8c28c867eafb",
  "outputs": {},
  "resources": [
    {
      "mode": "managed",
      "type": "cloudflare_notification_policy",
      "name": "shop_COMPANY_com",
      "provider": "provider[\"registry.terraform.io/cloudflare/cloudflare\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "account_id": "<account_id>",
            "alert_type": "health_check_status_notification",
            "created": "2022-07-08T12:44:02Z",
            "description": "",
            **"email_integration": [
              {
                "id": "email_1@COMPANY.com",
                "name": ""
              }
            "email_integration": [
              {
                "id": "email_2@COMPANY.com",
                "name": ""
              }**
            ],
            "enabled": true,
            "filters": [
              {
                "enabled": [],
                "event_source": [],
                "event_type": [],
                "health_check_id": [
                  "<healthcheck_id_1>",
                  "<healthcheck_id_2>"
                ],
                "input_id": [],
                "limit": [],
                "new_health": [],
                "packets_per_second": [],
                "pool_id": [],
                "product": [],
                "protocol": [],
[...]

PR #2248, opened 2023-02-21
Fix multiple emails in notification policy

Qantas94Heavy:
Setting `data["id"]` only saves the last value as it keeps writing to the same map. Instead a new map should be created for each value.

Fixes: https://github.com/cloudflare/terraform-provider-cloudflare/issues/1917

### DIFF
--- a/.changelog/2248.txt
+++ b/.changelog/2248.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_notification_policy: ensure all emails are saved if multiple `email_integration` values specified
+```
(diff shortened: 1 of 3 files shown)


jacobbednarz:
thanks for this @Qantas94Heavy. can you please add an acceptance test that will demonstrate this fixes the issue?

jacobbednarz:
acceptance tests all looking good

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareNotificationPolicy_" -timeout 120m -parallel 1
=== RUN   TestAccCloudflareNotificationPolicy_Basic
--- PASS: TestAccCloudflareNotificationPolicy_Basic (26.83s)
=== RUN   TestAccCloudflareNotificationPolicy_WithFiltersAttribute
--- PASS: TestAccCloudflareNotificationPolicy_WithFiltersAttribute (24.59s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	52.330s
```

thanks @Qantas94Heavy 💯

PR #2255, opened 2023-02-23
CUSTESC-25878 added audit_ssh to gateway actions

rexscaria:
Adding anew action to gateway actions and a little more documentation on existing actions


jacobbednarz:
thanks for the update here @rexscaria but the files you've updated are generated artifacts. you can have a read of https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md to see how to make changes to the available fields.

PR #2267, opened 2023-03-01
DLP-810: Add support for DLP payload logging

OrionNebula:
Adds support for the DLP payload logging feature, both enabling it for an individual rule and setting the account-level public key setting. Both new options can be added into existing resources. While you need an entitlement to _set_ the public key, you don't need one to read it. Setting the key isn't attempted unless the `public_key {}` block is specified on the `cloudflare_teams_account` resource, so this shouldn't be a breaking change.

### DIFF
--- a/.changelog/2267.txt
+++ b/.changelog/2267.txt
@@ -1,0 +1,7 @@
+```release-note:enhancement
+resource/cloudflare_teams_account: Add support for DLP payload logging public key
+```
+
+```release-note:enhancement
+resource/cloudflare_teams_rule: Add support for enabling DLP payload logging per-rule
+```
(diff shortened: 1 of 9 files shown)


jacobbednarz:
acceptance tests all looking good

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^(TestAccCloudflareTeamsAccount|TestAccCloudflareTeamsRuleBasic)" -timeout 120m -parallel 1
=== RUN   TestAccCloudflareTeamsAccountConfigurationBasic
--- PASS: TestAccCloudflareTeamsAccountConfigurationBasic (18.26s)
=== RUN   TestAccCloudflareTeamsRuleBasic
--- PASS: TestAccCloudflareTeamsRuleBasic (12.74s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	31.448s
```

jacobbednarz:
thanks @OrionNebula! you rock 🤘

rodmatos:
Hi @jacobbednarz and @OrionNebula, this pull request seems to introduce a regression for the resource type `cloudflare_teams_account`. 
I do not have DLP payload logging enabled and I am not providing any DLP-related configuration to this resource but I am faced with the following error: `Error: error finding DLP Account config "<account_id>": not found (3307).

OrionNebula:
Hi @rodmatos, thanks for pointing this out and sorry for the trouble! I think this is due to the fact that the payload log settings endpoint will return 404 if you've never configured it before, but there's no check for this case in the read function for the resource. I'll try to have a fix for this today.

rodmatos:
Super, thank you so much @OrionNebula, let me know if you need some request logs or anything else from my side :-)

Issue #2225, opened 2023-02-08
Impossible to not specify an empty `cookie` block `cloudflare_page_rule` since `3.34.0`


lra:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.3.7
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v3.34.0

### Affected resource(s)

- cloudflare_page_rule

### Terraform configuration files

```hcl
resource "cloudflare_page_rule" "page_rule" {
    id       = "..."
    priority = 29
    status   = "active"
    target   = "..."
    zone_id  = "..."

    actions {
        always_use_https     = false
        browser_cache_ttl    = "1800"
        cache_level          = "bypass"
        disable_apps         = false
        disable_performance  = false
        disable_railgun      = false
        disable_security     = false
        disable_zaraz        = false
        edge_cache_ttl       = 604800
        host_header_override = "smcqa01eastasiacdn.blob.core.windows.net"

        cache_key_fields {

            header {
                check_presence = []
                exclude        = [
                    "origin",
                ]
                include        = []
            }

            host {
                resolved = false
            }

            query_string {
                exclude = []
                ignore  = true
                include = []
            }

            user {
                device_type = false
                geo         = false
                lang        = false
            }
        }
    }
}
```

### Link to debug output

https://gist.github.com/lra/5241499210278c1c1e8aca383529ebf3

### Expected output

No changes. Your infrastructure matches the configuration.

### Actual output

Terraform will perform the following actions:

  # module.seismic.cloudflare_page_rule.page_rule["zone_seismic-dev.com_rule_smc-qa01-cdn-IE"] will be updated in-place
  ~ resource "cloudflare_page_rule" "page_rule" {
        id       = "..."
        # (4 unchanged attributes hidden)

      ~ actions {
            # (10 unchanged attributes hidden)

          ~ cache_key_fields {
              + cookie {
                  + check_presence = (known after apply)
                  + include        = (known after apply)
                }

                # (4 unchanged blocks hidden)
            }
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.

### Steps to reproduce

1. Don't define a cookie block under `cloudflare_page_rule.cache_key_fields` or define an empty cookie block.
2. Run terraform apply
3. Run terraform plan -> you get a drift

[...]

PR #2269, opened 2023-03-02
Update doc for resource page_rule

junze-smg:
Today I bumped into https://github.com/cloudflare/terraform-provider-cloudflare/issues/2225 and according to https://github.com/cloudflare/terraform-provider-cloudflare/issues/2225#issuecomment-1427150424, the documentation should be updated, so here's my PR to avoid further confusions.

### DIFF
--- a/docs/resources/page_rule.md
+++ b/docs/resources/page_rule.md
@@ -129,10 +129,10 @@ resource "cloudflare_page_rule" "foobar" {
 }
 ```
 
-- `cookie` - (Required, but allowed to be empty) Controls what cookies go into Cache Key:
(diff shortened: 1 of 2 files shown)


junze-smg:
hey @jacobbednarz it looks like i don't have the permission to add the `workflow/skip-changelog-entry` label to this PR. could you please add it? Thanks!

jacobbednarz:
> it looks like i don't have the permission to add the `workflow/skip-changelog-entry` label to this PR.

yep, that's expected 👍 it's a maintainer only thing to prevent people trying to skip adding entries.

cheers for updating this one!

PR #2272, opened 2023-03-06
tests: swap to provider agnostic test plugin

jacobbednarz:
`github.com/hashicorp/terraform-plugin-testing` is backwards compatible with SDKv2, just has a new name and is split from the v2 provider


jacobbednarz:
blocked on hashicorp/terraform-plugin-testing#93 pending the next steps to resolve this one

Issue #2278, opened 2023-03-09
Support `sentinelone` type in the `cloudflare_device_posture_rule` schema


aborrello:
### Current Terraform and Cloudflare provider version

Terraform version: 1.2.3
Cloudflare provider version: 3.30.0

### Description

One undocumented component of the device posture API is the "support" for SentinelOne. This is currently implemented as the `sentinelone` type, which I believe itself is an alias for the `application` type. This request is to add `sentinelone` to the `cloudflare_device_posture_rule` schema validator.

A key difference is that existing checks created through the user interface already have the `sentinelone` type associated with them, making it difficult to import these rules into Terraform definitions. Additionally, the WARP Client UI displays SentinelOne checks and application checks differently, however, I suspect this is purely aesthetic.

### Use cases

If a team has an existing WARP client posture rule for SentinelOne, it is challenging to import the check without creating a new rule entirely (as changing the `type` field will cause the resource to be destroyed and recreated). The ideal end state would be to allow Terraform definitions of the `sentinelone` type, bringing the provider in parity with the UI for this specific rule.

### Potential Terraform configuration

```hcl
resource "cloudflare_device_posture_rule" "edr" {
  account_id  = "123456789"
  name        = "My name"
  description = "My description"

  type     = "sentinelone"
  schedule = "5m"

  match {
    platform = "linux"
  }

  input {
    running    = true
    path       = "/path/to/sentineld"
    thumbprint = "xxx"
  }
}
```

PR #2279, opened 2023-03-09
Support `sentinelone` type in the `cloudflare_device_posture_rule` schema

aborrello:
Follow up PR for issue https://github.com/cloudflare/terraform-provider-cloudflare/issues/2278.

Adds support for the `sentinelone` device posture rule type (which is undocumented, but used by some posture rules created in the dashboard). No additional `input` schema modifications need to be made as `sentinelone` appears to be an alias for / accept the same parameters as the `application` type.

### DIFF
--- a/.changelog/2279.txt
+++ b/.changelog/2279.txt
@@ -1,3 +1,3 @@
 ```release-note:enhancement
-resource/cloudflare_access_device_posture_rule: Add support for `sentinelone` type
+resource/cloudflare_device_posture_rule: Add support for `sentinelone` type.
 ```
--- a/.changelog/2279.txt
+++ b/.changelog/2279.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/cloudflare_access_device_posture_rule: Add support for `sentinelone` type
+```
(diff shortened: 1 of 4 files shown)


jacobbednarz:
thanks for the issue and subsequent PR! i've pinged the team internally to confirm the stabilty of `sentinelone` as a configuration option here. i can't find anything in our docs which suggests this application type isn't yet stable enough to be relied upon public SDKs (and [why this is important](https://github.com/cloudflare/cloudflare-go/blob/master/docs/public-api-documentation.md)). device posture rules are moving fast which may also mean i just don't have the search fu to find the updated documentation for this.

i'll mark this one as pending for now but hopefully have an answer soon that we can move this PR in either direction.

in the meantime, if this is something you need outside of the main line, there are [docs on running a custom version](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/using-non-released-versions.md) that can include your patch.

jacobbednarz:
we don't have documentation for this so the team will need to get those out before we land this one. the team mentioned they may be able to get to this next week or so.

PR #2290, opened 2023-03-15
Add support for Address Map resource

hunts:
Adds support for the [Proxy IPs](https://developers.cloudflare.com/byoip/proxy-ips/) feature, allow CRUD address maps.


internal/sdkv2provider/provider.go
```diff
@@ -357,7 +359,11 @@ func configure(version string, p *schema.Provider) func(context.Context, *schema
 		retryOpt := cloudflare.UsingRetryPolicy(int(retries), int(minBackOff), int(maxBackOff))
 		options := []cloudflare.Option{limitOpt, retryOpt, baseURL}
 
-		options = append(options, cloudflare.Debug(logging.IsDebugOrHigher()))
+		if logging.IsDebugOrHigher() {
```
jacobbednarz, line 362: does this need to stay? i can't say i've used it so unsure what the goal is here.

internal/sdkv2provider/resource_cloudflare_address_map_test.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/consts"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccCloudflareAddressMap(t *testing.T) {
```
jacobbednarz, line 12: just waiting to get this all hooked up on the acceptance testing account so these can run

internal/sdkv2provider/schema_cloudflare_address_map.go
```diff
@@ -0,0 +1,81 @@
...
+		},
+		"ips": {
+			Type:        schema.TypeSet,
+			Optional:    true,
+			Description: "The set of IPs on the Address Map.",
+			Elem: &schema.Resource{
+				SchemaVersion: 1,
+				Schema: map[string]*schema.Schema{
+					"ip": {
+						Type:         schema.TypeString,
```
jacobbednarz, line 49: will want to add a `Description` here (and to the other missing fields)

internal/sdkv2provider/schema_cloudflare_address_map.go
```diff
@@ -0,0 +1,81 @@
...
+		},
+		"memberships": {
+			Type:        schema.TypeSet,
+			Optional:    true,
+			Description: "Zones and Accounts which will be assigned IPs on this Address Map.",
+			Elem: &schema.Resource{
+				SchemaVersion: 1,
+				Schema: map[string]*schema.Schema{
+					"identifier": {
+						Type:     schema.TypeString,
```
jacobbednarz, line 64: same as https://github.com/cloudflare/terraform-provider-cloudflare/pull/2290/files#r1136505878

internal/sdkv2provider/schema_cloudflare_address_map.go
```diff
@@ -0,0 +1,81 @@
...
+						Required: true,
+					},
+					"kind": {
+						Type:         schema.TypeString,
+						Required:     true,
+						ValidateFunc: validation.StringInSlice([]string{"account", "zone"}, false),
```
jacobbednarz, line 70: same as https://github.com/cloudflare/terraform-provider-cloudflare/pull/2290/files#r1136505878

docs/resources/address_map.md
```diff
@@ -0,0 +1,83 @@
+---
```
jacobbednarz, line 1: will re-run this after the schema descriptions are updated.

jacobbednarz:
thanks for this @hunts! we're 95% of the way there. to summarise the inline comments:

- we need to update all fields to have a `Description` field.
- get the acceptance testing hooked up with the correct entitlements so we can run these tests.
- add a CHANGELOG entry 

after that, we should be good to go here.

internal/sdkv2provider/provider.go
hunts: I'm not sure about the `cloudflare.Debug()` thing, what I'd like to add is the HTTP logging capability, it's helpful on figuring out what goes wrong by looking at the request and response payloads.

jacobbednarz: gotcha, so the existing functionality already handles this. 

[`cloudflare.Debug()`](https://github.com/cloudflare/cloudflare-go/blob/64af3d2566da4be7093bc20562c92709628c42fa/options.go#L97) logs the HTTP transaction with some [sensitive values redacted](https://github.com/cloudflare/cloudflare-go/blob/64af3d2566da4be7093bc20562c92709628c42fa/cloudflare.go#L386-L414). the tricky bit to see here is that `logging.IsDebugOrHigher()` returns a boolean based on the setting provided to `TF_LOG`. if you want debugging enabled, you use the environment variable `TF_LOG=DEBUG terraform apply ...` to see all the logs. you can also control the [formats and flow using some other environment variables](https://developer.hashicorp.com/terraform/plugin/log/managing).

jacobbednarz: in short, i think we can revert this and still get all the debugging goodies you are after here.

hunts: Awesome, dropped the commit. I think I tried `TF_LOG_PROVIDER=DEBUG` and it gave nothing about the HTTP payloads, thought there is no HTTP logging in the provider. Apparently, it is related to `TF_LOG`.

jacobbednarz:
acceptance tests all passing

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareAddressMap" -timeout 120m -parallel 1
=== RUN   TestAccCloudflareAddressMap
--- PASS: TestAccCloudflareAddressMap (37.29s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	38.320s
```

Issue #2283, opened 2023-03-13
List Item Resource


rwhelan:
### Current Terraform and Cloudflare provider version

We need to be able to manage List Items individually

### Description

We would like to create a `cloudflare_list_item` resource type.

As a pre-req, we'd would also need a `cloudflare_list` data source.  This would allow us to look up the list id needed to create individual list items.

I started roughing out the data source, which would need to be followed by the list item resource type.

https://github.com/rwhelan/terraform-provider-cloudflare/compare/0f9af804800f8ea3be14455a683756d0d1336bb0...8795bc9fc8e5b8dd8c1e0cae101762da950cb5fc

(the above diff doesn't support filtering to a single list, and returns all the lists found in the account.  Its a WIP)

### Use cases

We currently need to manage list items as individual resources.  Having to generate the code needed to include them in a `list` resource is getting cumbersome. 

### Potential Terraform configuration

```hcl
data "cloudflare_list" "customer_redirects" {
  account_id = "01234567890123456789012345678901"
  name = "redirects"
}

resource "cloudflare_list_item" "some_redirect" {
  list_id = data.cloudflare_list.customer_redirects.id
  / * kind = ? * /
  redirect = {
    source_url = ...
    target_url = ...
    ......
  }
}

```

rwhelan:
I'm happy to continue working on this and submit a PR if helpful.  The Contributing Doc said I had to open an issue prior to submitting my PRs.

nboaldin:
I would like this to be implemented as well.

elibarb:
👍

jacobbednarz:
thanks for the suggestion! you're welcome to submit a PR for this if it's something you're after and it doesn't break the backwards compatibility of the provider.

i would recommend breaking this down into a couple of PRs though for sake of review and testing.

rwhelan:
> 

That was my plan.  I was going to submit one for the data source and one for the list_item resource. 

If you think more granulation would be helpful, let me know!

Issue #2294, opened 2023-03-16
`cloudflare_list` inline item


rwhelan:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.4.0
cloudflare provider v4.1.0

### Affected resource(s)

`cloudflare_list`

### Terraform configuration files

```hcl
resource "cloudflare_list" "list_name" {
  account_id  = "01234567890123456789012345678901"
  name        = "test_list"
  kind        = "ip"

// item {
//   value {
//     ip = "192.0.2.1"
//   }
//   comment = "A list item comment"
// }
}
```

### Link to debug output

N/A

### Expected output

The list get updated to have zero items

### Actual output

Any items on the list remain.

### Steps to reproduce

1. Create a list resource with in-line `item {`s
2. Apply the terraform
3. Comment out or remove the all the in-line `item` attributes
4. Apply the terraform

### Additional factoids

Terraform will correctly detect the change and plan to remove the items from the list.  However, on the cloudflare dashboard, the items will remain on the list.

Having zero item attributes will _not_ result in an empty list.  The list will retain however many items it contained prior to being configured with none. 

Subsequent  runs of terraform will continue to detect the drift, but will not clear the list content out.

jacobbednarz:
closed by #2304

PR #2295, opened 2023-03-16
Resource list item

rwhelan:
Resource: `cloudflare_list_item`
Addresses: https://github.com/cloudflare/terraform-provider-cloudflare/issues/2283

New `cloudflare_list_item` resource to allow managing list items independently of (not inline) the `cloudflare_list` resource.

### DIFF
--- a/internal/sdkv2provider/provider.go
+++ b/internal/sdkv2provider/provider.go
@@ -258,6 +258,7 @@ func New(version string) func() *schema.Provider {
 				"cloudflare_zone_lockdown":                          resourceCloudflareZoneLockdown(),
 				"cloudflare_zone_settings_override":                 resourceCloudflareZoneSettingsOverride(),
 				"cloudflare_zone":                                   resourceCloudflareZone(),
+				"cloudflare_list_item":                              resourceCloudflareListItem(),
 			},
 		}
 
--- a/internal/sdkv2provider/resource_cloudflare_list.go
+++ b/internal/sdkv2provider/resource_cloudflare_list.go
@@ -94,6 +94,10 @@ func resourceCloudflareListRead(ctx context.Context, d *schema.ResourceData, met
 	d.Set("description", list.Description)
 	d.Set("kind", list.Kind)
 
+	if d.Get("ignore_inline_items").(bool) {
+		return nil
+	}
+
 	items, err := client.ListListItems(ctx, cloudflare.AccountIdentifier(accountID), cloudflare.ListListItemsParams{
 		ID: d.Id(),
 	})
--- a/internal/sdkv2provider/resource_cloudflare_list_item.go
+++ b/internal/sdkv2provider/resource_cloudflare_list_item.go
@@ -1,0 +1,217 @@
+package sdkv2provider
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/MakeNowJust/heredoc/v2"
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/consts"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/pkg/errors"
+)
+
+func resourceCloudflareListItem() *schema.Resource {
+	return &schema.Resource{
+		Schema:        resourceCloudflareListItemSchema(),
+		CreateContext: resourceCloudflareListItemCreate,
+		ReadContext:   resourceCloudflareListItemRead,
+		UpdateContext: resourceCloudflareListItemUpdate,
+		DeleteContext: resourceCloudflareListItemDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceCloudflareListItemImport,
+		},
+		Description: heredoc.Doc(`
+			Provides individual list items (IPs, Redirects) to be used in Edge Rules Engine
+			across all zones within the same account.
+		`),
+	}
+}
+
+func resourceCloudflareListItemCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*cloudflare.API)
+	accountID := d.Get(consts.AccountIDSchemaKey).(string)
+	listID := d.Get("list_id").(string)
+	listItemType := listItemType(d)
+
+	list, err := client.GetList(ctx, cloudflare.AccountIdentifier(accountID), listID)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("unable to find list with id %s: %w", listID, err))
+	}
+
+	if list.Kind != listItemType {
+		return diag.FromErr(fmt.Errorf("items of type %s can not be added to lists of type %s", listItemType, list.Kind))
+	}
+
+	createListItemResponse, err := client.CreateListItem(ctx, cloudflare.AccountIdentifier(accountID), cloudflare.ListCreateItemParams{
+		ID:   listID,
+		Item: buildListItemCreateRequest(d),
+	})
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("unable to create list item on list id %s: %w", listID, err))
+	}
+
+	newestItem := mostRecentlyCreatedItem(createListItemResponse)
+	d.SetId(newestItem.ID)
+
+	return nil
+}
+
+func resourceCloudflareListItemImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	attributes := strings.SplitN(d.Id(), "/", 3)
+
+	if len(attributes) != 3 {
+		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"accountID/listID/itemID\"", d.Id())
+	}
+
+	accountID, listID, itemID := attributes[0], attributes[1], attributes[2]
+	d.SetId(itemID)
+	d.Set(consts.AccountIDSchemaKey, accountID)
+	d.Set("list_id", listID)
+
+	resourceCloudflareListItemRead(ctx, d, meta)
+
+	return []*schema.ResourceData{d}, nil
+}
+
+func resourceCloudflareListItemRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*cloudflare.API)
+	accountID := d.Get(consts.AccountIDSchemaKey).(string)
+	listID := d.Get("list_id").(string)
+
+	listItem, err := client.GetListItem(ctx, cloudflare.AccountIdentifier(accountID), listID, d.Id())
+	if err != nil {
+		if strings.Contains(err.Error(), "could not find item") {
+			tflog.Info(ctx, fmt.Sprintf("List item %s no longer exists", d.Id()))
+			d.SetId("")
+			return nil
+		}
+		return diag.FromErr(errors.Wrap(err, fmt.Sprintf("error reading List Item with ID %q", d.Id())))
+	}
+
+	d.Set("comment", listItem.Comment)
+
+	if listItem.IP != nil {
+		d.Set("ip", *listItem.IP)
+	}
+
+	if listItem.Redirect != nil {
+		optBoolToString := func(b *bool) string {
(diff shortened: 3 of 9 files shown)


rwhelan:
Will create a changelog entry file and publish shortly

.changelog/2295.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:new-resource
+resource/cloudflare_list_item: Create and manage list items as Terraform resources.
```
jacobbednarz, line 2:
```suggestion
cloudflare_list_item
```

internal/sdkv2provider/provider.go
```diff
@@ -258,6 +258,7 @@ func New(version string) func() *schema.Provider {
 				"cloudflare_zone_lockdown":                          resourceCloudflareZoneLockdown(),
 				"cloudflare_zone_settings_override":                 resourceCloudflareZoneSettingsOverride(),
 				"cloudflare_zone":                                   resourceCloudflareZone(),
+				"cloudflare_list_item":                              resourceCloudflareListItem(),
```
jacobbednarz, line 261: let's keep this list alphabetically ordered.

internal/sdkv2provider/resource_cloudflare_list.go
```diff
@@ -94,6 +94,10 @@ func resourceCloudflareListRead(ctx context.Context, d *schema.ResourceData, met
 	d.Set("description", list.Description)
 	d.Set("kind", list.Kind)
 
+	if d.Get("ignore_inline_items").(bool) {
```
jacobbednarz, line 97: this seems unnecessary given that during the operations we have access to the number of items present in the configuration.

internal/sdkv2provider/resource_cloudflare_list_item.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func resourceCloudflareListItemRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*cloudflare.API)
+	accountID := d.Get(consts.AccountIDSchemaKey).(string)
+	listID := d.Get("list_id").(string)
+
+	listItem, err := client.GetListItem(ctx, cloudflare.AccountIdentifier(accountID), listID, d.Id())
+	if err != nil {
+		if strings.Contains(err.Error(), "could not find item") {
```
jacobbednarz, line 88: instead of using string contains matches, let's us the `*cloudflare.NotFoundError` error directly. you can see an example of this at https://github.com/cloudflare/terraform-provider-cloudflare/blob/144b7908512e82e8415d039084d2519221d2008c/internal/sdkv2provider/resource_cloudflare_filter.go#L89-L90

internal/sdkv2provider/resource_cloudflare_list_item.go
```diff
@@ -0,0 +1,217 @@
...
+		d.Set("subpath_matching", optBoolToString(listItem.Redirect.SubpathMatching))
+		d.Set("preserve_path_suffix", optBoolToString(listItem.Redirect.PreservePathSuffix))
+	}
+
+	return nil
+}
+
+func resourceCloudflareListItemUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	// The API does not currently support in-place update of list items
+	dr := resourceCloudflareListItemDelete(ctx, d, meta)
```
jacobbednarz, line 129: this is probably a bad idea because there will be a period of time where there will be no item in the list when folks are expecting it to be. instead, i would recommend we make the schema for `ip` and `redirect` to force new resources instead as that will defer the decision to the end user.

internal/sdkv2provider/resource_cloudflare_list_item.go
```diff
@@ -0,0 +1,217 @@
...
+	request := cloudflare.ListItemCreateRequest{
+		Comment: d.Get("comment").(string),
+	}
+
+	if itemType == "ip" {
+		request.IP = cloudflare.StringPtr(d.Get("ip").(string))
+		return request
+	}
+
+	// itemType must be 'redirect'
```
jacobbednarz, line 180: let's remove this comment as it doesn't seem to be adding any value here. generally, we follow the guidelines at https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/ for when to keep comments in code.

internal/sdkv2provider/resource_cloudflare_list_item_test.go
```diff
@@ -0,0 +1,154 @@
...
+					resource.TestCheckResourceAttr(name, "comment", rnd+"-updated"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccCloudflareListItem_BadListItemType(t *testing.T) {
+	rnd := generateRandomResourceName()
+	// name := fmt.Sprintf("cloudflare_list_item.%s", rnd)
```
jacobbednarz, line 76: if this isn't used, let's remove it.

internal/sdkv2provider/resource_cloudflare_list_item_test.go
```diff
@@ -0,0 +1,154 @@
...
+				Config:      testAccCheckCloudflareBadListItemType(rnd, rnd, rnd, accountID),
+				ExpectError: regexp.MustCompile(" can not be added to lists of type "),
+			},
+		},
+	})
+}
+
+func testAccCheckCloudflareIPListItem(ID, name, comment, accountID string) string {
+	return fmt.Sprintf(`
+  resource "cloudflare_list" "list_name" {
```
jacobbednarz, line 96:
```suggestion
  resource "cloudflare_list" "%[1]s" {
```
it's a good practice to make these dynamic to avoid potential collisions.

internal/sdkv2provider/resource_cloudflare_list_item_test.go
```diff
@@ -0,0 +1,154 @@
...
+    account_id          = "%[4]s"
+    name                = "%[2]s"
+    description         = "list named %[2]s"
+    kind                = "ip"
+	ignore_inline_items = true
+  }
+  
+  resource "cloudflare_list_item" "%[1]s" {
+    account_id = "%[4]s"
+	list_id    = cloudflare_list.list_name.id
```
jacobbednarz, line 106:
```suggestion
	list_id    = cloudflare_list.%[1]s.id
```

internal/sdkv2provider/resource_cloudflare_list_item_test.go
```diff
@@ -0,0 +1,154 @@
...
+	ip         = "192.0.2.0"
+	comment    = "%[3]s"
+  } `, ID, name, comment, accountID)
+}
+
+func testAccCheckCloudflareListItemExists(n string, listItem *cloudflare.ListItem) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		accountID := os.Getenv("CLOUDFLARE_ACCOUNT_ID")
+		listRS := s.RootModule().Resources["cloudflare_list.list_name"]
```
jacobbednarz, line 115: you'll need to pass in the random ID value to use it here too.

internal/sdkv2provider/resource_cloudflare_list_item_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+		*listItem = foundList
+
+		return nil
+	}
+}
+
+func testAccCheckCloudflareBadListItemType(ID, name, comment, accountID string) string {
+	return fmt.Sprintf(`
+  resource "cloudflare_list" "list_name" {
```
jacobbednarz, line 140:
```suggestion
  resource "cloudflare_list" "%[1]s" {
```

internal/sdkv2provider/resource_cloudflare_list_item_test.go
```diff
@@ -0,0 +1,154 @@
...
+    account_id          = "%[4]s"
+    name                = "%[2]s"
+    description         = "list named %[2]s"
+    kind                = "redirect"
+	ignore_inline_items = true
+  }
+  
+  resource "cloudflare_list_item" "%[1]s" {
+    account_id = "%[4]s"
+	list_id    = cloudflare_list.list_name.id
```
jacobbednarz, line 150:
```suggestion
	list_id    = cloudflare_list.%[1]s.id
```

internal/sdkv2provider/resource_cloudflare_list_test.go
```diff
@@ -38,6 +38,7 @@ func TestAccCloudflareList_Exists(t *testing.T) {
 					resource.TestCheckResourceAttr(
 						name, "name", rnd),
 				),
+				Destroy: false,
```
jacobbednarz, line 41: any reason why we are setting this here? the test should be excercising deletion as a part of the lifecycle flow.

jacobbednarz (changes requested):

internal/sdkv2provider/resource_cloudflare_list.go
rwhelan: Are you suggesting that if the `cloudflare_list` resource do _not_ include any `item` attributes, we assume this is the desired behavior?  I think that would make sense for setting up a new list with externally managed list items, but it would prevent the terraform from being able to remove all the list items from a list which was populated via in-line items.  While [thats the current behavior](https://github.com/cloudflare/terraform-provider-cloudflare/issues/2294), I assume thats a bug as every terraform plan with an empty in-line item list will plan to remove all the items, which it does not, currently, do.

It would be great to not have to have another attribute on `cloudflare_list`; I feel like I'm being dense; sorry if I'm missing something obvious :-/

jacobbednarz: to better isolate these issues, let's split the two issues here and not conflate them.

the first is focusing on when _this resource_ should make HTTP calls to save the outcome to terraform state and how that is determined. since the `cloudflare_list` resource has a list (terraform type) of `lists`, we should be able to tell whether or not we need to make the GET request for all resources to save it against this resource using `d.Get("list").([]interface{})` and running that though a `len` check. if the value is over 0, it is assumed that we want to fetch all list items for this resource.

secondly, how we manage the list items outside of this resource without blowing them all away and having conflicts. we should be able to reuse the above logic and avoid saving/fetching if the list count is less than zero since that means we're not managing the list items in the `cloudflare_list` resource.

finally, if there is a bug with how we are clearing the API when no list items are present on this resource, that is something we can triage and address. if this issue is getting the way of understanding how to proceed with the resource split, i would recommend tackling that first and then circling back to this one to avoid any overlapping work/bugs.

does that help clarify the intention here?

rwhelan: If I am understanding, you are suggesting we check the number of [item](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/sdkv2provider/schema_cloudflare_list.go#L37)s defined in a `cloudflare_list`.  If it equals 0, we _assume_ we are not managing list items in-line with the list resource.  If I'm understanding that properly, that makes sense... Until the user wants to clear out a list with items managed in line.

I fear I'm missing something fundamental because with the above logic, you could manage list items inline as long as you didn't want to use terraform to delete all list items.  If you use the `cloudflare_list` to deploy items in-line, you'd be unable to use the `cloudflare_list` resource to delete all items as clearing out the in line item definitions would cause the resource to ignore any deployed list items.  It might not be necessary functionality, I think it would be un-intuitive (to clear out the in-line items and have them NOT get deleted)

...unless there is way to inspect state and know that there _was_ `items` defined that are no longer there

Sorry if I'm still not understanding or not being clear in what i'm thinking

jacobbednarz: > If it equals 0, we _assume_ we are not managing list items in-line with the list resource

correct and this assumption is *only* used for whether or not to persist to the state file. 

> [...] Until the user wants to clear out a list with items managed in line.
> ...unless there is way to inspect state and know that there was items defined that are no longer there

for the case where we would want to clear out the existing inline managed rules, you'll need to do a check using `d.GetChange` (similar to https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/sdkv2provider/resource_cloudflare_teams_list.go#L120-L134). this approach will allow you to determine if list items were previously managed and whether you should initate removing the inline managed items. at that point, if a user wanted to manage them separately, they will need to already have the resources ready to apply once the list is empty.

rwhelan: This is the piece i was missing; thank you for the direction.  

Update inbound shortly!

rwhelan:
Unsure how i managed to messup the rebase so badly

PR #2357, opened 2023-04-06
Updated error message on changelog entry creation when curl is not in…

pvitty:
Updated the change log creation script to give the correct error message when curl is not installed.


.changelog/2357.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:internal
+Fixed error message on changelog creation
```
jacobbednarz, line 2:
```suggestion
scripts/generate-changelog-entry: make error message match the executable we are expecting
```

PR #2358, opened 2023-04-06
WIP Support for Cloudflare Queue Consumers as a resource

OilyLime:
Parking this here for now, but will be returned to and somewhat overhauled following some forthcoming API changes.

### DIFF
--- a/internal/sdkv2provider/resource_cloudflare_queue_consumer.go
+++ b/internal/sdkv2provider/resource_cloudflare_queue_consumer.go
@@ -1,0 +1,168 @@
+package sdkv2provider
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/consts"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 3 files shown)


jacobbednarz:
closing this out for now since it hasn't progressed and for new resources, we should be targetting the `terraform-plugin-framework` and avoid a bunch of zero/nil issues within Terraform. do drop me a message if you plan to reopen this one and we can walk through it.

vpctorr:
Hey there! I was wondering if we could expect to see queue consumer anytime soon?

SteffTek:
**BUMP** Because I really need this for my next project. I'll need to create a bunch of queues for each customer and don't have the patience to click though the UI every time (╯°□°)╯︵ ┻━┻

wsierakowski:
We are also adopting queues in our TF-provisioned project and have just encountered the same obstacle. It would be incredibly useful to be able to consistently manage everything from the single Terraform script if the support for a queue consumer was added to the provider.

SteffTek:
@wsierakowski we "fixed" this problem by having a wrangler cli command after terraform, it's not pretty, and we need to delete the queue consumer and reapply it because otherwise it'll crash, but it works... Sadly

wsierakowski:
> @wsierakowski we "fixed" this problem by having a wrangler cli command after terraform, it's not pretty, and we need to delete the queue consumer and reapply it because otherwise it'll crash, but it works... Sadly

We temporarily implemented a script that imperatively provisions the queue using the API https://developers.cloudflare.com/api/operations/queue-v2-create-queue. Hopefully it won't take long to have the ability to manage the entire project's infra with TF!

PR #2381, opened 2023-04-16
Update origin_ca_certificate.md

vincentbockaert:
tls_cert_request's key_algorithm is a read-only attribute, you can't assign a value to it. See: https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/cert_request#read-only


jacobbednarz (changes requested):
thanks for the PR however, the file you've edited is a generated artifact. to modify the contents, you'll need to follow the process from https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md

Issue #2276, opened 2023-03-07
cloudflare_notification_policy does not support Tunnel Health Event filters


DDuarte:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```
Terraform v1.3.9
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v3.34.0
```

### Affected resource(s)

- cloudflare_notification_policy

### Terraform configuration files

```hcl
resource "cloudflare_notification_policy" "tunnel_health_alert" {
  account_id  = local.account_id
  name        = "Tunnel Health Alert"
  description = ""
  enabled     = true
  alert_type  = "tunnel_health_event"

  filters {
    new_status = ["TUNNEL_STATUS_TYPE_DOWN"]
    tunnel_id = null
  }
}
```

### Link to debug output

https://gist.github.com/DDuarte/51e21a8d4a05e73f1151822a536606ab

### Expected output

Importing a `cloudflare_notification_policy` of `alert_type = tunnel_health_event` with filters configured works cleanly.

Filter `new_status` string list and `tunnel_id` needs to be supported. On the API it looks like:

```json
    "filters": {
      "new_status": [
        "TUNNEL_STATUS_TYPE_DOWN"
      ],
      "tunnel_id": null
    }
```

### Actual output

`Error: failed to set filters: Invalid address to set: []string{"filters", "0", "new_status"}`

or

`Error: failed to set filters: Invalid address to set: []string{"filters", "0", "tunnel_id"}`

### Steps to reproduce

1. `terraform import cloudflare_notification_policy.tunnel_health_alert <account_id>/<policy_id>` for a notification policy of type `tunnel_health_event`

### References

- https://developers.cloudflare.com/fundamentals/notifications/notification-available/#tunnel
- https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/tunnel-monitoring/notifications/

DDuarte:
TF_LOG=DEBUG gist added

PR #2390, opened 2023-04-20
Adds support for new_status filter for cloudflare_notification_policy tunnel_health_event

stefanrusu-loctax:
Adds missing filter which is required for creating `tunnel_health_event` policies with `cloudflare_notification_policy`. Fixes https://github.com/cloudflare/terraform-provider-cloudflare/issues/2276.

### DIFF
--- a/.changelog/2390.txt
+++ b/.changelog/2390.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_notification_policy: Fix missing new_status filter required by tunnel_health_event policies
+```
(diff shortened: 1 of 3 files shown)


jacobbednarz:
are you able to point me to where this is mentioned in our public docs? i'm unable to find it.

jacobbednarz:
if we don't have public documentation that generally means either 1) the feature (or value) isn't stable enough yet to add to the SDKs and have an unreadable public contract or 2) the team has missed updating the docs. 

I'll poke the team on this one to check however, the policy is without public documentation, we don't merge and potentially put unstable APIs (or values) in people's hands.

PR #2394, opened 2023-04-24
Add support for max upload params in Logpush jobs

Evesy:
Documented here https://developers.cloudflare.com/logs/get-started/api-configuration/#max-upload-parameters

Requires release of `cloudflare-go` to add library support for this


jacobbednarz:
thanks @Evesy 👏

PR #2409, opened 2023-05-03
Added `auto` to the list of compression algorithms for the `compress_response` action.

orium:


orium:
Created as draft until the service is ready to accept `auto`.

PR #2413, opened 2023-05-06
Fix issues with turnstile widget resource

Cyb3r-Jak3:
Adds region and bot fight mode to be computed with a new unit test to check
Fix issue where widget could not be updated

Fixes #2412

PR #2424, opened 2023-05-09
Add valid values to cloudflare_turnstile_widget.mode doc

### LINKED ISSUE #2417: Turnstile `mode` valid values not documented
### Confirmation

```text
Terraform v1.4.6
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v4.5.0
+ provider registry.terraform.io/hashicorp/aws v4.62.0
```

### Affected resource(s)

- `cloudflare_turnstile_widget`

```hcl
resource "cloudflare_turnstile_widget" "loco_client" {
  account_id = "xxxxxxxxxxxxx"
  name       = "My thing"
  domains    = ["example.com"]

  mode = ?????
}
```

### Link to debug output

n/a

### Panic output

_No response_

### Expected output

I expected to be able to read in the documentation which values are supported for the `mode` value.

### Actual output

It's not documented.

### Steps to reproduce

1. https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/turnstile_widget#mode
2. Observe missing documentation

### Additional factoids

_No response_

### References

https://developers.cloudflare.com/api/operations/accounts-turnstile-widget-create#request-body

Seems like the valid values are:

- `non-interactive`
- `invisible`
- `managed`

LinusU:
Fixes #2417

### DIFF
--- a/docs/resources/turnstile_widget.md
+++ b/docs/resources/turnstile_widget.md
@@ -28,7 +28,7 @@ resource "cloudflare_turnstile_widget" "example" {
 
 - `account_id` (String) The account identifier to target for the resource.
 - `domains` (Set of String) Domains where the widget is deployed
-- `mode` (String) Widget Mode
+- `mode` (String) Widget Mode. Valid values: `"non-interactive"`, `"invisible"`, or `"managed"`.
 - `name` (String) Human readable widget name.
 
 ### Optional
--- a/docs/resources/turnstile_widget.md
+++ b/docs/resources/turnstile_widget.md
@@ -28,7 +28,7 @@ resource "cloudflare_turnstile_widget" "example" {
 
 - `account_id` (String) The account identifier to target for the resource.
 - `domains` (Set of String) Domains where the widget is deployed
-- `mode` (String) Widget Mode. Valid values: `"non-interactive"`, `"invisible"`, or `"managed"`.
+- `mode` (String) Widget Mode. Available values: `non-interactive`, `invisible`, `managed`
 - `name` (String) Human readable widget name.
 
 ### Optional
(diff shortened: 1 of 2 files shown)


LinusU:
@jacobbednarz could you add the `workflow/skip-changelog-entry` tag?

> #### Changes that should _not_ have a CHANGELOG entry
> 
> - Resource and provider documentation updates

LinusU:
Hmm, seems like the best thing would be if the documentation generator could look at the `stringvalidator.OneOf` validator and do this automatically? 🤔 

I've added the text to `MarkdownDescription` now 🚀

jacobbednarz:
> Hmm, seems like the best thing would be if the documentation generator could look at the `stringvalidator.OneOf` validator and do this automatically? 🤔

this is a known restriction of the plugin framework. it is already tracked upstream via https://github.com/hashicorp/terraform-plugin-framework/issues/625

jacobbednarz:
thanks for this one @LinusU 🏆

LinusU:
Thanks for the quick pull!

btw. do you know approximately when #2413 will be released? I ran into basically all issues covered by that one and had to `terraform state rm` to recover, and I would imagine that anyone else trying to use Turnstile via Terraform would have the same problems...

jacobbednarz:
the release process (including how to calculate the next one) is documented at https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/release-process.md.

if you need to use an unreleased version of the provider, there is also some steps you can use from https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/using-non-released-versions.md to run whatever version you're after.

Issue #2260, opened 2023-02-27
Resource of type cloudflare_notification_policy with alert_type "advanced_ddos_attack_l7_alert" can't be created


brunorubin:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```
Terraform v1.3.4
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v4.0.0
```
and
```
Terraform v1.3.4
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v3.35.0
```
### Affected resource(s)

- cloudflare_notification_policy

### Terraform configuration files

```hcl
resource "cloudflare_notification_policy" "ddos_notifications_type_l7" {
  account_id  = module.cloudflare_account.id
  name        = "DDoS Notification"
  enabled     = false
  alert_type  = "advanced_ddos_attack_l7_alert"

  filters {
    target_zone_name = ["sampledomain.com"]
    target_host      = ["sub.sampledomain.com"]
  }
}
```

### Link to debug output

https://gist.github.com/brunorubin/f4594b792961c51df6c8f136a906c8fc

### Expected output

```
cloudflare_notification_policy.ddos_notifications_type_l7: Creating...
cloudflare_notification_policy.ddos_notifications_type_l7: Creation complete after 3s [id=23c4901d88ce401e99a0a4b71b875917]
```

### Actual output

```
│ Error: error creating policy DDoS: Invalid object selections (17003)
│
│   with cloudflare_notification_policy.ddos_notifications_type_l7,
│   on ddos.tf line 25, in resource "cloudflare_notification_policy" "ddos_notifications_type_l7":
│   25: resource "cloudflare_notification_policy" "ddos_notifications_type_l7" {
│
```

### Steps to reproduce

1. Create a new `cloudflare_notification_policy`
2. Define `alert_type` to `advanced_ddos_attack_l7_alert`
3. Include a filter for hostname and zone:
```
  filters {
    target_host      = ["sampledomain.com"]
    target_zone_name = ["sub.sampledomain.com"]
  }
```
4. Run `terraform apply`

### Additional factoids

When removing `target_host` the resource gets created normally:
```
  filters {
    target_zone_name = ["sampledomain.com"]
  }
```
---
Looks like the actual Cloudflare Schema expects `target_hostname` instead
```
      {
[...]

PR #2425, opened 2023-05-09
fix: unexpected crashes when setting target_host_name with a filters attribute

atsushii:
fixed unexpected crashes when setting target_host_name with a filters attribute

Fixes: #2260


jacobbednarz:
thanks for this one @atsushii 🙇

PR #2434, opened 2023-05-16
Fix workers_kv import to properly split the id

### LINKED ISSUE #2432: Unable to import cloudflare_workers_kv
### Confirmation

Terraform v1.4.2
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v4.5.0

### Affected resource(s)

 - cloudflare_workers_kv

```hcl
resource "cloudflare_workers_kv" "this" {
  account_id   = data.cloudflare_accounts.this.accounts[0].id
  namespace_id = cloudflare_workers_kv_namespace.this.id
  key          = "example"
  value        = jsonencode({
            "headers" : [
              [
                "access-control-allow-origin",
                "*"
              ]
            ]})
}
```

### Link to debug output

https://gist.github.com/tweakster/8295c2196400c96a3fa2cd8f6edd5dff

### Panic output

_No response_

### Expected output

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.

### Actual output

╷
│ Error: error reading workers kv: identifier required but missing
│ 
│ 
╵

### Steps to reproduce

TF_LOG=DEBUG terraform import 'cloudflare_workers_kv.this' 'xxxxxxxxxxxxxxxxxxxxxxx/yyyyyyyyyyyyyyyyyyyyyy/zzzzzzzzzzzzzzzzzzzzzzz'

### Additional factoids

I think the problem is that Account ID is not set in `resourceCloudflareWorkersKVImport` so when [called for](https://github.com/cloudflare/terraform-provider-cloudflare/blob/f61a23ff084af0f756916980fb1baa021f950f7b/internal/sdkv2provider/resource_cloudflare_workers_kv.go#L37) in  `resourceCloudflareWorkersKVRead` it errors

I did manage to hack the provider locally and get it to work. This is my amended function:
```go
func resourceCloudflareWorkersKVImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
	accountID, id, err := parseId(d.Id())
	if err != nil {
		return nil, err
	}

	d.Set(consts.AccountIDSchemaKey, accountID)
	d.SetId(id)
	resourceCloudflareWorkersKVRead(ctx, d, meta)

	return []*schema.ResourceData{d}, nil
}
```

### References

_No response_

tweakster:
Fixes #2432.

The worker_kv import id has 3 parts:
   <account_id>/<namespace_id>/<key_name>

Account ID needs to be set before resourceCloudflareWorkersKVRead
is called and that function will then expects id to be
<namespace_id>/<key_name>

### DIFF
--- a/internal/sdkv2provider/resource_cloudflare_workers_kv.go
+++ b/internal/sdkv2provider/resource_cloudflare_workers_kv.go
@@ -101,14 +101,15 @@ func resourceCloudflareWorkersKVDelete(ctx context.Context, d *schema.ResourceDa
 }
 
 func resourceCloudflareWorkersKVImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-	namespaceID, key, err := parseId(d.Id())
-	if err != nil {
-		return nil, err
+	parts := strings.SplitN(d.Id(), "/", -1)
+	if len(parts) != 3 {
+		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"accountID/namespaceID/keyName\"", d.Id())
 	}
 
-	d.Set("namespace_id", namespaceID)
-	d.Set("key", key)
-
+	d.Set(consts.AccountIDSchemaKey, parts[0])
+	d.Set("namespace_id", parts[1])
+	d.Set("key", parts[2])
+	d.SetId(fmt.Sprintf("%s/%s", parts[1], parts[2]))
 	resourceCloudflareWorkersKVRead(ctx, d, meta)
 
 	return []*schema.ResourceData{d}, nil
--- a/.changelog/2434.txt
+++ b/.changelog/2434.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_workers_kv: Fix import to properly parse the id
+```


internal/sdkv2provider/resource_cloudflare_workers_kv.go
```diff
@@ -101,14 +101,13 @@ func resourceCloudflareWorkersKVDelete(ctx context.Context, d *schema.ResourceDa
 }
 
 func resourceCloudflareWorkersKVImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-	namespaceID, key, err := parseId(d.Id())
+	accountID, id, err := parseId(d.Id())
```
jacobbednarz, line 104: `parseId` no longer seems suitable here given the method only splits into 2 sections and the import format needs 3. i.e. `<account_id>/<namespace_id>/<key_name>`

jacobbednarz (changes requested):

internal/sdkv2provider/resource_cloudflare_workers_kv.go
```diff
@@ -101,14 +101,15 @@ func resourceCloudflareWorkersKVDelete(ctx context.Context, d *schema.ResourceDa
...
+		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"accountID/namespaceID/keyName\"", d.Id())
 	}
 
-	d.Set("namespace_id", namespaceID)
-	d.Set("key", key)
-
+	d.Set(consts.AccountIDSchemaKey, parts[0])
+	d.Set("namespace_id", parts[1])
+	d.Set("key", parts[2])
+	d.SetId(fmt.Sprintf("%s/%s", parts[1], parts[2]))
```
tweakster, line 112: I see a similarity with https://github.com/cloudflare/terraform-provider-cloudflare/blob/483e9b5e5dc294d3ad57ee1dd0cf976af5b6b67d/internal/sdkv2provider/resource_cloudflare_workers_kv.go#L74 and wonder if it's worth having a `CreateId` function. Thoughts?

internal/sdkv2provider/resource_cloudflare_workers_kv.go
tweakster: > `parseId` no longer seems suitable here given the method only splits into 2 sections and the import format needs 3. i.e. `<account_id>/<namespace_id>/<key_name>`

@jacobbednarz 

Sorry. Looking back it's not my proudest work. I rushed a bit.

I've done some more thorough testing and needed to add a couple of lines so the future `plan`s don't try and recreate the resource.

How is that?

internal/sdkv2provider/resource_cloudflare_workers_kv.go
tweakster: I did it anyway. It's in a separate commit so I can remove if you think it's inappropriate.

internal/sdkv2provider/resource_cloudflare_workers_kv.go
```diff
@@ -121,3 +122,7 @@ func parseId(id string) (string, string, error) {
 	}
 	return parts[0], parts[1], nil
 }
+
+func fmtId(namespaceID string, key string) string {
```
jacobbednarz, line 126: this method is what [`parseId`](https://github.com/cloudflare/terraform-provider-cloudflare/blob/22331413ae48c47860d450f37f13cb931ab1182b/internal/sdkv2provider/resource_cloudflare_workers_kv.go#L117-L123) already does. the issue for the import is that it is expecting 3 but would only get two. i think we can revert everything outside of the `Import` changes and this will be fine.

tweakster: @jacobbednarz No problem. Done.

jacobbednarz:
thanks for getting this over the line @tweakster 👏

PR #2436, opened 2023-05-17
Improve error handling in account delete function

### LINKED ISSUE #2437: Improve Error Handling of AccountDeletion

Terraform v1.4.3-dev
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v3.35.0

### Description

When debugging an issue with account deletion failures, I had to dig into the code to see what was oing on. I noticed that the resourceCloudflareAccountDelete could log more helpful information about the account that failed to delete.

### Use cases

I was deleting multiple accounts and it was only one of them that was failing to delete.

```hcl
N/A
...
```

### References

_No response_

therealdwright:
This commit surfaces some additional information (the account ID) on account 
deletion failure cases.

I encountered some strange behaviour (that ended up being my fault) that
led to me digging through this module. Thought I'd leave it in a better place than
I found it as I found a couple of minor issues.

Closes #2437

### DIFF
--- a/.changelog/2436.txt
+++ b/.changelog/2436.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/cloudflare_account: provide account ID for error handling in `resourceCloudflareAccountDelete`
+```
--- a/internal/sdkv2provider/resource_cloudflare_access_keys_configuration.go
+++ b/internal/sdkv2provider/resource_cloudflare_access_keys_configuration.go
@@ -93,7 +93,7 @@ func resourceCloudflareKeysConfigurationImport(ctx context.Context, d *schema.Re
 	d.Set(consts.AccountIDSchemaKey, accountID)
 
 	if err := resourceCloudflareAccessKeysConfigurationRead(ctx, d, meta); err != nil {
-		return nil, fmt.Errorf("diagnostic error occurred: %v", err)
+		return nil, errors.New(err[0].Summary)
 	}
 
 	return []*schema.ResourceData{d}, nil
--- a/internal/sdkv2provider/resource_cloudflare_access_keys_configuration.go
+++ b/internal/sdkv2provider/resource_cloudflare_access_keys_configuration.go
@@ -92,6 +92,9 @@ func resourceCloudflareKeysConfigurationImport(ctx context.Context, d *schema.Re
 	d.SetId(accountID)
 	d.Set(consts.AccountIDSchemaKey, accountID)
 
-	resourceCloudflareAccessKeysConfigurationRead(ctx, d, meta)
+	if err := resourceCloudflareAccessKeysConfigurationRead(ctx, d, meta); err != nil {
+		return nil, fmt.Errorf("diagnostic error occurred: %v", err)
+	}
+
 	return []*schema.ResourceData{d}, nil
 }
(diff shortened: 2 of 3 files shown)


internal/sdkv2provider/resource_cloudflare_account.go
```diff
@@ -66,8 +66,7 @@ func resourceCloudflareAccountRead(ctx context.Context, d *schema.ResourceData,
 
 	foundAcc, _, err := client.Account(ctx, accountID)
 	if err != nil {
-		var notFoundError *cloudflare.NotFoundError
-		if errors.As(err, &notFoundError) {
+		if errors.As(err, &cloudflare.NotFoundError{}) {
```
jacobbednarz, line 69: i can't recall what version of Go i had this issue in but it was required to have an initialised struct for `errors.As` -- using it in this fashion caused it to never match.

.changelog/2436.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+resource/cloudflare_account: Improve error handling of resourceCloudflareAccountDelete
```
jacobbednarz, line 2:
```suggestion
resource/cloudflare_account: provide account ID for error handling in `resourceCloudflareAccountDelete`
```

internal/sdkv2provider/resource_cloudflare_account.go
therealdwright: No problem, have backed out this part of the change.

internal/sdkv2provider/resource_cloudflare_access_keys_configuration.go
```diff
@@ -92,6 +92,9 @@ func resourceCloudflareKeysConfigurationImport(ctx context.Context, d *schema.Re
 	d.SetId(accountID)
 	d.Set(consts.AccountIDSchemaKey, accountID)
 
-	resourceCloudflareAccessKeysConfigurationRead(ctx, d, meta)
+	if err := resourceCloudflareAccessKeysConfigurationRead(ctx, d, meta); err != nil {
+		return nil, err
```
jacobbednarz, line 96:
```suggestion
		return nil, diag.FromErr(err)
```

internal/sdkv2provider/resource_cloudflare_access_keys_configuration.go
```diff
@@ -92,6 +92,9 @@ func resourceCloudflareKeysConfigurationImport(ctx context.Context, d *schema.Re
 	d.SetId(accountID)
 	d.Set(consts.AccountIDSchemaKey, accountID)
 
-	resourceCloudflareAccessKeysConfigurationRead(ctx, d, meta)
+	if err := resourceCloudflareAccessKeysConfigurationRead(ctx, d, meta); err != nil {
+		return nil, fmt.Errorf("diagnostic error occurred: %v", err)
```
jacobbednarz, line 96: since this will be dumping a `diag.Diagnostics` struct, this is going to be pretty unreadable.

```
[{0 My Summary Some more details here []}]
``` 
instead we probably want to pull out the first (and likely only `diag.Diagnostic`) to return that as the error.
```suggestion
		return nil, errors.New(err[0].Summary)
```

jacobbednarz:
thanks for following through on this one @therealdwright, appreciate it!

Issue #2439, opened 2023-05-17
Documentation for cloudflare_pages_project suggests a bad value for root_dir


jamielinux:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```console
$ terraform -v
Terraform v1.4.6
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v4.6.0
```

### Affected resource(s)

- `cloudflare_pages_project`

### Terraform configuration files

```hcl
# Manage build config
resource "cloudflare_pages_project" "build_config" {
  account_id        = "f037e56e89293a057740de681ac9abbe"
  name              = "this-is-my-project-01"
  production_branch = "main"
  build_config {
    build_command       = "npm run build"
    destination_dir     = "build"
    root_dir            = "/"
    web_analytics_tag   = "cee1c73f6e4743d0b5e6bb1a0bcaabcc"
    web_analytics_token = "021e1057c18547eca7b79f2516f06o7x"
  }
}
```

### Link to debug output

https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/pages_project

### Expected output

```hcl
root_dir            = ""
```

### Actual output

```hcl
root_dir            = "/"
```

### Steps to reproduce

1. Set `root_dir = "/"` as suggested in the docs.
2. A Pages project is created with `root_dir` of `//` (double slash).

jamielinux:
The dashboard advises:

> If your site is not a subdirectory, leave this path value empty.

![Screenshot from 2023-05-17 11-02-07](https://github.com/cloudflare/terraform-provider-cloudflare/assets/2598089/44c7887f-91e2-4885-a05f-1701b368922c)

PR #2440, opened 2023-05-17
Suggest a better default value for root_dir in pages_project docs

jamielinux:
Fixes https://github.com/cloudflare/terraform-provider-cloudflare/issues/2439

`root_dir = "/"` results in a Pages project that has `root_dir` of `//` (double slash).

### DIFF
--- a/docs/resources/pages_project.md
+++ b/docs/resources/pages_project.md
@@ -147,7 +147,7 @@ Optional:
 
 - `build_command` (String) Command used to build project.
 - `destination_dir` (String) Output directory of the build.
-- `root_dir` (String) Directory to run the command.
+- `root_dir` (String) Your project's root directory, where Cloudflare runs the build command. If your site is not in a subdirectory, leave this path value empty.
 - `web_analytics_tag` (String) The classifying tag for analytics.
 - `web_analytics_token` (String) The auth token for analytics.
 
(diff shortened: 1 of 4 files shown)


docs/resources/pages_project.md
jacobbednarz, line 1: this file is a generated artifact. check out https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md on how to change these files and not have them overridden on the next run.

jacobbednarz (changes requested):

docs/resources/pages_project.md
jamielinux: Sorry, I should have read that contributing readme! I've updated the PR.

jacobbednarz:
@WalshyDev does it also make sense to update the schema description with something like "omit the trailing slash"? https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/sdkv2provider/schema_cloudflare_pages_project.go#L24

WalshyDev:
Thank you for this fix @jamielinux 

@jacobbednarz yes definitely. The copy we use in the UI is "Your project’s root directory, where Cloudflare runs the build command. If your site is not in a subdirectory, leave this path value empty."

jacobbednarz:
thanks @WalshyDev - have updated the schema to use the same one. feel free to open a larger PR with any copy changes if you'd like it closer to the UI copy across the board.

PR #2442, opened 2023-05-17
FW-5543: Add deprecation message for Firewall rules and Filters

obbap1:
This adds deprecation messages for the firewall rules and filters resource to aid with the migration to custom rules.

Screenshot 2023-05-17 at 14 59 15
Screenshot 2023-05-17 at 14 59 36


jacobbednarz:
thanks for this @obbap1 . hope you don't mind but i went ahead and cleaned up the formatting a little and added this to the resource docs too so that folks coming to that see this as well.

obbap1:
Thanks @jacobbednarz 🎉

PR #2454, opened 2023-05-18
Adds additional validation for TTLs for cloudflare_ruleset action_parameters

iveelsm:
This adds new validation for the schema that enforces TTL values for specific modes within the cloudflare_ruleset action parameters. This applies to the browser_ttl and edge_ttl blocks within the action_parameters schema. The validation will serialize the existing values into a type that is a 1:1 match of the schema and then will determine if the default value for the ttl is greater than zero for the configured rule.

Implementation:

* Custom object validator for Browser TTL block within action_parameters for cloudflare_ruleset
* Custom object validator for Edge TTL block within action_parameters for cloudflare_ruleset

### DIFF
--- a/.changelog/2454.txt
+++ b/.changelog/2454.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_ruleset: Validation of ttls for action_parameters with edge_ttl or browser_ttl mode of override_origin
+```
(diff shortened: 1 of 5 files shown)


jacobbednarz:
at a high level, this looks fine. let's add some test coverage to ensure that these validators are covered. an acceptance test is likely the easiest to interface with.

PR #2492, opened 2023-06-01
Ignore invalid full wildcards for cache_rule cache_key includes/excludes

iveelsm:
Wildcards for the include and exclude values for cache keys need to be set using the ignore rules. Setting the value using include or exclude can cause unexpected behaviors. This addresses this outstanding bug by adding new validation to the schema.

Implementation:

* Adds new validation to the query_key schema for cache_key in cache_rules

### DIFF
--- a/.changelog/2492.txt
+++ b/.changelog/2492.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_ruleset: removes ability to set wildcards for include and exclude, provides guidance on proper values to use instead
+```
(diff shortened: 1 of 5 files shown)


jacobbednarz:
looks like this new validator is going to break expected functionality here - https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/framework/service/rulesets/resource.go#L524-L538

iveelsm:
This was mischaracterized as a problem that needed additional validation. It turns out everything is functioning exactly as it should. There is no need for this PR. Closing. :)

PR #2512, opened 2023-06-08
Update zone.md

karthik7678:
If the Zone Plan is not provided, it defaults to the "Free" plan.


jacobbednarz (changes requested):
thanks for the PR! unfortunately, the file you've edited is a generated artifact of the documentation proces. to see how to make changes to the documentation, you're best to check out https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md

PR #2525, opened 2023-06-14
Clarify documentation

hp197:


jacobbednarz:
thanks for the PR but this is already documented on the resource itself - https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/ruleset#import. we don't want to be duplicating that documentation given the resource documentation is automatically generated.

PR #2532, opened 2023-06-15
CUSTESC-29915: unify output for rules without rule settings

### LINKED ISSUE #2085: cloudflare_teams_rule without rule_settings create a diff on every apply
### Confirmation

```
Terraform v1.3.9
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v3.29.0
```
### Affected resource(s)

```
cloudflare_teams_rule
```

```hcl
resource "cloudflare_teams_rule" "allow_admin_all" {
  account_id  = var.account_id
  name        = "AllowAdminAll"
  description = "Allow VIN Team OpsNet Access"
  precedence  = 1000
  enabled     = true
  action      = "allow"
  filters     = ["l4"]
  identity    = "any(identity.groups.name[*] in {\"Admin\"})"
}
```

### Link to debug output

https://gist.github.com/wind0r/8eb1290b04f07e026dfd85bcd4b8ee10

### Panic output

No Panic

### Expected output

No Diff when the same config is applied again

### Actual output

  ```
  # cloudflare_teams_rule.allow_vin_opsnet will be updated in-place
  ~ resource "cloudflare_teams_rule" "allow_admin_all" {
        id          = "$ID"
        name        = "AllowAdminAll"
        # (8 unchanged attributes hidden)

      - rule_settings {
          - add_headers                        = {} -> null
          - block_page_enabled                 = false -> null
          - insecure_disable_dnssec_validation = false -> null
          - override_ips                       = [] -> null
        }
    }
   ```

### Steps to reproduce

1. Create Example Teams rule
2. Plan and Apply
3. Plan again.

### Additional factoids

IMO: Applying the same Input again should not result in a diff.

Looks like CF API return a rule_settings object which doesn't match the state of not having a rule_settings object.
Since rule_settings is optional I would like to not add it to all team rules, just for having a clean diff.

### References

_No response_

alyssamw:
Closes #2085


jacobbednarz:
thanks for the PR here @alyssamw. i went ahead and pushed a couple of commits to address some issues i found here. details are in the commits but at a high level:

- i moved the `empty` logic into the `flattenTeamsRuleSettings` method for two reasons. the first is that `Read`/`Create` need to use the same source for when to save the nested object in state. the second part is that we always want to call `d.Set` (regardless of whether we have values) to make sure we can clear the state if someone removes the values.
- i extended the test logic to cover the second part of the first point but also found a logic bug in the `empty` logic where a `!= nil` should have been a `== nil`.

acceptance tests are now passing

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareTeamsRule_" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareTeamsRule_Basic
--- PASS: TestAccCloudflareTeamsRule_Basic (14.99s)
=== RUN   TestAccCloudflareTeamsRule_NoSettings
--- PASS: TestAccCloudflareTeamsRule_NoSettings (21.98s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	38.044s
```

.changelog/2532.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_teams_rules.go: unify output for rules without rule settings
```
jacobbednarz, line 2:
```suggestion
resource/cloudflare_teams_rules.go: handle state correctly when `rules_setting` is empty
```

PR #2534, opened 2023-06-16
Fix healthcheck port example values

SebastiaanYN:
The type of `port` is a number but the example has it as strings


jacobbednarz:

PR #2539, opened 2023-06-19
Fix `access_ca_certificate` import

Cyb3r-Jak3:
Fixes importing access_ca_certifcates.

Reading an access application requires using the application ID
https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/sdkv2provider/resource_cloudflare_access_ca_certificate.go#L65-L70

This is not provided with the import as it only sets account ID and the access certificate ID
https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/sdkv2provider/resource_cloudflare_access_ca_certificate.go#L134-L135

It might be worth reconsidering how the IDs are handled with the `access_ca_certifcate` resource because the actual certificate ID is never used and only the application ID is.

### DIFF
--- a/internal/sdkv2provider/resource_cloudflare_access_ca_certificate.go
+++ b/internal/sdkv2provider/resource_cloudflare_access_ca_certificate.go
@@ -116,13 +116,13 @@ func resourceCloudflareAccessCACertificateDelete(ctx context.Context, d *schema.
 }
 
 func resourceCloudflareAccessCACertificateImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-	attributes := strings.SplitN(d.Id(), "/", 3)
+	attributes := strings.SplitN(d.Id(), "/", 4)
 
-	if len(attributes) != 3 {
-		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"account/accountID/accessCACertificateID\" or \"zone/zoneID/accessCACertificateID\"", d.Id())
+	if len(attributes) != 4 {
+		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"account/accountID/applicationID/accessCACertificateID\" or \"zone/zoneID/applicationID/accessCACertificateID\"", d.Id())
 	}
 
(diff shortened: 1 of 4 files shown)


Cyb3r-Jak3:
I would add an import test but I don't know how to extract the application ID during the test

internal/sdkv2provider/resource_cloudflare_access_ca_certificate.go
```diff
@@ -116,13 +116,13 @@ func resourceCloudflareAccessCACertificateDelete(ctx context.Context, d *schema.
...
-		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"account/accountID/accessCACertificateID\" or \"zone/zoneID/accessCACertificateID\"", d.Id())
+	if len(attributes) != 4 {
+		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"account/accountID/applicationID/accessCACertificateID\" or \"zone/zoneID/applicationID/accessCACertificateID\"", d.Id())
 	}
 
-	identifierType, identifierID, accessCACertificateID := attributes[0], attributes[1], attributes[2]
+	identifierType, identifierID, applicationID, accessCACertificateID := attributes[0], attributes[1], attributes[2], attributes[3]
 
 	if AccessIdentifierType(identifierType) != AccountType && AccessIdentifierType(identifierType) != ZoneType {
 		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"account/accountID/accessCACertificateID\" or \"zone/zoneID/accessCACertificateID\"", d.Id())
```
jacobbednarz, line 128:
```suggestion
		return nil, fmt.Errorf("invalid id (\"%s\") specified, should be in format \"account/accountID/applicationID/accessCACertificateID\" or \"zone/zoneID/applicationID/accessCACertificateID\"", d.Id())
```

jacobbednarz:
we also need to update the example import at https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/examples/resources/cloudflare_access_ca_certificate/import.sh (don't forgot to run `make docs`).

jacobbednarz:
thanks - we can revisit the import path if some of it isn't actually needed.

PR #2582, opened 2023-07-04
resource/cloudflare_tunnel_config: fix TunnelDuration order of magnitude

0x2b3bfa0:
> [_A TunnelDuration is a Duration that has custom serialization for JSON. JSON in Javascript assumes that int fields are 32 bits and Duration fields are deserialized assuming that numbers are in nanoseconds, which in 32bit integers limits to just 2 seconds. This type assumes that when serializing/deserializing from JSON, that the number is in seconds, while it maintains the YAML serde assumptions._](https://pkg.go.dev/github.com/cloudflare/cloudflare-go@v0.70.0#TunnelDuration)

### Workaround

```terraform
resource "cloudflare_tunnel_config" "example" {
  ...
  config {
    ...
    origin_request {
      # HACK: fix the default values
      connect_timeout = "30ns" # 30s
      keep_alive_timeout = "90ns" # 90s
      tcp_keep_alive = "30ns" # 30s
      tls_timeout = "10ns" # 10s
    }
  }
}
```

### DIFF
--- a/internal/sdkv2provider/resource_cloudflare_tunnel_config.go
+++ b/internal/sdkv2provider/resource_cloudflare_tunnel_config.go
@@ -36,15 +36,15 @@ func resourceCloudflareTunnelConfig() *schema.Resource {
 func buildTunnelOriginRequest(originRequest map[string]interface{}) (originConfig cloudflare.OriginRequestConfig) {
 	if v, ok := originRequest["connect_timeout"]; ok {
 		timeout, _ := time.ParseDuration(v.(string))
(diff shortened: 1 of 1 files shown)


jacobbednarz:
this PR should not be necessary since the underlying library addressed handling the values. you may however have to update your own state with either a new resource or the specific values you want to realign it.

0x2b3bfa0:
Thank you! My bad; I was using an outdated version of the provider. 🤦🏼‍♂️

Issue #2591, opened 2023-07-06
Can't set default browser ttl to 0


kleinmp:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.5.2
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v4.9.0

### Affected resource(s)

cloudflare_ruleset

### Terraform configuration files

```hcl
resource "cloudflare_ruleset" "terraform_managed_resource--http_request_cache_settings" {
  kind    = "zone"
  name    = "default"
  phase   = "http_request_cache_settings"
  zone_id = var.cloudflare_zone_id
  rules {
    action = "set_cache_settings"
    action_parameters {
      browser_ttl {
        default = 0
        mode    = "override_origin"
      }
      cache = true
      cache_key {
        custom_key {
          query_string {
            include = ["*"]
          }
        }
      }
      edge_ttl {
        default = 86400
        mode    = "override_origin"
        status_code_ttl {
          status_code = 302
          value       = -1
        }
        status_code_ttl {
          status_code = 307
          value       = -1
        }
      }
    }
    description = "Default Cache"
    enabled     = true
    expression  = "(http.host eq \"test.example.com\")"
  }
  rules {
    action = "set_cache_settings"
    action_parameters {
      cache = false
    }
    description = "No Cache Paths"
    enabled     = true
    expression  = "(starts_with(http.request.uri.path, \"/123/\")) or (starts_with(http.request.uri.path, \"/234/\")) or (starts_with(http.request.uri.path, \"/345/\"))"
  }
}
```

### Link to debug output

https://gist.github.com/kleinmp/6f9af873c760e687bc717ad218a99b03

### Expected output

The configuration is expected to be valid when default value is 0 for browser ttl on a cache rule.

### Actual output

│ Error: invalid configuration
│ 
│   with cloudflare_ruleset.terraform_managed_resource--http_request_cache_settings,
│   on main.tf line 42, in resource "cloudflare_ruleset" "terraform_managed_resource--http_request_cache_settings":
│   42:       browser_ttl {
│ 
│ using mode 'override_origin' requires setting a default for ttl
╵
╷
│ Error: Invalid Attribute Value
│ 
│   with cloudflare_ruleset.terraform_managed_resource--http_request_cache_settings,
[...]

PR #2592, opened 2023-07-06
#2591 address browser_ttl.default value issue of not being able to be set to `0`

nialdaly:
### Summary
Hi there, this is a possible resolution to issue [#2591](https://github.com/cloudflare/terraform-provider-cloudflare/issues/2591).

When the `Browser TTL` option is to set to `Override origin` and `No cache` the corresponding `default` value is `0`.

Screenshot 2023-07-06 at 22 05 11

I'm unable to run the acceptance tests locally, so any guidance would be greatly appreciated!


jacobbednarz:
@iveelsm can you clarify the expected behaviour here and if this was intentional?

iveelsm:
@jacobbednarz This was intentional. I will surface it more on the issue as well, but this is the underlying implementation leaking through here. The general problem here is that a Browser TTL of "Override Origin" set to `0` actually translates on the backend to "Respect Origin".

I have a new feature for Browser Bypass which is rolling out soon, but we should not allow people to specify `0` here until the legacy API contracts are sufficiently wrangled.

nialdaly:
Thanks for clarifying @iveelsm

iveelsm:
Sure thing @nialdaly! Thank you for the PR! This is just one of those unfortunate things where our API internally isn't quite as logical as we need it to be and requires a lot of backend work before this would work.

nialdaly:
That makes sense, thanks for the additional context!

PR #2431, opened 2023-05-12
Minor Update to Ruleset Documentation

dzeleniak:
Hello ! 

This is a quick update to the documentation for rulesets
I noticed while working with the provider that the example for Zone-level WAF with tag-based overrides is using the ```status``` field:

image

The schema documentation shows that this has been changed to use the ```enabled``` field instead.

```
        categories {
          category = "wordpress"
          action   = "block"
          enabled = true
        }
```

Fixes #2428

PR #2611, opened 2023-07-11
Remove a old syntax for rules.action_parameters.overrides.categories

kamillle:
Currently rules.action_parameters.overrides.categories accepts `enabled` field to define enable or disable the override, but the document has a old syntax that express we should use `status` field

### DIFF
--- a/docs/resources/ruleset.md
+++ b/docs/resources/ruleset.md
@@ -73,13 +73,13 @@ resource "cloudflare_ruleset" "zone_level_managed_waf_with_category_based_overri
         categories {
           category = "wordpress"
           action   = "block"
-          status   = "enabled"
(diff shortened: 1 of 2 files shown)


jacobbednarz:

kamillle:
@jacobbednarz Umm, ok. I'm not sure but [this commit](https://github.com/cloudflare/terraform-provider-cloudflare/pull/2611/commits/6c7f1344123fc7f6daac8fe2c4c0702f531a1704) makes sense for the rule?

jacobbednarz:
i'm not sure why, but this was already addressed in #2431 but hasn't stuck :/

for this PR, you needed to also run `make docs` which generated the static docs but other than that, this is all good. thanks!

Issue #2622, opened 2023-07-19
Spectrum Application "edge_ips" is required, but is marked as optional


witanred:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```
rj@C02FTAWRMD6R terraform % terraform -v
Terraform v1.5.3
on darwin_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v4.5.0
```

### Affected resource(s)

cloudflare_spectrum_application

### Terraform configuration files

```hcl
terraform {
  required_providers {
    cloudflare = {
      source = "cloudflare/cloudflare"
      version = "4.5.0"
    }
  }
}

provider "cloudflare" {
  email = "imagine an email"
  api_key = "imagine an api key"
}

variable "zone_id" {
  default = "86cb..."
}

variable "account_id" {
  default = "e745..."
}

variable "domain" {
  default = "tf.ross.gg"
}

############################################
#---------configure_cf_zone_spectrum-------#
############################################
resource "cloudflare_spectrum_application" "spectrum" {
  zone_id = var.zone_id
  protocol     = "tcp/22"
  traffic_type = "direct"

  dns {
    type = "CNAME"
    name = "ssh.tf.ross.gg"
  }

  origin_direct = [
    "tcp://2.2.2.2:22"
  ]
}
```

### Link to debug output

https://gist.github.com/witanred/7156a46546ceed89263ea76edd4b7752

### Panic output

NA see above

### Expected output

No failure, as edge_ips is marked optional

### Actual output

See https://gist.github.com/witanred/7156a46546ceed89263ea76edd4b7752

### Steps to reproduce

Run the config as is with valid auth, and you will get this error. Make sure the zone is entitled to create tcp spectrum applications.

### Additional factoids

Feel free to message me internally if you have any questions. I am happy to submit a PR.
I may well be misreading the error output here, but the example configuration should not error so easily without an optional field. So the error is in documentation somewhere.

### References

https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/spectrum_application#nestedblock--origin_dns

witanred:
Public Gist at https://gist.github.com/witanred/7156a46546ceed89263ea76edd4b7752 has been updated with DEBUG output

PR #2629, opened 2023-07-22
cloudflare_spectrum_application: Remove default values, make `edge_ip…

Nmishin:
…s` parameter optional

### DIFF
--- a/.changelog/2629.txt
+++ b/.changelog/2629.txt
@@ -1,0 +1,3 @@
+```release-note:breaking-change
+resource/cloudflare_spectrum_application: Remove default values, make `edge_ips` parameter optional.
+```
--- a/docs/resources/spectrum_application.md
+++ b/docs/resources/spectrum_application.md
@@ -47,16 +47,16 @@ resource "cloudflare_spectrum_application" "example" {
 
 ### Optional
 
-- `argo_smart_routing` (Boolean) Enables Argo Smart Routing. Defaults to `false`.
+- `argo_smart_routing` (Boolean) Enables Argo Smart Routing.
 - `edge_ips` (Block List, Max: 1) The anycast edge IP configuration for the hostname of this application. (see [below for nested schema](#nestedblock--edge_ips))
-- `ip_firewall` (Boolean) Enables the IP Firewall for this application. Defaults to `true`.
+- `ip_firewall` (Boolean) Enables the IP Firewall for this application.
 - `origin_direct` (List of String) A list of destination addresses to the origin. e.g. `tcp://192.0.2.1:22`.
 - `origin_dns` (Block List, Max: 1) A destination DNS addresses to the origin. (see [below for nested schema](#nestedblock--origin_dns))
 - `origin_port` (Number) Origin port to proxy traffice to. Conflicts with `origin_port_range`.
 - `origin_port_range` (Block List, Max: 1) Origin port range to proxy traffice to. When using a range, the protocol field must also specify a range, e.g. `tcp/22-23`. Conflicts with `origin_port`. (see [below for nested schema](#nestedblock--origin_port_range))
-- `proxy_protocol` (String) Enables a proxy protocol to the origin. Available values: `off`, `v1`, `v2`, `simple`. Defaults to `off`.
-- `tls` (String) TLS configuration option for Cloudflare to connect to your origin. Available values: `off`, `flexible`, `full`, `strict`. Defaults to `off`.
-- `traffic_type` (String) Sets application type. Available values: `direct`, `http`, `https`. Defaults to `direct`.
+- `proxy_protocol` (String) Enables a proxy protocol to the origin. Available values: `off`, `v1`, `v2`, `simple`.
+- `tls` (String) TLS configuration option for Cloudflare to connect to your origin. Available values: `off`, `flexible`, `full`, `strict`.
+- `traffic_type` (String) Sets application type. Available values: `direct`, `http`, `https`.
 
 ### Read-Only
 
--- a/internal/sdkv2provider/resource_cloudflare_spectrum_application.go
+++ b/internal/sdkv2provider/resource_cloudflare_spectrum_application.go
@@ -118,8 +118,10 @@ func resourceCloudflareSpectrumApplicationRead(ctx context.Context, d *schema.Re
 		}
 	}
 
-	if err := d.Set("edge_ips", flattenEdgeIPs(application.EdgeIPs)); err != nil {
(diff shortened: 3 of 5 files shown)


Nmishin:
Fix for https://github.com/cloudflare/terraform-provider-cloudflare/issues/2622

Default values deleted because they prevent users from create some different types of Spectrum applications.

For prevent errors like `json: unknown field "edge_ips`, all that parameters (`traffic_type`, `tls`, `ip_firewall`, `proxy_protocol`, `argo_smart_routing`) should be configured explicitly only for needed application types.

jacobbednarz:
these defaults were intentionally added as the API required them at one point in time. i'll send this to the team internally to confirm what behaviour we need here vs in the Go library.

Nmishin:
@jacobbednarz I fully agree, but we definitely need to to something with that, because for now I can't create spectrum app at all. For example, in my account available only ssh, rdp, and minecraft, and I can't create any of them because of default values (v4.10.0 provider):

```
resource "cloudflare_spectrum_application" "spectrum" {
  zone_id  = var.zone_id
  protocol = "rdp"

  dns {
    type = "CNAME"
    name = "rdp.mydomain"
  }

  origin_direct = [
    "tcp://95.163.248.95:3389"
  ]

}
```

```
terraform apply

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

  # cloudflare_spectrum_application.spectrum will be created
  + resource "cloudflare_spectrum_application" "spectrum" {
      + argo_smart_routing = false
      + id                 = (known after apply)
      + ip_firewall        = true
      + origin_direct      = [
          + "tcp://95.163.248.95:3389",
        ]
      + protocol           = "rdp"
      + proxy_protocol     = "off"
      + tls                = "off"
      + traffic_type       = "direct"
      + zone_id            = "df066cb84f969ab8f4bc51b7f2eb2aa6"

      + dns {
          + name = "rdp.mydomain"
          + type = "CNAME"
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

cloudflare_spectrum_application.spectrum: Creating...
╷
│ Error: error creating spectrum application for zone: json: unknown field "traffic_type" (10012)
│ 
│   with cloudflare_spectrum_application.spectrum,
│   on main.tf line 1, in resource "cloudflare_spectrum_application" "spectrum":
│    1: resource "cloudflare_spectrum_application" "spectrum" {
```

jacobbednarz:
this PR is on hold as we've found the zone mentioned in the original issue isn't correctly setup which may be the cause here.

aberglund-cf:
Hi @Nmishin, I've tested this PR and while it does fix the problem in the issue it still has some slightly strange behavior regarding the change to default values. 

I'm seeing that when `edge_ips` is not declared it will always show a diff trying to remove those elements:
```shell
~ resource "cloudflare_spectrum_application" "testapp" {
        id                 = "<id>"
      - proxy_protocol     = "off" -> null
      - tls                = "off" -> null
      - traffic_type       = "direct" -> null

      - edge_ips {
          - connectivity = "all" -> null
          - ips          = [] -> null
          - type         = "dynamic" -> null
        }

        # (2 unchanged blocks hidden)
    }
```
I believe this is due to the mismatched handling of default values between the provider and the API. The API is returning the default values but the provider isn't expecting them and will try to remove those values. Ideally we'd want to avoid this confusing behavior. I think a good approach would be keep your change for reading `edge_ips` config values and to align the defaults values in the schema with what's being produced by the API.

Maybe the `edge_ips` schema can be updated to show something like this?
```go
"edge_ips": {
       Type:     schema.TypeList,
       Optional: true,
       Computed: true,
       DefaultFunc: func() (interface{}, error) {
               defaults := map[string]interface{}{
                       "type":         cloudflare.SpectrumEdgeTypeDynamic,
                       "connectivity": cloudflare.SpectrumConnectivityAll,
                       "ips":          []string{},
               }
               return []map[string]interface{}{defaults}, nil
       },
       MaxItems:    1,
```
Please let us know if this helps with your issue.

Nmishin:
Hi @aberglund-cf could you please check/test my MR for now, because i don't have all config options in my subscription ...
I added "Computed: true" option to the properties, and as I can read, this can be enough for handle default values from API.
If it still work incorrectly - please fell free to add commits to my MR :)

jacobbednarz:
this looks good from an acceptance test perspective

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareSpectrumApplication_" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareSpectrumApplication_Basic
--- PASS: TestAccCloudflareSpectrumApplication_Basic (12.56s)
=== RUN   TestAccCloudflareSpectrumApplication_OriginDNS
--- PASS: TestAccCloudflareSpectrumApplication_OriginDNS (10.67s)
=== RUN   TestAccCloudflareSpectrumApplication_OriginPortRange
--- PASS: TestAccCloudflareSpectrumApplication_OriginPortRange (13.06s)
=== RUN   TestAccCloudflareSpectrumApplication_Update
--- PASS: TestAccCloudflareSpectrumApplication_Update (18.40s)
=== RUN   TestAccCloudflareSpectrumApplication_EdgeIPConnectivity
--- PASS: TestAccCloudflareSpectrumApplication_EdgeIPConnectivity (10.95s)
=== RUN   TestAccCloudflareSpectrumApplication_EdgeIPsMultiple
--- PASS: TestAccCloudflareSpectrumApplication_EdgeIPsMultiple (13.84s)
=== RUN   TestAccCloudflareSpectrumApplication_BasicSSH
--- PASS: TestAccCloudflareSpectrumApplication_BasicSSH (9.82s)
=== RUN   TestAccCloudflareSpectrumApplication_BasicRDP
--- PASS: TestAccCloudflareSpectrumApplication_BasicRDP (9.08s)
=== RUN   TestAccCloudflareSpectrumApplication_BasicMinecraft
--- PASS: TestAccCloudflareSpectrumApplication_BasicMinecraft (9.40s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	108.788s
```

once @aberglund-cf gives the 👍 , we can merge.

@Nmishin in the meantime, can you please run `make docs` to have the documentation update correctly?

aberglund-cf (commented):
:+1:  LGTM. Thanks for helping us sort out this issue!

PR #2638, opened 2023-07-27
Stop customhostnames from waiting for SSL if it's already active

cdloh:
This can happen if you are supplying a custom certificate.

The hostname will go active in SSL and never go pending. Resulting in the provider waiting forever.

### DIFF
--- a/.changelog/2638.txt
+++ b/.changelog/2638.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_custom_hostname: Fix customhostnames from waiting for SSL with `wait_for_ssl_pending_validation` if SSL status is `active`
+```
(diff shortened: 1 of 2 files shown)


.changelog/2638.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_custom_hostname: Fix customhostnames from waiting for SSL with `wait_for_ssl_pending_validation` if SSL status is `active`
```
jacobbednarz, line 2:
```suggestion
resource/cloudflare_custom_hostname: prevent infinite loop when `wait_for_ssl_pending_validation` is set if SSL status is already `active`
```

jacobbednarz:
i inverted the logic check here to return early as the `!= active` combined with the other expression was a little confusing to grok when it should kick in.

jacobbednarz:
thanks @cdloh 🙇

PR #2672, opened 2023-08-08
resource/cloudflare_bot_managment: add Bot Management resource

### LINKED ISSUE #693: Add support for Cloudflare Bot Management
Bot management is a service provided by CloudFlare that detects bots and exposes new variables you can use in custom firewall rules. For more information, see here: https://www.cloudflare.com/products/bot-management/

It is enabled/disabled in the Settings menu of a zone, but it is not exposed via the settings api so it can not be set using this provider at the moment.

I have made a feature request in the underlying cloudflare-go library: https://github.com/cloudflare/cloudflare-go/issues/467

cwlowder:
Cloudflare-Go has been updated to include bot_management, we should update terraform to also include this change.

Depends on https://github.com/cloudflare/cloudflare-go/pull/1363

Closes #693


cwlowder:
Cloudflare-go was updated, decided to do some more testing. Had to update the PR, but I keep getting this error when I run `terraform apply` in my test setup locally:

```
cloudflare_bot_management.bm_bots_config: Creating...
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to cloudflare_bot_management.bm_bots_config, provider "provider[\"registry.terraform.io/cloudflare/cloudflare\"]" produced an
│ unexpected new value: Root resource was present, but now absent.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

@jacobbednarz  thoughts?

jacobbednarz:
@cwlowder that looks like it removed the resource due to it missing from the provider. i just bumped the Go library here so you should be able to re-run these tests now and have them work. otherwise, i can take a look shortly.

jacobbednarz:
ah, i can see a couple of issues here:

- your test cases are referencing `cloudflare_api_shield`
- you're missing a `d.SetId` call which is used as a lookup anchor (tfproviderlint is also complaining about this) and why the test cannot find the resource despite it actually being created.

i'll push up a couple of commits addressing this and you can take a look if you'd like :)

jacobbednarz:
@cwlowder i went ahead and pushed up those fixes. would you mind eyeballing the unentitled test case as i _think_ i know what you were trying to achieve there but do want to be sure before we merge it that was the intention.

acceptance tests are passing

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareBotManagement_" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareBotManagement_SBFM
--- PASS: TestAccCloudflareBotManagement_SBFM (19.36s)
=== RUN   TestAccCloudflareBotManagement_Unentitled
--- PASS: TestAccCloudflareBotManagement_Unentitled (7.19s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	27.512s
```

cwlowder:
My intention is that some zones, like a free zone, will not be able to turn on SBFM or disable auto updating of the ml model. When you attempt to do so, the API will return an error. I want to test that behavior.

jacobbednarz:
perfect. that is the behaviour I accounted for here so I think we're good to merge. 

thanks!

PR #2691, opened 2023-08-17
Add new datasource to get current user info

Cyb3r-Jak3:
Adds a new data source to get the information of the current user


jacobbednarz:
can you strip this PR down to just the changes you're aiming to make here? the current state is going to cause some merge issues given you've put them on multiple branches.

Cyb3r-Jak3:
Done

PR #2744, opened 2023-09-06
Fix `cloudflare_access_identity_provider` incorrectly discards SCIM configuration secret

### LINKED ISSUE #2743: `scim_config.secret` in `cloudflare_access_identity_provider` not populated correctly
### Confirmation

Terraform v1.5.6
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v4.14.0

### Affected resource(s)

cloudflare_access_identity_provider

```hcl
resource "cloudflare_access_identity_provider" "azure_ad" {
  account_id = "REDACTED"
  name       = "AzureAD"
  type       = "azureAD"

  config {
    client_id      = "test"
    client_secret  = "test"
    directory_id   = "test"
    support_groups = true
  }
  scim_config {
    enabled = true
    group_member_deprovision = true
  }
}

output "scim_secret" {
  value = cloudflare_access_identity_provider.azure_ad.scim_config[0].secret
}
```

### Link to debug output

https://gist.github.com/F21/4e163a84f284987a904153931506514c

### Panic output

_No response_

### Expected output

Terraform should output the scim secret. If `scim_config.secret` is marked as sensitive in the resource schema, it should be displayed as `(sensitive)`, like so:

```
scim_secret =  (sensitive)
```

### Actual output

It was displayed as:
```
scim_secret = "**********************************"
```

### Steps to reproduce

1. Run terraform apply.

### Additional factoids

The reason for this bug is outlined in the API documentation: https://developers.cloudflare.com/api/operations/access-identity-providers-add-an-access-identity-provider#response-body

As stated under `result.scim_config.secret` in the response body, the `secret` parameter is only returned upon resource creation for the first time and is redacted in subsequent requests.

In the current implementation, the secret is not stored in the state after the creation of the identity provider. Rather, the resource is read back after creation, which literally reads the redacted secret: `**********************************` and then stores it in the state. This results in the actual SCIM secret being unaccessible and lost forever.

### References

_No response_

F21:
This PR fixes #2743 where the SCIM configuration secret in `cloudflare_access_identity_provider` is incorrectly discarded, rendering the secret unusable with other resources.


F21:
ping @jacobbednarz

jacobbednarz:
at a high level, i don't think this is the right approach to take. we don't want to be munging the state internally from what the API provides as you end up with a difference in user provided configuration and the remote has (which we are abusing here).

instead, we want a similar approach to https://github.com/cloudflare/terraform-provider-cloudflare/blob/e06a8652650db90083f037b4e7189da1cd82b0ec/internal/sdkv2provider/schema_cloudflare_access_identity_provider.go#L130 whereby we explicitly set the state to the concealed string value from the `StateFunc`. if this value is a read once operation according to the service, we want to reflect a similar pattern in the terraform resource itself. 

if a user wants to reuse the value, it should be stored as a local or fetched from a secure backend.

jacobbednarz:
i don't have a IdP to test this with but [c7c0cca](https://github.com/cloudflare/terraform-provider-cloudflare/pull/2744/commits/c7c0cca1d1ce0498acc5e0e62a6fab4d0f538117) should get us majority of the way there (unless there is something unique about this field from the other two fields we do this with)

F21:
Let me pull your changes and test it out.

jacobbednarz:
right, that makes this a little more interesting. 8f4fe7d7 should help then; if the value is present from the create response, add it to the state and ignore it everywhere else.

jacobbednarz:
i went back and forth with this but it looks like this may be the only way to have it conditionally updated (in SDKv2) without compromising on other issues. there is a better approach in the framework however, that is a little way off for most resources.

i did find a couple of `panic`s here though due to the `interface{}` usage so updating those to explicit strings have addressed them.

thanks for getting this one over the line.

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareAccessIdentityProvider_" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareAccessIdentityProvider_OneTimePin
--- PASS: TestAccCloudflareAccessIdentityProvider_OneTimePin (21.39s)
=== RUN   TestAccCloudflareAccessIdentityProvider_OAuth
=== PAUSE TestAccCloudflareAccessIdentityProvider_OAuth
=== RUN   TestAccCloudflareAccessIdentityProvider_OAuthWithUpdate
=== PAUSE TestAccCloudflareAccessIdentityProvider_OAuthWithUpdate
=== RUN   TestAccCloudflareAccessIdentityProvider_SAML
=== PAUSE TestAccCloudflareAccessIdentityProvider_SAML
=== RUN   TestAccCloudflareAccessIdentityProvider_AzureAD
    provider_test.go:217: Skipping acceptance test for default account (f037e56e89293a057740de681ac9abbe). Pending investigation into automating Azure IDP.
--- SKIP: TestAccCloudflareAccessIdentityProvider_AzureAD (0.00s)
=== RUN   TestAccCloudflareAccessIdentityProvider_OAuth_Import
=== PAUSE TestAccCloudflareAccessIdentityProvider_OAuth_Import
=== RUN   TestAccCloudflareAccessIdentityProvider_SCIM_Config_Secret
=== PAUSE TestAccCloudflareAccessIdentityProvider_SCIM_Config_Secret
=== RUN   TestAccCloudflareAccessIdentityProvider_SCIM_Secret_Enabled_After_Resource_Creation
=== PAUSE TestAccCloudflareAccessIdentityProvider_SCIM_Secret_Enabled_After_Resource_Creation
=== CONT  TestAccCloudflareAccessIdentityProvider_OAuth
--- PASS: TestAccCloudflareAccessIdentityProvider_OAuth (8.82s)
=== CONT  TestAccCloudflareAccessIdentityProvider_OAuth_Import
--- PASS: TestAccCloudflareAccessIdentityProvider_OAuth_Import (12.71s)
=== CONT  TestAccCloudflareAccessIdentityProvider_SCIM_Secret_Enabled_After_Resource_Creation
--- PASS: TestAccCloudflareAccessIdentityProvider_SCIM_Secret_Enabled_After_Resource_Creation (19.46s)
=== CONT  TestAccCloudflareAccessIdentityProvider_SCIM_Config_Secret
--- PASS: TestAccCloudflareAccessIdentityProvider_SCIM_Config_Secret (13.61s)
=== CONT  TestAccCloudflareAccessIdentityProvider_SAML
=== CONT  TestAccCloudflareAccessIdentityProvider_OAuthWithUpdate
--- PASS: TestAccCloudflareAccessIdentityProvider_SAML (9.46s)
--- PASS: TestAccCloudflareAccessIdentityProvider_OAuthWithUpdate (10.95s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	97.043s
```

PR #2750, opened 2023-09-07
Mark source owner and repo as force replacement for pages projects

### LINKED ISSUE #2737: Changing pages source has no effect
### Confirmation

Terraform v1.5.6
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v4.13.0

### Affected resource(s)

cloudflare_pages_project

```hcl
# Direct upload Pages project
resource "cloudflare_pages_project" "this" {
  account_id        = local.acct_id
  name              = local.app_ident
  production_branch = local.prod_branch
  build_config {
    destination_dir = "html"
  }
  source {
    type = "github"
    config {
      owner             = "level12"
      repo_name         = "juke-demo-static"
      production_branch = local.prod_branch
      # "*" is the API default, but the CF provider doesn't set it,
      # `plan` constantly shows a change is needed if not explicit.
      preview_branch_includes = ["*"]
    }
  }
```

### Link to debug output

https://gist.github.com/rsyring/074c880f915495c502a21a040957cac6

### Panic output

_No response_

### Expected output

When the plan shows that the github `repo_name` should be changed, and I run apply, I expect the repo_name to be updated in Cloudflare.  Subsequently, if I run plan again, I expect to see no changes.

### Actual output

When the plan shows that the github `repo_name` should be changed, and I run apply, the repo_name value in Cloudflare is not updated.  A subsequent `plan` shows the same change needing to be applied again.

### Steps to reproduce

1. Use a TF config similar to what I provided above.  
2. apply changes
3. Change the github repo_name
4. apply changes
5. Run plan

### Additional factoids

The root problem IMO is that the API accepts these values and ignores them.  A consequence of only PATCH being available for a pages project and not PUT?

Best bet would be to figure out what values the CF API will actually set on a project and, if anything else needs modification in the plan, indicate the resource will need to be destroyed and re-created.

### References

_No response_

Cyb3r-Jak3:
Fixes #2737 

Pages doesn't support changing the git source of a pages project, and you need to delete and recreate. PR marks project owner or repo name as ForceNew.

### DIFF
--- a/.changelog/2750.txt
+++ b/.changelog/2750.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/pages_project: force replace when changing pages source
+```
--- a/docs/resources/pages_project.md
+++ b/docs/resources/pages_project.md
@@ -344,13 +344,13 @@ Required:
 Optional:
 
 - `deployments_enabled` (Boolean) Toggle deployments on this repo. Defaults to `true`.
-- `owner` (String) Project owner username.
+- `owner` (String) Project owner username. **Modifying this attribute will force creation of a new resource.**
 - `pr_comments_enabled` (Boolean) Enable Pages to comment on Pull Requests. Defaults to `true`.
 - `preview_branch_excludes` (List of String) Branches will be excluded from automatic deployment.
 - `preview_branch_includes` (List of String) Branches will be included for automatic deployment.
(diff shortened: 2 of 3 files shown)


jacobbednarz:
if you can run `make docs` to update the registry docs, I'm 👍 here

Cyb3r-Jak3:
Done

PR #2769, opened 2023-09-15
Destroy behaviour for `cloudflare_tunnel_config` resources

samcook:
Hi,

We have noticed an issue when performing a `terraform destroy` of our Cloudflare Tunnel resources (which include `cloudflare_tunnel_config` "server side tunnel configuration") where, if any `cloudflared` processes are still running, terraform will error out complaining that there are active connections when it tries to destroy the `cloudflare_tunnel_config` resource.

The `cloudflare_tunnel` delete function calls `CleanupTunnelConnections` immediately prior to calling `DeleteTunnel`, however the `cloudflare_tunnel_config` delete function (which also just deletes the tunnel entirely) doesn't clean up tunnel connections beforehand.

~This PR adds a call to `CleanupTunnelConnections` to the delete function of `cloudflare_tunnel_config` to make its behaviour match that of `cloudflare_tunnel`.~

PR updated to apply an empty configuration when destroying `cloudflare_tunnel_config` resources.

As an aside, it seems odd that deleting the tunnel configuration in terraform is implemented as deleting the tunnel completely. I realise the Cloudflare REST API only has `GET` and `PUT` endpoints for tunnel configurations. Would it make sense to have it either be a no-op, or perhaps have the delete apply an "empty" configuration of some sort (and let the destruction of the `cloudflare_tunnel` object be responsible for cleaning up the resource)? 🤔


jacobbednarz:
this behaviour is intentional. you need to remove the tunnel configuration prior to deleting the tunnel to avoid potential outages. the assumption that removal of the configuration warrants deletion of the tunnel is also against how many people use this resource so I'm afraid I'm going to have to decline the PR in the current state.

jacobbednarz:
apologies, I just expanded the diff a little more and see what you mean in the last paragraph -- this may be an existing bug. 

i agree that is odd and somewhat surprising as it should only be cleaning up the config attached, not the tunnel.

jacobbednarz:
if you update this to instead remove the config via `UpdateTunnelConfiguration`, that should be enough to address the underlying issue here and make it function as expected.

samcook:
> if you update this to instead remove the config via `UpdateTunnelConfiguration`, that should be enough to address the underlying issue here and make it function as expected.

@jacobbednarz I've updated this PR to apply an empty configuration when destroying `cloudflare_tunnel_config` resources.

docs/resources/tunnel_config.md
```diff
@@ -9,8 +9,6 @@ description: |-
 
 Provides a Cloudflare Tunnel configuration resource.
 
-!> When you delete a tunnel configuration, the tunnel will be deleted. You need to make sure that the tunnel is not in use before deleting the configuration.
```
jacobbednarz, line 12: unfortunately, the file you've edited is a generated artifact of the documentation process. to see how to make changes to the documentation, you're best to check out https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md

jacobbednarz:
@obezuk do you recall why we didn't do this in the first place but instead favoured deleting the parent tunnel?

PR #2780, opened 2023-09-20
Docs: Correct import statement for list_item

rtizzy:
Ran across this while trying to import some existing lists. The documentation seems to indicate list instead of list_item.


rtizzy:
This does not seem to require a changlog entry @jacobbednarz, let me know if there is something else that needs changing here. 

I'd add the label but don't have permissions to do so.

jacobbednarz:

PR #2796, opened 2023-09-27
Fixes #2738

### LINKED ISSUE #2738: cloudflare_pages_project has drift in source.config.preview_branch_includes
### Confirmation

Terraform v1.5.6
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v4.13.0

### Affected resource(s)

cloudflare_pages_project

```hcl
resource "cloudflare_pages_project" "this" {
  account_id        = local.acct_id
  name              = local.app_ident
  production_branch = local.prod_branch
  build_config {
    destination_dir = "html"
  }
  source {
    type = "github"
    config {
      owner             = "level12"
      repo_name         = "juke-demo-static"
      production_branch = local.prod_branch
      # "*" is the API default, but the CF provider doesn't set it,
      # `plan` constantly shows a change is needed if not explicit.
      # preview_branch_includes = ["*"]
    }
  }
}
```

### Link to debug output

https://gist.github.com/rsyring/b337a5f4835e7ac9ea36012c5eea63ce

### Panic output

_No response_

### Expected output

When using config provided above and applying, on the next plan, I would expect to see no changes needed.

### Actual output

When using config provided above and applying, on the next plan, I would expect to see no changes needed.

```
Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  ~ update in-place

  # cloudflare_pages_project.this will be updated in-place
  ~ resource "cloudflare_pages_project" "this" {
        id                = "juke-integration-test"
        name              = "juke-integration-test"

      ~ source {

          ~ config {
              ~ preview_branch_includes       = [
                  - "*",
                ]
                # (8 unchanged attributes hidden)
            }
        }

        # (2 unchanged blocks hidden)
    }

```

### Steps to reproduce

1. Use config similar to what is tried above
2. apply
3. plan

### Additional factoids

_No response_

### References

_No response_

WalshyDev:


internal/sdkv2provider/schema_cloudflare_pages_project.go
```diff
@@ -93,6 +93,7 @@ func resourceCloudflarePagesProjectSchema() map[string]*schema.Schema {
 							Elem: &schema.Schema{
 								Type: schema.TypeString,
 							},
+							Default: []string{"*"},
```
jacobbednarz, line 96: more about S0008 - https://github.com/bflad/tfproviderlint/tree/main/passes/S008

WalshyDev: Oh that's fun... issue has been open since 2021 - https://github.com/hashicorp/terraform-plugin-sdk/issues/142

This was defaulted in the API because we wanted to ensure that with Terraform you still had reasonable defaults and didn't accidentally make a project which would never be triggered by Git.

Think we should remove the API default then?

jacobbednarz: i think you can still default it in the API however, the type should be `Computed` to use the API default if none is provided. if that doesn't work, it's probably just better that the API throws an error if this isn't provided (or mark it as `Required` in the schema).

PR #2814, opened 2023-10-04
Drop separate misspell

ferhatelmas:
Since golangci automatically covers it,
no need for separate installation.

Which is also older than the packaged one.


jacobbednarz:
the separate install was to make it accessible on the `PATH` - is golanglint-ci setup to do this now?

ferhatelmas:
it won't but what is the need to have it in `PATH`? Also, couple of other tools there, seems not used.

jacobbednarz:
we should keep that yes. the current tools end up usable thanks to being in the Go PATH.

jacobbednarz:
using the devcontainer config at https://github.com/cloudflare/terraform-provider-cloudflare/tree/master/.devcontainer

jacobbednarz:
here is an example of why we have the additional dependency.

without your PR

```
root@a70b2b382ed4:/workspaces/terraform-provider-cloudflare# which misspell
/go/bin/misspell
```

whichs means anyone can drop into it and kick off _just_ misspell or use it in their editor. however, with your PR, that no longer exists and you have to invoke it via golanci-lint (` golangci-lint run --disable-all -E misspell`).

```
root@236a1d598dc0:/workspaces/terraform-provider-cloudflare# which golangci-lint
/go/bin/golangci-lint
root@236a1d598dc0:/workspaces/terraform-provider-cloudflare# which misspell
root@236a1d598dc0:/workspaces/terraform-provider-cloudflare#
```

@Cyb3r-Jak3 you're the only other one i see contributing to the devcontainer setup externally, do you have a use case for keeping misspell on the PATH?

jacobbednarz:
@danquack you also just made a contribution to the devcontainer so curious if you have a preference/thoughts here either?

danquack:
Im not using the dev container, just using my system go to execute. I don't have a preference on this one.

Cyb3r-Jak3:
> @Cyb3r-Jak3 you're the only other one i see contributing to the devcontainer setup externally, do you have a use case for keeping misspell on the PATH?

I don't see a problem with removing from the path. I only use it via `make lint` which should still work with this.

jacobbednarz:
let's give this a go and see if it causes anyone pain.

ferhatelmas:
thanks for accepting and merging it

PR #2816, opened 2023-10-06
make user agent string configurable

jaxxstorm:
This allows the user agent string to be configurable, meaning providers that "bridge" this provider don't report Terraform as their origin

### DIFF
--- a/.changelog/2816.txt
+++ b/.changelog/2816.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+provider: Allow user agent string to be configurable via an environment variable
+```
--- a/internal/consts/provider.go
+++ b/internal/consts/provider.go
@@ -89,6 +89,9 @@ const (
 
 	UserAgentDefault = "terraform/%s terraform-plugin-sdk/%s terraform-provider-cloudflare/%s"
 
+	// Configurable user agent string
+	UserAgentEnvVarKey = "CLOUDFLARE_PROVIDER_USER_AGENT"
+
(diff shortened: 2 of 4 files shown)


jacobbednarz:
thanks for the PR! 

overwriting the user agent completely isn't something we want to support given it is used to assess impact of changes and this could potentially make that harder to find. i'm open to allowing appending to the UA though if there is additional context we can append so we still have the context of the internals we are relying on.

jacobbednarz:
also just realised the UA will always send `terraform-plugin-sdk` in the UA but we could be sending a resource from the framework plugin so we'll need to modify that too.

jaxxstorm:
@jacobbednarz this change is primarily so that calls from the pulumi-cloudflare provider are accurately identified. If you have a better mechanism to help identify those rather than via env var, happy to rework this.

jacobbednarz:
i'm fine with appending to the UA, just not allowing a complete overwrite. if you want to rework this to allow that, I'm 👍.

jaxxstorm:
@jacobbednarz if I rework this so only the first section of the UA is modified, would that work?

E.g.

```
UserAgentDefault = "%s/%s terraform-plugin-sdk/%s terraform-provider-cloudflare/%s"
```

Where that first %s is automatically inferred at terraform and it can be overriden as `pulumi` ?

jacobbednarz:
i've opened #2831 which should give us what we need here (as well as an upstream bug for the plugin version missing in some cases). check it out and let me know if that suits your needs!

PR #2831, opened 2023-10-11
Allow setting the User Agent Operator suffix

jacobbednarz:
Within the Terraform ecosystem, there are multiple ways of using the
providers. Generally, it is through the Terraform CLI however, there are
third party tools like Pulumi that rely on the schemas to build their
own integrations.

One issue with this is that the requests they make are lumped with the
label of `terraform/<version>` in the HTTP user agent which doesn’t
allow observablility into those specific requests. To enable better
visibility here, we are introducing a new schema field called a User
Agent Operator suffix. It is configurable in the schema
(`user_agent_operator_suffix`) and via the environment variable
(`CLOUDFLARE_USER_AGENT_OPERATOR_SUFFIX`) and will provide a way for
these third party integrations to denote the source of their requests.

⚠️This *is not* a recommended configuration change for most and when not
explicitly configured will default to the existing `terraform/<version>`
string (as it always has). You should only change this value if you know
what you are doing.

Supersedes #2816


internal/consts/provider.go
```diff
@@ -43,6 +43,12 @@ const (
 	// Default value for the API base path.
 	APIBasePathDefault = "/client/v4"
 
+	// Schema key for the User Agent operator suffix.
+	UserAgentOperatorSuffixSchemaKey = "user_agent_operator_suffix"
+
+	// Environment variable key for the User Agent operator suffix
```
jacobbednarz, line 49:
```suggestion
	// Environment variable key for the User Agent operator suffix.
```

internal/consts/provider.go
```diff
@@ -87,7 +93,7 @@ const (
 	// Deprecated: Use resource specific account ID values instead.
 	AccountIDEnvVarKey = "CLOUDFLARE_ACCOUNT_ID"
 
-	UserAgentDefault = "terraform/%s terraform-plugin-sdk/%s terraform-provider-cloudflare/%s"
+	UserAgentDefault = "terraform-provider-cloudflare/%s terraform-plugin-%s/%s"
```
jacobbednarz, line 97:
```suggestion
```

jacobbednarz:
fyi @jaxxstorm

PR #2841, opened 2023-10-12
resource/cloudflare_api_shield_schema_validation_settings

djhworld:
Adds new resource `resource/cloudflare_api_shield_schema_validation_settings`

Cloudflare-Go has been updated to include support for schema validation settings in API Shield Schema Validation 2.0

This PR updates the terraform provider to support managing these settings in API Shield

Depends on https://github.com/cloudflare/cloudflare-go/pull/1418


docs/resources/api_shield_schema_validation_settings.md
```diff
@@ -0,0 +1,37 @@
...
+- `validation_default_mitigation_action` (String) The default mitigation action used when there is no mitigation action defined on the operation.
+- `zone_id` (String) The zone identifier to target for the resource. **Modifying this attribute will force creation of a new resource.**
+
+### Optional
+
+- `validation_override_mitigation_action` (String) When set, this overrides both zone level and operation level mitigation actions. Defaults to `disable_override`.
+
+### Read-Only
+
+- `id` (String) The ID of this resource.
```
hc2116, line 35: Which ID are we talking about?

internal/sdkv2provider/schema_cloudflare_api_shield_schema_validation_settings.go
```diff
@@ -0,0 +1,28 @@
...
+			Required:    true,
+			ForceNew:    true,
+		},
+		"validation_default_mitigation_action": {
+			Description: "The default mitigation action used when there is no mitigation action defined on the operation",
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"validation_override_mitigation_action": {
+			Description: "When set, this overrides both zone level and operation level mitigation actions",
```
hc2116, line 22: Maybe add that this can only take values "none" and "disable_override" to kill SV 2.0?

internal/sdkv2provider/resource_cloudflare_api_shield_schema_validation_settings.go
```diff
@@ -0,0 +1,123 @@
...
+	if err := d.Set("validation_override_mitigation_action", *settings.OverrideMitigationAction); err != nil {
+		return diag.FromErr(err)
+	}
+
+	// Settings are configured at the zone level so using the zoneID as the ID
+	d.SetId(zoneID)
+	return nil
+}
+
+func resourceCloudflareAPIShieldSchemaValidationSettingsDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
hc2116, line 94: Why is this a delete? We are not deleting anything? Or do you have to give a delete-function here?

internal/sdkv2provider/resource_cloudflare_api_shield_schema_validation_settings_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+		result, err := client.GetAPIShieldSchemaValidationSettings(
+			context.Background(),
+			cloudflare.ZoneIdentifier(rs.Primary.Attributes[consts.ZoneIDSchemaKey]),
+		)
+		if err != nil {
+			return fmt.Errorf("encountered error getting schema validation settings: %w", err)
+		}
+
+		if result.DefaultMitigationAction != "none" {
```
hc2116, line 68: If we want to verify that we receive the default-values, maybe compare against `cloudflareAPIShieldSchemaValidationSettingsDefault().DefaultMitigationAction`?

internal/sdkv2provider/schema_cloudflare_api_shield_schema_validation_settings.go
djhworld: I'd rather not encode this into here, the API rejects invalid values.

internal/sdkv2provider/resource_cloudflare_api_shield_schema_validation_settings.go
djhworld: if a resource is removed from the terraform state, it deletes the resource - as we don't actually "delete" anything this just resets the state of the settings to default. 

It could be argued there should be a `DELETE` endpoint on the API, however this implementation will achieve the same.

jacobbednarz:
acceptance tests all looking good!

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareAPIShieldSchemaValidationSettings_" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareAPIShieldSchemaValidationSettings_Create
--- PASS: TestAccCloudflareAPIShieldSchemaValidationSettings_Create (14.01s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	15.002s
```

thanks!

PR #2849, opened 2023-10-13
Fix: CF list item ID duplication on creation

agesta23:
This PR addresses an issue where multiple cloudflare_list_item resources were assigned the same ID during creation. By refining the logic for determining the item's ID post-creation, we can ensure that each cloudflare_list_item gets its unique and correct ID, thus reflecting its actual state in Cloudflare.


jacobbednarz:
thanks for the time you've put into this PR! however, this really needs to be fixed in the service itself otherwise it will be a bug for other list item types _and_ we don't currently have a way to fetch just a single record. there is already work under way for this internally so once that lands, we'll raise a PR to fix this in the provider.

PR #2851, opened 2023-10-14
Make a note about logging only being supported with skip action

Cyb3r-Jak3:
A small docs update clarifying that logging options are only valid for the skip action. There is an API error of logging options only allowed in the skip action (20018) that is returned if you try with other rules.


.changelog/2851.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+resource/ruleset: Add note that logging is only supported with the skip action
```
jacobbednarz, line 2:
```suggestion
resource/cloudflare_ruleset: Add note that logging is only supported with the skip action
```

PR #2859, opened 2023-10-18
add `casb_findings` to allowed list of logpush datasets

webbgeorge:
The `casb_findings` logpush dataset is not currently supported by the `cloudflare_logpush_job` resource. This PR adds support for this dataset type.

The console support this, and I have also tested with the `/client/v4/accounts/{account_identifier}/logpush/jobs` endpoint to double check the API supports this

Docs about this logpush dataset can be found here: [https://developers.cloudflare.com/logs/reference/log-fields/account/casb_findings/](https://developers.cloudflare.com/logs/reference/log-fields/account/casb_findings/)

I have not raised an issue for this, as it seems fairly minor - happy to do so if requested.

### DIFF
--- a/.changelog/2859.txt
+++ b/.changelog/2859.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/cloudflare_logpush_job: add support for `casb_findings` dataset
+```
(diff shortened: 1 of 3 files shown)


jacobbednarz:
LGTM, thanks!

PR #2868, opened 2023-10-22
add device posture rules data source

### LINKED ISSUE #2864: List device posture rules

1.6.2

### Description

It would be great to have a data source to list device posture rules (or get a specific posture rule based on name) to use as an input rather than need to retrieve that value manually. This is especially important when there are implicit device posture rules created (such as serial number) as part of a teams list that is created.

It seems that this is available in the API now but I didn't see in upstream go code. 

https://developers.cloudflare.com/api/operations/device-posture-rules-list-device-posture-rules

### Use cases

- Get a device posture rule based on a given/known name
- List device posture rules as a set and return ids

```hcl
data "cloudflare_device_posture_rule" "example" {
   account_id = {account_id}
   name = var.name
}

data "cloudflare_device_posture_rules" "example" {
   account_id = {account_id}
}
# returns list of objects with attributes
```

### References

None that I see so far.

broswen:
Adds a data source to return matching device posture rules.
Resolves #2864


internal/sdkv2provider/data_source_device_posture_rules_test.go
```diff
@@ -0,0 +1,63 @@
...
+				Config: testAccCloudflareDevicePostureRulesConfig(rnd, accountID),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(name, consts.AccountIDSchemaKey, accountID),
+					resource.TestCheckResourceAttr(name, "name", rnd),
+					resource.TestCheckResourceAttrSet(name, "rules.0.id"),
+					resource.TestCheckResourceAttr(name, "rules.0.name", rnd),
+					resource.TestCheckResourceAttr(name, "rules.0.type", "file"),
+					resource.TestCheckResourceAttr(name, "rules.0.description", "check for /dev/random"),
+					resource.TestCheckResourceAttr(name, "rules.0.schedule", "1h"),
+					resource.TestCheckNoResourceAttr(name, "rules.0.expiration"),
```
jacobbednarz, line 32: looks like this is failing because the API is returning `null` and that is being stringified as a value.

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareDevicePostureRulesDataSource" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareDevicePostureRulesDataSource
    data_source_device_posture_rules_test.go:18: Step 1/1 error: Check failed: Check 8/8 error: data.cloudflare_device_posture_rules.kvfkelrflt: Attribute 'rules.0.expiration' found when not expected
--- FAIL: TestAccCloudflareDevicePostureRulesDataSource (8.10s)
FAIL
FAIL	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	9.059s
FAIL
make: *** [testacc] Error 1
```

jacobbednarz:
lgtm once we finalise that failing test case 🥇

internal/sdkv2provider/data_source_device_posture_rules_test.go
broswen: It looks like the struct in cloudflare-go isn't a string pointer, and the existing device posture rule resource doesn't handle the null either. Do you think it's ok to just assert it's an empty string?

jacobbednarz: i'm trying to remember the exact go-cty internals here but i _think_ that will do it. let me update the test and make sure that works and if so, roll with that for now and we can fix the struct to handle this better later on.

jacobbednarz: looks like that will do it 👍 

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareDevicePostureRulesDataSource" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareDevicePostureRulesDataSource
--- PASS: TestAccCloudflareDevicePostureRulesDataSource (12.59s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	13.509s
```

jacobbednarz:
acceptance tests all passing
```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareDevicePostureRulesDataSource" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareDevicePostureRulesDataSource
--- PASS: TestAccCloudflareDevicePostureRulesDataSource (12.59s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	13.509s
```

PR #2870, opened 2023-10-23
Add additional test function for vscode

danquack:
Not sure if y'all would want this, but this has made it helpful from my contributions in the AWS provider to test just the one function vs running things on the cli.

If not feel free to just close. I thought about opening an issue, but the code is fairy small its easier just to decline imo

Hashi docs: https://hashicorp.github.io/terraform-provider-aws/debugging/#set-up-launchjson


jacobbednarz:
thanks! happy to add it in if others are using it.

PR #2890, opened 2023-10-27
Document default for `proxied` on `cloudflare_record`

demon:
This was originally requested in #4/fixed in b09befa, but was lost in the move to autogen docs (0813d55).

### DIFF
--- a/docs/resources/record.md
+++ b/docs/resources/record.md
@@ -53,7 +53,7 @@ resource "cloudflare_record" "_sip_tls" {
 - `comment` (String) Comments or notes about the DNS record. This field has no effect on DNS responses.
 - `data` (Block List, Max: 1) Map of attributes that constitute the record value. Conflicts with `value`. (see [below for nested schema](#nestedblock--data))
 - `priority` (Number) The priority of the record.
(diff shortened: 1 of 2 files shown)


jacobbednarz:
thanks for the PR! unfortunately, there is a little confusion here as the value from the schema *does not* default to `false` which is why this doesn't exist in the autogen'd docs (if it did, it would be [automatically appended](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/sdkv2provider/provider.go#L32-L38)). the `proxied` field is not sent at all unless it is explicitly defined.

demon:
Ah, I see what you mean. I wonder if there's a way we can document this that makes sense as to the underlying behavior? When creating a record manually in your browser, the proxied field is checked by default (until you change record type to something that can't be proxied) -- so one could assume the Terraform provider would do the same barring docs to the contrary?

jacobbednarz:
the provider mimics the [API](https://developers.cloudflare.com/api/operations/dns-records-for-a-zone-create-dns-record), not the UI. the reason is that the UI has different behaviour to account for some of the complexities in the APIs and commonly covers sharp edges for users.

PR #2906, opened 2023-11-02
Add deprecation message to tiered_caching in Argo resource

foreseaz:
Deprecating tiered_caching attribute in [Argo resource](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/argo ) and in favor of [cloudflare_tiered_cache](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/tiered_cache) resource to manage tiered cache.

### DIFF
--- a/.changelog/2906.txt
+++ b/.changelog/2906.txt
@@ -1,3 +1,3 @@
 ```release-note:note
-add deprecation message to Argo resource tiered_caching attribute
+resource/cloudflare_argo: `tiered_caching` attribute is deprecated in favour of the dedicated `cloudflare_tiered_cache` resource.
 ```
--- a/.changelog/2906.txt
+++ b/.changelog/2906.txt
@@ -1,0 +1,3 @@
+```release-note:note
+add deprecation message to Argo resource tiered_caching attribute
+```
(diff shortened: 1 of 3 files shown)


internal/sdkv2provider/resource_cloudflare_argo.go
```diff
@@ -27,6 +27,11 @@ func resourceCloudflareArgo() *schema.Resource {
 			Cloudflare Argo controls the routing to your origin and tiered
 			caching options to speed up your website browsing experience.
 		`),
+		DeprecationMessage: heredoc.Doc(fmt.Sprintf(`
+			%s field in argo resource is in deprecation phase that will last for one year (Nov 1, 2024).
```
jacobbednarz, line 31: do you need a year here? or is the next major release fine?

foreseaz: Yes, that's what I'd like to confirm. Next release is good, so what should be the proper wording here?

jacobbednarz: usually something like https://github.com/cloudflare/terraform-provider-cloudflare/blob/03fc7df3bd98ddfc3f351ba1a06ffca1903ab001/internal/sdkv2provider/schema_cloudflare_zone_settings_override.go#L405 is fine. so

`X has been deprecated in favour of ...` is generally suffice. then in the next major release we remove it entirely.

jacobbednarz:
thanks @foreseaz 🥇

foreseaz:
Will we create a separated PR to remove the attribute before next release?

PR #2938, opened 2023-11-14
Update zone.md with secondary type

garrettgalow:
doc update


garrettgalow:
this isn't supported, so needs to be before the doc update

PR #2950, opened 2023-11-18
docs(cloudflare_certificate_pack): Fix import doc

AttyXY:
Specify params to pass to import command, like all [other docs](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/access_mutual_tls_certificate#import)

### DIFF
--- a/docs/resources/certificate_pack.md
+++ b/docs/resources/certificate_pack.md
@@ -96,7 +96,7 @@ Read-Only:
 Import is supported using the following syntax:
 
 ```shell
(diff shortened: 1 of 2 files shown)


AttyXY:
> Oops! It looks like no changelog entry is attached to this PR. Please include a release note as described in https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/changelog-process.md.
> 
> Example:
> 
> ```
> ```release-note:TYPE
> Release note
> ```
> ```
> 
> If you do not require a release note to be included and you have permission, please add the `workflow/skip-changelog-entry` label. Otherwise, a maintainer will add the label or ask you for one when they review the PR.

I don't have permission to add `workflow/skip-changelog-entry` label, but think should be applied here

PR #2955, opened 2023-11-21
DLP-1474: Predefined profiles must be imported for use

jongiddy:
Predefined DLP profiles cannot be created. They must be imported to use them. Modify the example code to show this.


jacobbednarz:

additionally, instead of including the `import` blocks (which are terraform 1.5+ only), can we add a note to the import example? or a note in the docs themselves?

jongiddy:
Updated to modify `resource.tf`

jacobbednarz:
thanks!

PR #2961, opened 2023-11-24
Create origin ca data source

### LINKED ISSUE #2526: Need a data source for `cloudflare_origin_ca_certificate`

```
Terraform v1.5.0
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v4.8.0
```

### Description

Since cloudflare v4.8.0, the `csr` is a required property of the `cloudflare_origin_ca_certificate` resource.
This is preventing us to keep track of our `cloudflare_origin_ca_certificate` with terraform as we don't have the CSR that we used to create those certs, and there is no way to recover them.
In the state, the csr property is set to null and any terraform plan leads to:
```
│ Error: Missing required argument
│ 
│   on ../../modules/seismic-global-service/cloudflare.tf line 259, in resource "cloudflare_origin_ca_certificate" "seismic_origin_server_cert":
│  259: resource "cloudflare_origin_ca_certificate" "seismic_origin_server_cert" {
│ 
│ The argument "csr" is required, but no definition was found.
```

### Use cases

To solve this issue, we would like to see a data source for `cloudflare_origin_ca_certificate` so we can keep track of and reference our `cloudflare_origin_ca_certificate` without having to state the csr. 

```hcl
data "cloudflare_origin_ca_certificate" "example" {
    id = "1234567890"
}
```

### References

The csr requirement was introduced in https://github.com/cloudflare/terraform-provider-cloudflare/pull/2496

danquack:
I love the upgrade to Framework! 

## testing notes
When creating the data source, I am unable to run the tests though because the provider isn't aware of the resource. Is there some sort of merge process? Do the resources have to be moved too? Happy to do the legwork, just looking for what’s you’ve done in the past. 

Close #2526


jacobbednarz:
> I am unable to run the tests though because the provider isn't aware of the resource

the framework structure is individual packages so you need to reference it appropriately for it to run them. examples:

- https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/.github/workflows/acceptance-tests.yml#L68
- `TEST=./internal/framework/service/rulesets TESTARGS='-run "^..." -count 1' make testacc` for filtering (i'll add this to the docs for clarity)

danquack:
> > I am unable to run the tests though because the provider isn't aware of the resource
> 
> the framework structure is individual packages so you need to reference it appropriately for it to run them. examples:
> 
> * https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/.github/workflows/acceptance-tests.yml#L68
> * `TEST=./internal/framework/service/rulesets TESTARGS='-run "^..." -count 1' make testacc` for filtering (i'll add this to the docs for clarity)

How does this work when we’re talking a resource in sdk and data source in framework? Does the resource have to be migrated?

internal/framework/service/origin_ca_certificate/data_source_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"encoding/pem"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/acctest"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+func TestAccCloudflareUserDataSource(t *testing.T) {
```
jacobbednarz, line 17:
```suggestion
func TestAccCloudflareOriginCACertificateDataSource(t *testing.T) {
```

jacobbednarz:
> How does this work when we’re talking a resource in sdk and data source in framework? Does the resource have to be migrated?

they don't have to be in the same plugin for actually using it. they are only treated differently internally and running tests.

i went ahead and updated the schema and tests here to get it working. on the schema side, `csr` and `requested_validity` don't exist (intentionally) and the tests were a bit out because of the wrong datasource name but it's all passing now.

```
TF_ACC=1 go test ./internal/framework/service/origin_ca_certificate -v -run "^TestAccCloudflareOriginCACertificateDataSource" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareOriginCACertificateDataSource
=== PAUSE TestAccCloudflareOriginCACertificateDataSource
=== CONT  TestAccCloudflareOriginCACertificateDataSource
--- PASS: TestAccCloudflareOriginCACertificateDataSource (13.24s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/framework/service/origin_ca_certificate	13.912s
```

danquack:
makes sense. I totally just copy pasta’s those so thanks for making the code not suck so much!

jacobbednarz:
thanks for getting this one over the line! 🏅

Issue #2941, opened 2023-11-15
Missing affected_components filter in cloudflare_notification_policy


avegao:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.5.7
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v4.19.0
+ provider registry.terraform.io/hashicorp/tls v4.0.4

### Affected resource(s)

- cloudflare_notification_policy

### Terraform configuration files

```hcl
resource "cloudflare_notification_policy" "incident_alert" {
  account_id  = cloudflare_account.weplan_analytics.id
  name        = "Incident alert"
  description = "Cloudflare is experiencing an incident"
  enabled     = true
  alert_type  = "incident_alert"

  filters {
    affected_components = ["analytics"] # Example, not implemented
  }

  dynamic "email_integration" {
    for_each = concat(tolist(var.superadmin_emails), tolist(var.admin_emails))

    content {
      id = email_integration.value
    }
  }
}
```

### Link to debug output

https://gist.github.com/avegao/d476f3f75ea58645ecc3fddc769e6270

### Expected output

Successful import or apply

### Actual output

```
Error: failed to set filters: Invalid address to set: []string{"filters", "0", "affected_components"}
```

### Steps to reproduce

1. Create new 'Incident Alert' notification from Dashboard with Affected component filter
2. Import to terraform with `terraform import`

![](https://github.com/cloudflare/terraform-provider-cloudflare/assets/560141/be4800db-cf3c-444b-9130-83408dc59a70)

### Additional factoids

I don't know because with `TF_LOG=DEBUG` returns different error. But is easy to reproduce.

### References

#2901

PR #3009, opened 2023-12-16
Implement the 'affected_components' option for cloudflare_notification_policy

Nmishin:
Fix for https://github.com/cloudflare/terraform-provider-cloudflare/issues/2941


Nmishin:
@jacobbednarz do we know - an affected components ID a persistent?
Because as I realised - I can't set simply `affected_components = ["analytics"]` i need to use id instead, like  `affected_components = ["4c231tkdlpcl"]`

jacobbednarz:
unsure sorry - @jasoncabot may know?

Nmishin:
@jacobbednarz  @jasoncabot  i thinking how to handle that better for end users.
maybe we can have a map somewhere inside provider like  analytics:4c231tkdlpcl and allow to use names instead of id ?

jacobbednarz:
@jasoncabot can we get this in a public API endpoint please? having random string mappings is definitely less than ideal from a testing and usage point of view.

jacobbednarz:
@Nmishin this should be a nicer DX now. the `affected_components` will take a friendly name ("API", "Cache Reserve") and remap it for API calls.

give it a whirl and let me know what you think.

Nmishin:
@jacobbednarz tested one more time in my env - looks super!

jacobbednarz (approved):

PR #3014, opened 2023-12-18
writing down my findings

mikemartin1090:


jacobbednarz:
thanks but this isn't the place for this sort of documentation to exist. the support ticket should take care of getting the documentation updated or constraints added to the schema.

PR #3026, opened 2023-12-23
Migrate list item

Cyb3r-Jak3:
Migrate `cloudflare_list_item` to the plugin framework.

**This is a breaking change because attributes which used `enabled` and `disabled` are now boolean values.**


Cyb3r-Jak3:
I've attempted to make it less breaking by adding a state migrator.

jacobbednarz:
if you can do it without a breaking change, we can include it in the normal release pipeline. otherwise, we'll need to wait for the next major version bump.

Cyb3r-Jak3:
Tested locally, and it just requires users to change the strings to bools in the attributes and the schema is migrated.

Used
```hcl
terraform {
  required_providers {
    cloudflare = {
      source  = "cloudflare/cloudflare"
      version = "4.21.0"
    }
  }
}

variable "account_id" {
  type = string
}

variable "api_token" {
  type = string
}

provider "cloudflare" {
  api_token = var.api_token
}

resource "cloudflare_list" "drift" {
  account_id  = var.account_id
  name        = "test_schema_change"
  description = "test_schema_change"
  kind        = "redirect"
}

resource "cloudflare_list_item" "name" {
  account_id = var.account_id
  list_id = cloudflare_list.drift.id
  redirect {
    source_url       = "https://source.example/"
    target_url       = "https://target.example"
    status_code      = 302
    subpath_matching = "enabled"
    preserve_path_suffix = "disabled"
    # subpath_matching = true
    # preserve_path_suffix = false
  }
}
```

then using the [.terraformrc](https://developer.hashicorp.com/terraform/cli/config/config-file#cli-configuration-file-terraformrc-or-terraform-rc) file to use the dev provider, I am able to switch to booleans with terraform reporting no changes needed and see a schema_version of 1 in the state.

PR #3080, opened 2024-01-26
cloudflare_device_settings_policy.auto_connect is in seconds

klyubin:
This fixes a bug in documentation. Specifically, the documentation states that cloudflare_device_settings_policy.auto_connect is in minutes. In reality though, that field is in seconds. This commit adjusts the documentation accordingly.


jacobbednarz:

jacobbednarz:
@TylerStanish could you confirm the change here is correct?

TylerStanish:
That is correct, it should be seconds

jacobbednarz:
thanks for confirming @TylerStanish 👏

Issue #2690, opened 2023-08-16
Drift in `cloudflare_ruleset` for `http_request_firewall_custom` phase for computed args `id`, `last_updated`, `ref`, `version`


nitrocode:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform 1.5.5
Cloudflare provider 4.12.0 latest

### Affected resource(s)

- cloudflare_ruleset

### Terraform configuration files

```hcl
resource "cloudflare_ruleset" "custom" {
  kind    = "zone"
  name    = "default"
  phase   = "http_request_firewall_custom"
  zone_id = local.zone_id

  # ... 30 rules ...

  # e.g. pre-existing rule
  rules {
    enabled     = true
    action      = "log"
    description = "<snip>"
    expression  = <<-EOT
              <snip>
            EOT
  }

  # added additional rule which caused a drift. could be any rule
  rules {
    enabled     = true
    action      = "log"
    description = "<snip>"
    expression  = <<-EOT
              any(http.request.headers["custom-header"][*] matches ".?")
            EOT
  }
```

### Link to debug output

none

### Panic output

N/A

### Expected output

No perma drift

### Actual output

Perma drift

### Steps to reproduce

1. Add a new rule or modify an existing one
2. terraform plan

### Additional factoids

This happened after we bumped the provider from 3.x to 4.x

We deleted the `cloudflare_ruleset` and reimported.

If we do not make a change, we see `No changes` as expected. Any change we make, and we see a bunch of drift between rules, all affecting arguments that we do not set such as `id` and `version`

We have 30 something rules and we see these computed arguments changing for every rule.

in 4.12.0 cloudflare provider

```hcl
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # cloudflare_ruleset.custom will be updated in-place
  ~ resource "cloudflare_ruleset" "custom" {
        id      = "<snip>"
        name    = "default"
        # (3 unchanged attributes hidden)

      ~ rules {
          ~ id           = "<snip>" -> (known after apply)
          ~ last_updated = "2023-08-18 16:36:15.176813 +0000 UTC" -> (known after apply)
          ~ ref          = "<snip>" -> (known after apply)
          ~ version      = "3" -> (known after apply)
[...]

PR #3082, opened 2024-01-26
Remove unneeded parameters from cloudflare_ruleset

Nmishin:
We start migration to the cloudflare_ruleset and faced the same issue as described here: https://github.com/cloudflare/terraform-provider-cloudflare/issues/2690

I checked `(known after apply)` parameters and realise that `version` and `last_updated` can't be set via [API](https://developers.cloudflare.com/api/operations/createZoneRuleset) and looks like no make sense to have them in the TF state.
 
 Also I added `PlanModifiers` for ID, because if I right understand rule ID shouldn't be changed from outside of terraform - only explicitly.

### DIFF
--- a/internal/framework/service/rulesets/model.go
+++ b/internal/framework/service/rulesets/model.go
@@ -14,7 +14,6 @@ type RulesetResourceModel struct {
 }
 
 type RulesModel struct {
-	Version                types.String                   `tfsdk:"version"`
 	Action                 types.String                   `tfsdk:"action"`
 	ActionParameters       []*ActionParametersModel       `tfsdk:"action_parameters"`
 	Description            types.String                   `tfsdk:"description"`
@@ -22,14 +21,12 @@ type RulesModel struct {
 	ExposedCredentialCheck []*ExposedCredentialCheckModel `tfsdk:"exposed_credential_check"`
 	Expression             types.String                   `tfsdk:"expression"`
 	ID                     types.String                   `tfsdk:"id"`
-	LastUpdated            types.String                   `tfsdk:"last_updated"`
 	Logging                []*LoggingModel                `tfsdk:"logging"`
 	Ratelimit              []*RatelimitModel              `tfsdk:"ratelimit"`
 	Ref                    types.String                   `tfsdk:"ref"`
 }
 
 type ActionParametersModel struct {
-	Version                  types.String                                 `tfsdk:"version"`
 	AdditionalCacheablePorts types.Set                                    `tfsdk:"additional_cacheable_ports"`
 	AutomaticHTTPSRewrites   types.Bool                                   `tfsdk:"automatic_https_rewrites"`
 	AutoMinify               []*ActionParameterAutoMinifyModel            `tfsdk:"autominify"`
--- a/.changelog/3082.txt
+++ b/.changelog/3082.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/cloudflare_ruleset: reduce noise when ruleset changed, by remowing unneeded options `version` and `last_updated`.
+```
--- a/docs/resources/ruleset.md
+++ b/docs/resources/ruleset.md
@@ -469,13 +469,11 @@ Optional:
 - `exposed_credential_check` (Block List) List of parameters that configure exposed credential checks. (see [below for nested schema](#nestedblock--rules--exposed_credential_check))
 - `logging` (Block List) List parameters to configure how the rule generates logs. Only valid for skip action. (see [below for nested schema](#nestedblock--rules--logging))
 - `ratelimit` (Block List) List of parameters that configure HTTP rate limiting behaviour. (see [below for nested schema](#nestedblock--rules--ratelimit))
-- `ref` (String) Rule reference.
+- `ref` (String) The reference of the rule (the rule ID by default). Example: `my_ref`
 
 Read-Only:
 
 - `id` (String) Unique rule identifier.
-- `last_updated` (String) The most recent update to this rule.
-- `version` (String) Version of the ruleset to deploy.
 
 <a id="nestedblock--rules--action_parameters"></a>
 ### Nested Schema for `rules.action_parameters`
@@ -532,7 +530,6 @@ Optional:
 - `status_code` (Number) HTTP status code of the custom error response.
 - `sxg` (Boolean) Turn on or off the SXG feature.
 - `uri` (Block List) List of URI properties to configure for the ruleset rule when performing URL rewrite transformations. (see [below for nested schema](#nestedblock--rules--action_parameters--uri))
-- `version` (String) Version of the ruleset to deploy.
 
 <a id="nestedblock--rules--action_parameters--algorithms"></a>
 ### Nested Schema for `rules.action_parameters.algorithms`
--- a/internal/framework/service/rulesets/resource.go
+++ b/internal/framework/service/rulesets/resource.go
@@ -8,7 +8,6 @@ import (
 	"reflect"
 	"sort"
 	"strings"
-	"time"
 
 	"github.com/cloudflare/cloudflare-go"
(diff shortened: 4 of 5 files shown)


jacobbednarz:
`ref` and `version` are both optionally controllable by end users. `last_updated` cannot be updated but are used by customers so we cannot remove it.

jacobbednarz (changes requested):
thanks for the PR but since these fields are present in the API response so based on the 1:1 policy and conventions, these need to exist in the schema.

internal/framework/service/rulesets/model.go
```diff
@@ -14,22 +14,19 @@ type RulesetResourceModel struct {
...
 	Expression             types.String                   `tfsdk:"expression"`
 	ID                     types.String                   `tfsdk:"id"`
-	LastUpdated            types.String                   `tfsdk:"last_updated"`
 	Logging                []*LoggingModel                `tfsdk:"logging"`
 	Ratelimit              []*RatelimitModel              `tfsdk:"ratelimit"`
 	Ref                    types.String                   `tfsdk:"ref"`
 }
 
 type ActionParametersModel struct {
-	Version                  types.String                                 `tfsdk:"version"`
```
jacobbednarz, line 32: from memory, the top level `version` is read only; this one is optionally provided. @zakcutner can confirm.

Nmishin: yes, top level version is read only, when I tried to explicitly set it i have an error:
```
│ Error: Invalid Configuration for Read-Only Attribute
│ 
│   with cloudflare_ruleset.terraform_managed_resource,
│   on main.tf line 11, in resource "cloudflare_ruleset" "terraform_managed_resource":
│   11:     version     = "3"
│ 
│ Cannot set value for this attribute as the provider has marked it as read-only. Remove the configuration line setting the value.
│ 
│ Refer to the provider documentation or contact the provider developers for additional information about configurable and read-only attributes that are
│ supported.
```

And when I tried to set version for `action_parameters` I have also error, but different:
```
  # cloudflare_ruleset.terraform_managed_resource will be updated in-place
  ~ resource "cloudflare_ruleset" "terraform_managed_resource" {
        id      = "9662a2ab33a84458a1382122723bc411"
        name    = "inspire_firewall_custom"

      ~ rules {
          ~ id           = "0834845279ee4892a27c36778ed48b88" -> (known after apply)
          ~ last_updated = "2024-01-26 15:05:38.398532 +0000 UTC" -> (known after apply)
          ~ ref          = "0834845279ee4892a27c36778ed48b88" -> (known after apply)
          ~ version      = "10" -> (known after apply)
            # (4 unchanged attributes hidden)
        }
      ~ rules {
          ~ id           = "b77ca99c7bef4ce2b902877e4fc4c80f" -> (known after apply)
          ~ last_updated = "2024-01-26 15:05:38.398532 +0000 UTC" -> (known after apply)
          ~ ref          = "b77ca99c7bef4ce2b902877e4fc4c80f" -> (known after apply)
          ~ version      = "1" -> (known after apply)
            # (4 unchanged attributes hidden)

          ~ action_parameters {
              + version = "3"
            }

          ~ logging {
              ~ enabled = true -> false
            }
        }
    }

cloudflare_ruleset.terraform_managed_resource: Modifying... [id=9662a2ab33a84458a1382122723bc411]
╷
│ Error: error updating ruleset with ID "9662a2ab33a84458a1382122723bc411"
│ 
│   with cloudflare_ruleset.terraform_managed_resource,
│   on main.tf line 1, in resource "cloudflare_ruleset" "terraform_managed_resource":
│    1: resource "cloudflare_ruleset" "terraform_managed_resource" {
│ 
│ invalid JSON: unknown field "version"
```

Nmishin:
About 1:1 with API response, i'm fully ok but it looks a little bit strange, and not in all resources we have the same behaviour, for example I randomly get `cloudflare_zone_settings_override` - in the [API response](https://developers.cloudflare.com/api/operations/zone-settings-get-all-zone-settings) we have `modified_on` field but in the [schema](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/sdkv2provider/schema_cloudflare_zone_settings_override.go) it isn't present.

From the other hand - how we can manage changes of `version` and `last_updated`? For version we can calculate a new version like existing version +1, thats ok, but what about `last_updated`? I don't think we can calculate it locally and update the state - it will be difference in milliseconds every time, and in TF plan we will still see the differences ...

Also we can ignore changes of this field with `PlanModifiers` or smth like this, but I can't find a working example ...
And again - if we will somehow ignore that fields and have some additional logic only for following 1:1 policy and conventions this is a make sense?

wdyt?

internal/framework/service/rulesets/model.go
jacobbednarz: yep, that looks right. you're trying to use the `version` attribute in the wrong phase so you get that cryptic error message (there is an internal ticket to address that).

jacobbednarz:
> About 1:1 with API response, i'm fully ok but it looks a little bit strange, and not in all resources we have the same behaviour, for example I randomly get `cloudflare_zone_settings_override` - in the [API response](https://developers.cloudflare.com/api/operations/zone-settings-get-all-zone-settings) we have `modified_on` field but in the [schema](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/sdkv2provider/schema_cloudflare_zone_settings_override.go) it isn't present.

`cloudflare_zone_settings_override` is a very old resource with it's own quirks so let's not use it as a great example of conventions. it is in the process of being deprecated and will be replaced with a better resource soon-ish. if you find others, we should definitely correct them as 1:1 is how we expect them to operate except in some very edge cases (like the bool => string remapped fields for accounting for zero values).

> From the other hand - how we can manage changes of `version` and `last_updated`? For version we can calculate a new version like existing version +1, thats ok, but what about `last_updated`? I don't think we can calculate it locally and update the state - it will be difference in milliseconds every time, and in TF plan we will still see the differences ...

the changes in `version` and `last_updated` are marked as optionally computed so providing you don't set them, the values will reflect what is in the API. it's a matter of finding out what exactly is triggering the behaviour you are seeing here and whether it is something we can address.

> Also we can ignore changes of this field with `PlanModifiers` or smth like this, but I can't find a working example ...

tbh, we shouldn't need to ignore them, `Computed: true` should handle that correctly since the remote updated it and we just need to sync it into the state.

Nmishin:
thanks for the detailed answer!

Ok, i will update MR for remove only
```          
action_parameters {
    version = (known after apply)
}
```
for now, let's wait API/docks update from Cloudflare side.

For the changes in `version` and `last_updated` I will try to find out how to address this, and will create different MR.

jacobbednarz:
👍 to updating `version` to use the `stringplanmodifier`.

when i get some free time, i can also look into the causes of #2690 to see what the HTTP interactions are doing to shed light on what is happening here.

jacobbednarz:
chatting with @zakcutner today about this, i had a 💡 moment and have a better idea on what is happening here. in short, it's two separate issues that end up causing similar behaviour.

the first issue is that for the truly computed fields (such as `ref`), we are relying on `ValueString` which returns the known string value. if string is null or unknown, returns `""`. the problem with this is that it _always_ defines a value whereas terraform should be marking it as unknown during the plan phase and ignoring it. this is a relatively straight forward fix because we can conditionally set this and then the terraform plan will do the right thing.

the second issue is slightly more complex. since we maintain the rules as part of a ruleset as one, whenever we modify one, we end send them all back to the server with any updates found. even if the rule hasn't changed, it gets an updated `version` and `last_updated` values and that is what terraform is showing in the drift (correctly so in this case). given this is a terraform <> service behaviour, i'll have a chat with some folks internally and see what we can come up with as a shorter term solution given the ideal here would be to manage the rules as individual resources but that is still a way off.

jacobbednarz:
superseded by #3091

Issue #1981, opened 2022-10-19
add a data source with list of DNS records


pznamensky:
### Current Terraform and Cloudflare provider version

```
$ terraform -v
Terraform v1.2.6
+ provider registry.terraform.io/cloudflare/cloudflare v3.25.0
...
```

### Description

It would be very useful to have a data source with DNS records to be able to iterate over them in terraform.
Corresponding API call: https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records

### Use cases

We use terraform to manage most of our DNS records in Cloudflare.
For some of them, we'd like to be sure we have enabled Zero Trust and for some of add an external monitoring (e.g. using [this](https://registry.terraform.io/providers/StatusCakeDev/statuscake/latest/docs/resources/uptime_check) provider)
At the moment we have to manually add DNS records to multiple places. But it would be much easier to iterate over DNS records and add monitoring/ZeroTrust/...

### Potential Terraform configuration

```hcl
data "cloudflare_records" "example" {
  zone_id  = var.zone_id
  filter {
    type    = "A"
    proxied = "true"
  }
}
```

jacobbednarz:
the example above is already achievable with https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/data-sources/record for a single record. do you need one where multiple records are returned together? what is the use case here?

pznamensky:
Indeed, the data source `cloudflare_record` already can return a record, but it's not very convenient in case you'd like to add an external monitoring for example for each domain with "A" record or enforce Zero Trust for all of these domains.
I mean, I'd like to iterate over the list of records from the data source instead of querying each record or using terraform variables with predefined list of records.

jacobbednarz:
it's possible but I'm unsure if you'd really want to monitor every A/AAAA record (for example) that a zone has. my guess is that you'll get to the point where you would have records excluded or iterate over later to remove before referencing in your monitoring. explicit is better than implicit comes to mind here. 

I'm happy to accept the request however, it's definitely not at the top of the priority list for now given the niche usage. you're welcome to submit a PR if you'd like to see it sooner.

bikerpatch:
With the recent release of terraform [checks](https://developer.hashicorp.com/terraform/language/checks), the use case of this type of data source is to have Terraform check whether records have been created outside of Terraform, indicating potentially undesired (or potentially malicious) unmanaged configuration

juneeighteen:
[...]

PR #3084, opened 2024-01-29
Add the ability to filter records on the content field

tfalkenberg1:
This PR solves [my problem](https://github.com/cloudflare/terraform-provider-cloudflare/issues/1981#issuecomment-1910473742) described in #1981. 

With that change, I can now filter on the content of my NS records, so I only get one result back and can retrieve the id of the record to use it in my import statement. 

I ran `make tests` and `make docs` but did not execute the acceptance test. However I tested the behavior successfully by building the provider and using it for my use case. 

Please let me know if I missed anything for the PR process.

### DIFF
--- a/internal/sdkv2provider/data_source_record.go
+++ b/internal/sdkv2provider/data_source_record.go
@@ -102,7 +102,6 @@ func dataSourceCloudflareRecordRead(ctx context.Context, d *schema.ResourceData,
 
 	if len(records) != 1 && !contains([]string{"MX", "URI"}, searchRecord.Type) {
 		return diag.Errorf("only wanted 1 DNS record. Got %d records", len(records))
-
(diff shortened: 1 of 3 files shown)


internal/sdkv2provider/data_source_record.go
```diff
@@ -96,6 +102,7 @@ func dataSourceCloudflareRecordRead(ctx context.Context, d *schema.ResourceData,
 
 	if len(records) != 1 && !contains([]string{"MX", "URI"}, searchRecord.Type) {
 		return diag.Errorf("only wanted 1 DNS record. Got %d records", len(records))
+
```
jacobbednarz, line 105:
```suggestion
```

jacobbednarz:
thanks!

PR #3088, opened 2024-01-30
Deku shrub patch 1

Deku-shrub:
After spending 30 minutes wondering why my data source couldn't fetch my CNAME record, it turns out that it's because it defaults to 'A'. There is no support for 'any type' which would be pain even if the functionality existed.

Given how popular CNAME records are, especially in the cloud, I didn't find that A-defaulting behaviour intuitive so I believe it's reasonable to add 'type' into the example here so it can be explicitly considered most times, even if it is an optional attribute.


Deku-shrub:
@jacobbednarz I believe no further changelog info is needed so if you approve, please add the relevant label :)

jacobbednarz:
thanks for the PR but this is already called out in the documentation, see https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/data-sources/record#type. 

given the field is optional, it doesn't make sense to force people to define it if they are using this data source.

PR #3105, opened 2024-02-06
Add Images Variant Resource

### LINKED ISSUE #3104: Cloudflare Images Variants

```
➜  terraform-provider-cloudflare git:(cloudflare_images_variants) terraform -v
Terraform v1.1.5
on darwin_arm64
```

### Description

Support for the Cloudflare Images Variant resource. 

### Use cases

Would like to manage image variant resources alongside other resources (e.g. users). 

```hcl
resource "cloudflare_images_variant" "example" {
  account_id                = "f037e56e89293a057740de681ac9abbe"
  variant_id                = "example"
  never_require_signed_urls = true
  options {
    fit      = "scale-down"
    metadata = "none"
    width    = 500
    height   = 500
  }
}
```

### References

_No response_

hwrdprkns:
Fixes #3104 .


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/agext/levenshtein v1.2.3 // indirect
-	github.com/cloudflare/cloudflare-go v0.87.0
+	github.com/cloudflare/cloudflare-go v0.87.1-0.20240205234129-cca23db4beb9
```
hwrdprkns, line 7: Once new golang library is released i can scrap these changes

jacobbednarz: please scrap them before hand. we can test locally by shimming in the newer version without potentially causing merge conflicts.

jacobbednarz (changes requested):
thanks for the PR here but as the message above where you bumped the maximum allowed resources, the sdkv2provider is now considered frozen and we're only accepting new resources using the `terraform-plugin-framework`. this is especially important for resources that rely on true/false values like you have here to avoid zero value issues with terraform internals.

Issue #3109, opened 2024-02-08
Support for Build Cache of Pages Project


daku10:
### Current Terraform and Cloudflare provider version

terraform: 1.7.3
terraform-provider-cloudflare: 4.24.0

### Description

Build Cache of Cloudflare Pages has released, but the provider doesn't support the feature.
The go API has also released already.
https://github.com/cloudflare/cloudflare-go/issues/1488

### Use cases

Make it possible to configure Build Cache.

### Potential Terraform configuration

```hcl
resource "cloudflare_pages_project" "api-test" {
  account_id = var.account_id
  name       = "example"

  build_config {
    // introduce here
    build_caching   = true
    // ...
  }
  // ...other properties
}
```

PR #3110, opened 2024-02-08
Add `build_caching` on cloudflare_pages_project

daku10:
Closes: #3109 

`d.GetOk` doesn't handle explicit `false`, therefore I use `d.Get`.


internal/sdkv2provider/resource_cloudflare_pages_project_test.go
```diff
@@ -204,6 +205,7 @@ func TestAccCloudflarePagesProject_BuildConfig(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(name, "name", rnd),
 					resource.TestCheckResourceAttr(name, consts.AccountIDSchemaKey, accountID),
+					resource.TestCheckResourceAttr(name, "build_config.0.builld_caching", "true"),
```
jacobbednarz, line 208:
```suggestion
					resource.TestCheckResourceAttr(name, "build_config.0.build_caching", "true"),
```

daku10: Oh, thanks!

PR #3185, opened 2024-03-14
Support MSP_BIZ rate plan

ch0wm3in:
As the title says, support the rate plan called MSP_BIZ


ch0wm3in:
```release-note:note
resource/cloudflare_zone: Add MSP_BIZ rate plan as supported parameter for `plan` .
```

jacobbednarz:
thanks for the PR! unfortunately this isn't a publicly documented subscription nor is the internal/external plan mapping correct so at this point, we cannot accept it without having the public API contract established. 

for now, I will make this as pending documentation and I'd recommend submitting a support ticket to get this documented so we can accept it here.

ch0wm3in:
@jacobbednarz Any news regarding this?

PR #3191, opened 2024-03-16
Fix import for cloudflare_list_item

mbarrien:
There was a small typo introduced in the migration of cloudflare_list_item to the plugin framework in #3026, that breaks imports. The expected id format is a 3 part string separated by slashes, but the import code performs a check at the beginning that checks that it is only 2 parts long (even though the code right after expects 3 parts). This 1-character change fixes this issue.

The current workaround is to pin to a provider version at 4.22 or below.


jacobbednarz:
thanks!

PR #3201, opened 2024-03-20
Add is_deleted filter option to tunnel data source

### LINKED ISSUE #3069: cloudflare_tunnel data source returning inactive tunnel with same name
### Confirmation

Terraform v1.5.7
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v4.20.0
+ provider registry.terraform.io/hashicorp/google v5.12.0
+ provider registry.terraform.io/hashicorp/random v3.6.0
+ provider registry.terraform.io/hashicorp/vault v3.24.0

### Affected resource(s)

data cloudflare_tunnel
https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/data-sources/tunnel

```hcl
data "cloudflare_tunnel" "example" {
  account_id = "f037e56e89293a057740de681ac9abbe"
  name       = "my-tunnel"
}
```

### Link to debug output

None

### Panic output

_No response_

### Expected output

The data lookup should return the active tunnel, not an older inactive and deleted tunnel

### Actual output

The API call used by the data resource returns multiple results, and a single one is selected offering no way to further filter (by status etc)

### Steps to reproduce

Create and delete a tunnel multiple times with the same name, then review the tunnel ID returned in a data resource

### Additional factoids

_No response_

### References

_No response_

MerlinDMC:
This PR solves a problem we're having onboarding our services to Cloudflare. The problematic behaviour is already described in #3069.

With the addition of a `is_deleted` filter we can now ignore soft-deleted entities and get the correct tunnel back from the API.
The filter is utilising the already available [IsDeleted](https://github.com/cloudflare/cloudflare-go/blob/6f98ca9e9077ea6943b9cdc66f2dd7389cc58ade/tunnel.go#L222) parameter of the ListTunnel API.

`make test` and `TESTARGS='-run ^TestAccCloudflareTunnel_Match' make testacc` have been running successfully.

I have left the original functionality intact to not force anyone to adopt setting the parameter.

Closes #3069.

### DIFF
--- a/internal/sdkv2provider/data_source_tunnel.go
+++ b/internal/sdkv2provider/data_source_tunnel.go
@@ -28,6 +28,12 @@ func dataSourceCloudflareTunnel() *schema.Resource {
 				Description: "Name of the tunnel.",
 				ForceNew:    true,
 			},
+			"is_deleted": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "If true, only include deleted tunnels. If false, exclude deleted tunnels. If empty, all tunnels will be included.",
+				ForceNew:    true,
+			},
 			"id": {
 				Type:        schema.TypeString,
 				Computed:    true,
@@ -59,7 +65,13 @@ func dataSourceCloudflareTunnelRead(ctx context.Context, d *schema.ResourceData,
 	accID := d.Get(consts.AccountIDSchemaKey).(string)
 
 	name := d.Get("name").(string)
-	tunnels, _, err := client.ListTunnels(ctx, cloudflare.AccountIdentifier(accID), cloudflare.TunnelListParams{Name: name})
+	params := cloudflare.TunnelListParams{Name: name}
+
+	if v, ok := d.GetOkExists("is_deleted"); ok {
+		params.IsDeleted = cloudflare.BoolPtr(v.(bool))
+	}
+
+	tunnels, _, err := client.ListTunnels(ctx, cloudflare.AccountIdentifier(accID), params)
 	if err != nil {
 		return diag.FromErr(fmt.Errorf("failed to fetch Tunnel: %w", err))
 	}
@@ -71,6 +83,7 @@ func dataSourceCloudflareTunnelRead(ctx context.Context, d *schema.ResourceData,
 
 	d.SetId(tunnel.ID)
 	d.Set("status", tunnel.Status)
+	d.Set("is_deleted", tunnel.DeletedAt != nil)
 	d.Set("id", tunnel.ID)
 	d.Set("tunnel_type", tunnel.TunnelType)
 	d.Set("remote_config", tunnel.RemoteConfig)
--- a/.changelog/3201.txt
+++ b/.changelog/3201.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+datasource/cloudflare_tunnel: Add the option to filter deleted tunnels
+```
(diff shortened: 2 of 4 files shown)


MerlinDMC:
I'm unsure about the deletion `make docs` created for `docs/resources/hyperdrive_config.md`.
Should I drop them from my PR to make it simpler?

internal/sdkv2provider/data_source_tunnel.go
```diff
@@ -59,7 +65,14 @@ func dataSourceCloudflareTunnelRead(ctx context.Context, d *schema.ResourceData,
 	accID := d.Get(consts.AccountIDSchemaKey).(string)
 
 	name := d.Get("name").(string)
-	tunnels, _, err := client.ListTunnels(ctx, cloudflare.AccountIdentifier(accID), cloudflare.TunnelListParams{Name: name})
+	params := cloudflare.TunnelListParams{Name: name}
+
+	if v, ok := d.GetOkExists("is_deleted"); ok {
+		isDeleted := v.(bool)
+		params.IsDeleted = &isDeleted
```
jacobbednarz, line 72:
```suggestion
		params.IsDeleted = cloudflare.BoolPtr(v.(bool))
```

jacobbednarz:
> I'm unsure about the deletion `make docs` created for `docs/resources/hyperdrive_config.md`.
> Should I drop them from my PR to make it simpler?

that looks to be because your branch is out of date with the upstream. if you update it, it should be present.

jacobbednarz:
thanks for the contribution here! 

looking at your test though, it doesn't actually do anything. if i comment out your newly added lines, the tests are still passing. i suspect it's related to https://github.com/cloudflare/terraform-provider-cloudflare/pull/3201/files#diff-3efcea7c71b17ecec6cfec3258b6a7d07026abc92fae723c69bad61941b1faeaR59 not actually doing anything.

internal/sdkv2provider/data_source_tunnel.go
MerlinDMC: I was looking for exactly that and did not find it - thx.

MerlinDMC:
Even after a rebase the `make docs` task does not change for me. Since this looks to be a problem for me locally I will remove the change.
The test case is weird. In theory the 3rd step removes the resource as well which is when you comment out the 2nd step you still see the same result. I added the 2nd step to ensure that the 3rd step runs after the resource has been fully removed from state and got deleted from the API. It basically should guard against a race condition for the delete - I hate flaky tests 😄

MerlinDMC:
I have reworked the tests if you don't mind giving it another glance over.

jacobbednarz:
thanks - i went ahead and tidied up the test assertions here as much of it wasn't needed for a minimal test case and was a little confusing as to what was trying to be achieved there.

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareTunnel_Match" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareTunnel_MatchName
--- PASS: TestAccCloudflareTunnel_MatchName (9.63s)
=== RUN   TestAccCloudflareTunnel_MatchIsDeleted
--- PASS: TestAccCloudflareTunnel_MatchIsDeleted (22.83s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	33.431s
```

Issue #3250, opened 2024-04-17
Updating existing/imported Hyperdrive fails as no ID passed through on Update


siame:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```
tofu -v
OpenTofu v1.6.0
on darwin_arm64

provider registry.opentofu.org/cloudflare/cloudflare v4.29.0
```

### Affected resource(s)

`cloudflare_hyperdrive_config`

### Terraform configuration files

```hcl
resource "cloudflare_hyperdrive_config" "hyperdrives" {
  account_id = "my-id"
  name       = "name"
  origin = {
    database = "postgres"
    password = "password"
    host     = "host"
    port     = 5432
    scheme   = "postgres"
    user     = "user"
  }
}
```

### Link to debug output

None

### Expected output

Should update existing/imported hyperdrive resource with values provided in the `cloudflare_hyperdrive_config` resource block.

### Actual output

```
╷
│ Error: Error updating hyperdrive config
│
│   with cloudflare_hyperdrive_config.hyperdrives,
│   on cloudflare.tf line 24, in resource "cloudflare_hyperdrive_config" "hyperdrives":
│   24: resource "cloudflare_hyperdrive_config" "hyperdrives" {
│
│ required hyperdrive config id is missing
╵
```

### Steps to reproduce

1. Create hyperdrive outside of terraform
2. Write `cloudflare_hyperdrive_config` resource block for that above created hyperdrive
3. Import hyperdrive created in step 1 for the resource block created in step 2 using command `terraform import cloudflare_hyperdrive_config.example <account_id>/<hyperdrive_config_id>`
4. `terraform apply` to apply changes to reconcile between imported hyperdrive and `cloudflare_hyperdrive_config` resource
5. Should error

### Additional factoids

The error message `required hyperdrive config id is missing` originates from https://github.com/cloudflare/cloudflare-go/blob/master/hyperdrive.go#L13, which in the context of our update call, starts here: https://github.com/cloudflare/cloudflare-go/blob/master/hyperdrive.go#L176-L178.

This is happening because our call to `Update` within the terraform-provider does not pass through a `HyperdriveID` parameter: https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/framework/service/hyperdrive_config/resource.go#L130-L143.

[...]

PR #3251, opened 2024-04-17
Include HyperdriveID in Update payload

siame:
Implements #3250

Summarising the issue above:

`terraform apply` using an existing `resource cloudflare_hyperdrive_config {}` returns an error message 

```
required hyperdrive config id
```

This message originates from https://github.com/cloudflare/cloudflare-go/blob/master/hyperdrive.go#L13, which in the context of our update call, starts here: https://github.com/cloudflare/cloudflare-go/blob/master/hyperdrive.go#L176-L178.

This is happening because our call to Update within the terraform-provider does not pass through a HyperdriveID parameter: https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/framework/service/hyperdrive_config/resource.go#L130-L143.

### DIFF
--- a/.changelog/3251.txt
+++ b/.changelog/3251.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_hyperdrive_config: Fix 'HyperdriveID' not included in Update call
+```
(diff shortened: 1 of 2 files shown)


jacobbednarz:
thanks!

Issue #3001, opened 2023-12-11
logpush ownership_challenge: was null, but now has a value


aegiacometti:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

1.5.7

### Affected resource(s)

resource "cloudflare_logpush_job" "job_http_requests" 

### Terraform configuration files

```hcl
This is my setup

- var.cloudflare_zones is a key:value with zone_name:zone_id

- the module construct

module "cloudflare_logpush_job" {
source = "../../../../.terraform/cs_modules/tf-cf-logpush"
for_each = jsondecode(var.cloudflare_zones)

cloudflare_zone_id = each.value
lougpush_name = "tf-logpush-${each.key}"
gcp_bucket_destination_name = module.cloudflare_logpush_gcp_bucket.name
gcp_bucket_destination_url = module.cloudflare_logpush_gcp_bucket.bucket_url
logpull_job_options = local.logpull_job_options
}

- module code

resource "cloudflare_logpush_ownership_challenge" "ownership_challenge" {
zone_id = var.cloudflare_zone_id
destination_conf = var.gcp_bucket_destination_url
}

data "google_storage_bucket_object_content" "challenge_file" {
depends_on = [cloudflare_logpush_ownership_challenge.ownership_challenge]
bucket = var.gcp_bucket_destination_name
name = cloudflare_logpush_ownership_challenge.ownership_challenge.ownership_challenge_filename
}

resource "cloudflare_logpush_job" "job_http_requests" {
depends_on = [data.google_storage_bucket_object_content.challenge_file]
enabled = true
zone_id = var.cloudflare_zone_id
name = var.lougpush_name
destination_conf = var.gcp_bucket_destination_url
logpull_options = var.logpull_job_options
ownership_challenge = data.google_storage_bucket_object_content.challenge_file.content
dataset = "http_requests"
}
```

### Link to debug output

https://gist.github.com/aegiacometti/248460cc6d7d86eece05491333b1bd65

### Expected output

There no other error, since in the 2nd run it works fine. I don't have the log for that one.

### Actual output

│ Error: Provider produced inconsistent final plan
│
│ When expanding the plan for
│ module.cloudflare_logpush_job["paack.io"].cloudflare_logpush_job.job_http_requests
│ to include new values learned so far during apply, provider
[...]

Issue #2794, opened 2023-09-25
Provider produced inconsistent final plan with automated logpush ownership challenge


Arnall:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.5.1
on linux_amd64

    provider registry.terraform.io/cloudflare/cloudflare v4.14.0
    provider registry.terraform.io/hashicorp/google v4.81.0

### Affected resource(s)

-  cloudflare_logpush_job
-  cloudflare_logpush_ownership_challenge

### Terraform configuration files

```hcl
resource "cloudflare_logpush_ownership_challenge" "ownership_challenge" {
  zone_id = module.cloudflare_id.zone_id[0]

  destination_conf = local.destination_conf
}
data "google_storage_bucket_object_content" "challenge_file" {
  bucket = "<bucket_name>"
  name   = cloudflare_logpush_ownership_challenge.ownership_challenge.ownership_challenge_filename
}
resource "cloudflare_logpush_job" "http_logs" {
  zone_id = module.cloudflare_id.zone_id[0]

  dataset                     = "http_requests"
  destination_conf            = local.destination_conf
  enabled                     = true
  filter                      = var.filter
  frequency                   = var.frequency
  logpull_options             = "fields=${local.http_fields}&timestamps=unixnano&CVE-2021-44228=${var.log4j}"
  max_upload_bytes            = var.max_upload_bytes
  max_upload_interval_seconds = var.max_upload_interval_seconds
  ownership_challenge         = data.google_storage_bucket_object_content.challenge_file.content
}
```

### Link to debug output

https://gist.github.com/Arnall/5a5823566c96191a4cbb1ec5ae7f90b4

### Expected output

no error during apply

### Actual output

"cloudflare_logpush_job.http_logs" error: Provider produced inconsistent final plan

When expanding the plan for cloudflare_logpush_job.http_logs to include new
values learned so far during apply, provider
"registry.terraform.io/cloudflare/cloudflare" produced an invalid new value
for .ownership_challenge: was null, but now
cty.StringVal("[redacted]..[redacted]").

This is a bug in the provider, which should be reported in the provider's
own issue tracker.

### Steps to reproduce

Just try to create cloudflare_logpush_job and cloudflare_logpush_ownership_challenge resources with one plan/apply.

### Additional factoids

[...]

Issue #2752, opened 2023-09-08
Provider produced inconsistent final plan with automated logpush ownership challenge


Arnall:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.5.1
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v4.14.0
+ provider registry.terraform.io/hashicorp/google v4.81.0

### Affected resource(s)

- cloudflare_logpush_job
- cloudflare_logpush_ownership_challenge

### Terraform configuration files

```hcl
resource "cloudflare_logpush_ownership_challenge" "ownership_challenge" {
  zone_id = module.cloudflare_id.zone_id[0]

  destination_conf = local.destination_conf
}
data "google_storage_bucket_object_content" "challenge_file" {
  bucket = "<bucket_name>"
  name   = cloudflare_logpush_ownership_challenge.ownership_challenge.ownership_challenge_filename
}
resource "cloudflare_logpush_job" "http_logs" {
  zone_id = module.cloudflare_id.zone_id[0]

  dataset                     = "http_requests"
  destination_conf            = local.destination_conf
  enabled                     = true
  filter                      = var.filter
  frequency                   = var.frequency
  logpull_options             = "fields=${local.http_fields}&timestamps=unixnano&CVE-2021-44228=${var.log4j}"
  max_upload_bytes            = var.max_upload_bytes
  max_upload_interval_seconds = var.max_upload_interval_seconds
  ownership_challenge         = data.google_storage_bucket_object_content.challenge_file.content
}
```

### Link to debug output

https://gist.github.com/Arnall/5a5823566c96191a4cbb1ec5ae7f90b4

### Expected output

no error during apply

### Actual output

"cloudflare_logpush_job.http_logs" error: Provider produced inconsistent final plan

When expanding the plan for cloudflare_logpush_job.http_logs to include new
values learned so far during apply, provider
"registry.terraform.io/cloudflare/cloudflare" produced an invalid new value
for .ownership_challenge: was null, but now
cty.StringVal("[redacted]..[redacted]").

This is a bug in the provider, which should be reported in the provider's
own issue tracker.

### Steps to reproduce

Just try to create  `cloudflare_logpush_job` and `cloudflare_logpush_ownership_challenge` resources with one plan/apply.

### Additional factoids

[...]

PR #3259, opened 2024-04-22
Depend explicitly on the challenge to prevent some errors

gdubicki:
like
```
"registry.terraform.io/cloudflare/cloudflare" produced an invalid new value
for .ownership_challenge: was null, but now cty.StringVal("[redacted]..[redacted]")
```

As recommended by @jacobbednarz in
https://github.com/cloudflare/terraform-provider-cloudflare/issues/2752#issuecomment-1712190497

Workaround for issues like
https://github.com/cloudflare/terraform-provider-cloudflare/issues/2794 https://github.com/cloudflare/terraform-provider-cloudflare/issues/2752 https://github.com/cloudflare/terraform-provider-cloudflare/issues/3001

### DIFF
--- a/docs/resources/logpush_job.md
+++ b/docs/resources/logpush_job.md
@@ -74,6 +74,9 @@ data "aws_s3_bucket_object" "challenge_file" {
 }
 
 resource "cloudflare_logpush_job" "example_job" {
+  depends_on = [
+    cloudflare_logpush_ownership_challenge.ownership_challenge, data.aws_s3_bucket_object.challenge_file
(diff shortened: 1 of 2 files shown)


gdubicki:
> 👋 the `data.aws_s3_bucket_object.challenge_file` in the `depends_on` is redundant as the `challenge_ownership` value already references the datasource value. if you want to update that, we can get this merged

Thanks, corrected that!

jacobbednarz:
thanks!

PR #3260, opened 2024-04-22
FLPROD-1301: Fixing managed headers update bug

WowVeryLogin:
Currently managed headers provider won't change the state of resource not present in the tf state. This means customers can't disable managed headers rules by deleting them from tf config. Fixing it with test.


jacobbednarz:
i believe what you're seeing here is a side effect of zero value handling in the legacy SDKv2. while this works, it may have unintended side effects and bugs down the road. check out #2170 for more in depth details of this as the final mitigation option may be to upgrade to the newer implementation. for now, we're 👌 

acceptance tests are passing
```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareManagedHeaders" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareManagedHeaders
=== PAUSE TestAccCloudflareManagedHeaders
=== CONT  TestAccCloudflareManagedHeaders
--- PASS: TestAccCloudflareManagedHeaders (21.27s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	22.191s
```

PR #3280, opened 2024-05-01
turnstile: rename id to sitekey

### LINKED ISSUE #3093: terraform cloudflare_turnstile_widget Error reading challenge widget when offlabel=true
### Confirmation

Terraform **v1.7.1**

Provider cloudflare/cloudflare **v4.16.0** and **v4.23.0**

### Affected resource(s)

cloudflare_turnstile_widget

```hcl
resource "cloudflare_turnstile_widget" "main" {
  account_id = "05frrr5f66s5ss5568fe"
  name       = "Example Widget"
  domains    = "[mydomain.com]"
  mode       = "non-interactive"
  offlabel   = "true"
}
```

### Link to debug output

https://gist.github.com/MagonBC/6b05a7006e792f9a59e0db31bde5da6f

### Panic output

_No response_

### Expected output

  # cloudflare_turnstile_widget.main will be updated in-place
  ~ resource "cloudflare_turnstile_widget" "main" {
      ~ bot_fight_mode = false -> (known after apply)
        id             = "0xthesite_key"
        name           = "Example Widget"
      ~ offlabel       = false -> true
      ~ region         = "world" -> (known after apply)
      ~ secret         = (sensitive value)
    }

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

cloudflare_turnstile_widget.main: Modifying... [id=thesite_key]
cloudflare_turnstile_widget.main: Modifications complete after 1s [id=thesite_key]

### Actual output

╷
│ Error: Error reading challenge widget
│
│   with cloudflare_turnstile_widget.main,
│   on [main.tf](http://main.tf/) line 1, in resource "cloudflare_turnstile_widget" "main":
│    1: resource "cloudflare_turnstile_widget" "main" {
│
│ required site key missing
╵

### Steps to reproduce

With Turnstile Widget license free: (https://www.cloudflare.com/products/turnstile/)

1. set offlabel=true
2. terraform apply
3. reset offlabel=false
4. terraform apply

With Enterprise License:

5. terraform apply
6. set offlabel=true
7. terraform apply

### Additional factoids

_No response_

### References

_No response_

vasilevalex:
Fixes #3093

### DIFF
--- a/.changelog/3280.txt
+++ b/.changelog/3280.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_turnstile_widget: rename `id` to `sitekey`
+```
--- a/docs/resources/turnstile_widget.md
+++ b/docs/resources/turnstile_widget.md
@@ -34,7 +34,7 @@ resource "cloudflare_turnstile_widget" "example" {
 ### Optional
 
 - `bot_fight_mode` (Boolean) If bot_fight_mode is set to true, Cloudflare issues computationally expensive challenges in response to malicious bots (Enterprise only).
-- `id` (String) The identifier of this resource. This is the site key value.
+- `sitekey` (String) The identifier of this resource. This is the site key value.
 - `offlabel` (Boolean) Do not show any Cloudflare branding on the widget (Enterprise only).
 - `region` (String) Region where this widget can be used.
 
--- a/internal/framework/service/turnstile/model.go
+++ b/internal/framework/service/turnstile/model.go
@@ -4,7 +4,7 @@ import "github.com/hashicorp/terraform-plugin-framework/types"
 
 type TurnstileWidgetModel struct {
 	AccountID    types.String `tfsdk:"account_id"`
-	ID           types.String `tfsdk:"id"`
+	SiteKey      types.String `tfsdk:"sitekey"`
 	Domains      types.Set    `tfsdk:"domains"`
 	Name         types.String `tfsdk:"name"`
 	Secret       types.String `tfsdk:"secret"`
--- a/internal/framework/service/turnstile/resource.go
+++ b/internal/framework/service/turnstile/resource.go
@@ -92,7 +92,7 @@ func (r *TurnstileWidgetResource) Read(ctx context.Context, req resource.ReadReq
 		return
 	}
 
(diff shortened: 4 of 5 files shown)


jacobbednarz (changes requested):
thanks for the PR! unfortunately, it can't be accepted as is. all resources must have the `id` field as part of their schema. where it overlaps with another identifier, behind the scenes we can just reuse that value.

vasilevalex:
Hi @jacobbednarz
Yes, sorry I don't have experience of terraform provider development. I just wanted to use widget for my small Hugo blog hosted on Cloudflare Pages. And with this patch I at least created the widget.

So if you can help me with some questions:
1. What should regular user expect to receive from terraform? 'id' with several comments everywhere in the examples and in the documentation explaining that 'id' is actually sitekey. Or just use 'sitekey' the same way as it is implemented in API?
2. As I can see the code just remapps 'id' <-> 'sitekey'.  But this document says that this "should not" be done: https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/provider-design.md  I'm bit confused.

Thanks!

jacobbednarz:
> 1. What should regular user expect to receive from terraform? 'id' with several comments everywhere in the examples and in the documentation explaining that 'id' is actually sitekey. Or just use 'sitekey' the same way as it is implemented in API?

the `id` is what we want here since it is used as an anchor for terraform operations and it is aliased to `sitekey` in the API calls. can you elaborate on what errors (with debug logs) drew you to make the change? @punkeel might be able to help diagnose how you got there with more context.

> 2. As I can see the code just remapps 'id' <-> 'sitekey'.  But this document says that this "should not" be done: https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/provider-design.md  I'm bit confused.

which part are you referring to? i'm not sure which part you mean however, using the `sitekey` (primary) value as the `id` something we do in most resources since they are unique.

vasilevalex:
Thanks for the answers @jacobbednarz,  I'll make changes.

punkeel:
I suspect this sitekey/id is a red herring, and the real issue is around error handling in Terraform... at least, from what I've seen, every time the API returns an error, the state gets corrupted (which leads to a similar "id/sitekey is missing" error message).

Let me file a small PR, see if that helps

PR #3284, opened 2024-05-03
Fix error handling in cloudflare_turnstile

punkeel:
When we receive an error from the API, we would record an error (not return) and write an invalid State. For some reason, Terraform accepts it... yet it seems to corrupt the tf state.

Add some returns to handle error more cleanly.

Related: #3093 (but unsure if that's the only fix required)

### DIFF
--- a/.changelog/3284.txt
+++ b/.changelog/3284.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_turnstile: Fix error handling corrupting state
+```
--- a/internal/framework/service/turnstile/resource.go
+++ b/internal/framework/service/turnstile/resource.go
@@ -73,6 +73,7 @@ func (r *TurnstileWidgetResource) Create(ctx context.Context, req resource.Creat
 		})
 	if err != nil {
 		resp.Diagnostics.AddError("Error creating challenge widget", err.Error())
+		return
 	}
 
 	data = buildChallengeModelFromWidget(
@@ -96,6 +97,7 @@ func (r *TurnstileWidgetResource) Read(ctx context.Context, req resource.ReadReq
 
 	if err != nil {
 		resp.Diagnostics.AddError("Error reading challenge widget", err.Error())
+		return
 	}
 
 	data = buildChallengeModelFromWidget(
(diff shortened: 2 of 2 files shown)


punkeel:
cc @vasilzhigilei
kudos to @kiannh for the help

lbarthon (approved):

vasilzhigilei:
Did you mean to cc me?

punkeel:
My bad, I meant to cc @vasilevalex.

jacobbednarz:
> When we receive an error from the API, we would record an error (not return) and write an invalid State. For some reason, Terraform accepts it... yet it seems to corrupt the tf state.

terraform [doesn't prevent the state](https://developer.hashicorp.com/terraform/plugin/framework/diagnostics#how-errors-affect-state) from being written as it cannot make decisions about what values should and should not be modified with the error response so it's up to the provider to determine the correct course of action.

i'm on the fence here if we want to return early here as it basically negates anything that happens after this in the resource operation (including the good values). however, given the `id` is used as an anchor, this seems appropriate for the failure mode.

jacobbednarz:
acceptance tests all passing

```
TF_ACC=1 go test ./internal/framework/service/turnstile -v -run "^TestAccCloudflareTurnstileWidget_" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareTurnstileWidget_Basic
=== PAUSE TestAccCloudflareTurnstileWidget_Basic
=== RUN   TestAccCloudflareTurnstileWidget_Minimum
=== PAUSE TestAccCloudflareTurnstileWidget_Minimum
=== RUN   TestAccCloudflareTurnstileWidget_NoDomains
=== PAUSE TestAccCloudflareTurnstileWidget_NoDomains
=== CONT  TestAccCloudflareTurnstileWidget_Basic
--- PASS: TestAccCloudflareTurnstileWidget_Basic (12.13s)
=== CONT  TestAccCloudflareTurnstileWidget_NoDomains
--- PASS: TestAccCloudflareTurnstileWidget_NoDomains (9.71s)
=== CONT  TestAccCloudflareTurnstileWidget_Minimum
--- PASS: TestAccCloudflareTurnstileWidget_Minimum (9.93s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/framework/service/turnstile	33.164s
```

punkeel:
@jacobbednarz thanks & thanks for the link!

The weird part to me is: we declare a `Schema`, yet Terraform accepts to write invalid states with `State.Set`. Why isn't the schema enforced.
This wouldn't prevent the bad error-handling, but it would make it obvious _at the time it is (wrongly) set_, not _used again_. 😕 

See #3285 for other places where we may want to return on errors (with a bonus semgrep)

PR #3286, opened 2024-05-03
Remove deprecated SRV name data fields

janik-cloudflare:
References: <https://developers.cloudflare.com/fundamentals/api/reference/deprecations/#name-related-data-fields-on-srv-dns-records>

### DIFF
--- a/internal/sdkv2provider/schema_cloudflare_record.go
+++ b/internal/sdkv2provider/schema_cloudflare_record.go
@@ -73,10 +73,6 @@ func resourceCloudflareRecordSchema() map[string]*schema.Schema {
 						Type:     schema.TypeString,
 						Optional: true,
 					},
-					"service": {
-						Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


internal/sdkv2provider/schema_cloudflare_record.go
```diff
@@ -73,10 +73,6 @@ func resourceCloudflareRecordSchema() map[string]*schema.Schema {
 						Type:     schema.TypeString,
 						Optional: true,
 					},
-					"service": {
-						Type:     schema.TypeString,
-						Optional: true,
-					},
```
janik-cloudflare, line 79: Moved this one to the NAPTR section since that's now the only type using it

janik-cloudflare:
I believe I'll have to write up a migration at some point (if this is even still relevant then) but why are you interested in this change @markmartirosian? It should make no difference either way

PR #3300, opened 2024-05-10
Add support for cloudchamber_logs Logpush dataset

jseba:
This adds the `cloudchamber_logs` dataset to Logpush as a valid option, as well as updating the docs to list the dataset.


jseba:
cc @jacobbednarz @irvinebroque for review

jacobbednarz:
pending API schema documentation. once that lands, we can get this merged.

PR #3334, opened 2024-05-30
Mirror API record_delimiter defaults

jamesgol:
According to https://developers.cloudflare.com/logs/reference/log-output-options/ the default options for record_suffix are "}\n" and not "}" The defaults cause malformed JSON when there is more than a single record in a time interval.

Perhaps a better default would be to set record_delimiter to "\n" and keep field_delimiter "}" but this change is intended to mirror the published documentation

### DIFF
--- a/.changelog/3334.txt
+++ b/.changelog/3334.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_logpush_job: Mirror API defaults for record_delimiter to include newline
+```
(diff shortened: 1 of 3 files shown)


jacobbednarz:
cc @nuggien @cmackenzie1

cmackenzie1:
This looks good to me and matches public docs for NDJSON. 

Screenshot 2024-05-30 at 1 48 45 PM

cmackenzie1 (approved):

PR #3335, opened 2024-05-31
resource/cloudflare_list_item: handle overlapping redirect `source_url`

jacobbednarz:
To determine the individual list item identifier, we perform an additional HTTP search against the endpoint filtering for the value. In the case of IPs, we use a "starts with" match whereas for redirects, it is a contains. This creates an edge whereby if you have a multiple redirects whereby they share a substring, you cannot differentiate them.

This introduces an additional loop performing an exact match against what the user has provided with all found search results.

NB: For a small dataset, this is manageable however, for 10+ similar items, you are very likely to exhaust your API quota due to additional requests. In the future, the service will improve the search to provide different search match types which we can lean on here instead.

### DIFF
--- a/.changelog/3335.txt
+++ b/.changelog/3335.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_list_item: handle overlapping redirect `source_url`
+```
--- a/internal/framework/service/list_item/resource.go
+++ b/internal/framework/service/list_item/resource.go
@@ -312,6 +312,13 @@ func createListItem(ctx context.Context, client *muxclient.Client, data *ListIte
 			break
 		}
 
+		for _, item := range items {
+			if item.Redirect.SourceUrl == searchTerm {
+				items = []cfv1.ListItem{item}
+				break
+			}
+		}
+
 		//lintignore:R018
 		time.Sleep(time.Duration(attempts) * time.Second)
 	}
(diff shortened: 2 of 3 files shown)


jacobbednarz:
acceptance tests all passing

```
TF_ACC=1 go test ./internal/framework/service/list_item -v -run "^TestAccCloudflareListItem_" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareListItem_Basic
=== PAUSE TestAccCloudflareListItem_Basic
=== RUN   TestAccCloudflareListItem_MultipleItems
--- PASS: TestAccCloudflareListItem_MultipleItems (20.99s)
=== RUN   TestAccCloudflareListItem_Update
--- PASS: TestAccCloudflareListItem_Update (31.46s)
=== RUN   TestAccCloudflareListItem_BadListItemType
--- PASS: TestAccCloudflareListItem_BadListItemType (5.72s)
=== RUN   TestAccCloudflareListItem_ASN
--- PASS: TestAccCloudflareListItem_ASN (14.67s)
=== RUN   TestAccCloudflareListItem_Hostname
--- PASS: TestAccCloudflareListItem_Hostname (25.33s)
=== RUN   TestAccCloudflareListItem_Redirect
--- PASS: TestAccCloudflareListItem_Redirect (16.42s)
=== RUN   TestAccCloudflareListItem_RedirectWithOverlappingSourceURL
--- PASS: TestAccCloudflareListItem_RedirectWithOverlappingSourceURL (33.41s)
=== CONT  TestAccCloudflareListItem_Basic
--- PASS: TestAccCloudflareListItem_Basic (16.86s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/framework/service/list_item	166.078s
```

PR #3351, opened 2024-06-10
Fix importing waiting room events

frankywahl:
There is currently a bug when importing a waiting room event because the
value is out of range. The 3rd element of an array is indexed with value
2, not value 3


jacobbednarz:
thanks!

PR #3356, opened 2024-06-13
chore(stale): set reasonable issue stale days

nitrocode:


jacobbednarz:
thanks but we intentionally want this to 30 days, not 1 year.

nitrocode:
Thanks for the quick response

If you set it lower, then you have fewer issues to close manually /s 😁

If you look at the referenced issue, a number of times users have commented just to remove the stale tag. Is that the workflow desired for users? 

We appreciate all your hard work for what it's worth. It's just discouraging when spending time to write an issue, get community upvotes, only to see it closed due to "activity" which only takes into consideration issue comments. I hope you can understand where I'm coming from.

Please reconsider the 30 days with the team because a 30 day window isn't a friendly number to the community unless issues are getting fixed rapidly.

jacobbednarz:
the linked issue is a great example of why longer term improvements don't make sense in GitHub issues. while i totally agree it is an annoyance in UX, it doesn't result in broken configurations, just difficult to read plans in some scenarios. if you check the now buried PR (as well as #3082 and #3091), there was an attempted fix that uncovered we can't actually do a whole heap on the Terraform side to address this without first having the service team evaluate underlying design changes. parts of the architecture that underpin this system cannot be easily changed. 

service teams and their PMs don't use GitHub for roadmapping prioritisation; the reactions are there to prevent the spam "+1" or "me too" comments so that the issues stay readable and are not used for weighting. we rely on the internal feature request process for building roadmaps not GitHub votes. again, i agree it's not ideal to track them externally but it's currently what is available and doesn't require one person to chase tickets full time instead of actually fixing issues. the ideal would be a shared view of the internal and external requests but that is unfortunately a way off. 

the 30 day window for stale issues is perfectly fine for _bugs_ which are addressed via a different pipeline by service teams. to preempt the next question, no it doesn't exist for features 🙂

PR #3373, opened 2024-06-21
docs: add a skip rule on managed ruleset example

madhifallah:
#### Summary
This PR addresses the need to explicitly highlight the logging parameter in our configuration.

#### Context
Even though the logging parameter is optional, omitting it currently leads to the following error:

```
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to cloudflare_ruleset.dummy_org, provider "provider[\"[registry.terraform.io/cloudflare/cloudflare\](http://registry.terraform.io/cloudflare/cloudflare%5C)"]" produced an unexpected
│ new value: .rules[0].logging: block count changed from 0 to 1.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```
#### Details
To improve clarity and prevent potential configuration issues, this PR:
1. **Documents the logging parameter**: Clearly indicates the presence and role of the logging parameter in the configuration.
2. **Prevents common errors**: By highlighting this parameter, users are less likely to encounter the aforementioned error due to its omission.

#### Testing
- Verified that specifying the logging parameter works as expected.

#### Impact
This change should enhance user experience by reducing the likelihood of configuration errors related to the logging parameter.

### DIFF
--- a/docs/resources/ruleset.md
+++ b/docs/resources/ruleset.md
@@ -38,13 +38,29 @@ resource "cloudflare_ruleset" "magic_transit_example" {
   }
 }
 
-# Zone-level WAF Managed Ruleset
+# Zone-level WAF Managed Ruleset with exceptions
 resource "cloudflare_ruleset" "zone_level_managed_waf" {
   zone_id     = "0da42c8d2132a9ddaf714f9e7c920711"
   name        = "managed WAF"
-  description = "managed WAF ruleset description"
+  description = "managed WAF ruleset with exceptions description"
   kind        = "zone"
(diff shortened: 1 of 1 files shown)


jacobbednarz:
thanks for the PR. we intentionally don't have kitchen sink style (showing all available options and permutations) examples for all our resources and unfortunately, this change falls into that category so we're unable to accept this as a change. 

the field is listed as `Optional` because of the way the schema is represented. it doesn't mean it is necessarily optional for this permutation but instead, it is available for certain configurations and not others.

PR #3413, opened 2024-06-30
feat: add a basic workflow for gritql v5 migration

morgante:
This contains the basic codegen workflow for the v5 upgrade.

Currently it handles inspecting an old Terraform provider schema and generating the required GritQL query. The GritQL query is checked into the repo so it can be run entirely offline.

We still need to add support for attribute/resource renames, but I want to get early feedback on the workflow and tests.

### DIFF
--- a/.grit/.gitignore
+++ b/.grit/.gitignore
@@ -1,0 +1,2 @@
+.gritmodules*
+*.log
--- a/.grit/grit.yaml
+++ b/.grit/grit.yaml
@@ -1,0 +1,3 @@
+version: 0.0.1
+patterns:
+  - name: github.com/getgrit/stdlib#*
--- a/.grit/patterns/terraform_cloudflare_v5.grit
+++ b/.grit/patterns/terraform_cloudflare_v5.grit
@@ -1,0 +1,199 @@
+
+language hcl
+
+pattern terraform_cloudflare_v5() {
+  or {
+  `cors_headers { $block }` => `cors_headers = {
+  $block
+}` where { $block <: within `resource "cloudflare_access_application" $_ { $_ }` },
+  `landing_page_design { $block }` => `landing_page_design = {
+  $block
+}` where { $block <: within `resource "cloudflare_access_application" $_ { $_ }` },
+  `saas_app { $block }` => `saas_app = {
+  $block
+}` where { $block <: within `resource "cloudflare_access_application" $_ { $_ }` },
+  `scim_config { $block }` => `scim_config = {
+  $block
+}` where { $block <: within `resource "cloudflare_access_application" $_ { $_ }` },
+  `exclude { $block }` => `exclude = {
+  $block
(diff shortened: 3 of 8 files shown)


jacobbednarz:
thanks @morgante! i'll take a deeper look at this tomorrow when i'm in front of a computer.

i started the v5 migration on thursday at https://github.com/cloudflare/terraform-provider-cloudflare/blob/next/templates/guides/version-5-upgrade.md so i'll work to integrate these together in the coming days and get the pipeline together so these changes can easily be combined with our manual ones.

morgante:
>  i'll work to integrate these together in the coming days and get the pipeline together so these changes can easily be combined with our manual ones.

Sounds good. The way I set it up is for the auto-generated pattern to be [here](https://github.com/cloudflare/terraform-provider-cloudflare/pull/3413/files#diff-a18d1b11bfa93c7073c4c0764c779548dba863a9e08fb1464dc6a0a4c4b97843R4) but the actual pattern users would invoke composes it [here](https://github.com/cloudflare/terraform-provider-cloudflare/pull/3413/files#diff-0be604bbe4be5bbfe229a53c254fdffae535adcb93b6e3fa5d938c5b21e28308R14).

jacobbednarz:
i ended up creating a new PR for this one (in #3414) as i couldn't edit this branch on the fork which caused some issues. 

i dropped the JSON to avoid repository bloat and the upgrade notes markdown file which we can add in once the registry one is finalised.

morgante:
> i dropped the JSON to avoid repository bloat and the upgrade notes markdown file which we can add in once the registry one is finalised.

FYI the upgrade notes markdown I was using actually also functions as tests for `grit patterns test` to sanity check / validate the final migration. I think you'll probably want to have a wrapper GritQL markdown like that.

jacobbednarz:
i'm happy to include it but don't want two versions of the upgrade notes in parallel. are we able to point it at the existing upgrade notes? i'm working in your two improvements to it now.

morgante:
> i'm happy to include it but don't want two versions of the upgrade notes in parallel. are we able to point it at the existing upgrade notes? i'm working in your two improvements to it now.

Right now it has to be located inside `.grit/patterns` for the CLI to pick it up.

jacobbednarz:
alright - i'll have a think about our options here to not duplicate/missing items between them but still remain compatible with testing.

morgante:
Since it looks like Hashicorp [requires](https://developer.hashicorp.com/terraform/registry/providers/docs#directory-structure) the directory structure you're using, I think it's probably a reasonable enhancement for Grit to be able to pick up patterns elsewhere. I'll try to add that this week: https://github.com/getgrit/gritql/issues/392

jacobbednarz:
thanks - that would be helpful. i was also thinking maybe just a symlink from the registry generated artifact too which should work around this and not have duplication in the docs.

PR #3415, opened 2024-07-01
build(grit): invoke pattern at the end of the file

jacobbednarz:
cc @morgante

### DIFF
--- a/.grit/patterns/terraform_cloudflare_v5.grit
+++ b/.grit/patterns/terraform_cloudflare_v5.grit
@@ -197,3 +197,5 @@ pattern terraform_cloudflare_v5() {
 }` where { $block <: within `resource "cloudflare_zone_settings_override" $_ { $_ }` }
   }
 }
+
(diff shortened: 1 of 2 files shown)


.grit/patterns/terraform_cloudflare_v5.grit
```diff
@@ -196,4 +196,6 @@ pattern terraform_cloudflare_v5() {
   $block
 }` where { $block <: within `resource "cloudflare_zone_settings_override" $_ { $_ }` }
   }
-}
\ No newline at end of file
+}
+
+terraform_cloudflare_v5()
```
morgante, line 201: FYI the way I set it up before was intended for this generated file *not* to be invoked. The markdown file would be actual pattern that users invoke, but we can adjust later once that is restored.

jacobbednarz: ah got it. makes sense 👍

PR #3423, opened 2024-07-02
feat: add a PoC pattern for extracting tests

morgante:
This might require some adjustment, but should mostly do what you need.


jacobbednarz:
superseded by #3431

PR #3443, opened 2024-07-07
Add gateway_categories datasource

brennoo:
This PR is trying to add `cloudflare_gateway_categories` data source, using the SDK v2 

for some reason I could not understand in my `build-dev` is hardcoded to a wrong API base path, example:

```
Error: Failed to fetch Gateway Categories
│
│   with data.cloudflare_gateway_categories.this,
│   on data.tf line 6, in data "cloudflare_gateway_categories" "this":
│    6: data "cloudflare_gateway_categories" "this" {
│
│ GET
│ "https://api.cloudflare.com/client/accounts/f037e56e89293a057740de681ac9abbegateway/categories":
│ 404 Not Found {"success":false,"errors":[{"code":7000,"message":"No route
│ for that URI"}]}
```

other resources working fine, setting `CLOUDFLARE_API_BASE_PATH` doesn't affect `cloudflare_gateway_categories`


internal/framework/provider/provider.go
```diff
@@ -194,6 +196,9 @@ func (p *CloudflareProvider) Configure(ctx context.Context, req provider.Configu
 	} else {
 		basePath = utils.GetDefaultFromEnv(consts.APIBasePathEnvVarKey, consts.APIBasePathDefault)
 	}
+	// Adding a trailing slash for ResolveReference to work
+	basePath = strings.TrimSuffix(basePath, "/") + "/"
```
jacobbednarz, line 200: this shouldn't be necessary; we already have resources and datasources that rely on this functionality.

brennoo: hello Jacob, thanks for reviewing my first PR here as well

do I need to change something at the datasource?  without this I'm getting the wrong basepath, example:

╷
│ Error: Failed to fetch Gateway Categories
│
│   with data.cloudflare_gateway_categories.this,
│   on data.tf line 6, in data "cloudflare_gateway_categories" "this":
│    6: data "cloudflare_gateway_categories" "this" {
│
│ GET
│ "https://api.cloudflare.com/client/accounts/f037e56e89293a057740de681ac9abbe/gateway/categories":
│ 404 Not Found {"success":false,"errors":[{"code":7000,"message":"No route
│ for that URI"}]}

jacobbednarz: that looks like you're setting the value in your environment or provider incorrectly. i would recommend insetting all the CLOUDFLARE_* environment variables. 

the existing resources work fine for me with the correct base path with nothing else set.

brennoo: it seems this is the only one using `client.V2` - without this change all resource works except this one

`CLOUDFLARE_API_BASE_PATH=/client/v4 terraform plan # still the same`

`CLOUDFLARE_API_BASE_PATH=/client/v4/ terraform plan # this works`

jacobbednarz: good point on the SDK v2 here. however, we cannot add this here without making the other API calls return a double slash in the URLs (which we don't want). for compatibility, we will need to restrict this only to the `cfv2Options` on L203 instead (not the entire `basePath` variable).

brennoo: would it be OK to just move it after L202 as this is the last time `basePath` is used for v1 or add a `basePathV2` variable for that?

brennoo: I've added `basePathV2` to be on the safe side

jacobbednarz:
acceptance tests all passing

```
TF_ACC=1 go test ./internal/framework/service/gateway_categories -v -run "^TestAccCloudflareGatewayCategories_" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareGatewayCategories_DataSource
=== PAUSE TestAccCloudflareGatewayCategories_DataSource
=== CONT  TestAccCloudflareGatewayCategories_DataSource
--- PASS: TestAccCloudflareGatewayCategories_DataSource (3.71s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/framework/service/gateway_categories	4.798s
```

thanks @brennoo 👏

PR #3463, opened 2024-07-15
Fix partial risk behavior configuration bug

inanna-malick:
Previously, if a user's account supported risk behaviors A, B and C but only behaviors A and B were configured, a 'provider produced inconsistent results after apply' error occurred.

This is because the cloudflare API call for user risk behaviors returns the status of all behaviors.

This PR changes the user risk behavior terraform provider to - on read, create and update calls - prune the API call results to only include the behaviors the terraform config includes

A test confirming that the fix works has been included

### DIFF
--- a/.changelog/3463.txt
+++ b/.changelog/3463.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-fix resource/cloudflare_risk_behavior bug where partial definition of risk behaviors resulted in a provider error
+resource/cloudflare_risk_behavior: fix bug where partial definition of risk behaviors resulted in a provider error
 ```
--- a/.changelog/3463.txt
+++ b/.changelog/3463.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+fix resource/cloudflare_risk_behavior bug where partial definition of risk behaviors resulted in a provider error
+```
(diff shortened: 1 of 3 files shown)


jacobbednarz:
this seems like a case where someone is only partially managing the resource in terraform. is that the case?

inanna-malick:
> this seems like a case where someone is only partially managing the resource in terraform. is that the case?

Yes, but even in the case where someone is fully managing the resource in terraform, we sometimes add new risk behavior types, which would put them in a partially-managed state (which we want to handle gracefully).

jacobbednarz:
acceptance tests passing

```
TF_ACC=1 go test ./internal/framework/service/risk_behavior/ -v -run "^TestAccCloudflareRiskBehavior_" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareRiskBehavior_Partial
=== PAUSE TestAccCloudflareRiskBehavior_Partial
=== RUN   TestAccCloudflareRiskBehavior_Basic
=== PAUSE TestAccCloudflareRiskBehavior_Basic
=== CONT  TestAccCloudflareRiskBehavior_Partial
--- PASS: TestAccCloudflareRiskBehavior_Partial (2.82s)
=== CONT  TestAccCloudflareRiskBehavior_Basic
--- PASS: TestAccCloudflareRiskBehavior_Basic (1.97s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/framework/service/risk_behavior	6.140s
```

PR #3486, opened 2024-07-22
refactor(grit): move around Grit patterns to support centralised entrypoint

jacobbednarz:
Moves the grit patterns around to better organise the various manual and automatic migrations we have for v5.

cc @morgante for the upcoming integration of the workflow.


morgante (commented):
Looks good, for the workflow you just need to change [this](https://github.com/cloudflare/terraform-provider-cloudflare/blob/next/.grit/workflows/stainless/workflow.ts#L144) and [this](https://github.com/cloudflare/terraform-provider-cloudflare/blob/next/.grit/workflows/stainless/workflow.ts#L135).

jacobbednarz:
thanks - i also removed the call to the pattern now that we're calling them all in the one place.

PR #3498, opened 2024-07-24
fix: change mTLS associated hostnames to a set

### LINKED ISSUE #3436: Perpetual drift from Access mTLS cert associated hostname reordering
### Confirmation

- [X] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [X] I have searched the issue tracker and my issue isn't already found.

Terraform v1.9.1
on darwin_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v4.36.0

### Affected resource(s)

- cloudflare_access_mutual_tls_certificate

```hcl
# Configure the Cloudflare provider using the required_providers stanza
# required with Terraform 0.13 and beyond. You may optionally use version
# directive to prevent breaking changes occurring unannounced.
terraform {
  required_providers {
    cloudflare = {
      source  = "cloudflare/cloudflare"
      version = "~> 4.0"
    }
  }
}

variable "account_id" {
  type    = string
  default = "<account id>"
}

variable "certificate" {
  type    = string
  default = <<EOF
<self signed cert generated with openssl>
  EOF
}

resource "cloudflare_access_mutual_tls_certificate" "test" {
  name        = "test"
  account_id  = var.account_id
  certificate = var.certificate
  associated_hostnames = [
    "<hostname 1>",
    "<hostname 2>",
  ]
}
```

### Link to debug output

https://gist.github.com/bporter816/a339f01e2c2ea6107279e589bb5d21b5

### Panic output

_No response_

### Expected output

When the ordering of associated hostnames returned from the Cloudflare API differs from the ordering in Terraform, we expect one of two outcomes:
1. Applying Terraform (or doing a PUT against the API) with the same list of hostnames in a different order should actually reorder them on the server side so that future runs don't produce any drift.
2. Terraform should not care about the order of the returned hostnames and report no drift if the order differs from what the API returns.

### Actual output

If the ordering differs, we get a permanent drift, like so:
```

  # cloudflare_access_mutual_tls_certificate.test will be updated in-place
  ~ resource "cloudflare_access_mutual_tls_certificate" "test" {
      ~ associated_hostnames = [
          - "<hostname 1>",
            "<hostname 2>",
          + "<hostname 1>",
        ]
        id                   = "<id>"
        name                 = "test"
    }

```
Both applying this change through Terraform and directly through the Cloudflare API does not actually reflect the new ordering on the server side, and the drift continues to show up in subsequent runs. We could work around this if the ordering was deterministic and something we could account for in our Terraform config (if it is, please advise and we can take that approach) but we notice that even for the same list of hostnames, they are returned in a different order on different certs.

### Steps to reproduce

1. Create an Access mTLS cert with a multiple associated hostnames
2. Change the ordering of the hostnames in the Terraform configuration to a different order than the Cloudflare API returns them in
3. This results in drift which will show up in subsequent plan/apply and will not go away upon successful apply

### Additional factoids

_No response_

### References

_No response_

BSFishy:
The API doesn't order mTLS associated hostnames in any meaningful way. This can cause drift as seen in #3436. We can mitigate this by ignoring associated hostname changes, when the only thing that has changed is the ordering.

I tried a number of other approaches including sorting the list that is returned from the API and sorting the incoming associated hostname list, but none of that type of stuff worked.

fixes #3436


ajholland:
Looks good to me!

.changelog/3498.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_access_mutual_tls_certificate: sort associated hostnames alphabetically
```
bporter816, line 2: Thanks for the fix! Just a nit that this changelog still says the hostnames are being sorted alphabetically, which isn't the case anymore.

BSFishy: Nice catch! Fixed that :)

internal/sdkv2provider/schema_cloudflare_access_mutual_tls_certificate.go
```diff
@@ -37,7 +48,31 @@ func resourceCloudflareAccessMutualTLSCertificateSchema() map[string]*schema.Sch
...
+					// size discrepency (an entry was added or removed)
+					return false
+				}
+
+				for _, hostname := range oldHostnames {
+					if !includes(newHostnames, hostname) {
+						// size is the same, but before and after contain different entries (an entry was renamed)
+						return false
+					}
+				}
```
BSFishy, line 71: This operation is only performed on the `oldHostnames` slice. I think that since we know their lengths are the same and if old contains something that new doesn't, then that inherently means an entry was renamed and therefore we don't need to perform this operation on `newHostnames`, but I want to call this out to make sure my thinking is correct :P

jacobbednarz (changes requested):
the changes you're making here are basically what [`TypeSet`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typeset) is built for. i'd suggest we swap to that instead and remove the additional schema functions here.

BSFishy:
> the changes you're making here are basically what [`TypeSet`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typeset) is built for. i'd suggest we swap to that instead and remove the additional schema functions here.

I thought what I was doing felt wrong! Thanks for pointing me in the right direction. Updated and ready for re-review :)

bporter816:
Hi @jacobbednarz, can we get this reviewed for the next release?

jacobbednarz:
the acceptance tests are failing here (and on `master`) so we'll need to work out what needs to change to support it.

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareAccessMutualTLS" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareAccessMutualTLSBasic
    resource_cloudflare_access_mutual_tls_certificate_test.go:75: Step 1/2 error: Error running apply: exit status 1
        
        Error: error creating Access Mutual TLS Certificate for accounts "f037e56e89293a057740de681ac9abbe": error from makeRequest: access.api.error.certificate_required (12032)
        
          with cloudflare_access_mutual_tls_certificate.jndisxqick,
          on terraform_plugin_test.tf line 12, in resource "cloudflare_access_mutual_tls_certificate" "jndisxqick":
          12: resource "cloudflare_access_mutual_tls_certificate" "jndisxqick" {
        
--- FAIL: TestAccCloudflareAccessMutualTLSBasic (0.86s)
=== RUN   TestAccCloudflareAccessMutualTLSBasicWithZoneID
    resource_cloudflare_access_mutual_tls_certificate_test.go:118: Step 1/2 error: Error running apply: exit status 1
        
        Error: error creating Access Mutual TLS Certificate for zones "0da42c8d2132a9ddaf714f9e7c920711": error from makeRequest: access.api.error.certificate_required (12032)
        
          with cloudflare_access_mutual_tls_certificate.sjkcssabds,
          on terraform_plugin_test.tf line 12, in resource "cloudflare_access_mutual_tls_certificate" "sjkcssabds":
          12: resource "cloudflare_access_mutual_tls_certificate" "sjkcssabds" {
        
--- FAIL: TestAccCloudflareAccessMutualTLSBasicWithZoneID (0.72s)
FAIL
FAIL	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	2.365s
FAIL
make: *** [testacc] Error 1
```

BSFishy:
I think I agree with @bporter816 that it looks like the `CLOUDFLARE_MUTUAL_TLS_CERTIFICATE` variable isn't being set. Apart from that, the file mentioned seems to be fine to me. To my understanding, `TestCheckResourceAttrSet` checks if an attribute is set, rather than if it contains a specific value (which for the certificate is fine I think), and lists should automatically be coerced to sets where necissary, according to [these docs](https://developer.hashicorp.com/terraform/language/functions/toset). Let me know if any of this sounds wrong and I can take another look

bporter816:
Ah yes, I think I misread the docs on that function. Another thing I wasn't sure about is if things like this indexing still work: https://github.com/cloudflare/terraform-provider-cloudflare/blob/4708b6283f183eea0b9b8220ef0ea36558bb04a9/internal/sdkv2provider/resource_cloudflare_access_mutual_tls_certificate_test.go#L131

jacobbednarz:
nice find. after fixing that (and adding a check in the tests for future runners), this is all now passing

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareAccessMutualTLS" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareAccessMutualTLSBasic
--- PASS: TestAccCloudflareAccessMutualTLSBasic (9.57s)
=== RUN   TestAccCloudflareAccessMutualTLSBasicWithZoneID
--- PASS: TestAccCloudflareAccessMutualTLSBasicWithZoneID (9.22s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider	19.298s
```

PR #3581, opened 2024-08-08
update workflow images to pin to ubuntu-24.04 to fix out of storage errors

jafowler:
there is currently a bug with ubuntu-latest workflow image that is causing our workflows to run out of storage
image
https://github.com/actions/runner-images/issues/10386#issuecomment-2271433971

I've verified that the same go binaries are available in 24.04 as 22.04 (which latests is pointed at currently).
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#go
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#go

Looks like @jacobbednarz was looking into this yesterday as well but the cleanups didn't help:
https://github.com/cloudflare/terraform-provider-cloudflare/commit/2dade830a00b4629d07b33ce65c87d146e8110c4

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -19,7 +19,7 @@ on:
 jobs:
   acceptance-tests:
     name: Acceptance Tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
 
     env:
       CLOUDFLARE_ACCOUNT_ID: f037e56e89293a057740de681ac9abbe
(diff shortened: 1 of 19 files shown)


mgirouard (commented):
lgtm

mgirouard (approved):

jafowler:
Jk, @jacobbednarz changed DID help. The job we were pointed at that failed was the CHANGELOG action which seems to be failing for quite some some. Can we clean this up if its not being used Jacob?

Issue #1919, opened 2022-09-21
[3.24.0] Regression - Cannot specify both api_user_service_key and api_token


uncycler:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.2.9
Cloudflare provider v3.24.0

### Affected resource(s)

- cloudflare_origin_ca_certificate

### Terraform configuration files

```hcl
resource "cloudflare_origin_ca_certificate" "test" {}
```

### Link to debug output

N/A

### Expected output

Resource `cloudflare_origin_ca_certificate` requires `api_user_service_key` but not other resources.

### Actual output

│ Error: Invalid combination of arguments
│ 
│   with provider["registry.terraform.io/cloudflare/cloudflare"],
│   on <empty> line 0:
│   (source code not available)
│ 
│ "api_user_service_key": only one of
│ `api_key,api_token,api_user_service_key` can be specified, but
│ `api_token,api_user_service_key` were specified.

### Steps to reproduce

1. Try to use `cloudflare_origin_ca_certificate` and `cloudflare_record` in the same Terraform configuration.

jacobbednarz:
this is not a regression. you only need to provide one of these auth methods when making API calls and sending both makes the internals difficult (and time consuming) to debug.

uncycler:
Hmm, ok. Quite a breaking change. So, I guess we need to create an alias provider for `cloudflare_origin_ca_certificate`.

btw, documentation still state: `This resource requires you use your Origin CA Key as the [api_user_service_key](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/index#api_user_service_key), in conjunction with an api_token or email and api_key.`

jacobbednarz:
an alias provider will also work with separate credentials. 

feel free to send over a PR updating the docs and I'll get it merged. check out https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/contributing/registry-documentation.md on how the documentation generation works.

eyedvabny:
[...]

PR #3636, opened 2024-08-14
Allow empty api_token for ENV authentication

petero-dk:
A PR to implement the fix described here:
https://github.com/cloudflare/terraform-provider-cloudflare/issues/1919#issuecomment-2132204631

This would allow us to specify BOTH `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_API_USER_SERVICE_KEY` and then force "remove" the token from the aliased provider that we choose like this:

```tf
provider "cloudflare" {
  alias                = "user_service_key"
  api_token = "" #explicitly set to ""
}

provider "cloudflare" {
  api_user_service_key = "" #explicitly set to ""
}
```

When using variables for our Cloudflare api_token the values are stored in the state file. Using environment variables like this means the api token never leaves the build environment.

### DIFF
--- a/internal/framework/provider/provider.go
+++ b/internal/framework/provider/provider.go
@@ -112,7 +112,7 @@ func (p *CloudflareProvider) Schema(ctx context.Context, req provider.SchemaRequ
 				MarkdownDescription: fmt.Sprintf("The API Token for operations. Alternatively, can be configured using the `%s` environment variable. Must provide only one of `api_key`, `api_token`, `api_user_service_key`.", consts.APITokenEnvVarKey),
 				Validators: []validator.String{
 					stringvalidator.RegexMatches(
-						regexp.MustCompile(`[A-Za-z0-9-_]{40}`),
+						regexp.MustCompile(`[A-Za-z0-9-_]{40}|^$`),
 						"API tokens must be 40 characters long and only contain characters a-z, A-Z, 0-9, hyphens and underscores",
 					),
 				},
--- a/internal/sdkv2provider/provider.go
+++ b/internal/sdkv2provider/provider.go
@@ -107,7 +107,7 @@ func New(version string) func() *schema.Provider {
 					Type:         schema.TypeString,
 					Optional:     true,
 					Description:  fmt.Sprintf("The API Token for operations. Alternatively, can be configured using the `%s` environment variable. Must provide only one of `api_key`, `api_token`, `api_user_service_key`.", consts.APITokenEnvVarKey),
-					ValidateFunc: validation.StringMatch(regexp.MustCompile("[A-Za-z0-9-_]{40}"), "API tokens must be 40 characters long and only contain characters a-z, A-Z, 0-9, hyphens and underscores"),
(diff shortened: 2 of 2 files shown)


jacobbednarz:
thanks for the PR but we're going to pass on this one. we intentionally don't want to be mixing auth schemes due to the difficulty in debugging auth problems as a user of the provider and service owner.

petero-dk:
Can I please get you to rethink this one? Because it forces me to put a secret in the terraform plan. And I am already mixing having the ability to use the CLOUDFLARE_API_TOKEN but not for the service key.

Adding the ability to simply pass "" would be the absolute simplest addition to the code base, it will be backwards compatible and would throw the normal error for most cases where people enter garbage. 

It would also still work if no keys are added, as the code later checks for empty strings for all auth methods.

Which means very specifically it would solve the issue of having multiple Environment Variables with Cloudflare access -> which would ALREADY give errors, and be hard to debug. This would introduce NO additional problems but actually solve one.

If you are firm on this one, would you be open for a boolean switch so I can choose which env variable it will use?

For context:
This is my current configuration (that works) which is also not really easy to debug also. I really don't like that I don't have control of my origin CA key and that it will be stored outside of my build system in TF state.

```tf
provider "cloudflare" {
    
}

provider "cloudflare" {
  alias  = "cloudflare-origin"
  #"origin ca key here. the one from the very bottom of api tokens page. called service key nowhere on the internet"
  api_user_service_key = var.cloudflare_origin_token
}
```

jacobbednarz:
sorry but this isn't something we want to support for the reasons mentioned in my last comment. while this may solve your specific use case, it makes the provider test cases, auth scheme handling and service debugging much harder in order to support it. all of which needs to be contextually loaded when making any changes or investigating issues. the ROI here is very low and increases the potential maintenance overhead which isn't a good trade off. 

the great thing about open source projects is that you can fork and run whatever suits your needs -- even if the upstream doesn't accept it :) that would be my suggestion here if this makes your specific use case a better experience for you. the repository already has documentation on how to do this if you want to take that path.

petero-dk:
I don't think you understand - it is not my specific use case. It is a terrible use experience that you only support ENV based authentication, in some instances. And once multiple providers are required debugging why one works and the other does not has taken very very long time. 

Additionally, the regex check is only provided on one of the authentication methods so clearly it is not fully baked through.

I understand that the proposed fix might seem counterintuitive (explicitly setting something to blank) however I am also asking for input to create a better method that is inline with your structure. I will of course be willing to implement that. Very good ROI on people helping a company for free.

But saying you do not want to support fully ENV based auth when specifying multiple providers is bad for ROI is simply not true for a company that prides on security above all else. (P.S we are doing an ISO and DORA cert now and this was one of the flags that came up, so we cannot be the only ones noticing this, maybe we are just the only ones stubborn enough to want to fix it)

One of the great things about open source projects is that I can fork it, create improvements, discuss better ways to do those improvements and then contribute back to the original project so everyone can benefit, and I don't have to run my own specific code and keep that up to date.

Issue #3590, opened 2024-08-09
Confusing error message pertaining to deprecation of "value" attribute of cloudflare_record


mccartney:
### Confirmation

- [X] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [X] I have searched the issue tracker and my issue isn't already found.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```
terraform -v
Terraform v1.5.7
on darwin_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v4.39.0
+ provider registry.terraform.io/hashicorp/aws v5.62.0
```

### Affected resource(s)

cloudflare_record

### Terraform configuration files

```hcl
resource "cloudflare_record" "CNAME_my" {
  zone_id = cloudflare_zone.my.id
  name    = "o-c.kr"
  type    = "CNAME"
  ttl     = 7200
  value   = "REDACTED.amazonaws.com."
}
```

### Link to debug output

N/A

### Expected output

- If `value` is deprecated, I assume it's still legal to use it. So I assume the code should work, i.e. `terraform apply` should succeed.
- Also the error message is misleading. It claims I have provided value for `content`, which I didn't.

### Actual output

│ Error: either 'content' (present: true) or 'data' (present: false) must be provided

### Steps to reproduce

terraform apply

mccartney:
> This issue has been marked with triage/needs-information and is unlikely to receive maintainer attention until the log file is provided making this a complete bug report.

Fair enough.
I did my best to provide you valuable input. I don't think `TF_LOG=DEBUG` is needed in this specific case.

mccartney:
Related to #3509 
(which is BTW funny as it's an anagram of this issue id)

mccartney:
Works fine after downgrading to `4.38.0`.

kofshikz:
It appears that just changing `value` to `content` also helps

mccartney:
> Works fine after downgrading to 4.38.0.

Watch out when/if downgrading. If you had some runs with `4.39.0` it might have changed `schema_version` from `2` to `3` in your Terraform state file for existing records. In such a case, Terraform will refuse to work with your resources after downgrading with messages like:
```
│ The current state of cloudflare_record.REDACTED was created by a newer provider version than is currently selected. Upgrade the cloudflare provider to work with this state.
```
[...]

PR #3674, opened 2024-08-18
Safely deprecate value

mgirouard:
This patch follows up #3590 by allowing `value` to continue be set while still preserving it's deprecated status.

`value` and `content` are mutually exclusive. Setting both will return an error.

    │ Error: 'content' (present: true) must not be set with 'value' (present: true)

No state changes will take place when replacing `value` with `content` provided the value is identical.


bruceoutdoors:
LGTM!

musa-cf (approved):
👍

PR #3704, opened 2024-08-21
Fix issue with ID not being populated in Hyperdrive config update

OilyLime:


jacobbednarz:
can we add a regression test to cover this change and make sure it's fixing the issue here?

OilyLime:
> can we add a regression test to cover this change and make sure it's fixing the issue here?

Done!

PR #3730, opened 2024-08-22
Remove frequency from cloudflare_logpush_job

soheiokamoto:
This removes `frequency` from `cloudflare_logpush_job`, as it has been deprecated.


jacobbednarz:
👋 if this is now deprecated, we can instead add the `Deprecated` attribute to the schema and then in the next major release we can rip it out completely.

soheiokamoto:
> 👋 if this is now deprecated, we can instead add the `Deprecated` attribute to the schema and then in the next major release we can rip it out completely.

Ah, cool. Closing this PR in favor of https://github.com/cloudflare/terraform-provider-cloudflare/pull/3745

PR #3833, opened 2024-08-29
Adding support for creating queue consumers

mephmanx:


.gitignore
```diff
@@ -35,3 +35,5 @@ dist/*
 .envrc
 .flox/
 .vscode/private.env
+
+setvars.sh
```
jacobbednarz, line 39: let's move this into your personal gitignore; we don't use this file anywhere in the project.

internal/sdkv2provider/schema_cloudflare_queue.go
```diff
@@ -17,5 +17,68 @@ func resourceCloudflareQueueSchema() map[string]*schema.Schema {
...
+			Type:        schema.TypeList,
+			Optional:    true,
+			Description: "Array of consumers.",
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"created_on": {
+						Description: "Created on date",
+						Type:        schema.TypeString,
+						Optional:    true,
+						ForceNew:    true,
```
jacobbednarz, line 30: you don't want `ForceNew` everywhere otherwise you'll constantly be recreating the resource despite not having changes.

jacobbednarz:
what do you specifically want review for the approach here? it's largely copy/paste of what we've done for other products so that seems fine.

as for the queue consumer specific stuff, i've pinged some folks internally as to how this should operate/be used since i'm not familiar with the product.

.gitignore
mephmanx: ok, will do

internal/sdkv2provider/schema_cloudflare_queue.go
mephmanx: ok, removing them

mephmanx:
@jacobbednarz 

Still working on it.  If you think the approach is fine, that is all I need.  I am having some difficulty with getting the schemaa right for the consumers array that I want to add to the Q object...  Once I can get that working I think the rest will be pretty easy.

mephmanx:
@jacobbednarz 

Im trying to get this schema to work for the consumers:

consumers = [
			{
				script_name = "%[1]s"
				settings {
					batch_size = 1
				}
			}
		]

as it looks like there can be multiple consumers and I want to support that.  If you have any advice for how to define the schema or a better way to adjust it, let me know.  Ill keep fighting with it until I get it right.

jacobbednarz:
closing this out as v4 is has moved out of active development and v5 is automatically generated. feel free to checkout v5 and see if your change is there. if not, a request through to CSUP/your account team is the way to go.

PR #3526, opened 2024-07-31
Support admin code override disable_for_time field

da-cf:
Adds support for the disable_for_time field for the [admin override feature
](https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp/configure-warp/warp-settings/#admin-override)

Relies on a [cloudflare-go PR](https://github.com/cloudflare/cloudflare-go/pull/2797)

PR #3931, opened 2024-09-09
Add missing docs for admin override

da-cf:
Add missing schema examples for disable_for_time admin override feature for WARP clients.

Strangely, [previously merged docs](https://github.com/cloudflare/terraform-provider-cloudflare/pull/3526) for this feature do not appear in [the Terraform Registry](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/teams_account). Wondering if this is possibly due the field not being present in the schema example?

### DIFF
--- a/.changelog/3931.txt
+++ b/.changelog/3931.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/zero_trust_gateway_settings: Add missing disable_for_time example field
+```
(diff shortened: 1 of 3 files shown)


jacobbednarz:
i suspect this was caught up in the rename of the ZT resources. worth calling out here, this won't change the examples for `cloudflare_teams_account` given you're only modifying the newer resource.

da-cf:
> i suspect this was caught up in the rename of the ZT resources. worth calling out here, this won't change the examples for `cloudflare_teams_account` given you're only modifying the newer resource.

What do I need to do to get the `cloudflare_teams_account` example updated?

jacobbednarz:
it will go in https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/examples/resources/cloudflare_teams_account/resource.tf however, it looks like that already has the key. perhaps the confusion here was new vs old resource?

PR #3981, opened 2024-09-12
Add image_notification to notification policy

dochne:
This adds the new image_notification to notification policy


jacobbednarz:
looks this has already landed in v5 too via #3979

dochne:
Fantastic! Thanks for the heads up Jacob!

jacobbednarz:
we can still add it here as well for v4. i was just pointing out we have this automatically in v5 🙂

PR #4007, opened 2024-09-14
Add resource cloudflare_page_shield

joshuamsager:
Existing API: https://github.com/cloudflare/cloudflare-go/blob/8743749c8ce71e82a506005135ac90f1120fc061/page_shield.go#L12
API documentation: https://developers.cloudflare.com/api/operations/page-shield-get-settings

Adds support for managing Page Shield configuration via terraform.

### DIFF
--- a/internal/sdkv2provider/provider.go
+++ b/internal/sdkv2provider/provider.go
@@ -276,6 +276,7 @@ func New(version string) func() *schema.Provider {
 				"cloudflare_observatory_scheduled_test":                      resourceCloudflareObservatoryScheduledTest(),
 				"cloudflare_origin_ca_certificate":                           resourceCloudflareOriginCACertificate(),
 				"cloudflare_page_rule":                                       resourceCloudflarePageRule(),
+				"cloudflare_page_shield":                                     resourceCloudflarePageShield(),
 				"cloudflare_pages_domain":                                    resourceCloudflarePagesDomain(),
 				"cloudflare_pages_project":                                   resourceCloudflarePagesProject(),
 				"cloudflare_queue":                                           resourceCloudflareQueue(),
--- a/internal/sdkv2provider/resource_cloudflare_page_shield.go
+++ b/internal/sdkv2provider/resource_cloudflare_page_shield.go
@@ -1,0 +1,93 @@
+package sdkv2provider
+
+import (
+	"context"
+	"fmt"
+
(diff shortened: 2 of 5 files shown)


internal/sdkv2provider/provider.go
```diff
@@ -276,6 +276,7 @@ func New(version string) func() *schema.Provider {
 				"cloudflare_observatory_scheduled_test":                      resourceCloudflareObservatoryScheduledTest(),
 				"cloudflare_origin_ca_certificate":                           resourceCloudflareOriginCACertificate(),
 				"cloudflare_page_rule":                                       resourceCloudflarePageRule(),
+				"cloudflare_page_shield":                                     resourceCloudflarePageShield(),
```
joshuamsager, line 279: Willing to change the name to: `cloudflare_page_shield_settings`

internal/sdkv2provider/resource_cloudflare_page_shield.go
```diff
@@ -0,0 +1,93 @@
...
+	client := meta.(*cloudflare.API)
+
+	tflog.Info(ctx, "Disabling Page Shield")
+
+	// There is no DELETE endpoint for schema validation settings,
+	// so terraform should reset the state to default settings
+	params := cloudflare.UpdatePageShieldSettingsParams{
+		Enabled:                        cloudflare.BoolPtr(false),
+		UseCloudflareReportingEndpoint: cloudflare.BoolPtr(true),
+		UseConnectionURLPath:           cloudflare.BoolPtr(false),
```
joshuamsager, line 76: Determined these based on how my zone is currently configured (and I believe we hadn't enabled Page Shield yet...)

internal/sdkv2provider/provider.go
jacobbednarz: we should definitely rename this resource to be more specific however, we also aren't accepting more resources added to the SDKv2 so we'll need to port this into the `terraform-plugin-framework` instead. we have [a ratchet](https://github.com/cloudflare/terraform-provider-cloudflare/pull/4007/files#diff-4ae7bd7759a37ef1ead791fcae9d7d3a02156d7f9fd95c7c614f3067e8a132bfR23) in place that notifies you when you try and run the acceptance test suite (acc tests won't run on PRs due to creating real resources and taking forever to run).

PR #4015, opened 2024-09-16
Implement "import" for cloudflare_api_token

ankon:
This is a pretty mechanical clone of other importers, and I want to see how it fares on the CI/CD.

Note that this is against v4.41.0, because master is right now broken:

```
==> Running go vet .
# github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider
internal/sdkv2provider/resource_cloudflare_record.go:57:3: unknown field ZoneID in struct literal of type cloudflare.CreateDNSRecordParams
internal/sdkv2provider/resource_cloudflare_record.go:109:119: newRecord.ZoneID undefined (type cloudflare.CreateDNSRecordParams has no field or method ZoneID)
internal/sdkv2provider/resource_cloudflare_record.go:142:77: newRecord.ZoneID undefined (type cloudflare.CreateDNSRecordParams has no field or method ZoneID)
internal/sdkv2provider/resource_cloudflare_record.go:418:53: record.ZoneName undefined (type cloudflare.DNSRecord has no field or method ZoneName)
# github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider
# [github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider]
vet: internal/sdkv2provider/resource_cloudflare_record.go:57:3: unknown field ZoneID in struct literal of type cloudflare.CreateDNSRecordParams

Vet found suspicious constructs. Please check the reported constructs
and fix them if necessary before submitting the code for review.
make: *** [GNUmakefile:63: vet] Error 1
```

### DIFF
--- a/.changelog/4015.txt
+++ b/.changelog/4015.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/cloudflare_api_token: Allow importing
+```
--- a/internal/sdkv2provider/resource_cloudflare_api_token.go
+++ b/internal/sdkv2provider/resource_cloudflare_api_token.go
@@ -21,6 +21,9 @@ func resourceCloudflareApiToken() *schema.Resource {
 		ReadContext:   resourceCloudflareApiTokenRead,
 		UpdateContext: resourceCloudflareApiTokenUpdate,
 		DeleteContext: resourceCloudflareApiTokenDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceCloudflareApiTokenImport,
+		},
 		Description: heredoc.Doc(`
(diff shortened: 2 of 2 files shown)


jacobbednarz:
API tokens don't allow fetching the secret after the initial creation so supporting `Import` is leaves the resource rather unusable. Instead, you should aim to recreate API tokens.

ankon:
Right, that indeed would be a problem. We're trying to come up with a nice way of managing these tokens, and passing them along to other parts of the system, but it's not this direction then.

Maybe an idea: Adding an explicit discussion on the `cloudflare_api_token` documentation that import is "intentionally not possible" or such?

PR #4120, opened 2024-09-24
Fix load_balancer tests

mgirouard:
This set of patches updates the `load_balancer` test cases for the v5 provider. This may not fix all issues as there still may be some upstream work to do in the schemas and/or code generator, but it should allow for a majority of tests to pass.


jacobbednarz:
just updated the version references to be `cloudflare-go/v3` and added the grit pattern for the renames.

PR #4149, opened 2024-10-01
Handle number types in encoder and decoder

dackerman:


jacobbednarz (approved):
fixes the failing example, nice one!

PR #4166, opened 2024-10-02
Add sweeper for dns_firewall acceptance tests

janik-cloudflare:


internal/services/dns_firewall/resource_test.go
```diff
@@ -1,14 +1,46 @@
...
 )
 
+func init() {
+	resource.AddTestSweepers("cloudflare_dns_firewall", &resource.Sweeper{
+		Name: "cloudflare_dns_firewall",
+		F: func(region string) error {
+			accountID := os.Getenv("CLOUDFLARE_ACCOUNT_ID")
+			ctx := context.Background()
+
+			client, err := acctest.SharedV1Client() // TODO(terraform): replace with SharedV2Clent
```
Fattouche, line 22: This isn't the generated client right? We should use that.

internal/services/dns_firewall/resource_test.go
```diff
@@ -1,14 +1,46 @@
...
+			if err != nil {
+				return fmt.Errorf("error establishing client: %w", err)
+			}
+
+			clusters, err := client.ListDNSFirewallClusters(ctx, cfv1.AccountIdentifier(accountID), cfv1.ListDNSFirewallClustersParams{})
```
jacobbednarz, line 27: i'd recommend using the newer SDK here to avoid the pains with the older library.

internal/services/dns_firewall/resource_test.go
jacobbednarz, line 22:
```suggestion
			client := acctest.SharedV2Client()
```

internal/services/dns_firewall/resource_test.go
janik-cloudflare: I gave that a try, please let me know how it looks!

Issue #4412, opened 2024-10-24
Unexpected Validation Error for TTL Settings of cloudflare_ruleset


digglife:
### Confirmation

- [X] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [X] I have searched the issue tracker and my issue isn't already found.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

- terraform 1.9.8
- terraform-provider-cloudflare 4.44.0

### Affected resource(s)

cloudflare_ruleset

### Terraform configuration files

```hcl
terraform {
  required_version = "~> 1.9.8"
  required_providers {
    cloudflare = {
      source  = "cloudflare/cloudflare"
      version = "4.44.0"
    }
  }
}

variable "my_ttl" {
  default = 86400
  type = number
}

provider "cloudflare" {
  api_key = "api_key"
  email   = "bug_report@cloudflare.com"
}
resource "cloudflare_ruleset" "caches" {
  kind    = "zone"
  name    = "bug-reproduce-rules"
  phase   = "http_request_cache_settings"
  zone_id = "zone_a"

  rules {
    action      = "set_cache_settings"
    description = "bug_reproduce"
    enabled     = true
    expression  = "(http.host eq \"hello\")"

    action_parameters {
      cache                      = true
      origin_error_page_passthru = true
      respect_strong_etags       = false

      edge_ttl {
        mode       = "override_origin"
        default    = var.my_ttl
      }
    }
  }
}
```

### Link to debug output

https://gist.github.com/digglife/1b4d145d95d9a4e7ca744c0d55f43eac

### Expected output

When the default of `edge_ttl` is set with variable,  terraform plan should be executed successfully.

```terraform
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # cloudflare_ruleset.caches will be created
  + resource "cloudflare_ruleset" "caches" {
      + description = (known after apply)
      + id          = (known after apply)
      + kind        = "zone"
      + name        = "bug-reproduce-rules"
      + phase       = "http_request_cache_settings"
      + zone_id     = "zone_a"

      + rules {
          + action       = "set_cache_settings"
          + description  = "bug_reproduce"
          + enabled      = true
          + expression   = "(http.host eq \"hello\")"
          + id           = (known after apply)
[...]

PR #4414, opened 2024-10-24
fix: unexpected validation error when ttl is set with variable

digglife:
When `default` is set with variable, it will be `unknown` in the request sent to the plugin, so the validation will fail. With this change, the plugin can handle this case correctly.

The validation still works when mode is `override_origin` and default is not set(which will be `null`) or `<=0`.

fixes: #4412

### DIFF
--- a/.changelog/4414.txt
+++ b/.changelog/4414.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_ruleset: unexpected validation error when ttl default is set with variable 
+```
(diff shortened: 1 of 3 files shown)


digglife:
Hi @jacobbednarz 
Could you please kindly review this PR? Thanks!

Issue: #4412

MAN98:
@jacobbednarz - Waiting for this PR to fix our issue with edge_ttl being a variable in cloudflare_ruleset cache rule?

PR #4447, opened 2024-10-28
Reset generated files to resync with the 'generated' branch

musa-cf:
Some generated files have some custom code that is out of sync with what the generator generates, and this caused the Terraform schema/models to be out of sync with the API schema/models.


mgirouard (approved):
lgtm

PR #4452, opened 2024-10-28
Fix: zt service access token

vaishakdinesh:
Fix the assertions for zero trust service access tokens tests


mgirouard (approved):
lgtm

internal/services/zero_trust_access_service_token/resource_test.go
```diff
@@ -54,7 +54,7 @@ func TestAccCloudflareAccessServiceToken_Basic(t *testing.T) {
 					resource.TestCheckResourceAttr(name, consts.AccountIDSchemaKey, accountID),
 					resource.TestCheckResourceAttr(name, "name", resourceName+"-updated"),
 					resource.TestCheckResourceAttrSet(name, "client_id"),
-					resource.TestCheckResourceAttrSet(name, "client_secret"),
+					resource.TestCheckNoResourceAttr(name, "client_secret"),
```
jacobbednarz, line 57: i think this is a regression; we need to have the client_secret in the state file which here looks to be failing.

internal/services/zero_trust_access_service_token/resource.go
```diff
@@ -151,13 +155,20 @@ func (r *ZeroTrustAccessServiceTokenResource) Update(ctx context.Context, req re
 		return
 	}
 	data = &env.Result
+	env.Result.ClientSecret = secret
+	data.ClientSecret = secret
+	resp.State.SetAttribute(ctx, path.Root("client_secret"), &secret)
+	resp.Diagnostics.Append(diags...)
```
jacobbednarz, line 161: i don't think this is doing anything as the `data.ClientSecret` assignment above is passed into the `resp.State.Set` below doing this exact thing again.

internal/services/zero_trust_access_service_token/resource.go
```diff
@@ -191,6 +202,14 @@ func (r *ZeroTrustAccessServiceTokenResource) Read(ctx context.Context, req reso
...
 		return
 	}
+
+	if !secret.IsNull() && secret.String() != "" {
+		env.Result.ClientSecret = secret
+		data.ClientSecret = secret
+		resp.State.SetAttribute(ctx, path.Root("client_secret"), &secret)
+		resp.Diagnostics.Append(diags...)
+	}
+
```
jacobbednarz, line 212: i think this can be simplified.

```suggestion
	data.ClientSecret = env.Result.ClientSecret
```

vaishakdinesh: I think it has to be set to the secret we get from the state? I've simplified the logic. Can you please take a look again?

internal/services/zero_trust_access_service_token/resource.go
```diff
@@ -150,14 +154,19 @@ func (r *ZeroTrustAccessServiceTokenResource) Update(ctx context.Context, req re
...
+	data.ClientSecret = secret
 
 	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
 }
 
 func (r *ZeroTrustAccessServiceTokenResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
 	var data *ZeroTrustAccessServiceTokenModel
 
+	secret := types.StringNull()
+	_ = req.State.GetAttribute(ctx, path.Root("client_secret"), &secret)
```
jacobbednarz, line 168:
```suggestion
	req.State.GetAttribute(ctx, path.Root("client_secret"), &secret)
```

internal/services/zero_trust_access_service_token/resource.go
```diff
@@ -102,6 +104,8 @@ func (r *ZeroTrustAccessServiceTokenResource) Create(ctx context.Context, req re
 
 func (r *ZeroTrustAccessServiceTokenResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
 	var data *ZeroTrustAccessServiceTokenModel
+	secret := types.StringNull()
+	_ = req.State.GetAttribute(ctx, path.Root("client_secret"), &secret)
```
jacobbednarz, line 108:
```suggestion
	req.State.GetAttribute(ctx, path.Root("client_secret"), &secret)
```

internal/services/zero_trust_access_service_token/resource.go
jacobbednarz: yep, that's correct however, we don't need to `SetAttribute` if we use `data` as a few lines down (L215), it gets passed to `resp.State.Set` anyway and ends up setting it for us.

jacobbednarz:
lgtm 

```
TF_ACC=1 go test ./internal/services/zero_trust_access_service_token/ -run "^TestAcc" -count 1 -v
=== RUN   TestAccCloudflareAccessServiceToken_Basic
--- PASS: TestAccCloudflareAccessServiceToken_Basic (9.09s)
=== RUN   TestAccCloudflareAccessServiceToken_Delete
--- PASS: TestAccCloudflareAccessServiceToken_Delete (4.61s)
=== RUN   TestAccCloudflareAccessServiceToken_WithDuration
--- PASS: TestAccCloudflareAccessServiceToken_WithDuration (8.06s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/services/zero_trust_access_service_token	23.023s
```

PR #4465, opened 2024-10-29
resource/cloudflare_zone_setting_override: add support for `ssl_automatic_mode`

jacobbednarz:
Pending the API fix to name the `id` key to the same as the route.


pznamensky:
Hi @jacobbednarz, it seems that you've done everything needed to merge `ssl_automatic_mode` into master branch.
Wdyt about merging it?

PR #4506, opened 2024-11-04
Fix ZT access application acceptance tests and test data

musa-cf:
Fixes acceptance tests for:
- `zero_trust_access_application`
- `zero_trust_access_identity_provider`


internal/services/zero_trust_access_application/testdata/accessapplicationconfigwithadefinedidp.tf
```diff
@@ -1,8 +1,8 @@
-
 resource "cloudflare_zero_trust_access_identity_provider" "%[1]s" {
   account_id = "%[4]s"
   name = "%[1]s"
   type = "onetimepin"
+  config = {}
```
jacobbednarz, line 5: we shouldn't need to put empty blocks anywhere. this should be handled without it being present.

musa-cf: I fixed that test in a second commit that I hadn't pushed up yet. The config block is non-empty in the working test.

internal/services/zero_trust_access_application/testdata/accessapplicationconfigwithautoredirecttoidentity.tf
```diff
@@ -1,8 +1,8 @@
-
 resource "cloudflare_zero_trust_access_identity_provider" "%[1]s" {
   zone_id = "%[2]s"
   name    = "%[1]s"
   type    = "onetimepin"
+  config = {}
```
jacobbednarz, line 5:
```suggestion
```

PR #4514, opened 2024-11-05
Added optional routing field to regional hostname

tarnfeld:
(WIP)

Introducing a new, and very optional `routing` field into the Regional Hostname API.
Schema and cloudflare-go library changes are not yet released.

I'm not sure if this empty checks are required, but since this field is very optional right now, I would prefer we don't generate a load of `routing = ""` config in terraform objects when it's redundant.


jacobbednarz:
moving to draft to keep stalebot of it. feel free to flip it back when you're ready here.

jacobbednarz:
closing this out as v4 is has moved out of active development and v5 is automatically generated. feel free to grab me internally to get this automatically generating.

PR #4496, opened 2024-11-03
Remove region from turnstile update

Cyb3r-Jak3:
Removes region from update as it is not listed in the API docs. Marks it as force recreate on update to it.

Fixes: #4217

Issue #4217, opened 2024-10-06
Issue updating domains list for cloudflare_turnstile_widget


amolsingh74:
### Confirmation

- [X] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [X] I have searched the issue tracker and my issue isn't already found.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform: 1.5.7
Provider: 4.43.0

### Affected resource(s)

cloudflare_turnstile_widget

### Terraform configuration files

```hcl
resource "cloudflare_turnstile_widget" "ex" {
  account_id      = "myaccountid"
  name            = "TFwidget"
  bot_fight_mode  = false
  mode            = "managed"
  offlabel        = false
  #region          = "world"

  domains = [
          "mydomain.com",
          "mydomain2.org",
        ] 

}
```

### Link to debug output

github.com/hashicorp/terraform-plugin-go@v0.24.0/tfprotov6/internal/diag/diagnostics.go:58

### Panic output

│ Error: Error reading challenge widget
│ 
│   with cloudflare_turnstile_widget.ex,
│   on cf-tmp.tf line 1, in resource "cloudflare_turnstile_widget" "ex":
│    1: resource "cloudflare_turnstile_widget" "ex" {
│ 
│ error from makeRequest: json: unknown field "region"

### Expected output

  # cloudflare_turnstile_widget.ex will be updated in-place
  ~ resource "cloudflare_turnstile_widget" "ex" {
      ~ domains        = [
          - "mydomain2.org",
            # (5 unchanged elements hidden)
        ]
        id             = "myaccountit"
        name           = "TFwidget"
      ~ secret         = (sensitive value)
        # (5 unchanged attributes hidden)
    }

### Actual output

│ Error: Error reading challenge widget
│ 
│   with cloudflare_turnstile_widget.ex,
│   on cf-tmp.tf line 1, in resource "cloudflare_turnstile_widget" "ex":
│    1: resource "cloudflare_turnstile_widget" "ex" {
│ 
│ error from makeRequest: json: unknown field "region"

### Steps to reproduce

im using provider version `"4.43.0"`

im creating a turnstile widget

```
resource "cloudflare_turnstile_widget" "ex" {
  account_id      = "myaccountid"
  name            = "TFwidget"
  bot_fight_mode  = false
  mode            = "managed"
[...]

PR #4521, opened 2024-11-06
Remove region reference from turnstile docs

HeroProtagonist:
I am facing the same error mentioned in https://github.com/cloudflare/terraform-provider-cloudflare/issues/4217.

I see `region` was removed in this PR https://github.com/cloudflare/terraform-provider-cloudflare/pull/4496.

Removing reference to `region` from the docs to match the current state of things


jacobbednarz:

jacobbednarz:
> I see `region` was removed in this PR #4496.

`region` isn't removed entirely. only for updates with the associated validation to force recreation if the `region` value changes.

i don't think we need this PR but thanks anyway!

PR #4603, opened 2024-11-18
Feat: api token data source tests

vaishakdinesh:


internal/services/api_token/data_source_test.go
```diff
@@ -0,0 +1,41 @@
...
+					resource.TestCheckResourceAttrSet(name, "status"),
+					resource.TestCheckResourceAttr(name, "status", "active"),
+					resource.TestCheckResourceAttr(name, "token_id", tokenID),
+				),
+			},
+		},
+	})
+}
+
+func testAccCloudflareZoneConfigBasic() string {
```
jacobbednarz, line 35: as part of this acceptance test, you should create the API token and then reference it within the datasource instead of using a static one.

vaishakdinesh: updated! @jacobbednarz

PR #4604, opened 2024-11-18
Add 'zero_trust_device_default_profile_certificates' acceptance tests

musa-cf:
The tests won't pass until an internal PR gets merged and released.

### DIFF
--- a/internal/services/zero_trust_device_default_profile_certificates/resource_test.go
+++ b/internal/services/zero_trust_device_default_profile_certificates/resource_test.go
@@ -1,0 +1,51 @@
+package zero_trust_device_default_profile_certificates_test
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


internal/services/zero_trust_device_default_profile_certificates/resource_test.go
```diff
@@ -0,0 +1,51 @@
...
+			},
+		},
+	})
+}
+
+func testCloudflareDevicePolicyCertificates(rnd, zoneID string, enable bool) string {
+	return fmt.Sprintf(`
+resource "cloudflare_zero_trust_device_default_profile_certificates" "%[1]s" {
+	zone_id = "%[2]s"
+	enabled = "%[3]t"
```
jacobbednarz, line 48: you don't need to quote the booleans

```suggestion
	enabled = %[3]t
```

musa-cf: Ah, fixed! Thanks

Issue #4648, opened 2024-11-22
Destroy on cloudflare_authenticated_origin_pulls disables instead of deleting


PhilipSkinner:
### Confirmation

- [X] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [X] I have searched the issue tracker and my issue isn't already found.
- [X] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform: 1.6.6
Provider: 4.46.0

### Affected resource(s)

cloudflare_authenticated_origin_pulls

### Terraform configuration files

```hcl
resource "cloudflare_authenticated_origin_pulls" "my_per_hostname_aop" {
  zone_id                                = "..."
  authenticated_origin_pulls_certificate = "..."
  hostname                               = "..."
  enabled                                = true
}
```

### Link to debug output

[N/A](https://gist.github.com/PhilipSkinner/dd83aefd0fa3884379967d30750e197c)

### Expected output

Enabled flag set to null as per the documentation: https://developers.cloudflare.com/api/operations/per-hostname-authenticated-origin-pull-enable-or-disable-a-hostname-for-client-authentication

### Actual output

Enabled flag set to false, which disables the association but does not destroy it - see documentation: https://developers.cloudflare.com/api/operations/per-hostname-authenticated-origin-pull-enable-or-disable-a-hostname-for-client-authentication

### Steps to reproduce

* Create a cloudflare_authenticated_origin_pulls via tf plan/apply etc
* Destroy the resource using tf destroy
* Query the cloudflare API and verify that the association has not been removed:

```curl -X GET "https://api.cloudflare.com/client/v4/zones/${cloudflare_zone}/origin_tls_client_auth/hostnames?page=${page}" --header "X-Auth-Email: $CLOUDFLARE_USER" --header "X-Auth-Key: $CLOUDFLARE_KEY"```

### Additional factoids

See: https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/sdkv2provider/resource_cloudflare_authenticated_origin_pulls.go#L126

PR #4649, opened 2024-11-22
Fixes destroy on cloudflare_authenticated_origin_pulls

PhilipSkinner:
According to the documentation the enabled flag must be set to null (…or nil) in order for the association to be destroyed. Setting this to false keeps the association in a disabled state instead of destroying the resource.

Documentation on this API endpoint can be found here:

https://developers.cloudflare.com/api/operations/per-hostname-authenticated-origin-pull-enable-or-disable-a-hostname-for-client-authentication

Referencing issue raised on provider:

https://github.com/cloudflare/terraform-provider-cloudflare/issues/4648


jacobbednarz:
thanks for the PR. can you please add test coverage for this change to confirm it actually does what we intend here?

PhilipSkinner:
will do, this is blocked by the cloudflare API client library not supporting this API endpoint correctly so I'll raise a PR to fix that first then come back to this once its been released

PhilipSkinner:
PR raised for fixing the API client: https://github.com/cloudflare/cloudflare-go/pull/3680

Once this has been approved and released I'll update the terraform provider to use the fixed library.

jacobbednarz:
handled via https://github.com/cloudflare/terraform-provider-cloudflare/pull/4661

PR #4655, opened 2024-11-24
fix(terraform): updated `data.cloudflare_zone` block to use `filter`

cpeliciari:


jacobbednarz:
this is incorrect for v4.x, `filter` is not an attribute.

cpeliciari:
@jacobbednarz this is in 5.0.0-alpha1

PR #4660, opened 2024-11-25
feat: add r2 oc region to accepted location hints

zackpollard:
This adds the new R2 Oceania region to the list of supported location hints in the provider


jacobbednarz:
i'm unable to push changes for you here so you'll need to run `make docs` and push up the result.

zackpollard:
> i'm unable to push changes for you here so you'll need to run `make docs` and push up the result.

Have run that and pushed those changes, apologies for missing that.

As a side note there was an unrelated docs change in the notification_policy file, I didn't push that here but presumably it was missed in another PR so it's out of sync.

PR #4722, opened 2024-12-05
page rules on off actions

jafowler:
This PR aims to add custom support for sending `browser_check` and `cache_by_device_type` to the generated code.


internal/services/page_rule/resource.go
```diff
@@ -110,7 +110,7 @@ func (r *PageRuleResource) Update(ctx context.Context, req resource.UpdateReques
 		return
 	}
 
-	dataBytes, err := data.MarshalJSONForUpdate(*state)
+	dataBytes, err := data.MarshalJSON()
```
mgirouard, line 113: What are the consequences of using this other method?

internal/services/page_rule/resource_test.go
```diff
@@ -332,6 +332,76 @@ func TestAccCloudflarePageRule_MinifyAction(t *testing.T) {
...
+		ProtoV6ProviderFactories: acctest.TestAccProtoV6ProviderFactories,
+		CheckDestroy:             testAccCheckCloudflarePageRuleDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckCloudflarePageRuleConfigOnOff(rnd, zoneID, target, "cache_by_device_type", "on"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCloudflarePageRuleExists(resourceName, &pageRule),
+					resource.TestCheckResourceAttr(resourceName, consts.ZoneIDSchemaKey, zoneID),
+					resource.TestCheckResourceAttr(resourceName, "target", fmt.Sprintf("%s", target)),
+                    resource.TestCheckResourceAttr(resourceName, "actions.cache_by_device_type", "on"),
```
mgirouard, line 389: Please `fmt`

internal/services/page_rule/resource.go
jafowler: using `MarshalJSONForUpdate(*state)` gives us incorrect PUT values. 

example: 
```
{
    "actions": {
        "cache_by_device_type": "off"
    },
    "priority": 1,
    "status": "disabled",
    "target": "ppxzakepze.terraform.cfapi.net"
}

```
target / actions are incorrect in this case, it should be
```
{
    "priority": 1,
    "status": "disabled",
    "targets": [
        {
            "constraint": {
                "operator": "matches",
                "value": "mudxsiynjr.terraform.cfapi.net"
            },
            "target": "url"
        }
    ],
    "actions": [
        {
            "id": "cache_by_device_type",
            "value": "off"
        }
    ]
}

```

jacobbednarz: `MarshalJSONForUpdate` should be used in the updates as it takes into considerations the field type (required, optional, computed) to ensure we always have the correct known values. this seems like the fields or schemas may be incorrect defined if it's not picking them.

mgirouard: Ah, I see. I wonder if instead we should also give the `PageRuleActionsModel.MarshalJSONForUpdate` method the same treatment we did for `PageRuleActionsModel.MarshalJSON`. This would help keep the amount of code changes to codegen'd files to a minimum.
https://github.com/cloudflare/terraform-provider-cloudflare/blob/ecea695686810d5c7d88001c45f66087ac2712bf/internal/services/page_rule/model.go#L28

jafowler: I'm looking at how the schema is defined here https://github.com/cloudflare/terraform-provider-cloudflare/pull/4722/files#diff-7ff72210193722a4e79ee559520597a59a76770873c73c8040a411320478cef3L62 

and i think Jacob is correct. We probably need `actions` to have  `id` and `value` instead of all the page rules?

mgirouard (approved):
lgtm

vaishakdinesh:
lgtm too! Don't have an approve button though :)

PR #4728, opened 2024-12-06
Remove minify tests

mgirouard:
Minify is deprecated and is no longer available. This patch removes unnecessary tests and their related data.


musa-cf (approved):

jacobbednarz:
looks like we've got a merge conflict here. could you resolve that one?

PR #4733, opened 2024-12-09
Update issue sync owners of Pages/Workers

WalshyDev:
Deploy & Config own Pages/Workers APIs and should be the owners on these escalated issues.


jacobbednarz:
thanks!

PR #4697, opened 2024-12-03
resource/cloudflare_ruleset: Improve diffs when only some rules are changed

zakcutner:
#### Remove version in ruleset rule action parameters

This action parameter is used within the execute action, but cannot
actually be set to anything other than `latest`.

Removing this field from Terraform prevents an issue where it shows up
in diffs, due to it being a computed attribute.

#### Prevent ruleset rule IDs incorrectly showing as changed in diffs

In the Rulesets API, it is possible to prevent rule IDs from changing
across updates by using rule refs. If a rule's ref does not change as
part of an update, then neither will its ID.

However, Terraform does not know this and always reports that rule IDs
may change, which creates very large diffs. In this change, we use a
resource plan modifier to "teach" Terraform this rule.

#### Remove `version` and `last_updated` fields from ruleset rules

These fields frequently change, leading to verbose diffs in plans. Since
these fields are not really useful within Terraform, we remove them from
the schema completely to prevent them from "polluting" diffs.

This also matches the behavior we have for the ruleset-level `version`
and `last_updated` fields, which are also not represented in the
Terraform schema.

#### Remove logic to preserve ruleset rule refs

This logic tries to detect which rules are the same as before during an
update, and preserve the refs (and therefore IDs) of these rules.

However, this logic is fallible, since we cannot know the user's true
intention, and makes it difficult for us to remove the `version` and
`last_updated` fields from rules.

If users want to preserve rule refs across updates, they can now
actually use the `ref` field directly—and there is no need for this
remapping logic anymore.

Closes #2690

PR #4748, opened 2024-12-10
feat(rulesets): improve handling of `id` and `ref` unknown values

jacobbednarz:
Updates the schema definition to apply
`stringplanmodifier.UseStateForUnknown` on the values as once they are set, they should not be changing and we can safely rely on the values.

This is the last piece of the rulesets drift issue.

### DIFF
--- a/internal/framework/service/rulesets/schema.go
+++ b/internal/framework/service/rulesets/schema.go
@@ -109,10 +109,16 @@ func (r *RulesetResource) Schema(ctx context.Context, req resource.SchemaRequest
 						consts.IDSchemaKey: schema.StringAttribute{
 							Computed:            true,
 							MarkdownDescription: "Unique rule identifier.",
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.UseStateForUnknown(),
+							},
 						},
 						"ref": schema.StringAttribute{
-							Optional:            true,
-							Computed:            true,
+							Optional: true,
+							Computed: true,
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.UseStateForUnknown(),
+							},
 							MarkdownDescription: "Rule reference.",
 						},
 						"enabled": schema.BoolAttribute{


internal/framework/service/rulesets/schema.go
```diff
@@ -108,10 +108,16 @@ func (r *RulesetResource) Schema(ctx context.Context, req resource.SchemaRequest
...
+								stringplanmodifier.UseStateForUnknown(),
+							},
 						},
 						"ref": schema.StringAttribute{
-							Optional:            true,
-							Computed:            true,
+							Optional: true,
+							Computed: true,
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.UseStateForUnknown(),
```
jacobbednarz, line 119: @zakcutner we're pushing folks to use this however, it cannot be defined once the rule is created. should we also mark this as [`RequiresReplace()`](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#requiresreplace) as well?

zakcutner: We cannot do this unfortunately, because it's only the rule (and not the whole ruleset resource) that will be replaced if the `ref` changes

zakcutner:
We actually shouldn't mark either of the `id` or `ref` fields as `UseStateForUnknown`. This is because they _will_ change across updates to the ruleset, so adding this will cause the provider to emit an incorrect plan. This is because the API cannot know which rules in the new version of the ruleset map to which rules in the previous version (i.e., rules might have been reordered, rules might have been added or deleted, etc.).

[The way around this](https://developers.cloudflare.com/terraform/troubleshooting/rule-id-changes/) is to give each rule an explicit `ref` (by default, the `ref` is set to the same as the `id`). This way, the API knows which rule `id`s to keep the same in the new version, by looking at which `ref`s have remained the same. The custom plan modifier that was added in #4697 will show the correct diff in this case, by matching rule `ref`s in the state and the plan against each other.

internal/framework/service/rulesets/schema.go
jacobbednarz: does the `UseStateForUnknown()` make sense?

PR #4790, opened 2024-12-17
Adds Terraform 1.5 import syntax

gavin-truework:
Terraform 1.5 and later supports importing resources in code, and it's nice to have the syntax in the documentation.

This adds a line to the existing import section that shows how to import each resource via the new import block.

### DIFF
--- a/.changelog/4790.txt
+++ b/.changelog/4790.txt
@@ -1,0 +1,3 @@
+```release-note:dependency
+Improves documentation to include new import options for Terraform 1.5.0
+```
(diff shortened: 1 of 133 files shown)


jacobbednarz:

gavin-truework:
@jacobbednarz Got it, thank you for that link. I'll see if I can get that automation working.

jacobbednarz:
given v5 is just around the corner, i wouldn't stress _too_ much about this. we can look to autogenerate it there if this is something we need to extend. given the current syntax still works, i'm also not too phased if this doesn't land right now since it's got full support in all versions.

PR #4805, opened 2024-12-19
feat: add bYOIPv6 + endpoint + Destination IP support at locations

rexscaria:


jacobbednarz:
what is the plan with `policy_ids` here? we can't just remove it without a migration path.

rexscaria:
The feature was discontinued a while ago. It’s not coming back.

It’s also removed from docs.
What should we do here?


*Rex Scaria | | ***@***.*** ***@***.***> |
+1-480-628-9203*

jacobbednarz:
if it's no longer usable via the API or the product, we can remove it here but need to be sure no one is using it internally. 

if it's still usable (although deprecated), we'll need to continue to have it here but marked as deprecated.

rexscaria:
It’s no longer usable. We should remove it.


*Rex Scaria | | ***@***.*** ***@***.***> |
+1-480-628-9203*

jacobbednarz:
cool. i'll PR the removal before i cut the next release.

rexscaria:
I have updated this PR to reflect the current gateway API. This is ready to merge. Pls let me know if any addition is needed here.

jacobbednarz:
acceptance tests all passing

```
TF_ACC=1 go test ./internal/sdkv2provider -v -run "^TestAccCloudflareTeamsLocation" -count 1 -timeout 120m -parallel 1
=== RUN   TestAccCloudflareTeamsLocationBasic
--- PASS: TestAccCloudflareTeamsLocationBasic (2.89s)
PASS
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/sdkv2provider      3.463s
```

PR #4810, opened 2024-12-20
ruleset bug import if ServeStale present but empty

stboissdev:
if  ServeStale is present but empty during import, the provider crash


jacobbednarz:
fixed via #4814

PR #4811, opened 2024-12-20
Record add content even if data empty

stboissdev:
if data is empty , content isn't add during import


jacobbednarz:
can you elaborate on why this is needed? or what issue it is solving? right now, it's very bare and doesn't contain any context as to the driver.

the conditional here is correct as when you define `data`, the value ends up stringified into `content` as well.

PR #4816, opened 2024-12-24
Add aegis option to zone settings

zaidoon1:
### Aegis setting

Aegis provides dedicated egress IPs (from Cloudflare to your origin) for your layer 7 WAF and CDN services. The egress IPs are reserved exclusively for your account so that you can increase your origin security by only allowing traffic from a small list of IP addresses.

The api endpoint and expected input/output can be seen below:

### Get Aegis setting value

```
curl --request GET \
 --url https://api.cloudflare.com/client/v4/zones/zone_id/settings/aegis \
 --header 'Content-Type: application/json' \
 --header 'X-Auth-Email: xxx'
{
  "result": {
    "editable": true,
    "id": "aegis",
    "modified_on": "2024-12-20T19:18:58.145335Z",
    "value": {
      "pool_id": "some-pool-id-you-own"
    }
  },
  "success": true,
  "errors": [],
  "messages": []
}
```

### Change aegis setting value

```
curl --request PATCH \
 --url https://api.cloudflare.com/client/v4/zones/zone_id/settings/aegis \
 --header 'Content-Type: application/json' \
 --header 'X-Auth-Email: xxx' \
 --data '{
 "value": {
   "pool_id": "some-pool-id-you-own"
 }
}'
{
  "result": {
    "editable": true,
    "id": "aegis",
    "modified_on": "2024-12-20T19:18:58.145335Z",
    "value": {
      "pool_id": "some-pool-id-you-own"
    }
  },
  "success": true,
  "errors": [],
  "messages": []
}

```

### Delete Aegis setting value (i.e disable Aegis)

```
curl --request DELETE \
 --url https://api.cloudflare.com/client/v4/zones/zone_id/settings/aegis \
 --header 'Content-Type: application/json' \
 --header 'X-Auth-Email: xxx'


{
  "result": {
    "editable": true,
    "id": "aegis",
    "modified_on": "2024-12-20T19:33:48.830534Z"
  },
  "success": true,
  "errors": [],
  "messages": []
}
```


zaidoon1:
there is a bunch of failures related to ` location.PolicyIDs undefined (type cloudflare.TeamsLocation has no field or method PolicyIDs) ` which I didn't touch here. Is this a known issue?

zaidoon1:
I believe this should work fine with GET/PATCH. the problem is with the DELETE since that's not standard for other zone settings, usually they are on/off that are set via PATCH. @jacobbednarz any idea on how I should go about this?

jacobbednarz:
> there is a bunch of failures related to location.PolicyIDs undefined (type cloudflare.TeamsLocation has no field or method PolicyIDs) which I didn't touch here. Is this a known issue?

yep - being worked on and doesn't have anything to do with this.

> I believe this should work fine with GET/PATCH. the problem is with the DELETE since that's not standard for other zone settings, usually they are on/off that are set via PATCH. @jacobbednarz any idea on how I should go about this?

you have two options. either update https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/internal/sdkv2provider/resource_cloudflare_zone_settings_override.go#L397 to handle this special case or 2) update the service to support turning it off via PATCH. given this is breaking the conventions, i would probably go with option 2 as that will also allow it to work on the code gen version.

zaidoon1:
>  2) update the service to support turning it off via PATCH

if we were to go with this option, what is the convention for doing this with a feature that is not "on/off" / "true/false". 

For example, should I make it so that setting pool_id to `null` would delete the zone setting value? Does terraform schema support having two data types for a field like "string" OR null as a type? This way I can update the PATCH endpoint to allow something like:
```
curl --request PATCH \
 --url https://api.cloudflare.com/client/v4/zones/zone_id/settings/aegis \
 --header 'Content-Type: application/json' \
 --header 'X-Auth-Email: xxx' \
 --data '{
 "value": {
   "pool_id": null
 }
}'
 ```
 
 to delete the value, or maybe even:
 
 ```
curl --request PATCH \
 --url https://api.cloudflare.com/client/v4/zones/zone_id/settings/aegis \
 --header 'Content-Type: application/json' \
 --header 'X-Auth-Email: xxx' \
 --data '{
 "value": null
}' 
```

jacobbednarz:
yep! usually, it is the falsey value of the attribute. 

ideally, we'd support straight DELETE methods instead however, given the age of the zone setting conventions i'm unsure what value that would bring today.

PR #4819, opened 2024-12-27
Add acceptance test generator

musa-cf:
## Summary:
- add `go.work` files to gitignore list
- add test generation tool
- add some guides relevant to acceptance tests and tool usage

## Tool usage examples
Happy path example:
```
> go run cmd/acctestgen/main.go new_service

To be created:
- file: internal/services/new_service/data_source_test.go
- file: internal/services/new_service/resource_test.go
- dir: internal/services/new_service/testdata/
- file: internal/services/new_service/testdata/resourcebasicconfig.tf
- file: internal/services/new_service/testdata/datasourcebasicconfig.tf

Do you confirm that the file paths above are correct? [y|n]: y

The files have been created. To run the acceptance tests, you will need to first
set the necessary environment variables. See https://github.com/cloudflare/terraform-provider-cloudflare/blob/v4.39.0/contributing/environment-variable-dictionary.md
for the list of variables. The values will vary depending on the account and zone being used for
the tests. Once the environment variables are set, you can run the tests using the command:

TF_ACC=1 go test ./internal/services/new_service -run "^TestAccCloudflare" -v -count 1s
```

With `-dryrun` flag:
```
> go run cmd/acctestgen/main.go -dryrun new_service

To be created:
- file: internal/services/new_service/data_source_test.go
- file: internal/services/new_service/resource_test.go
- dir: internal/services/new_service/testdata/
- file: internal/services/new_service/testdata/resourcebasicconfig.tf
- file: internal/services/new_service/testdata/datasourcebasicconfig.tf


Dry-run mode is enabled.

To create the files, run this again without the 'dryrun' flag

Example: go run cmd/acctest/main.go new_service
```

Error case:
```
go run cmd/acctestgen/main.go -dryrun fwaefawef

ERROR: service 'fwaefawef' not found

exit status 1
```


cmd/acctestgen/main.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		fmt.Printf("\nERROR: %s\n\n", err)
+		os.Exit(1)
+	}
+
+	datasourcePath := fmt.Sprintf("internal/services/%s/data_source_test.go", service)
+	resourcePath := fmt.Sprintf("internal/services/%s/resource_test.go", service)
+	testdataDirPath := fmt.Sprintf("internal/services/%s/testdata/", service)
+	resourceDataPath := fmt.Sprintf("internal/services/%s/testdata/resourcebasicconfig.tf", service)
+	datasourceDataPath := fmt.Sprintf("internal/services/%s/testdata/datasourcebasicconfig.tf", service)
```
jacobbednarz, line 31: we can probably refine the names here. given they are packaged now whereas the migration needed to keep the names for the purposes of tracking.

```suggestion
	resourceDataPath := fmt.Sprintf("internal/services/%s/testdata/basic.tf", service)
	datasourceDataPath := fmt.Sprintf("internal/services/%s/testdata/datasource_basic.tf", service)
```

cmd/acctestgen/main.go
```diff
@@ -0,0 +1,287 @@
...
+
+	// print preview of files/directories
+	fmt.Println("To be created:")
+	fmt.Printf("- file: %s\n", datasourcePath)
+	fmt.Printf("- file: %s\n", resourcePath)
+	fmt.Printf("- dir: %s\n", testdataDirPath)
+	fmt.Printf("- file: %s\n", resourceDataPath)
+	fmt.Printf("- file: %s\n", datasourceDataPath)
+	fmt.Println()
```
jacobbednarz, line 40: nit: i would go with a multi line string here instead for readability.

```suggestion
	fmt.Println(fmt.Sprintf(`To be created:
- file:       %[1]s
- file:       %[2]s
- directory:  %[3]s
- file:       %[4]s
- file:       %[5]s
`, datasourcePath, resourcePath, testdataDirPath, resourceDataPath, datasourceDataPath)
```

cmd/acctestgen/main.go
```diff
@@ -0,0 +1,287 @@
...
+						return errors.New("test not implemented")
+					},
+					resource.TestCheckResourceAttr(name, "some_string_attribute", "string_value"),
+				),
+			},
+		},
+	})
+}
+
+func test{{.PascalCase}}Config(rnd string) string {
```
jacobbednarz, line 243:
```suggestion
func testAcc{{.PascalCase}}Config(rnd string) string {
```

cmd/acctestgen/main.go
```diff
@@ -0,0 +1,287 @@
...
+						return errors.New("test not implemented")
+					},
+					resource.TestCheckResourceAttr(name, "some_string_attribute", "string_value"),
+				),
+			},
+		},
+	})
+}
+
+func test{{.PascalCase}}DataSourceConfig(rnd string) string {
```
jacobbednarz, line 283:
```suggestion
func testAcc{{.PascalCase}}DataSourceConfig(rnd string) string {
```

jacobbednarz (changes requested):
at a high level, i'm 👍 . two things we should do though to make the UX more consistent here are:

1) we move this into `tools/` given we aren't expecting the consumer of the library to use this. it's internal only and should be stripped from the end package that terraform consumes. check out the current `master` version of this for some inspiration.
2) this should be wrapped in a `scripts/...` execution. the goal of the `scripts` directory is to have a unified interface with how we expose tooling to operators. a deeper read on this is at https://github.com/github/scripts-to-rule-them-all but in essence, it means we maintain this pattern for all of the repositories.

musa-cf:
@jacobbednarz Thanks for the feedback! I've pushed a new commit with your suggestions.

scripts/generate-acctests
jacobbednarz, line 1: why is this a binary file?

musa-cf: Ah, fixed. It's definitely not a binary anymore. 👍

PR #4884, opened 2025-01-14
Fix typo in Workers example

irvinebroque:
Seems like examples here aren't autogen


jacobbednarz:
yep, v4.x does not have autogenerated examples. 5.x onwards is.

two things with this:

1. this is the legacy resource (worker_script, missing the s in workers) and it's around for documentation purposes if folks are still using it.
2. if we don't want this, we need to update https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/examples/resources/cloudflare_worker_script/resource.tf as the `docs` directory is generated from the schema + examples to build these pages.

PR #4885, opened 2025-01-14
Update worker_script.md

irvinebroque:
Without this I get:

```
➜  terraform-test git:(main) ✗ terraform plan
╷
│ Error: Unsupported argument
│
│   on cloudflare.tf line 20, in resource "cloudflare_workers_script" "terraform-test":
│   20:   name       = "terraform-test"
│
│ An argument named "name" is not expected here.
```


irvinebroque:
These docs differ from:

https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/workers_script

jacobbednarz:
i think we're talking about two different versions here. latest in the registry is 5.0.0-rc1 whereas this example is for 4.x. does that make this clearer?

PR #4894, opened 2025-01-16
migrate api_shield_operation to framework

BrendanBall:
This PR migrates `cloudflare_api_shield_operation` to the `terraform-plugin-framework` to fix a bug caused by the API returning a normalized version of a specified attribute that causes mismatches between resource config and state, resulting in the resource having to be recreated on every `terraform apply`. This is due to not following the expectations of terraform mentioned in https://github.com/hashicorp/terraform/blob/main/docs/resource-instance-change-lifecycle.md:

> Normalization: the remote API has returned some data in a different form than was recorded in the Previous Run State, but the meaning is unchanged.
> In this case, the provider should return the exact value from the Previous Run State, thereby preserving the value as it was written by the user in the configuration and thus avoiding unwanted cascading changes to elsewhere in the configuration.

Migrating to the framework is necessary to implement semantic equality between the unnormalized config and the normalized state returned from the API which is used during `terraform plan/apply`.


maciej (commented):
lgtm, left one comment, though.

(I can't approve because of lack access)

internal/framework/service/api_shield_operation/endpoint.go
```diff
@@ -0,0 +1,106 @@
...
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+var varMatch = regexp.MustCompile(`{([\w\d]+)}`)
```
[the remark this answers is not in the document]
BrendanBall: sure

internal/framework/provider/provider.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/consts"
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/framework/muxclient"
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/framework/service/access_mutual_tls_hostname_settings"
+	apishieldoperation "github.com/cloudflare/terraform-provider-cloudflare/internal/framework/service/api_shield_operation"
```
jacobbednarz, line 19: you shouldn't need to alias this given the package name is `apishieldoperation`

internal/framework/service/api_shield_operation/endpoint.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+var varMatch = regexp.MustCompile(`{([\w\d-_]+)}`)
+
+// EndpointType implemented based on https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom
+
+// Ensure the implementation satisfies the expected interfaces
```
jacobbednarz, line 19:
```suggestion
// Ensure the implementation satisfies the expected interfaces.
```

internal/framework/provider/provider.go
jacobbednarz, line 19:
```suggestion
	"github.com/cloudflare/terraform-provider-cloudflare/internal/framework/service/api_shield_operation"
```

jacobbednarz:
acceptance tests all passing

```
TF_ACC=1 go test ./internal/framework/service/api_shield_operation/ -run "^TestAcc" -count 1 -v
=== RUN   TestAccCloudflareAPIShieldOperation_Create
--- PASS: TestAccCloudflareAPIShieldOperation_Create (5.73s)
=== RUN   TestAccCloudflareAPIShieldOperation_ForceNew
--- PASS: TestAccCloudflareAPIShieldOperation_ForceNew (8.44s)
PASS
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/framework/service/api_shield_operation     15.876s
```

BrendanBall:
Thanks @jacobbednarz

PR #4935, opened 2025-01-24
Update field reference: data.HyperdriveID.ValueString() -> data.ID.ValueString()

AJMosley345:
The method 'data.HyperdriveID.ValueString()' was used, but the 'HyperdriveConfigModel' struct doesn't have this field. The method call has been updated to use 'data.ID.ValueString()' to correctly reference the existing field in the struct.


jacobbednarz:
thanks for the PR but this is a generated artifact and will be overridden on the next run. will take this internally for the team to review.

Issue #4982, opened 2025-01-30
Migrate to v5 when remote state contains incompatible v4 schema


peterhirn:
Migrating this very simple terraform setup: https://github.com/phi-ag/rvt-app/pull/175

I'm unable to plan/apply because the state obviously contains the old schema.

`terraform apply`

```
Running apply in HCP Terraform. Output will stream here. Pressing Ctrl-C
will cancel the remote apply if it's still pending. If the apply started it
will stop streaming the logs, but will not stop the apply running remotely.

Preparing the remote apply...

To view this run in a browser, visit:
https://app.terraform.io/app/phi/rvt-app/runs/run-tLNtwbNfm8Roz1WG

Waiting for the plan to start...

Terraform v1.9.8
on linux_amd64
Initializing plugins and modules...
╷
│ Error: no schema available for cloudflare_record.production while reading state; this is a bug in Terraform and should be reported
│
│
╵
╷
│ Error: no schema available for cloudflare_record.preview while reading state; this is a bug in Terraform and should be reported
│
│
╵
╷
│ Error: Unable to Read Previously Saved State for UpgradeResourceState
│
│   with cloudflare_pages_project.page,
│   on main.tf line 59, in resource "cloudflare_pages_project" "page":
│   59: resource "cloudflare_pages_project" "page" {
│
│ There was an error reading the saved resource state using the current
│ resource schema.
│
│ If this resource state was last refreshed with Terraform CLI 0.11 and
│ earlier, it must be refreshed or applied with an older provider version
│ first. If you manually modified the resource state, you will need to
│ manually modify it to match the current resource schema. Otherwise, please
│ report this to the provider developer:
│
│ AttributeName("build_config"): invalid JSON, expected "{", got "["
╵
Failed generating plan JSON
Exit code: 1

Failed to marshal plan to json: error marshaling prior state: unsupported attribute "domain"
Operation failed: 2 errors occurred:
        * failed running terraform plan (exit 1)
        * failed generating plan JSON: failed running command (exit 1)
```

Any recommendations how to handle this?

jacobbednarz:
have you seen the [migration guide](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/guides/version-5-upgrade)? if so, which steps did you take?

105ron:
Hi Jacob, I am having similar issues with the migration, and have been in contact with support opening case `01356987` but my support engineer is on a different time zone so there's delays between replies.

[...]

PR #5005, opened 2025-01-31
Clarify that state migration is covered in the renamed resources guide

ei-grad:
The previous wording could mislead users into thinking that state migration is not addressed in the guide at all. This change highlights that both resource renaming and state migration are covered.


ei-grad:
#4982 related

jacobbednarz:
thanks but we opted for this phrasing because there are multiple migration paths outlined and it's not all about state (nor will it be in the future if it is automatically handled).

PR #5029, opened 2025-02-03
docs: update deprecation date

pedrosousa:

## Changes being requested

Update deprecation dates, based on the changes already done for v4.x in #4785.

Not sure if this fix should be done in this repo or not, but we should update these dates to match what we have in Dev Docs and in v4.x docs.

### DIFF
--- a/docs/resources/filter.md
+++ b/docs/resources/filter.md
@@ -9,7 +9,7 @@ description: |-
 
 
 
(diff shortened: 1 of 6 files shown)


jacobbednarz:
ah apologies @pedrosousa. these would have been copied over before you made those changes in v4 and as they existed, we didn't overwrite them with the newer content.

Issue #4966, opened 2025-01-29
v5.0.0 Grit migration not working


MattJeanes:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.10.3
on windows_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v5.0.0
+ provider registry.terraform.io/hashicorp/azurerm v4.16.0

### Affected resource(s)

- cloudflare_dns_record

But affects everything from the migration

### Terraform configuration files

```hcl
locals {
  dns_records = [
    "hello-world",
    "@"
  ]
  proxied_records = {
    "hello-world" = true,
    "@"           = true
  }
}

resource "cloudflare_record" "dns" {
  for_each = toset(local.dns_records)

  zone_id = var.cloudflare_zone_id
  name    = each.key
  type    = "CNAME"
  content = azurerm_public_ip.abyss_public.fqdn
  ttl     = 1
  proxied = lookup(local.proxied_records, each.key, false)
}
```

### Link to debug output

N/A

### Panic output

N/A

### Expected output

The Grit migration will run successfully

### Actual output

Thie Grit migration is not found

### Steps to reproduce

https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/guides/version-5-upgrade#automatic-migration
 
1. Install grit
2. CD to terraform directory
3. grit apply cloudflare_terraform_v5 

ERROR (code: 200) - pattern definition not found: cloudflare_terraform_v5. Try running grit init.

(NOTE: `grit init` did not help)

### Additional factoids

I also tried adding this repo to the Grit config file and it nows shows other migrations but not the V5 upgrade one:

```yaml
version: 0.0.1
patterns:
  - name: github.com/getgrit/stdlib#*
  - name: github.com/cloudflare/terraform-provider-cloudflare#*
```

```
C:\_git\Abyss-Infrastructure\terraform [renovate/cloudflare-5.x ≡ +1 ~0 -0 !]>  grit list --language hcl

STANDARD LIBRARY PATTERNS


HCL patterns
  ✔ cloudflare_record_deprecate_value_for_content (source: cloudflare/terraform-provider-cloudflare)
[...]

PR #5030, opened 2025-02-03
fix: use correct name for Grit patterns

morgante:

## Changes being requested
Fixes https://github.com/cloudflare/terraform-provider-cloudflare/issues/4966 to use the correct pattern name.


jacobbednarz:
thanks @morgante! we've already landed this one on `next` (pending the incoming release). 

given this is custom code in the pipeline, we should be able to also land this now and it will just self resolve later but i'll keep an eye on it to avoid conflicts with our upcoming release.

PR #5139, opened 2025-02-13
[Examples] Fix cloudflare_dns_record example: changed ttl to 1 when record is marked as proxied

jacksonpradolima:

## Changes being requested

According to the provider: when a DNS record is marked as `proxied` the TTL must be 1 as Cloudflare will control the TTL internally.


jacobbednarz:
thanks but these examples are automatically generated based on the schemas. we're aware of some conflicts at the moment and will see what is possible going forward.

PR #5227, opened 2025-02-28
Change source field of pages_project to optional

dylanmtaylor:
I think this might be the fix needed for the error

│ Cannot set value for this attribute as the provider has marked it as read-only. Remove the configuration line setting the value. │ 

when setting the source property of a pages_project


jacobbednarz:
thanks for the PR, however, this is controlled via the OpenAPI schemas so the internal teams need to triage and address it. i'll pass this along however, there isn't an ETA when it may land.

PR #5367, opened 2025-03-21
Support module usage in grid state patterns

kub3let:

## Changes being requested
Support modules for grid patterns

## Additional context & links
If using modules with terraform the state file contains `"module": "module.*"` before `"mode": "managed",` which currently does not match the grid patterns


jacobbednarz:
thanks, however, we [intentionally don't support modules (or the migration)](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/guides/version-5-upgrade) of them as the abstractions are not guaranteed. if this works for you, that's great however, we cannot provide it for all customers.

PR #5471, opened 2025-04-08
docs: add missing rename for cloudflare_worker_route

MasseGuillaume:


jacobbednarz:
👋 this isn't present in the v4 => v5 migration notes as this was a part of the v4 migration which folks should be performing before the major upgrade.

PR #5539, opened 2025-05-01
docs: indicate cloudflare_workers_secret is removed in v5

markjmiller:


1000hz:
`cloudflare_worker_secret` / `cloudflare_workers_secret` in v4 actually corresponded to the Workers (not Workers for Platforms) Secrets endpoints. We should include guidance to specify secrets in `cloudflare_workers_script.bindings` instead when using v5.

Issue #5116, opened 2025-02-12
Panic migrating CAA record from v4 to v5


cmbuckley:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Before upgrade:
```
Terraform v1.10.5
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v4.52.0
```

After upgrade:
```
Terraform v1.10.5
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v5.0.0
```

### Affected resource(s)

* `cloudflare_dns_record` (when using `data` block/attribute for CAA record)

### Terraform configuration files

```hcl
# v4 resource
resource "cloudflare_record" "caa" {
  zone_id  = "..."
  name     = "@"
  ttl      = 1
  type     = "CAA"

  data {
    flags = "0"
    tag   = "issue"
    value = "letsencrypt.org"
  }
}

# v5 resource
resource "cloudflare_dns_record" "caa" {
  zone_id  = "..."
  name     = "@"
  ttl      = 1
  type     = "CAA"

  data = {
    flags = "0"
    tag   = "issue"
    value = "letsencrypt.org"
  }
}
```

### Link to debug output

https://gist.github.com/cmbuckley/9de33ff763db667b54a34eed1811db88

### Panic output

```
panic: interface conversion: attr.Value is basetypes.NumberValue, not basetypes.StringValue
```

### Expected output

Successful no-op or resource updated in-place

### Actual output

Panic as described above

### Steps to reproduce

1. Create CAA record using v4 of the provider using a data block, note that flag is a string (`flag = "0"`)
2. Follow [upgrade guide](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/guides/version-5-upgrade) to remove the resource from state, upgrade provider version and re-import as a `cloudflare_dns_record`
3. Run `terraform apply`, note that the resource will be updated in-place including the change `flags = 0 -> "0"`
4. Panic

### Additional factoids

The API returns a number per the documentation: https://developers.cloudflare.com/api/resources/dns/subresources/records/models/caa_record/#(schema)

[...]

PR #5579, opened 2025-05-13
fix(dns_record): align flags data type for datasource to resource and api response

jracollins:

## Changes being requested

Updating the `cloudflare_dns_record` and `cloudflare_dns_records` data_sources to align to the API response of a `Number` type (Instead of `Dynamic`) and also to the same that of the resource itself. 

This current difference prevents trying to compare any data sources to imports/resource definitions containing any CAA records with:

```
!!!!!!!!!!!!!!!!!!!!!!!!!!! TERRAFORM CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!
...
panic: inconsistent list element types
```

Where source is is a `cty.Number` and the other is `cty.DynamicPseudoType` 

## Additional context & links

The underlying type in the go library is `float64` so this aligns to that too. 
https://github.com/cloudflare/cloudflare-go/blob/1219741028ab564f1307be5a2d463a136a0df815/dns/record.go#L1785

It looks like there was a similar issue with the actual resource earlier in the 5.x versions here https://github.com/cloudflare/terraform-provider-cloudflare/issues/5116

### DIFF
--- a/docs/data-sources/dns_record.md
+++ b/docs/data-sources/dns_record.md
@@ -133,7 +133,7 @@ Read-Only:
 - `digest` (String) Digest.
 - `digest_type` (Number) Digest Type.
 - `fingerprint` (String) fingerprint.
-- `flags` (Dynamic) Flags for the CAA record.
+- `flags` (Number) Flags for the CAA record.
 - `key_tag` (Number) Key Tag.
 - `lat_degrees` (Number) Degrees of latitude.
 - `lat_direction` (String) Latitude direction.
--- a/docs/data-sources/dns_records.md
+++ b/docs/data-sources/dns_records.md
@@ -161,7 +161,7 @@ Read-Only:
 - `digest` (String) Digest.
 - `digest_type` (Number) Digest Type.
 - `fingerprint` (String) fingerprint.
-- `flags` (Dynamic) Flags for the CAA record.
+- `flags` (Number) Flags for the CAA record.
 - `key_tag` (Number) Key Tag.
 - `lat_degrees` (Number) Degrees of latitude.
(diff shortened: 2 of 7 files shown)


jacobbednarz:
thanks but this is generated by the internal schemas and the dynamic type is intended to support multiple variants of the `flags` which can be strings or integers.

jracollins:
Sure thing, I will re-raise an issue referencing this - thought a PR highlighting the required changes would be be just as clear even if not merged. 

Trying to work out why the autogenerated code must not be aligning for both data_source and resource I realise this fix is the wrong way, and instead the resource `data.flags` needs to be made dynamic rather than changing the data_source.

The `CAA` Record Type `flags` is a number, but `NAPTR` is a string, and so currently you would also be unable to manage any NAPTR records with the Terraform provider `resource`.

https://developers.cloudflare.com/api/resources/dns/subresources/records/models/caa_record/#(schema)%20%3E%20(property)%20data%20%3E%20(property)%20flags

https://developers.cloudflare.com/api/resources/dns/subresources/records/models/naptr_record/#(schema)%20%3E%20(property)%20data%20%3E%20(property)%20flags

It looks like it was actually this way before it was broken in https://github.com/cloudflare/terraform-provider-cloudflare/commit/d91aee1ef6b403177eb2bdfb5c73a8a3d36b79c3#diff-6498c19177df5133551e76105222248f721c4984851dd898a11b13c873e14bdaL126 (seemingly looking to resolve #5116), and I aligned these changes to the record based on your comments within that PR. 

I suspect the type will need to be dynamic in the resource to support both and the flags input validation need to be updated to check based on the record type.

PR #5588, opened 2025-05-15
TUN-9171: Fix vnet resource by mapping IsDefaultNetwork to IsDefault

jcsf:
## Summary
The virtual network resource uses a different field  to set and to read the same value. Therefore, to have the resource properly working on terraform v5, we need to map the value read to the value set. For this case the value used to write is `isDefault` and the value read, is `isDefaultNetwork`.

### DIFF
--- a/internal/services/zero_trust_tunnel_cloudflared_virtual_network/model.go
+++ b/internal/services/zero_trust_tunnel_cloudflared_virtual_network/model.go
@@ -6,6 +6,7 @@ import (
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/apijson"
 	"github.com/hashicorp/terraform-plugin-framework-timetypes/timetypes"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 )
 
 type ZeroTrustTunnelCloudflaredVirtualNetworkResultEnvelope struct {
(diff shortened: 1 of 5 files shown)


lmpn (approved):

chungthuang (approved):

lmpn (approved):

internal/services/zero_trust_tunnel_cloudflared_virtual_network/model.go
```diff
@@ -15,16 +16,26 @@ type ZeroTrustTunnelCloudflaredVirtualNetworkResultEnvelope struct {
...
 	DeletedAt        timetypes.RFC3339 `tfsdk:"deleted_at" json:"deleted_at,computed" format:"date-time"`
 }
 
+type newZeroTrustTunnelCloudflaredVirtualNetworkModel struct {
+	Name      types.String `json:"name,required"`
+	Comment   types.String `json:"comment,optional"`
+	IsDefault types.Bool   `json:"is_default,required"`
+}
+
 func (m ZeroTrustTunnelCloudflaredVirtualNetworkModel) MarshalJSON() (data []byte, err error) {
```
jacobbednarz, line 32: we shouldn't be modifying the marshaler given it's all codegen'd and makes resolving conflicts far more difficult. instead, if you need an alias, you should manually add the fields to the CRUD operations after the `data = &env.Result` bits to correctly populate the values.

Issue #5524, opened 2025-04-20
Optional fields cause crash in `cloudflare_zero_trust_tunnel_cloudflared(s)` when left null


aurimasniekis:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.11.4
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v5.3.0

### Affected resource(s)

data.cloudflare_zero_trust_tunnel_cloudflared
data.cloudflare_zero_trust_tunnel_cloudflareds

(Probably more than these as it's seams it's related to RFC3339 parsing)

### Terraform configuration files

```hcl
data "cloudflare_zero_trust_tunnel_cloudflareds" "this" {
  account_id = ""
}
```

### Link to debug output

Not needed

### Panic output

```
╷
│ Error: RFC3339 ValueRFC3339Time Error
│ 
│   with data.cloudflare_zero_trust_tunnel_cloudflareds.this,
│   on main.tf line 11, in data "cloudflare_zero_trust_tunnel_cloudflareds" "this":
│   11: data "cloudflare_zero_trust_tunnel_cloudflareds" "this" {
│ 
│ RFC3339 string value is null
╵
```

### Expected output

Optional fields should be… well, optional. The resource should return all tunnels when no filters are applied, instead of crashing due to null values.

### Actual output

Fails with an RFC3339 parsing error when optional fields are unset. This also means it’s currently impossible to fetch all tunnels without applying time filters. Even when both applied it doesn't work because both filters filter out results...

### Steps to reproduce

	1.	Use the data source without setting was_active_at or was_inactive_at
	2.	Run terraform plan or apply
	3.	Observe the crash

jhutchings1:
I've opened an internal issue for the feature team to work on this. Thanks!

musa-cf:
This should be fixed now in `v5.6.0`. Thank you for reporting this! Feel free to re-open this if you experience this issue again.

PR #5589, opened 2025-05-15
TUN-9253: Fix list cloudflared tunnels optional fields

jcsf:
## Summary
Previously, fields were parsed before checking if they were set, which could lead to empty value parsing. This change moves parsing after validation to ensure the fields are present.

Closes https://github.com/cloudflare/terraform-provider-cloudflare/issues/5524

### DIFF
--- a/internal/services/zero_trust_tunnel_cloudflared/data_source_model.go
+++ b/internal/services/zero_trust_tunnel_cloudflared/data_source_model.go
@@ -45,11 +45,6 @@ func (m *ZeroTrustTunnelCloudflaredDataSourceModel) toReadParams(_ context.Conte
 }
 
 func (m *ZeroTrustTunnelCloudflaredDataSourceModel) toListParams(_ context.Context) (params zero_trust.TunnelCloudflaredListParams, diags diag.Diagnostics) {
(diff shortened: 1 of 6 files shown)


lmpn (approved):

internal/services/zero_trust_tunnel_cloudflared/data_source_model.go
```diff
@@ -76,10 +71,18 @@ func (m *ZeroTrustTunnelCloudflaredDataSourceModel) toListParams(_ context.Conte
 		params.UUID = cloudflare.F(m.Filter.UUID.ValueString())
 	}
 	if !m.Filter.WasActiveAt.IsNull() {
-		params.WasActiveAt = cloudflare.F(mFilterWasActiveAt)
+		mFilterWasActiveAt, errs := m.Filter.WasActiveAt.ValueRFC3339Time()
+		diags.Append(errs...)
+		if errs != nil {
```
chungthuang: Should this be if error is not nil?

jcsf: You are right. Fixed

Issue #5641, opened 2025-06-04
cloudflare_notification_policy_webhooks loses fields after refresh


psapezhka:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.12.1
on linux_amd64

provider registry.terraform.io/cloudflare/cloudflare v5.5.0

### Affected resource(s)

cloudflare_notification_policy_webhooks

### Terraform configuration files

```hcl
resource "cloudflare_notification_policy_webhooks" "slack_cloudflare_alerts" {
  account_id = var.cloudflare_account_id
  name       = "Slack alerts"
  url        = "<<redacted>>"
  secret     = "<<redacted>>"
}
```

### Link to debug output

confidential

### Panic output

-

### Expected output

-

### Actual output

-

### Steps to reproduce

1. define `cloudflare_notification_policy_webhooks`
2. `import` it
3. verify `plan` and find the next
```
     # module.cloudflare.cloudflare_notification_policy_webhooks.slack_cloudflare_alerts will be updated in-place
  ~ resource "cloudflare_notification_policy_webhooks" "slack_cloudflare_alerts" {
      ~ created_at   = "2021-07-20T14:53:04Z" -> (known after apply)
        id           = "<<redacted>>"
      + last_failure = (known after apply)
      + last_success = (known after apply)
        name         = "Slack alerts"
      + secret       = (sensitive value)
      ~ type         = "slack" -> (known after apply)
        # (2 unchanged attributes hidden)
    }
```
4. despite this is already fail, run `apply`, 'cause "why not?", it's just a hook config
5. verify plan, and see "no changes"
6. run `refresh`
7. verify plan and see the same (step 3)
8. optional: inspect state before and after and notice the `secret` field equal to `null` after import/refresh 

### Additional factoids

part of the state after import/refresh

```
    {
      "module": "module.cloudflare",
      "mode": "managed",
      "type": "cloudflare_notification_policy_webhooks",
      "name": "slack_cloudflare_alerts",
      "provider": "module.cloudflare.provider[\"registry.terraform.io/cloudflare/cloudflare\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
[...]

Issue #5640, opened 2025-06-04
cloudflare_argo_smart_routing importing doesn't read the actual value


psapezhka:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.12.1
on linux_amd64
+ provider registry.terraform.io/cloudflare/cloudflare v5.5.0

### Affected resource(s)

cloudflare_argo_smart_routing

### Terraform configuration files

```hcl
resource "cloudflare_argo_smart_routing" "easybrain_argo" {
  zone_id = cloudflare_zone.easybrain.id
  value   = "on"
}
```

### Link to debug output

confidential

### Panic output

-

### Expected output

-

### Actual output

-

### Steps to reproduce

1. define `cloudflare_argo_smart_routing` resource
2. `import` it
3. `apply` shows changing of the `value` 
4. inspect state, find `"value": null`

### Additional factoids

part of state after the import

```
    {       
      "module": "module.cloudflare",
      "mode": "managed",
      "type": "cloudflare_argo_smart_routing",
      "name": "easybrain_argo",
      "provider": "module.cloudflare.provider[\"registry.terraform.io/cloudflare/cloudflare\"]",
      "instances": [
        {   
          "schema_version": 0,
          "attributes": {
            "id": "<<redacted>>",
            "value": null,
            "zone_id": "<<redacted>>"
          },
          "sensitive_attributes": [],
          "identity_schema_version": 0,
          "dependencies": [
            "module.cloudflare.cloudflare_zone.easybrain"
          ]
        }
      ]
    },
```

Moreover, `refresh` gives the same result, while argo routing is definitely enable with the UI

jhutchings1:
We have opened an issue for the team to investigate internally.

vaishakdinesh:
@psapezhka  This is fixed in 5.8.4!

Issue #5622, opened 2025-05-22
cloudflare_zero_trust_device_custom_profile always shows changes and errors after apply


nr-nrobledo:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

terraform -version
Terraform v1.11.3
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v5.5.0

### Affected resource(s)

cloudflare_zero_trust_device_custom_profile

### Terraform configuration files

```hcl
resource "cloudflare_zero_trust_device_custom_profile" "test_warp_policy" {
  account_id  = var.cloudflare_account_id
  name        = "Test Include Tunnel"
  description = "Test WARP client settings"
  precedence  = 99
  match                 = "identity.email == \"foo@domain.com\""
  enabled               = true
  allow_mode_switch     = false
  allow_updates         = false
  allowed_to_leave      = true
  auto_connect          = 0
  captive_portal        = 180
  disable_auto_fallback = false
  exclude_office_ips    = false
  include = [
    {address  = "192.168.200.0/24"},
    {host     = "ifconfig.io"}
  ]
  lan_allow_minutes              = 5
  register_interface_ip_with_dns = true
  support_url                    = ""
  switch_locked                  = false
  service_mode_v2 = {
    mode = "warp"
    # port = 0
  }
  tunnel_protocol = "masque"
}
```

### Link to debug output

.

### Expected output

I expect no changes if no changes were made.  But it always says there are changes with fallback_domains and sccm_vpn_boundary_support.

If I do actually make a change, let's say changing allow_mode_switch from false to true, I see the change I expect to see plus changes to fallback_domains and sccm_vpn_boundary_support.

If I apply it, then it errors and is unable to apply.

### Actual output

# cloudflare_zero_trust_device_custom_profile.test_warp_policy will be updated in-place
  ~ resource "cloudflare_zero_trust_device_custom_profile" "test_warp_policy" {
      ~ default                        = false -> (known after apply)
      ~ fallback_domains               = [
          - {
              - suffix = "home.arpa" -> null
            },
          - {
              - suffix = "intranet" -> null
            },
          - {
              - suffix = "internal" -> null
            },
          - {
[...]

Issue #5608, opened 2025-05-20
cloudflare_zero_trust_access_policy in 5.5.0 provider gives error upon apply unexpected new value: .app_count: was cty.NumberIntVal(0), but now cty.NumberIntVal(1)


anordby:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

PS C:\Users\b050ann\IdeaProjects\cloudflare> terraform -v
Terraform v1.12.0
on windows_amd64
+ provider app.terraform.io/mollerdigital/speed v0.2.0
+ provider registry.terraform.io/carlpett/sops v1.2.0
+ provider registry.terraform.io/cloudflare/cloudflare v5.5.0
+ provider registry.terraform.io/datadog/datadog v3.62.0
+ provider registry.terraform.io/hashicorp/azuread v3.4.0
+ provider registry.terraform.io/hashicorp/azurerm v4.29.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.3.7
+ provider registry.terraform.io/hashicorp/http v3.5.0
+ provider registry.terraform.io/hashicorp/random v3.7.2
+ provider registry.terraform.io/hashicorp/tfe v0.65.2
+ provider registry.terraform.io/hashicorp/time v0.13.1
+ provider registry.terraform.io/hashicorp/tls v4.1.0

### Affected resource(s)

cloudflare_zero_trust_access_application

### Terraform configuration files

```hcl
resource "cloudflare_zero_trust_access_policy" "ecom_moller_no" {
  account_id = local.account_ids["moller"]
  name       = "MollerEcomAdminPolicy"
  decision   = "allow"
  exclude    = []
  require    = []
  include = [
    for group in local.access_rule_groups["ecom_admin"] : {
      azure_ad = {
        identity_provider_id = local.access_provider_id_azure_ad
        id                   = local.azure_ad_group_ids[group]
      }
    }
  ]
}
```

### Link to debug output

https://gist.github.com/anordby/4203f48ec47009ba9d29584e79605086

### Expected output

No error

### Actual output

[...]

PR #5644, opened 2025-06-04
Update README.md

psapezhka728839:
Hi, please consider updating the README to inform users that the provider is far from ready for production use.

Zones, DNS records, Zero Trust, Cloudflare lists, page rules, rulesets barely work. API rate limiting is not implemented at all, which is obviously critical for large infra.

Here's a **partial** list of issues related to importing or modifying resources. These issues make version 5 nearly unusable:

https://github.com/cloudflare/terraform-provider-cloudflare/issues/5472
https://github.com/cloudflare/terraform-provider-cloudflare/issues/5517
https://github.com/cloudflare/terraform-provider-cloudflare/issues/5505
https://github.com/cloudflare/terraform-provider-cloudflare/issues/5135
https://github.com/cloudflare/terraform-provider-cloudflare/issues/5641
https://github.com/cloudflare/terraform-provider-cloudflare/issues/5640
https://github.com/cloudflare/terraform-provider-cloudflare/issues/5622
https://github.com/cloudflare/terraform-provider-cloudflare/issues/5608
https://github.com/cloudflare/terraform-provider-cloudflare/issues/5590
https://github.com/cloudflare/terraform-provider-cloudflare/issues/5577
https://github.com/cloudflare/terraform-provider-cloudflare/issues/5523

With all due respect, issues like these should not be happening. When using an **official** Terraform provider from a market leader, it’s reasonable to expect that, five minor versions after a major upgrade, everything should at least be **importable**.

I understand that "hello-world"-style configs built from scratch probably work fine. But real-world infrastructures — which can't simply be recreated — are suffering.

In the migration documentation, you wrote:

> We highly recommend reviewing this guide, make necessary changes and move to 5.x branch, as further 4.x releases are unlikely to happen outside of critical security fixes.

So we're just following that advice — and ending up spending dozens of hours mid-migration trying to manually fix a Terraform state with hundreds of thousand lines, just to make it somewhat usable.

Again, with all due respect, please consider adding a simple notice in the README to save engineers from wasting time.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Cloudflare Terraform Provider
+# Cloudflare Terraform Provider (WIP: NOT READY FOR PRODUCTION, PLEASE USE v4)
 
 The [Cloudflare Terraform provider](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs) provides convenient access to
 the [Cloudflare REST API](https://developers.cloudflare.com/api) from Terraform.


RichardoC:
Sounds like this matches their official stance <https://github.com/cloudflare/terraform-provider-cloudflare/issues/5573>

ic4-y:
Yes, it would seem this provider is a bit of an early access version for something great to come. But then, somebody has to test it, right? And all the engineers around here filing all these useful bug reports for free have been well-worth Cloudflare's time.

virtualjack:
> Sounds like this matches their official stance #5573

The written message downplays the reality.  Verbally, I've been told don't use 5.X unless you have to.

Thank you for putting this list together.  I was looking to put a similar list based so this is a time saver.

PR #5647, opened 2025-06-05
APISHI-4659: Add acceptance tests for schema validation

janrueth:

## Changes being requested
This change adds acceptance tests for schema validation resources

### DIFF
--- a/internal/services/schema_validation_operation_settings/resource_test.go
+++ b/internal/services/schema_validation_operation_settings/resource_test.go
@@ -1,0 +1,58 @@
+package schema_validation_operation_settings_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/acctest"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/consts"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/utils"
+	"github.com/hashicorp/go-uuid"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+func TestAccCloudflarePerOperationSetting(t *testing.T) {
+	rndResourceName := utils.GenerateRandomResourceName()
+
+	// resourceID is resourceIdentifier . resourceName
+	resourceID := "cloudflare_schema_validation_operation_settings." + rndResourceName
+	zoneID := os.Getenv("CLOUDFLARE_ZONE_ID")
+
+	resource.Test(t, resource.TestCase{
+		IsUnitTest:               false,
+		PreCheck:                 func() { acctest.TestAccPreCheck(t) },
+		ProtoV6ProviderFactories: acctest.TestAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			// this sets up a schema and a block per-operation mitigation action (note: we're not testing log as it might require further permissions)
+			{
+				Config: testAccCloudflareSchemaValidationWithOperationMitigationAction(rndResourceName, zoneID, "block"),
+
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceID, consts.ZoneIDSchemaKey, zoneID),
+					resource.TestCheckResourceAttrWith(resourceID, "operation_id", func(value string) error {
+						_, err := uuid.ParseUUID(value)
+						return err
+					}),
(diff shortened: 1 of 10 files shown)


alexpovel (commented):
lgtm

internal/services/schema_validation_operation_settings/resource_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// resourceID is resourceIdentifier . resourceName
+	resourceID := "cloudflare_schema_validation_operation_settings." + rndResourceName
+	zoneID := os.Getenv("CLOUDFLARE_ZONE_ID")
+
+	resource.Test(t, resource.TestCase{
+		IsUnitTest:               false,
+		PreCheck:                 func() { acctest.TestAccPreCheck(t) },
+		ProtoV6ProviderFactories: acctest.TestAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
```
vaishakdinesh, line 25: Its good to add a step which does a state import and a step to just plan after an apply so you can be sure that you don't any drift after a successful apply.

internal/services/schema_validation_settings/testdata/schema.tf
```diff
@@ -0,0 +1,19 @@
...
+  host = "api.example.com"
+  method = "GET"
+}
+
+resource "cloudflare_schema_validation_schemas" "%[1]s" {
+  zone_id = "%[2]s"
+  kind = "openapi_v3"
+  name = "test_schema.yaml"
+  source = file("testdata/test_schema.yaml")
+  validation_enabled = true
```
vaishakdinesh, line 13: what happens if this is false? Is that a default value? In that case will the service return the default value?

internal/services/schema_validation_operation_settings/testdata/schema.tf
```diff
@@ -0,0 +1,20 @@
...
+  host = "api.example.com"
+  method = "GET"
+}
+
+resource "cloudflare_schema_validation_schemas" "%[1]s" {
+  zone_id = "%[2]s"
+  kind = "openapi_v3"
+  name = "test_schema.yaml"
+  source = file("testdata/test_schema.yaml")
+  validation_enabled = true
```
vaishakdinesh, line 13: what happens if this is false? Is that a default value? In that case will the service return the default value?

internal/services/schema_validation_settings/resource_test.go
```diff
@@ -0,0 +1,84 @@
...
+		override := *validation_override_mitigation_action
+		if override != "null" {
+			override = fmt.Sprintf(`"%s"`, override)
+		}
+		block += fmt.Sprintf("\n  validation_override_mitigation_action = %s", override)
+	}
+	return acctest.LoadTestCase("schema.tf", resourceName, zone, block)
+}
+
+func ptrTo[T any](v T) *T {
```
vaishakdinesh: `nit` not a big deal. but if its useful maybe export and move to the acctest pkg? Else its fine where it is

internal/services/schema_validation_operation_settings/testdata/schema.tf
janrueth: yeah, if this is left out or false it will return a default value. There is a testcase in the schema_validation resource for taht

internal/services/schema_validation_settings/testdata/schema.tf
janrueth: Yeah part is tested explicity in the tests for cloudflare_schema_validation_schemas.
Here this resource simply exists to reference it for the setting which is the thing under the test here

internal/services/schema_validation_settings/resource_test.go
janrueth: sure, can move!

internal/services/schema_validation_operation_settings/resource_test.go
janrueth: Added but discovered a small issue on the API that I'm fixing

vaishakdinesh: @janrueth is this being worked on?

djhworld: @vaishakdinesh fix has been rolled out, I've ran these acceptance tests locally and they now pass.

vaishakdinesh (approved):

Issue #5488, opened 2025-04-09
cloudflare_zero_trust_tunnel_cloudflared cannot be imported


chris-pinola-rf:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```bash
$ terraform -version
Terraform v1.11.3
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v5.3.0
```

### Affected resource(s)

* cloudflare_zero_trust_tunnel_cloudflared

### Terraform configuration files

```hcl
import {
  to = cloudflare_zero_trust_tunnel_cloudflared.tunnel
  id = "{account_id}/{tunnel_id}"
}

resource "cloudflare_zero_trust_tunnel_cloudflared" "tunnel" {
  account_id    = "..."
  name          = "..."
  tunnel_secret = "..."
  config_src    = "cloudflare"
}
```

### Link to debug output

Not comfortable doing this publicly

### Expected output

The tunnel resource to be imported without needing replacement.

### Actual output

The provider wants to destroy and recreate the tunnel resource because it detects a change in the `config_src` property:

```hcl
  # cloudflare_zero_trust_tunnel_cloudflared.tunnel must be replaced
  # (imported from "{account_id}/{tunnel_id}")
  # Warning: this will destroy the imported resource
-/+ resource "cloudflare_zero_trust_tunnel_cloudflared" "tunnel" {
...
      + config_src        = "cloudflare" # forces replacement
```

### Steps to reproduce

1. Create a `cloudflare_tunnel` resource with version 4.52.0 of the provider with `config_src` set to `"cloudflare"`
2. Remove the `cloudflare_tunnel` resource from the Terraform state
3. Upgrade provider to 5.3.0
4. Run `terraform plan` with the HCL from above (`import {}` block + `cloudflare_zero_trust_tunnel_cloudflared`)

### Additional factoids

Was expecting this bug to be fixed in version 5.3.0 from #5432 but unfortunately it still persists.

### References

Related issues:
* #5363 
* #5450
* #5432

jhutchings1:
Thanks for letting us know. @musa-cf will take a look

gil-tohar-cyera:
[...]

PR #5663, opened 2025-06-13
Fix zero_trust_tunnel_cloudflared import

adrawbond-alto:

## Changes being requested
This fixes https://github.com/cloudflare/terraform-provider-cloudflare/issues/5488
## Additional context & links
`GET /accounts/{account_id}/cfd_tunnel/{tunnel_id}` returns a boolean `remote_config` instead of the `config_src` that is used in the `POST`. My understanding from the API docs is that they mean the same thing. This PR converts `remote_config` to `config_src` when importing a `cloudflare_zero_trust_tunnel_cloudflared`, as `config_src` is the property used in the Terraform resource. I have tested this change against my own infrastructure and it resolves the problem of recreating an imported `cloudflare_zero_trust_tunnel_cloudflared`. I'm not sure how to fix this with the code auto-generator, but I thought I'd put this up as a reference. 

https://developers.cloudflare.com/api/resources/zero_trust/subresources/tunnels/subresources/cloudflared/methods/get/
https://developers.cloudflare.com/api/resources/zero_trust/subresources/tunnels/subresources/cloudflared/methods/create/


hradaideh:
i can verify that
when you import cloudflare_zero_trust_tunnel_cloudflared resource, the `config_src` is set as null alongside `tunnel_secret` causing terraform to recreate the tunnel

plus, `policy`'s ID and `service_auth_401_redirect` are always shown as state diff

adrawbond-alto:
This was fixed in v5.8.4

PR #5671, opened 2025-06-17
fix for magic wan resources

tiagolicanton:
- fix V5 support for magic_wan_ipsec_tunnel, magic_wan_gre_tunnel, and magic_wan_static_route
 - remove extra wrapper field for specific Get/Put responses
 - remove extra fields from resource schema
 - add x-magic-new-hc-target:true header for endpoints accepting new health check format


internal/services/magic_wan_gre_tunnel/model.go
```diff
@@ -14,21 +14,18 @@ type MagicWANGRETunnelResultEnvelope struct {
 }
 
 type MagicWANGRETunnelModel struct {
```
vaishakdinesh: Give the amount of change in this model, I think its better to create a new `custom.go` file with a new Model and unmarshaller and use that in resource.go. We will definitely get merge conflicts here which the team may not be able to resolve easily. 
thoughts @musa-cf @mgirouard

internal/services/magic_wan_gre_tunnel/resource_test.go
```diff
@@ -30,19 +31,15 @@ func TestAccCloudflareGRETunnelExists(t *testing.T) {
 		ProtoV6ProviderFactories: acctest.TestAccProtoV6ProviderFactories,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccCheckCloudflareGRETunnelSimple(rnd, rnd, rnd, accountID),
+				Config: testAccCheckCloudflareGRETunnelSimple(rnd, rnd, rnd, accountID, cfIP),
```
vaishakdinesh: add steps to check no drift and import?

internal/services/magic_wan_gre_tunnel/schema.go
```diff
@@ -130,222 +130,11 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				Computed:    true,
 				CustomType:  timetypes.RFC3339Type{},
 			},
-			"modified": schema.BoolAttribute{
```
vaishakdinesh: Perhaps here too, similar to the `custom.go`, perhaps `custom_schema.go` and then something like 

`func customResourceSchema(ctx context.Context) schema.Schema {
    return schema.Schema{}
}
`
and in `schema.go`
`func ResourceSchema(ctx context.Context) schema.Schema {
    return customResourceSchema(ctx)
}
`

vaishakdinesh:
similar thoughts for the other places where model and schema have changed significantly. Can we also increase the testing? Add a little more steps for plan only and maybe import. Having different use cases, different attributes, would also greatly help.

vaishakdinesh:
@tiagolicanton once the tests are updated and are passing, can you add it to the inclusion list so we can run them everytime `scripts/run-ci-acceptance-tests`

internal/services/magic_wan_gre_tunnel/resource.go
```diff
@@ -174,6 +184,13 @@ func (r *MagicWANGRETunnelResource) Read(ctx context.Context, req resource.ReadR
 		return
 	}
 	bytes, _ := io.ReadAll(res.Body)
+	// remove extra wrapper field
+	bytes, err = utils.StripWrapper(bytes, "gre_tunnel")
+	if err != nil {
+		resp.Diagnostics.AddError("failed to parse response", err.Error())
+		return
+	}
+
```
vaishakdinesh: Could this be done as a part of a custom umarshaller defined on the custom model? Perform this action and then use the `apijson.Unmarshal`.

Page rules does something similar http://github.com/cloudflare/terraform-provider-cloudflare/blob/main/internal/services/page_rule/custom.go

Not big a deal, just lesser places for conflict :) @musa-cf @mgirouard

vaishakdinesh: same suggestion of the other places in the other resources.

internal/utils/magic_utils.go
```diff
@@ -0,0 +1,55 @@
...
+	switch accountID {
+	case "5e51c34a8fffd8b96136e87be3af7110":
+		cfIP = "162.159.68.213"
+	case "f037e56e89293a057740de681ac9abbe":
+		cfIP = "162.159.73.109"
+	default:
+		t.Fatalf("need to specify an anycast IP for account %s", accountID)
+	}
+	return cfIP
+}
```
vaishakdinesh, line 70: Just worried about the public aspect of this repo. If we can use variables or secrets to read these values, it might be better? 
Thoughts @tiagolicanton @musa-cf @mgirouard

internal/services/magic_wan_gre_tunnel/resource_test.go
```diff
@@ -120,45 +131,53 @@ func TestAccCloudflareGRETunnelUpdateMulti(t *testing.T) {
...
+func lookupCfIP(t *testing.T, accountID string) string {
+	var cfIP string
+	// CF anycast IPs are allocated per account
+	switch accountID {
+	case "5e51c34a8fffd8b96136e87be3af7110":
+		cfIP = "162.159.68.213"
+	default:
+		t.Fatalf("need to specify anycast IP for account %s", accountID)
+	}
+	return cfIP
```
vaishakdinesh: Just worried about the public aspect of this repo. If we can use variables or secrets to read these values, it might be better? 
Thoughts @tiagolicanton @musa-cf @mgirouard

vaishakdinesh (approved):
lgtm just need a resolution with the comments.

vaishakdinesh:
I would  like to test this going from 5.6 to 5.7. I know it might be broken in 5.6 but would really recommend an upgrade path test. even if manual. from last working version to this. @tiagolicanton @musa-cf @mgirouard

internal/services/magic_wan_gre_tunnel/resource_test.go
musa-cf: Shouldn't be an issue. The tags aren't sensitive information on their own.

musa-cf:
> I would like to test this going from 5.6 to 5.7. I know it might be broken in 5.6 but would really recommend an upgrade path test. even if manual. from last working version to this. @tiagolicanton @musa-cf @mgirouard

I agree with this, given the scope of the changes. If at all possible, can you test this flow:
1. Create the resources on v5.6.0
2. Set `TF_CLI_CONFIG_FILE=~/.terraform.rc` to point to your local provider with your changes
3. Run `terraform plan -refresh-only` to see if any changes are detected

Otherwise, the changes look good to me. Thank you for improving the tests!

internal/services/magic_wan_gre_tunnel/resource_test.go
tiagolicanton: good catch! this block is obsolete, will remove.

internal/utils/magic_utils.go
tiagolicanton: `f037e56e89293a057740de681ac9abbe` is already in the repo, and the cfIPs are public info https://www.cloudflare.com/ips/

musa-cf (approved):

PR #5683, opened 2025-06-23
AUTH-7265 Fix zero_trust_access_applications

GreenStage:
- Fixes `zero_trust_access_applications` errors and warnings that completely break the resource in TF v5
- Adds API normalization, plan modifications, and new validators for `zero_trust_access_applications`
- Partially fixes  `zero_trust_access_policy` & `zero_trust_access_identity_provider` just enough to allow the `zero_trust_access_applications` to pass. A complete fix for this resources is still needed in future PRs
- Re-enables all  `zero_trust_access_applications` acceptance tests, and adds some more
- This will likely break auto-generation for Access, but that is ok. We need to fix the provider first to unblock users, we can keep working in the future to move whatever logic we can from here into the API docs, and make smooth it out with the generator.

## Changes being requested
- Resource `zero_trust_access_application`
- Resource `zero_trust_access_policy`
- Resource `zero_trust_access_identity_provider`

### DIFF
--- a/internal/services/zero_trust_access_application/model.go
+++ b/internal/services/zero_trust_access_application/model.go
@@ -24,10 +24,10 @@ type ZeroTrustAccessApplicationModel struct {
 	CustomDenyMessage           types.String                                                               `tfsdk:"custom_deny_message" json:"custom_deny_message,optional"`
 	CustomDenyURL               types.String                                                               `tfsdk:"custom_deny_url" json:"custom_deny_url,optional"`
 	CustomNonIdentityDenyURL    types.String                                                               `tfsdk:"custom_non_identity_deny_url" json:"custom_non_identity_deny_url,optional"`
-	Domain                      types.String                                                               `tfsdk:"domain" json:"domain,optional"`
+	Domain                      types.String                                                               `tfsdk:"domain" json:"domain,computed_optional"`
 	HeaderBgColor               types.String                                                               `tfsdk:"header_bg_color" json:"header_bg_color,optional"`
 	LogoURL                     types.String                                                               `tfsdk:"logo_url" json:"logo_url,optional"`
-	Name                        types.String                                                               `tfsdk:"name" json:"name,optional"`
+	Name                        types.String                                                               `tfsdk:"name" json:"name,computed_optional"`
 	OptionsPreflightBypass      types.Bool                                                                 `tfsdk:"options_preflight_bypass" json:"options_preflight_bypass,optional"`
 	ReadServiceTokensFromHeader types.String                                                               `tfsdk:"read_service_tokens_from_header" json:"read_service_tokens_from_header,optional"`
 	SameSiteCookieAttribute     types.String                                                               `tfsdk:"same_site_cookie_attribute" json:"same_site_cookie_attribute,optional"`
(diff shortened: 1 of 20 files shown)


GreenStage:
Needs this to be fixed first https://github.com/cloudflare/terraform-provider-cloudflare/issues/5684
Proposal fix in https://github.com/cloudflare/terraform-provider-cloudflare/pull/5685

internal/services/zero_trust_access_application/model.go
```diff
@@ -41,21 +40,19 @@ type ZeroTrustAccessApplicationModel struct {
...
+	Tags                        customfield.List[types.String]                                             `tfsdk:"tags" json:"tags,optional"`
 	Destinations                customfield.NestedObjectList[ZeroTrustAccessApplicationDestinationsModel]  `tfsdk:"destinations" json:"destinations,computed_optional"`
-	LandingPageDesign           customfield.NestedObject[ZeroTrustAccessApplicationLandingPageDesignModel] `tfsdk:"landing_page_design" json:"landing_page_design,computed_optional"`
-	Policies                    customfield.NestedObjectList[ZeroTrustAccessApplicationPoliciesModel]      `tfsdk:"policies" json:"policies,computed_optional"`
-	SaaSApp                     customfield.NestedObject[ZeroTrustAccessApplicationSaaSAppModel]           `tfsdk:"saas_app" json:"saas_app,computed_optional"`
+	LandingPageDesign           customfield.NestedObject[ZeroTrustAccessApplicationLandingPageDesignModel] `tfsdk:"landing_page_design" json:"landing_page_design,optional"`
+	Policies                    *[]ZeroTrustAccessApplicationPoliciesModel                                 `tfsdk:"policies" json:"policies,optional"`
```
jroyal, line 54: Are policies actually optional?

internal/services/zero_trust_access_application/model.go
```diff
@@ -41,21 +40,19 @@ type ZeroTrustAccessApplicationModel struct {
...
+	SaaSApp                     customfield.NestedObject[ZeroTrustAccessApplicationSaaSAppModel]           `tfsdk:"saas_app" json:"saas_app,optional"`
```
jroyal, line 55: I thought we always return a saas_app block on apps?

internal/services/zero_trust_access_application/model.go
```diff
@@ -523,24 +520,22 @@ type ZeroTrustAccessApplicationPoliciesRequireServiceTokenModel struct {
...
+	AccessTokenLifetime           types.String                                                    `tfsdk:"access_token_lifetime" json:"access_token_lifetime,computed_optional"`
 	AllowPKCEWithoutClientSecret  types.Bool                                                      `tfsdk:"allow_pkce_without_client_secret" json:"allow_pkce_without_client_secret,optional"`
 	AppLauncherURL                types.String                                                    `tfsdk:"app_launcher_url" json:"app_launcher_url,optional"`
-	ClientID                      types.String                                                    `tfsdk:"client_id" json:"client_id,computed_optional"`
-	ClientSecret                  types.String                                                    `tfsdk:"client_secret" json:"client_secret,computed_optional"`
+	ClientID                      types.String                                                    `tfsdk:"client_id" json:"client_id,computed"`
+	ClientSecret                  types.String                                                    `tfsdk:"client_secret" json:"client_secret,computed,no_refresh"`
```
jroyal: is no_refresh something we will have to maintain going forward? Not sure if we can include that in the api docs.

internal/services/zero_trust_access_application/schema.go
```diff
@@ -1248,28 +1344,32 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+				Validators: []validator.Object{
+					customvalidator.RequiresOtherStringAttributeToBeOneOf(path.MatchRoot("type"), saasAppTypes...),
+					customvalidator.RequiredWhenOtherStringIsOneOf(path.MatchRoot("type"), saasAppTypes...),
+				},
 				Attributes: map[string]schema.Attribute{
 					"auth_type": schema.StringAttribute{
-						Computed:    true,
 						Description: "Optional identifier indicating the authentication protocol used for the saas app. Required for OIDC. Default if unset is \"saml\"\nAvailable values: \"saml\", \"oidc\".",
-						Optional:    true,
+						Required:    true,
```
jroyal: Seems weird to have this marked as required when the auth_type says that its optional.

internal/services/zero_trust_access_application/schema.go
```diff
@@ -1326,52 +1426,86 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+						},
+						PlanModifiers: []planmodifier.String{
+							stringplanmodifier.UseStateForUnknown(),
 						},
 					},
 					"name_id_transform_jsonata": schema.StringAttribute{
 						Description: "A [JSONata](https://jsonata.org/) expression that transforms an application's user identities into a NameID value for its SAML assertion. This expression should evaluate to a singular string. The output of this expression can override the `name_id_format` setting.",
 						Optional:    true,
+						Validators: []validator.String{
+							customvalidator.RequiresOtherStringAttributeToBeOneOf(path.MatchRoot("saas_app").AtName("auth_type"), "saml"),
```
jroyal: Is name id transform only available on saml?

internal/services/zero_trust_access_policy/model.go
```diff
@@ -27,9 +27,7 @@ type ZeroTrustAccessPolicyModel struct {
 	Exclude                      customfield.NestedObjectList[ZeroTrustAccessPolicyExcludeModel] `tfsdk:"exclude" json:"exclude,computed_optional"`
 	Include                      customfield.NestedObjectList[ZeroTrustAccessPolicyIncludeModel] `tfsdk:"include" json:"include,computed_optional"`
 	Require                      customfield.NestedObjectList[ZeroTrustAccessPolicyRequireModel] `tfsdk:"require" json:"require,computed_optional"`
-	AppCount                     types.Int64                                                     `tfsdk:"app_count" json:"app_count,computed"`
 	CreatedAt                    timetypes.RFC3339                                               `tfsdk:"created_at" json:"created_at,computed" format:"date-time"`
-	Reusable                     types.Bool                                                      `tfsdk:"reusable" json:"reusable,computed"`
```
jroyal, line 32: Why remove reusable?

internal/services/zero_trust_access_application/schema.go
ajholland: correct, I never got around to adding it for oidc (and we'd name it something else probably)

internal/services/zero_trust_access_application/schema.go
GreenStage: oh, in the API I think its required? CC @ajholland if I'm reading it wrong.
Either way we change it here or in the API docs

internal/services/zero_trust_access_application/model.go
GreenStage: no, just saas_apps

internal/services/zero_trust_access_application/model.go
GreenStage: yeah, you can make an app with no policies

internal/services/zero_trust_access_policy/model.go
GreenStage: because all policies through this resource have to be reusable, it doesnt support legacy ones

internal/services/zero_trust_access_application/schema.go
ajholland: It shouldn't be required - customers can create a saml saas_app without specifying it. It is required if they want to create an oidc saas_app. (this is all because of backwards compatibility with existing saml api configs)

internal/services/zero_trust_access_application/model.go
GreenStage: yeah I dont think we need it afte the custom code, ill delete it

internal/services/zero_trust_access_application/schema.go
GreenStage: Ok, I'll have a look

jroyal (approved):

Issue #5548, opened 2025-05-05
cloudflare_api_token resource always shows changes (drift)


boillodmanuel:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```
Terraform v1.9.8
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v5.3.0
```

### Affected resource(s)

- cloudflare_api_token

### Terraform configuration files

```hcl
terraform {
  required_version = ">= 1.0"
  required_providers {
    cloudflare = {
      source  = "cloudflare/cloudflare"
      version = "~>5.3"
    }
  }
  backend "local" {}
}

provider "cloudflare" {
  email   = local.CLOUDFLARE_EMAIL
  api_key = local.CLOUDFLARE_API_KEY
}


data "cloudflare_api_token_permission_groups_list" "all" {
}

locals {
  api_token_zone_permissions_groups_map = {
    for perm in data.cloudflare_api_token_permission_groups_list.all.result :
    perm.name => perm.id
    if contains(perm.scopes, "com.cloudflare.api.account.zone")
  }
}

resource "cloudflare_api_token" "test_api_token" {
  name   = "test-cf-v5-api-token"
  status = "active"

  policies = [{
    effect = "allow"
    permission_groups = [
      { "id" = local.api_token_zone_permissions_groups_map["DNS Write"] },
      { "id" = local.api_token_zone_permissions_groups_map["Zone Read"] },
    ]
    resources = {
      "com.cloudflare.api.account.${local.ACCOUNT_ID}" = "*"
    }
  }]
}
```

### Link to debug output

no

### Expected output

No changes detected

### Actual output

```bash
Terraform will perform the following actions:

  # cloudflare_api_token.test_api_token will be updated in-place
  ~ resource "cloudflare_api_token" "test_api_token" {
      + condition    = (known after apply)
        id           = "de493ea5bbd25d3b48f6f66abe35fdde"
      ~ issued_on    = "2025-05-05T09:22:28Z" -> (known after apply)
      + last_used_on = (known after apply)
      ~ modified_on  = "2025-05-05T09:37:52Z" -> (known after apply)
        name         = "test-cf-v5-api-token"
[...]

PR #5689, opened 2025-06-24
api token permission groups order does not matter

ghost:

## Changes being requested

API Token permission group order does not matter and permission group IDs should be unique so we can treat as a set to resolve unintended drift.

## Additional context & links

https://github.com/cloudflare/terraform-provider-cloudflare/issues/5548


mgirouard (approved):
There _should_ be a way to control this through an OAS extension, but that doesn't currently exist today. It's best to have this as custom code for now.

PR #5695, opened 2025-06-25
chore(tests): Add acceptance tests for WARP Connector resource

jcsf:
## Summary
This commit adds the acceptance tests for the new WARP Connector resource that was added to the Terraform v5


internal/services/zero_trust_tunnel_warp_connector/data_source_test.go
```diff
@@ -1,36 +1,40 @@
...
 
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/acctest"
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/utils"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
-	"github.com/hashicorp/terraform-plugin-testing/terraform"
 )
 
-func TestAccCloudflareZeroTrustTunnelWarpConnectorDataSource_Basic(t *testing.T) {
+func TestAccWARPConnectorDatasourceBasic(t *testing.T) {
+	if os.Getenv("CLOUDFLARE_API_TOKEN") != "" {
```
chungthuang, line 15: Is this unset for the same reason? https://github.com/cloudflare/terraform-provider-cloudflare/pull/5695/files

jcsf: Yes, but I don't think the reason is true. The issue is that in our openapi spec we only have the authentication mechanism of the API Key and not the API Token. So, the code assumed we don't support it. When we actually support. Since it doesn't hurt us, I just kept it as it is.

internal/services/image/resource.go
```diff
@@ -226,7 +227,7 @@ func (r *ImageResource) ImportState(ctx context.Context, req resource.ImportStat
 	}
 
 	data.AccountID = types.StringValue(path_account_id)
-	data.ID = types.StringValue(path_image_id)
+	data.ID = jsontypes.NewNormalizedValue(path_image_id)
```
vaishakdinesh: Would this be needed? @musa-cf @mgirouard 
i feel this could be generated?

jcsf: I had to manually change it because the code wasn't compiling without it.

musa-cf (approved):

PR #5691, opened 2025-06-24
setting api_token default status

ghost:
- [X] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested

Adding a default status for api_token. 

As it is today, users must provide a `status = "active"`. If the value is not set, then subsequent updates fail:

```
➜  steves-tf-test terraform apply
var.cf_token
  Enter a value: XXXXX

data.cloudflare_api_token_permission_groups_list.all: Reading...
data.cloudflare_api_token_permission_groups_list.all: Read complete after 0s
cloudflare_api_token.terraform: Refreshing state... [id=20584c36811ec4357e04261c653fb428]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # cloudflare_api_token.terraform will be updated in-place
  ~ resource "cloudflare_api_token" "terraform" {
        id           = "20584c36811ec4357e04261c653fb428"
      ~ issued_on    = "2025-06-24T20:53:40Z" -> (known after apply)
      + last_used_on = (known after apply)
      ~ modified_on  = "2025-06-24T20:53:40Z" -> (known after apply)
        name         = "terraform"
      ~ policies     = [
          ~ {
              ~ id                = "0c327b7d68c640c99d0b0f4df8dc3613" -> (known after apply)
              ~ permission_groups = [
                  ~ {
                        id   = "3030687196b94b638145a3953da2b699"
                      ~ name = "Zone Settings Write" -> (known after apply)
                    },
                  ~ {
                        id   = "e6d2666161e84845a636613608cee8d5"
                      ~ name = "Zone Write" -> (known after apply)
                    },
                  ~ {
                      ~ id   = "c03055bc037c4ea9afb9a9f104b7b721" -> "4755a26eedb94da69e1066d98aa820be"
                      ~ name = "SSL and Certificates Write" -> (known after apply)
                    },
                  ~ {
                      ~ id   = "3e0b5820118e47f3922f7c989e673882" -> "c03055bc037c4ea9afb9a9f104b7b721"
                      ~ name = "Logs Write" -> (known after apply)
                    },
                  ~ {
                      ~ id   = "4755a26eedb94da69e1066d98aa820be" -> "3e0b5820118e47f3922f7c989e673882"
                      ~ name = "DNS Write" -> (known after apply)
                    },
                ]
                # (2 unchanged attributes hidden)
            },
        ]
      - status       = "active" -> null
[...]

PR #5697, opened 2025-06-25
fixing api token tests

ghost:

## Changes being requested

Fixing API token tests.

Existing test suites were broken. 4 major issues:
1. all tests were affected by the absence of the "status" field which currently results in unexpected drift. I have a separate PR to address the drift bug https://github.com/cloudflare/terraform-provider-cloudflare/pull/5691
2. some tests were expecting an incorrect schema for the IP condition object. I think it's possible we changed that some time in the last year.
3. Existing tests claim that CLOUDFLARE_API_TOKEN is not supported for these actions, but it is now
4. I am not sure what the history of the TestAccAPIToken_AllowDeny test is, but it seems to be redundant now. It does not truly test any allow vs deny functionality, it also does not make use of the zone parameter that the code is referencing. It looks like it in effect is retesting what TestAccAPIToken_Basic was testing, so I removed the test and the test .tf file


internal/services/api_token/resource_test.go
```diff
@@ -40,41 +33,7 @@ func TestAccAPIToken_Basic(t *testing.T) {
 	})
 }
 
-func TestAccAPIToken_AllowDeny(t *testing.T) {
```
vaishakdinesh, line 43: Is this test not relevant anywhere or testing something incorrectly?

vaishakdinesh:
I see 135 commits, probably have to resync and then rebase?
but lgtm

vaishakdinesh (approved):

ghost:
shit. bad rebase. i will have to do some git surgery to this

internal/services/api_token/resource_test.go
ghost: Yeah this is covered in point #4 on the PR description. Really not clear what this was originally intending to test.

ghost:
Ok git surgery successful

PR #5724, opened 2025-07-07
set ssl attribute to optional

ryanmilne-digicert:
Change the SSL attribute of `cloudflare_custom_hostname` to be optional to match the behavior of the API.

```
curl -v https://api.cloudflare.com/client/v4/zones/<zone_id>/custom_hostnames \
    -H 'Content-Type: application/json' \
    -H "Authorization: Bearer $API_TOKEN" \
    -d '{
      "hostname": "somedomain.example.com"
    }'
```


ryanmilne-digicert:
Did not mean to make a PR into Cloudflare Main branch.

PR #5740, opened 2025-07-10
Fix failing CI parity tests

mgirouard:

## Changes being requested

Skips parity tests for page rules and access apps. Since there's custom code, we can't expect parity.


musa-cf:
This should target `next`

PR #5741, opened 2025-07-11
Skip failing acceptance tests for future follow-up

mgirouard:
Account needs to be entitled. Unclear what that entitlement is.

### DIFF
--- a/internal/services/zone_lockdown/resource_test.go
+++ b/internal/services/zone_lockdown/resource_test.go
@@ -11,7 +11,6 @@ import (
 )
 
 func TestAccCloudflareZoneLockdown(t *testing.T) {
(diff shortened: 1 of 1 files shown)


vaishakdinesh (approved):

internal/services/api_token/data_source_test.go
```diff
@@ -10,6 +10,7 @@ import (
 )
 
 func TestAccCloudflareAPITokenData(t *testing.T) {
+	t.Skip(`FIXME: Attribute 'id' expected to be set`)
```
vaishakdinesh: I think this is an assertion error. updating the assertion passes this test.

mgirouard:
I'll proceed with merging this now. All remaining skipped tests don't represent any regression and all have follow up work scheduled.

PR #5776, opened 2025-07-22
Fix magic transit connector (MCONN)

yihuaf:
- [ x ] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested
Fix the current MCONN resources and added acceptance test.

## Additional context & links

Working with the code-gen is proven to be challenging. As a result, the team is making the decision to create custom models and schemas for the time being. While some of the custom changes can be done through the code-gen, we are not taking chances for further time sink to fight the code-gent at this moment.


internal/services/magic_transit_connector/model.go
```diff
@@ -15,11 +15,11 @@ type MagicTransitConnectorModel struct {
...
-	InterruptWindowDurationHours types.Float64                     `tfsdk:"interrupt_window_duration_hours" json:"interrupt_window_duration_hours,optional"`
-	InterruptWindowHourOfDay     types.Float64                     `tfsdk:"interrupt_window_hour_of_day" json:"interrupt_window_hour_of_day,optional"`
-	Notes                        types.String                      `tfsdk:"notes" json:"notes,optional"`
-	Timezone                     types.String                      `tfsdk:"timezone" json:"timezone,optional"`
+	Activated                    types.Bool                        `tfsdk:"activated" json:"activated,computed_optional"`
+	InterruptWindowDurationHours types.Float64                     `tfsdk:"interrupt_window_duration_hours" json:"interrupt_window_duration_hours,computed_optional"`
+	InterruptWindowHourOfDay     types.Float64                     `tfsdk:"interrupt_window_hour_of_day" json:"interrupt_window_hour_of_day,computed_optional"`
+	Notes                        types.String                      `tfsdk:"notes" json:"notes,computed_optional"`
+	Timezone                     types.String                      `tfsdk:"timezone" json:"timezone,computed_optional"`
```
vaishakdinesh: These changes should be able to be done in the openapi schema/stainless config and have it trickle down. 
@musa-cf

internal/services/magic_transit_connector/model.go
```diff
@@ -34,6 +34,6 @@ func (m MagicTransitConnectorModel) MarshalJSONForUpdate(state MagicTransitConne
 }
 
 type MagicTransitConnectorDeviceModel struct {
-	ID           types.String `tfsdk:"id" json:"id,optional"`
-	SerialNumber types.String `tfsdk:"serial_number" json:"serial_number,optional"`
+	ID           types.String `tfsdk:"id" json:"id,computed_optional"`
+	SerialNumber types.String `tfsdk:"serial_number" json:"serial_number,computed_optional"`
```
vaishakdinesh: Think same for these.

internal/services/magic_transit_connector/resource.go
```diff
@@ -112,20 +112,31 @@ func (r *MagicTransitConnectorResource) Update(ctx context.Context, req resource
...
-	if err != nil {
-		resp.Diagnostics.AddError("failed to serialize http request", err.Error())
-		return
-	}
 	res := new(http.Response)
 	env := MagicTransitConnectorResultEnvelope{*data}
-	_, err = r.client.MagicTransit.Connectors.Update(
+
+	// The model doesn't map to the edit params directly, so we need to do a
+	// little mapping here.
```
vaishakdinesh: How come? Is the unmarshaller not setting it correctly?

vaishakdinesh:
> Should remove last_heartbeat and last_seen_version from the field
This would technically be a breaking change. I think we just need to tell terraform to not care about this fields since they dynamically change all the time.

internal/services/magic_transit_connector/resource.go
yihuaf: We don't allow editing for the device field in the request, so the edit (patch) can't include the device field from the model. The generated code is just marshal the `MagicTransitConnectorModel` directly into the params, which is incorrect. The API will reject.

yihuaf:
> This would technically be a breaking change. I think we just need to tell terraform to not care about this fields since they dynamically change all the time.

I am less worried about the breaking change, since this resource is only added in 5.7 and is broken since its release. So if there is a time to make decision like this, it is now.

How should I tell terraform to ignore these fields in the state? Is this going to be through the openapi?

internal/services/magic_transit_connector/model.go
yihuaf: compute_optional is fine. I am more concerned about how to add the PlanModifier in the right places.

internal/services/magic_transit_connector/model.go
yihuaf: The main question here is how to add PlanModifier for these two fields in the schema

internal/services/magic_transit_connector/model.go
vaishakdinesh: Planmodifiers might still have to be custom code. IE, manually added.

internal/services/magic_transit_connector/model.go
vaishakdinesh: Planmodifiers might still have to be custom code. IE, manually added

vaishakdinesh:
> How should I tell terraform to ignore these fields in the state? Is this going to be through the openapi?
i think adding the tag `no_refresh` should prevent this from being detected by terraform. I'll get you an example of it. 
https://github.com/cloudflare/terraform-provider-cloudflare/blob/main/internal/services/r2_bucket_event_notification/model.go#L20

internal/services/magic_transit_connector/resource.go
vaishakdinesh: Ah gotcha.

vaishakdinesh: Ah gotcha.

vaishakdinesh: Could we move this into a function?

yihuaf: Yes :)

yihuaf:
@vaishakdinesh I added more tests. Can you give the approval to run CI? I tested locally with our own lab account, but I want to see if CI account will also work.

internal/services/magic_transit_connector/resource.go
```diff
@@ -112,20 +135,13 @@ func (r *MagicTransitConnectorResource) Update(ctx context.Context, req resource
...
-	dataBytes, err := data.MarshalJSONForUpdate(*state)
-	if err != nil {
-		resp.Diagnostics.AddError("failed to serialize http request", err.Error())
-		return
-	}
 	res := new(http.Response)
 	env := MagicTransitConnectorResultEnvelope{*data}
-	_, err = r.client.MagicTransit.Connectors.Update(
+
+	_, err := r.client.MagicTransit.Connectors.Edit(
```
yihuaf: @vaishakdinesh The API to use needs to be `Edit`, not `Update`. (patch vs. put)

vaishakdinesh: Still looking into this. Seems like the config update didn't work?

yihuaf: It worked... sort of. The JSON patch update works. Although this is because of the side effect that `device` is marked as required and never change in a `Update` call, so it won't include the `device` when making the JSON edit params in this function. 

Rebased... Now waiting on the compute_optional to land.

yihuaf:
@vaishakdinesh while we wait on the compute_optional change to land, can you allow this PR to run CI.

yihuaf:
PTAL. Given that we have not been able to make the code-gen produce the expected code, we decided to change strategy and pursue a custom model and schema implementation. This should allow us to move forward with the fix while minimizing the effort to resolve merge conflicts, while providing our customers with a good TF experiences. I think this is the best compromise for the time being.

yihuaf:
@musa-cf Looks like I still need approval to run CI.

PR #5771, opened 2025-07-19
DS-15222: Add LogpushJob_Full test to test all fields

soheiokamoto:
This adds `LogpushJob_Full` test to test all fields.

This is to demonstrate what changes need to be made on `model.go` and `schema.go`, which are manually done here to make the t

Notable changes are the following:
 1. Remove `Computed` and `Default` from some fields to match with Logpush API.
 2. Added `Computed` for `Enabled` field, as Logpush API always return value (default `false`).
 2. Change `max_upload_*` to be integer, with range validator.
 3. Adding `no_refresh` attribute for the fields that cause "drift" due to Logpush API json `omitempty`.

Some changes would need to be applied to Logpush API OpenAPI schema to make same changes on the next SDK generation.

PR #5782, opened 2025-07-23
DS-15222: Fix Logpush job schema and model

soheiokamoto:
* DS-15222: Add no_refresh attribute for the fields that cause "drift" due to Logpush API json omitempty
    
    This adds `no_refresh` attribute for the fields that cause "drift" due to Logpush API json `omitempty`.

---
* DS-15222: Add validators to Logpush job max_upload_* fields, as SDK generator doesn't recognize multiple definitions by anyOf
    
    This adds validators to Logpush job `max_upload_*` fields, as SDK generator doesn't recognize multiple definitions by `anyOf`.


soheiokamoto:
Tests are failing for unknown reason (there are failures from unrelated files).

soheiokamoto:
This is a subset of https://github.com/cloudflare/terraform-provider-cloudflare/pull/5771, in case it needs to be done separately, but closing this, as the main PR includes acceptance tests, and they should go together ideally.

Issue #5578, opened 2025-05-13
cloudflare_logpush_job plan always suggests changes


joshuamsager:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.10.3
cloudflare/cloudflare v5.3.0

### Affected resource(s)

* cloudflare_logpush_job

### Terraform configuration files

```hcl
resource "cloudflare_logpush_job" "kitchensink" {
  name             = "HTTP-requests-kitchensink"
  zone_id          = var.zone_id
  kind             = var.kind
  dataset          = "http_requests"
  destination_conf = "https://${var.destination}?header_Authorization=Basic%20${var.auth_token}"

  # Used only for "Core" logging (kind is null)
  max_upload_interval_seconds = 30
  max_upload_records          = 1000

  output_options = {
    field_names      = local.fields
    timestamp_format = "unixnano"
    sample_rate      = var.sample
    output_type      = "ndjson"
    record_prefix    = "{"
    record_suffix    = "}\n"
    field_delimiter  = ","
  }

  enabled = var.enabled
}
```

### Link to debug output

Shared in Cloudflare Support ticket

### Expected output

Plan after apply should not propose changes

### Actual output

  # module.observe_apps.cloudflare_logpush_job.kitchensink will be updated in-place
  ~ resource "cloudflare_logpush_job" "kitchensink" {
      + error_message               = (known after apply)
        id                          = 677037
      - kind                        = "" -> null
      ~ last_complete               = "2025-04-15T14:03:29Z" -> (known after apply)
      + last_error                  = (known after apply)
        name                        = "HTTP-requests-kitchensink"
      ~ output_options              = {
          + cve_2021_44228   = false
            # (7 unchanged attributes hidden)
        }
        # (7 unchanged attributes hidden)
    }

### Steps to reproduce

1. Provision a logpush job without a _kind_ attribute
2. Configure output_options
3. Apply configuration
4. terraform plan

anordby:
I have this issue also. Please keep us posted @jhutchings1

joshuamsager:
Still broken in v5.5.0

dduportal:
[...]

PR #5785, opened 2025-07-24
Marks both logpull_options and output_options as computed

Cyb3r-Jak3:

## Changes being requested

Fixes an issue where the v4 provider has perma-drift due to the API setting both `logpull_options` and `output_options` but the current provider only expecting one to be set.

I also removed some err checks that weren't being used, as they were checking errors that would be already checked.

## Additional context & links
Similar to https://github.com/cloudflare/terraform-provider-cloudflare/issues/5578, but this fix is for the v4 provider.

### DIFF
--- a/.changelog/5785.txt
+++ b/.changelog/5785.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/cloudflare_logpush_job: mark both logpull_options and output_options as computed
+```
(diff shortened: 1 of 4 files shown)


Cyb3r-Jak3:
Not stale. Still waiting on review

vaishakdinesh:
Hi @Cyb3r-Jak3 
we're currently not actively releasing the v4 provider bar any critical/security fixes.
@KaydeeDee @rupalims

PR #5786, opened 2025-07-24
Fix zero_trust_organization resource drift & add import fn

GreenStage:

## Additional context & links

AUTH-7420

### DIFF
--- a/internal/services/zero_trust_organization/model.go
+++ b/internal/services/zero_trust_organization/model.go
@@ -4,7 +4,6 @@ package zero_trust_organization
 
 import (
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/apijson"
-	"github.com/hashicorp/terraform-plugin-framework-timetypes/timetypes"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
@@ -26,8 +25,6 @@ type ZeroTrustOrganizationModel struct {
 	CustomPages                    *ZeroTrustOrganizationCustomPagesModel `tfsdk:"custom_pages" json:"custom_pages,optional"`
 	LoginDesign                    *ZeroTrustOrganizationLoginDesignModel `tfsdk:"login_design" json:"login_design,optional"`
 	AutoRedirectToIdentity         types.Bool                             `tfsdk:"auto_redirect_to_identity" json:"auto_redirect_to_identity,computed_optional"`
-	CreatedAt                      timetypes.RFC3339                      `tfsdk:"created_at" json:"created_at,computed" format:"date-time"`
-	UpdatedAt                      timetypes.RFC3339                      `tfsdk:"updated_at" json:"updated_at,computed" format:"date-time"`
 }
 
 func (m ZeroTrustOrganizationModel) MarshalJSON() (data []byte, err error) {
--- a/internal/services/zero_trust_organization/resource.go
+++ b/internal/services/zero_trust_organization/resource.go
@@ -5,6 +5,8 @@ package zero_trust_organization
 import (
 	"context"
 	"fmt"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/importpath"
+	"github.com/hashicorp/terraform-plugin-framework/types"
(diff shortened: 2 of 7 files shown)


internal/services/zero_trust_organization/resource.go
```diff
@@ -194,13 +204,62 @@ func (r *ZeroTrustOrganizationResource) Read(ctx context.Context, req resource.R
...
+
+	resp.Diagnostics.Append(normalizeReadZeroTrustOrganizationAPIData(ctx, data, stateData)...)
 	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
 }
 
 func (r *ZeroTrustOrganizationResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
 
 }
 
+func (r *ZeroTrustOrganizationResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
```
mgirouard, line 221: This should be fine, but just wanted to call out that imports _should_ be automatic. If they don't exist automatically, it's usually an indication of a mis-alignment in the OAS of path parameters and content in the response body.

scripts/run-ci-acceptance-tests
```diff
@@ -138,6 +138,7 @@ go test -run "^TestAcc" -count 1 \
     ./internal/services/zero_trust_gateway_logging \
     ./internal/services/zero_trust_gateway_proxy_endpoint \
     ./internal/services/zero_trust_list \
+    ./internal/services/zero_trust_organization \
```
mgirouard, line 141: Thank you! 🙏

mgirouard (approved):

internal/services/zero_trust_organization/model.go
```diff
@@ -26,8 +25,6 @@ type ZeroTrustOrganizationModel struct {
 	CustomPages                    *ZeroTrustOrganizationCustomPagesModel `tfsdk:"custom_pages" json:"custom_pages,optional"`
 	LoginDesign                    *ZeroTrustOrganizationLoginDesignModel `tfsdk:"login_design" json:"login_design,optional"`
 	AutoRedirectToIdentity         types.Bool                             `tfsdk:"auto_redirect_to_identity" json:"auto_redirect_to_identity,computed_optional"`
-	CreatedAt                      timetypes.RFC3339                      `tfsdk:"created_at" json:"created_at,computed" format:"date-time"`
```
vaishakdinesh, line 29: Can this not be done through the schema?

internal/services/zero_trust_organization/resource.go
GreenStage: The /organizations API is anything but restful

like to get a single API, you call /organizations, which only returns a single item and not a list

Anyway, we should fix in TF as we're not gonna be able to change the API now

internal/services/zero_trust_organization/model.go
GreenStage: No, we need these in the schema.
We don't want them in terraform, specially the UpdatedAt as it can change through time without any apply.

vaishakdinesh: Hmm. the computed fields wont be a cause of the an update in place or drift if you're seeing that. There's likely something else which is changing. When that is the case, all computed fields will be shown as known after apply. 
I dont think this should be done. Unless im missing something. What is the exact issue which this solves?

GreenStage: Showing meaningless 'known after apply' fields in planning is enough of a reason and only adds noise to customers' experience.
We can discuss the theory and long term plan internally, but right now we need to stop the bleeding caused by auto generated code, and I'm not taking any chances with unneeded stuff.

musa-cf (approved):

internal/services/zero_trust_organization/model.go
vaishakdinesh: I'll merge this. But we should discuss this internally and find a path forward.

PR #5800, opened 2025-07-27
Rex/fix gateway settings version

rexscaria:
Somehow recently in this commit(or merge resolved), one of our version argument was removed from the generated code, despite it is still being present in the api spec [here](https://gitlab.cfdata.org/cloudflare/ares/access-resolver-account-service/-/blob/main/schemas/client/account/teams/accounts.yaml?ref_type=heads#L564)

This breaking the gateway setting resource completely and this PR should fix the issue, before we draw a release.

### DIFF
--- a/internal/services/zero_trust_gateway_settings/model.go
+++ b/internal/services/zero_trust_gateway_settings/model.go
@@ -31,7 +31,7 @@ func (m ZeroTrustGatewaySettingsModel) MarshalJSONForUpdate(state ZeroTrustGatew
 
 type ZeroTrustGatewaySettingsSettingsModel struct {
 	Antivirus             *ZeroTrustGatewaySettingsSettingsAntivirusModel                             `tfsdk:"antivirus" json:"antivirus,optional"`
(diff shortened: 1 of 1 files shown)


vaishakdinesh (approved):

musa-cf:
I see in the OpenAPI schema, `version` is `computed_optional`. This PR has it as `computed`. Is this intentional?

vaishakdinesh:
> I see in the OpenAPI schema, `version` is `computed_optional`. This PR has it as `computed`. Is this intentional?

Good catch. I think its an optional field? @rexscaria

rexscaria:
> > I see in the OpenAPI schema, `version` is `computed_optional`. This PR has it as `computed`. Is this intentional?
> 
> Good catch. I think its an optional field? @rexscaria

I can control this from the docs, so i can change that later as long as the version is added back.

but computed optional is fine if we have to pick one

vaishakdinesh:
Would that field be set on the service side? Or can a user set it as well? When i was trying it out. I was able to omit the version field entirely. Which then would make it an optional field yeah?

vaishakdinesh (approved):

PR #5812, opened 2025-07-29
Add `workers_script` import verification checks

1000hz:

## Changes being requested
This PR adds import verification steps to the acceptance tests of the `cloudflare_workers_script` resource.

It also replaces the legacy `Check` functions in these tests with more modern State Checks.

## Additional context & links
Refs IAC-117

### DIFF
--- a/internal/services/workers_script/resource_test.go
+++ b/internal/services/workers_script/resource_test.go
@@ -15,7 +15,10 @@ import (
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/acctest"
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/utils"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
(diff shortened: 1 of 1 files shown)


internal/services/workers_script/resource.go
```diff
@@ -304,9 +304,8 @@ func (r *WorkersScriptResource) Read(ctx context.Context, req resource.ReadReque
 		content = string(bytes)
 	}
 
-	// only update `content` if it was already present in state
-	// which might not be the case if `content_file` is used instead
-	if !data.Content.IsNull() {
+	// only update `content` if `content_file` isn't being used instead
+	if data.ContentFile.IsNull() {
```
1000hz, line 308: Previously, the implementation assumed that `content` would only be null in state if `content_file` is being used instead. However, when importing a resource both `content` and `content_file` are null in state. Instead, we should just populate `content` _unless_ `content_file` is present.

PR #5814, opened 2025-07-29
Rex/fix gateway settings

rexscaria:
This will make inspection + host selector values optional.

I tried to drive these from the api docs. But it looks like this was accidentally changed in a merge conflict. This PR should  revert that.


vaishakdinesh:
@rexscaria can you branch from next and target next?

Issue #5770, opened 2025-07-17
cloudflare_zero_trust_access_application type warp diff on every apply


silazare:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.10.5
on darwin_arm64

Cloudflare Terraform provider 5.7.1

### Affected resource(s)

cloudflare_zero_trust_access_application
with `type = "warp"` only

### Terraform configuration files

```hcl
resource "cloudflare_zero_trust_access_application" "warp" {
  name                        = "Warp Login App"
  domain                      = "${var.auth_domain}/warp"
  account_id                  = var.cloudflare_account_id
  allowed_idps                = [var.cloudflare_identity_provider_google_workspace_id]
  type                        = "warp"
  auto_redirect_to_identity   = true
  allow_authenticate_via_warp = false
  policies = [{
    id         = cloudflare_zero_trust_access_policy.warp.id
    precedence = 1
  }]
}
```

### Link to debug output

none

### Expected output

No change detected

### Actual output

```
  # cloudflare_zero_trust_access_application.warp will be updated in-place
  ~ resource "cloudflare_zero_trust_access_application" "warp" {
      + allow_authenticate_via_warp = false
        id                          = "xxx"
        name                        = "Warp Login App"
      - session_duration            = "24h" -> null
        # (7 unchanged attributes hidden)
    }
```

### Steps to reproduce

1. terraform apply
2. terraform plan

### Additional factoids

It tries to remove session duration which is set as 24h in UI
Session duration should be allowed in WARP type as well.
When try to revert it back, there is an error in atrribute:
```
│ Error: Invalid Attribute Combination
│
│   with cloudflare_zero_trust_access_application.warp,
│   on cloudflare-zero-trust-tunnel.tf line 54, in resource "cloudflare_zero_trust_access_application" "warp":
│   54:   session_duration            = "24h"
│
│ "session_duration" can only be set if "type" is one of: "saas", "dash_sso", "self_hosted", "ssh", "vnc", "rdp", "app_launcher"
```

KaydeeDee:
Thank you for submitting an issue! I've sent it over to the team to investigate.

bckg:
[...]

PR #5817, opened 2025-07-30
Fix zero_trust_access_application constant diffs for warp type apps

GreenStage:
- warp type apps can not use allow_authenticate_via_warp
- Fixed an issue when setting policies to null would show constant drift as the API interprets null values for the policies array as a no-op

Addresses  https://github.com/cloudflare/terraform-provider-cloudflare/issues/5770

### DIFF
--- a/internal/services/zero_trust_access_application/resource.go
+++ b/internal/services/zero_trust_access_application/resource.go
@@ -63,7 +63,11 @@ func (r *ZeroTrustAccessApplicationResource) Create(ctx context.Context, req res
 		return
 	}
 
-	resp.Diagnostics.Append(loadConfigSensitiveValuesForWriting(ctx, data, &req.Config)...)
+	res := new(http.Response)
+	env := ZeroTrustAccessApplicationResultEnvelope{*data}
+	params := zero_trust.AccessApplicationNewParams{}
+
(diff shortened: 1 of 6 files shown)


internal/services/zero_trust_access_application/resource.go
```diff
@@ -63,7 +63,11 @@ func (r *ZeroTrustAccessApplicationResource) Create(ctx context.Context, req res
 		return
 	}
 
-	resp.Diagnostics.Append(loadConfigSensitiveValuesForWriting(ctx, data, &req.Config)...)
+	res := new(http.Response)
+	env := ZeroTrustAccessApplicationResultEnvelope{*data}
+	params := zero_trust.AccessApplicationNewParams{}
+
+	resp.Diagnostics.Append(normalizeWriteZeroTrustApplicationAPIData(ctx, data, &req.Config)...)
```
GreenStage, line 70: had to move this from line 66 to here because the request normalization was side-affecting the response object as well.

Stainless is unpacking the API response object into the existing request 'data' object instead of unpacking it into a new empty one. I believe stainless does this for a reason, so I'm not risking a change to that logic.

By moving the API request normalization until after the response envelope is instantiated, we avoid the issue without affecting other logic

musa-cf (approved):

PR #5851, opened 2025-08-04
chore: modernize zero_trust_access_policy acceptance tests

tamas-jozsa:
This MR modernizes zero_trust_access_policy tests by using V5 syntax. I had to skip `TestAccCloudflareAccessPolicy_IsolationRequired` because it depends on ZT Gateway Settings resource, which has some bugs that I need to investigate first.


internal/services/zero_trust_access_policy/normalizations.go
```diff
@@ -39,3 +40,21 @@ func normalizeReadZeroTrustAccessPolicyAPIData(ctx context.Context, data, source
...
+// Specialized normalization for import operations where API omits false boolean values
+func normalizeImportZeroTrustAccessPolicyAPIData(ctx context.Context, data *ZeroTrustAccessPolicyModel) diag.Diagnostics {
+	diags := make(diag.Diagnostics, 0)
+
+	// Set boolean fields to false if they are null (API omits false values during import)
+	if data.ApprovalRequired.IsNull() {
+		data.ApprovalRequired = types.BoolValue(false)
+	}
+	if data.IsolationRequired.IsNull() {
+		data.IsolationRequired = types.BoolValue(false)
```
GreenStage, line 53: these are going to show diffs if the value in the state is null, no?

internal/services/zero_trust_access_policy/testdata/accesspolicyreusableconfig.tf
```diff
@@ -5,5 +5,8 @@
       include = [{
         email = ["a@example.com"]
       }]
+      approval_required              = "false"
+      isolation_required             = "false"
+      purpose_justification_required = "false"
     }
```
GreenStage, line 11: can we also have a test with these ommited? its a valid config

tamas-jozsa: added TestAccCloudflareAccessPolicy_OptionalBooleans

GreenStage:
lgtm

PR #5854, opened 2025-08-04
fix(cloudflare_list_item) fix bulk operation polling and update

broswen:

## Changes being requested
Creating, updating, or deleting list items results in an asynchronous bulk job that must be polled to know when it has completed and the new list item is visible. 

It is not possible to update an existing list item by ID, so we must create a new list item and delete the old one.

If the "key" value is the same, they will be merged into the same list item. So we can optimize updates by only requiring replacement if the "key" value has changed.

## Additional context & links

Bulk operation status
https://developers.cloudflare.com/api/resources/rules/subresources/lists/subresources/bulk_operations/methods/get/

### DIFF
--- a/internal/services/list_item/resource.go
+++ b/internal/services/list_item/resource.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"strconv"
+	"time"
 
 	"github.com/cloudflare/cloudflare-go/v5"
 	"github.com/cloudflare/cloudflare-go/v5/option"
@@ -97,6 +98,12 @@ func (r *ListItemResource) Create(ctx context.Context, req resource.CreateReques
 		return
 	}
 
+	err = pollBulkOperation(ctx, data.AccountID.ValueString(), createEnv.Result.OperationID.ValueString(), r.client)
+	if err != nil {
+		resp.Diagnostics.AddError("list item bulk operation failed", err.Error())
+		return
+	}
+
 	searchTerm := getSearchTerm(data)
 	findItemRes := new(http.Response)
 	_, err = r.client.Rules.Lists.Items.List(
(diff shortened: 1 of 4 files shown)


internal/services/list_item/resource.go
```diff
@@ -97,6 +98,12 @@ func (r *ListItemResource) Create(ctx context.Context, req resource.CreateReques
 		return
 	}
 
+	err = pollBulkOperation(ctx, data.AccountID.ValueString(), createEnv.Result.OperationID.ValueString(), r.client)
```
mgirouard, line 104: I don't know if we own the `ctx` instance, so I'm unclear what timeout exists already. Given that, should we have a new context `WithTimeout`?

(Ditto other calls to `pollBulkOperation`)

broswen: I don't see a `context.WithTimeout` used anywhere currently in the provider.

I'm adding a `RequestOption` onto the client calls.  `option.WithRequestTimeout(time.Second*3)`

broswen: It looks like in general Terraform has a 20min timeout set for each of the resource operations.
https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/retries-and-customizable-timeouts#default-timeouts-and-deadline-exceeded-errors

jeffa5:
@mgirouard @vaishakdinesh Could we get a review on this please, or merge it. If there are any changes I'll have to make a new PR (can't push to this branch).

jeffa5:
@mgirouard @vaishakdinesh This is superseded by https://github.com/cloudflare/terraform-provider-cloudflare/pull/5881 which has more acceptance tests and fixes the behaviour implemented here. The problem was that the search was not reliable at returning the expected item at the first index, and we don't really have a way of updating items so they should just be recreated.

jeffa5:
This can be closed now that the other PR #5881  has been merged.

PR #5865, opened 2025-08-05
chore: modernize zero_trust_access_application tests

tamas-jozsa:
The MR modernizes tests from V4 to V5 syntax and increases testing coverage.


tamas-jozsa:
> lgtm - we do have an issue with some boolean fields like allow_iframe where setting them to false results in persistent plan changes. Do you think that is also worth adding tests for? I can do so since I need to post a change to fix it.

sure I can do that. Can you give me a list of those settings that you experience issues with?

PR #5866, opened 2025-08-05
Use `customfield` package within `cloudflare_ruleset` resource

zakcutner:


vaishakdinesh:
Can you branch from next and target next please?

internal/services/ruleset/data_source_model.go
```diff
@@ -21,10 +21,10 @@ type RulesetDataSourceModel struct {
 	RulesetID   types.String                                              `tfsdk:"ruleset_id" path:"ruleset_id,optional"`
 	AccountID   types.String                                              `tfsdk:"account_id" path:"account_id,optional"`
 	ZoneID      types.String                                              `tfsdk:"zone_id" path:"zone_id,optional"`
-	Description types.String                                              `tfsdk:"description" json:"description,computed"`
 	Kind        types.String                                              `tfsdk:"kind" json:"kind,computed"`
 	Name        types.String                                              `tfsdk:"name" json:"name,computed"`
 	Phase       types.String                                              `tfsdk:"phase" json:"phase,computed"`
+	Description types.String                                              `tfsdk:"description" json:"description,computed"`
```
vaishakdinesh, line 27: just checking, is description really computed?

zakcutner: I believe it is, because the API will return a default description of `""` if one is not provided. In any case though, since this is the data source, surely everything is computed? I'm mainly focussing on the resource (and not the data source) for now, so I don't think this is a blocker either way.

PR #5872, opened 2025-08-06
Add rule list acceptance tests

jeffa5:

## Changes being requested

Add acceptance tests for rule lists resources and data sources.


internal/services/list/resource_test.go
```diff
@@ -6,16 +6,21 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"strconv"
+	"strings"
 	"testing"
 
-	"github.com/cloudflare/cloudflare-go"
+	"github.com/cloudflare/cloudflare-go/v4"
```
mgirouard: Is there any reason why we're using the previous cf-go? AFAIK this should be using the latest version.

jeffa5: No, I must have been on an old version of the repo or something which meant it worked with the v4 interface. Fixing that and the data source model which seem to have gotten lost in my changes.

jeffa5: Fixed now

vaishakdinesh:
@jeffa5 a lot of the test file are removed? Any particular reason? Do we not need them anymore?

jeffa5:
> @jeffa5 a lot of the test file are removed? Any particular reason? Do we not need them anymore?

@vaishakdinesh The files were not really needed, we can just have a parameter for the list kind and use a single file for all of the cases.

PR #5874, opened 2025-08-06
fix(list_item): Data source

jeffa5:

## Changes being requested

Fixes the data source models for list items.

### DIFF
--- a/internal/services/list_item/data_source.go
+++ b/internal/services/list_item/data_source.go
@@ -5,14 +5,15 @@ package list_item
 import (
 	"context"
 	"fmt"
-	"github.com/cloudflare/cloudflare-go/v5/option"
-	"github.com/cloudflare/terraform-provider-cloudflare/internal/apijson"
-	logging "github.com/cloudflare/terraform-provider-cloudflare/internal/logging"
 	"io"
 	"net/http"
 
 	"github.com/cloudflare/cloudflare-go/v5"
+	"github.com/cloudflare/cloudflare-go/v5/option"
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
+
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/apijson"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/logging"
 )
 
 type ListItemDataSource struct {
--- a/internal/services/list_item/data_source.go
+++ b/internal/services/list_item/data_source.go
@@ -5,6 +5,11 @@ package list_item
 import (
 	"context"
 	"fmt"
+	"github.com/cloudflare/cloudflare-go/v5/option"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/apijson"
+	logging "github.com/cloudflare/terraform-provider-cloudflare/internal/logging"
+	"io"
+	"net/http"
 
 	"github.com/cloudflare/cloudflare-go/v5"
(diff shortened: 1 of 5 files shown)


internal/services/list_item/data_source.go
```diff
@@ -46,39 +52,39 @@ func (d *ListItemDataSource) Configure(ctx context.Context, req datasource.Confi
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	res := new(http.Response)
+	env := ListItemResultDataSourceEnvelope{*data}
+	_, err := d.client.Rules.Lists.Items.Get(
+		ctx,
+		data.ListID.ValueString(),
+		data.ItemID.ValueString(),
```
vaishakdinesh, line 72: @jeffa5 the PR @broswen open this was data.ID.ValueString(). Can you confirm which one it should be?

internal/services/list_item/data_source_model.go
```diff
@@ -13,28 +18,29 @@ type ListItemResultDataSourceEnvelope struct {
 
 type ListItemDataSourceModel struct {
 	AccountID  types.String                                              `tfsdk:"account_id" path:"account_id,required"`
-	ItemID     types.String                                              `tfsdk:"item_id" path:"item_id,required"`
 	ListID     types.String                                              `tfsdk:"list_id" path:"list_id,required"`
+	ItemID     types.String                                              `tfsdk:"item_id" path:"item_id,required"`
```
vaishakdinesh, line 22: Depending on the resolution of the above comment, this can probably be removed?

internal/services/list_item/data_source_schema.go
```diff
@@ -18,13 +22,89 @@ func DataSourceSchema(ctx context.Context) schema.Schema {
 				Description: "The Account ID for this resource.",
 				Required:    true,
 			},
+			"list_id": schema.StringAttribute{
+				Description: "The unique ID of the list.",
+				Required:    true,
+			},
 			"item_id": schema.StringAttribute{
```
vaishakdinesh, line 28: if the model attribute is removed, lets remove it from here too.

internal/services/list_item/data_source.go
jeffa5: From the docs I think it should be `ItemID` https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/data-sources/list_item

jeffa5:
Also, the builds are failing due to an issue in another area of the codebase, is this a known problem with the `next` branch? 

> Error: internal/services/stream/list_data_source_model.go:71:10: params.VideoName undefined (type "github.com/cloudflare/cloudflare-go/v5/stream".StreamListParams has no field or method VideoName)

internal/services/list_item/data_source.go
vaishakdinesh: I would recommend to try defining the data source with the latest release to confirm if thats the case.

jeffa5: Is that not the case? That link seems to be the latest release. Also, this seems to follow the standard terraform data source pattern of having the id field (that is computed) be separate from how you specify it for the data source.

Happy to have a link to the latest release for checking against otherwise.

jeffa5:
@vaishakdinesh rebased.

vaishakdinesh (approved):

PR #5876, opened 2025-08-06
Fixes for managed transforms

orium:


internal/services/managed_transforms/data_source_schema.go
```diff
@@ -34,16 +33,6 @@ func DataSourceSchema(ctx context.Context) schema.Schema {
...
-						"has_conflict": schema.BoolAttribute{
-							Description: "Whether the Managed Transform conflicts with the currently-enabled Managed Transforms.",
-							Computed:    true,
-						},
-						"conflicts_with": schema.ListAttribute{
-							Description: "The Managed Transforms that this Managed Transform conflicts with.",
-							Computed:    true,
-							CustomType:  customfield.NewListType[types.String](ctx),
-							ElementType: types.StringType,
-						},
```
vaishakdinesh: Do these have to be removed?

internal/services/managed_transforms/model.go
```diff
@@ -30,13 +29,9 @@ func (m ManagedTransformsModel) MarshalJSONForUpdate(state ManagedTransformsMode
 type ManagedTransformsManagedRequestHeadersModel struct {
 	ID            types.String                   `tfsdk:"id" json:"id,required"`
 	Enabled       types.Bool                     `tfsdk:"enabled" json:"enabled,required"`
-	HasConflict   types.Bool                     `tfsdk:"has_conflict" json:"has_conflict,computed"`
-	ConflictsWith customfield.List[types.String] `tfsdk:"conflicts_with" json:"conflicts_with,computed"`
```
vaishakdinesh: Are these required to be removed? They look like computed fields.

orium: See other comment.

internal/services/managed_transforms/data_source_schema.go
orium: Yes: if they exist terraform will see differences in set elements and won't understand that how a planned value correlates to a state value (I think that's not an issue with lists because it correlates the items based on the order, but that's not true for sets).

These fields should not be relevant for terraform anyway: I think they are only useful for the www dashboard. I've marked them as "stainless ignore" in the schema, so stainless should remove them soon.

vaishakdinesh (approved):

PR #5884, opened 2025-08-07
Get cloudflare_snippet resource working

WowVeryLogin:
- Update schema to support multiple files and to include the name of each file.
- Fix multipart uploading of files by using the correct form field name, file name, and content type.
- Use the get content endpoint to fetch the snippet's files and metadata when refreshing state.
- Mark the created on attribute as UseStateForUnknown to prevent plan drift.
- Update apiform to not marshal computed attributes. This matches the behavior of apijson, and prevents the snippet's created on and modified on attributes being sent in update requests.


zakcutner (approved):

internal/services/snippets/model.go
musa-cf, line 1: Does this file need modifications? The `snippets` resource will not work anyway.

WowVeryLogin: Yes, missed that, fixed

PR #5886, opened 2025-08-07
wip: fix zero_trust_access_application and policy import

ajholland:


internal/services/zero_trust_access_application/schema.go
```diff
@@ -54,6 +55,8 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 			"allow_iframe": schema.BoolAttribute{
 				Description: "Enables loading application content in an iFrame.",
 				Optional:    true,
+				Computed:    true,
```
tamas-jozsa, line 58: This is something you could define in your OpenAPI schema, which would be the preferred approach to avoid code conflicts between the generated code and these custom codes.

Issue #5811, opened 2025-07-29
Reusable policies are imported as inline type for cloudflare_zero_trust_access_application


dmx-on:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Cloudflare Terraform provider: 5.7.1
OpenTofu: v1.10.2

### Affected resource(s)

When importing an existing `cloudflare_zero_trust_access_application` that uses reusable policies into Terraform, the `plan` and `apply` commands incorrectly attempt to add inline policy fields such as `decision`, `include`, and others.

However, for reusable policies, the policies block should only contain `id` and `precedence` fields. After manually removing the extra fields from the state file, Terraform correctly detects no changes.

Additionally, the Cloudflare REST API (`https://api.cloudflare.com/client/v4/accounts/:account_id/access/apps`) responds with these inline policy fields even for applications that are configured to use reusable policies. This may be contributing to the incorrect behavior during import and plan.

### Terraform configuration files

```hcl
resource "cloudflare_zero_trust_access_application" "application" {
  for_each = var.applications

  account_id                 = var.account_id
  name                       = coalesce(each.value.name, each.key)
  domain                     = each.value.domain
  type                       = each.value.type
  session_duration           = each.value.session_duration
  auto_redirect_to_identity  = each.value.auto_redirect_to_identity
  enable_binding_cookie      = each.value.enable_binding_cookie
  custom_deny_message        = each.value.custom_deny_message
  custom_deny_url            = each.value.custom_deny_url
  http_only_cookie_attribute = each.value.http_only_cookie_attribute
  same_site_cookie_attribute = each.value.same_site_cookie_attribute
  logo_url                   = each.value.logo_url
  skip_interstitial          = each.value.skip_interstitial
  app_launcher_visible       = each.value.app_launcher_visible
  service_auth_401_redirect  = each.value.service_auth_401_redirect
  custom_pages               = each.value.custom_pages
  tags                       = each.value.tags

  allowed_idps = []

  destinations = each.value.destinations
  policies     = each.value.policies

  cors_headers             = each.value.cors_headers
  options_preflight_bypass = each.value.options_preflight_bypass
}

variable "applications" {
[...]

PR #5889, opened 2025-08-08
zt access application: fix reusable policy importing 

ajholland:

## Changes being requested

API returns all field for reusable policies, instead of just id and precedence, in the terraform import we need to null those so they don't appear in diffs. 

## Additional context & links

https://github.com/cloudflare/terraform-provider-cloudflare/issues/5811


internal/services/zero_trust_access_group/normalizations.go
```diff
@@ -28,3 +29,21 @@ func normalizeReadZeroTrustAccessGroupAPIData(ctx context.Context, data, sourceD
 
 	return diags
 }
+
+func normalizeImportZeroTrustAccessGroupAPIData(ctx context.Context, data *ZeroTrustAccessGroupModel) diag.Diagnostics {
```
ajholland, line 33: Include/Require/Exclude are provided by the API as [], but we don't want to require users to specify empty list in their terraform config. This change is applied to Group/Policy/Application code

internal/services/zero_trust_access_application/plan_modifiers.go
```diff
@@ -136,5 +137,35 @@ func modifyPlan(ctx context.Context, req resource.ModifyPlanRequest, res *resour
...
+	// Handle tags order normalization - API returns alphabetically sorted tags
+	// but we want to preserve the user's configuration order
+	if stateApp != nil && !planApp.Tags.IsNull() && !stateApp.Tags.IsNull() && !planApp.Tags.IsUnknown() {
+		normalizeTagsOrder(ctx, &planApp.Tags, stateApp.Tags)
+	}
+
 	res.Plan.Set(ctx, &planApp)
 }
+
+func normalizeTagsOrder(ctx context.Context, planTags *customfield.List[types.String], stateTags customfield.List[types.String]) {
```
ajholland, line 149: tags are a set, but specified in the API as an array. treat like a set here

internal/services/zero_trust_access_application/normalizations.go
```diff
@@ -201,3 +212,44 @@ func normalizeWriteZeroTrustApplicationAPIData(ctx context.Context, data *ZeroTr
...
+		data.AllowedIdPs = nil
+	}
+	if data.Policies != nil && len(*data.Policies) == 0 {
+		data.Policies = nil
+	}
+
+	if data.Policies != nil {
+		for i := range *data.Policies {
+			policy := &(*data.Policies)[i]
+			if !policy.ID.IsNull() && !policy.ID.IsUnknown() {
```
ajholland, line 228: reusable policies include an ID and all other attributes, we only need to include name and ID to match what the user provided

internal/services/zero_trust_access_policy/testdata/accesspolicyauthmethodconfig.tf
```diff
@@ -14,7 +14,4 @@ resource "cloudflare_zero_trust_access_policy" "%[1]s" {
       auth_method = "hwk"
     }
   }]
-  approval_required              = "false"
```
ajholland, line 17: These bool fields were handled in this branch by [internal/services/zero_trust_access_policy/normalizations.go](https://github.com/cloudflare/terraform-provider-cloudflare/pull/5889/files#diff-89fa7ecb9d7bf3c86b705e3c840efcc125ccbb781b16690f9a164caf44a53edc)

We don't want users to be required to set the value to false. On import the value is set to null. If a user wants to explicitly set the value to false after importing, they will need to run an apply again, but that seems fine given the tradeoffs.

internal/services/zero_trust_access_application/plan_modifiers.go
GreenStage: why not just change the schema to a set

Issue #5898, opened 2025-08-08
(cloudflare_worker_script) Durable Object migrations not applied


1000hz:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

5.8.0 – 5.8.2

### Affected resource(s)

- cloudflare_workers_script

### Terraform configuration files

```hcl
resource "cloudflare_workers_script" "worker" {
  account_id = var.account_id
  script_name = "tf-durable-object-migrations"
  content = <<-EOT
    import {DurableObject} from "cloudflare:workers"
    export class MyDurableObject extends DurableObject {}
    export default { fetch() {return new Response()} } 
  EOT
  main_module = "worker.js"
  bindings = [{
    name = "MY_DO"
    class_name = "MyDurableObject"
    type = "durable_object_namespace"
  }]
  migrations = {
    new_tag = "v1"
    new_sqlite_classes = ["MyDurableObject"]
  }
}
```

### Link to debug output

-

### Expected output

Worker script is successfully created.

### Actual output

```
│ Error: failed to make http request
│
│ PUT "https://api.cloudflare.com/client/v4/accounts/redacted/workers/scripts/tf-durable-object-migrations": 400 Bad Request {
│   "result": null,
│   "success": false,
│   "errors": [
│     {
│       "code": 10061,
│       "message": "Cannot create binding for class 'MyDurableObject' because it is not currently configured to implement Durable Objects. Ensure a --new-class migration has been applied to it."
│     }
│   ],
│   "messages": []
│ }
│ 
```

### Steps to reproduce

Try to apply the above Terraform config.

### Additional factoids

To address #5701, we marked `cloudflare_workers_script.migrations` as `WriteOnly: true` in the provider.

However, Terraform omits `WriteOnly` attributes from both state _and plans_. Since the resource implementation sources its data from the plan rather than directly from config, migrations are being omitted from Create/Update requests.

Issue #5892, opened 2025-08-08
cloudflare_workers_script secret_text environment variable gets replaced on every deploy


MartinHignett:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

**Terraform version** 1.7.3

**Cloudflare provider version** 5.8.2

### Affected resource(s)

- cloudflare_workers_script

### Terraform configuration files

```hcl
resource "cloudflare_workers_script" "files-worker" {
  account_id  = data.terraform_remote_state.cloudflare_tf.outputs.account_id
  script_name = "${terraform.workspace}-files"
  content     = file("dist/files.js")
  main_module = "worker.js"
  usage_model = "standard"

  bindings = [    
    {
      name = "PRIVATE_KEY"
      type = "secret_text"
      text = "PLACEHOLDER - REPLACE WITH PRIVATE KEY AFTER DEPLOYING"
    },
    {
      name = "DEBUG"
      type = "plain_text"
      text = "false"
    },
    {
      name         = "KEY_VALUE_STORE"
      type         = "kv_namespace"
      namespace_id = data.terraform_remote_state.cloudflare_tf.outputs.kv_namespace.id
    }
  ]
}
```

### Link to debug output

N/A

### Panic output

None

### Expected output

When I run the plan a second time, the secret_text value is unchanged (i.e. the private key I pasted in is still there).

### Actual output

The secret_text value is replaced with my placeholder value and the worker fails.

### Steps to reproduce

- Create a `cloudflare_workers_script` resource with a secret_text environment variable. A simple worker that prints the value of the secret to the logs will be needed to test the value
- Apply the plan
- Change the value in the secret
- Run the worker, you should see the new value
- Apply the plan again
- Run the worker
- Secret value has changed back to the placeholder

1000hz:
> * Change the value in the secret

To be clear, are you changing this value outside the scope of Terraform, i.e. from the Cloudflare dashboard?

MartinHignett:
Yes - for understandable reasons, I don't want to check my private key into git, so I am deploying a placeholder and putting the actual key via the console.

MartinHignett:
I've tried adding this

```
lifecycle {
    ignore_changes = [
[...]

PR #5899, opened 2025-08-08
Updates to `cloudflare_workers_script`

1000hz:

## Changes being requested
This PR includes the following updates and fixes for the `cloudflare_workers_script` resource:

> **[fix(workers_script): Revert treating cloudflare_workers_script.bindings as a Set](https://github.com/cloudflare/terraform-provider-cloudflare/commit/17c8defb79d71e316b043d6e8b23ac7e0797a554)**
>
> This was originally intended to prevent diffs due to ordering inconsistency between API and Terraform config, but resulted in a loss of fidelity in diffs as the entire `bindings` attribute was now being shown in diffs as `(sensitive)` due to a nested attribute being as such.

> **[chore(workers_script): use resourcevalidator.ExactlyOneOf() to ensure `content` or `content_file` is provided](https://github.com/cloudflare/terraform-provider-cloudflare/commit/9b445632016eb7cac8c36304a37a0b1acb97d8a4)**

> **[fix(workers_script): ignore unmanaged secret_text bindings](https://github.com/cloudflare/terraform-provider-cloudflare/commit/7c80a3d91bc7c36f0b1f5d916af1a21944abf8ff)**
>
> Fixes https://github.com/cloudflare/terraform-provider-cloudflare/issues/5892

> **[fix(workers_script): Obtain migrations directly from config instead of plan](https://github.com/cloudflare/terraform-provider-cloudflare/commit/d48f08cea6f0a843b77e47cc318df3d3f8773085)**
>
> Fixes https://github.com/cloudflare/terraform-provider-cloudflare/issues/5898

### DIFF
--- a/internal/services/workers_script/model.go
+++ b/internal/services/workers_script/model.go
@@ -89,18 +89,18 @@ func (r WorkersScriptModel) MarshalMultipart() (data []byte, formDataContentType
 }
 
 type WorkersScriptMetadataModel struct {
-	Assets             *WorkersScriptMetadataAssetsModel                               `tfsdk:"assets" json:"assets,optional"`
-	Bindings           customfield.NestedObjectSet[WorkersScriptMetadataBindingsModel] `tfsdk:"bindings" json:"bindings,computed_optional"`
-	BodyPart           types.String                                                    `tfsdk:"body_part" json:"body_part,optional"`
-	CompatibilityDate  types.String                                                    `tfsdk:"compatibility_date" json:"compatibility_date,computed_optional"`
-	CompatibilityFlags customfield.Set[types.String]                                   `tfsdk:"compatibility_flags" json:"compatibility_flags,computed_optional"`
-	KeepAssets         types.Bool                                                      `tfsdk:"keep_assets" json:"keep_assets,optional"`
-	KeepBindings       *[]types.String                                                 `tfsdk:"keep_bindings" json:"keep_bindings,optional"`
-	Logpush            types.Bool                                                      `tfsdk:"logpush" json:"logpush,computed_optional"`
-	MainModule         types.String                                                    `tfsdk:"main_module" json:"main_module,optional"`
-	Migrations         customfield.NestedObject[WorkersScriptMetadataMigrationsModel]  `tfsdk:"migrations" json:"migrations,optional"`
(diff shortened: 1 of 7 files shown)


internal/services/workers_script/model.go
```diff
@@ -89,18 +89,18 @@ func (r WorkersScriptModel) MarshalMultipart() (data []byte, formDataContentType
...
-	KeepBindings       *[]types.String                                                 `tfsdk:"keep_bindings" json:"keep_bindings,optional"`
-	Logpush            types.Bool                                                      `tfsdk:"logpush" json:"logpush,computed_optional"`
-	MainModule         types.String                                                    `tfsdk:"main_module" json:"main_module,optional"`
-	Migrations         customfield.NestedObject[WorkersScriptMetadataMigrationsModel]  `tfsdk:"migrations" json:"migrations,optional"`
-	Observability      *WorkersScriptMetadataObservabilityModel                        `tfsdk:"observability" json:"observability,optional"`
-	Placement          customfield.NestedObject[WorkersScriptMetadataPlacementModel]   `tfsdk:"placement" json:"placement,computed_optional"`
+	Assets             *WorkersScriptMetadataAssetsModel                                `tfsdk:"assets" json:"assets,optional"`
+	Bindings           customfield.NestedObjectList[WorkersScriptMetadataBindingsModel] `tfsdk:"bindings" json:"bindings,computed_optional"`
```
vaishakdinesh, line 93: I think with a list, the order might matter. So just confirming thats intended? @1000hz

1000hz: Yeah, we switched to using a Set in the last minor release to avoid diffs resulting solely from the API returning bindings in a different order than they were declared in config. Unfortunately, this also resulted in Terraform losing the ability to track differences on an element-by-element basis. To make matters worse, because `bindings.*.text` is marked as a sensitive attribute in the schema, the entire set of bindings was being obscured in plan diffs as `bindings = (sensitive)`.

I plan to come up with a better solution in a future release to avoid element order being significant but without losing diff fidelity.

vaishakdinesh: Ah gotcha. Thank you for the explanation.

vaishakdinesh (approved):

PR #5901, opened 2025-08-08
feat(apijson): add `decode_null_to_zero` tag option

zakcutner:
Add support for a new `decode_null_to_zero` tag option in JSON struct tags to decode null and omitted JSON values as zero values instead of leaving them unset (which usually results in them being set to null).

This is particularly useful for resources where the API does not include certain optional attributes in responses when they are the default value. For example, when the API uses Go's `omitempty` or `omitzero` with a non-pointer type.

In this case, the attribute cannot simply be marked as `Optional`, because Terraform will complain that the API does not include the attribute in responses when it is explicitly set to the default value in a configuration.

Marking the attribute as `Computed` and `Optional` with a `Default` also does not work without adding the new `decode_null_to_zero` tag, because Terraform expects the API to always include computed attributes in responses, which it will not for the default value.

Using the new `decode_null_to_zero` tag solves this issue, since from Terraform's perspective, the API now always returns a value for the attribute, even if it is just the default.

### DIFF
--- a/internal/apijson/decoder.go
+++ b/internal/apijson/decoder.go
@@ -66,6 +66,7 @@ type decoderBuilder struct {
 	// Whether or not this is the first element and called by [UnmarshalRoot], see
 	// the documentation there to see why this is necessary.
 	root bool
+
 	// The dateFormat (a format string for [time.Format]) which is chosen by the
 	// last struct tag that was seen.
 	dateFormat string
@@ -76,6 +77,11 @@ type decoderBuilder struct {
 	// This is used to control decoding behavior for computed and computed_optional
 	// fields.
 	updateBehavior TerraformUpdateBehavior
+
+	// decodeZeroValueWhenNull indicates whether null and omitted values should
+	// be decoded as the zero value of the field type instead of leaving the
+	// field unset.
+	decodeZeroValueWhenNull bool
 }
 
 // decoderState contains the 'run-time' state of the decoder.
@@ -112,10 +118,11 @@ type decoderField struct {
 
 type decoderEntry struct {
 	reflect.Type
-	dateFormat            string
-	root                  bool
-	unmarshalComputedOnly bool
-	tfSkipBehavior        TerraformUpdateBehavior
+	dateFormat              string
+	root                    bool
+	unmarshalComputedOnly   bool
+	tfSkipBehavior          TerraformUpdateBehavior
+	decodeZeroValueWhenNull bool
 }
 
 func (d *decoderBuilder) unmarshal(raw []byte, to any) error {
@@ -129,11 +136,12 @@ func (d *decoderBuilder) unmarshal(raw []byte, to any) error {
 
 func (d *decoderBuilder) typeDecoder(t reflect.Type) decoderFunc {
 	entry := decoderEntry{
-		Type:                  t,
-		dateFormat:            d.dateFormat,
-		root:                  d.root,
-		unmarshalComputedOnly: d.unmarshalComputedOnly,
-		tfSkipBehavior:        d.updateBehavior,
+		Type:                    t,
+		dateFormat:              d.dateFormat,
+		root:                    d.root,
+		unmarshalComputedOnly:   d.unmarshalComputedOnly,
+		tfSkipBehavior:          d.updateBehavior,
+		decodeZeroValueWhenNull: d.decodeZeroValueWhenNull,
 	}
 
 	if fi, ok := decoders.Load(entry); ok {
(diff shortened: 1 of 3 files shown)


ms-jpq:
hi @zakcutner, taking a look now, but before we dive too deeply would you mind outlining oh which resources / data sources attributes would this be used for? since that this PR currently does not include any model changes where the annotation might be used.

I am a bit hesitant to merge / upstream this because terraform semantics != golang semantics, and often the services that the Terraform provider is targeting in fact are not written in a language that have the `empty is zero value` semantics either.

zakcutner:
Thank you for reviewing this @ms-jpq! This new option would currently only be for the `cloudflare_ruleset` resource and data source. However, I believe several other Cloudflare APIs use `omitempty`/`omitzero` quite heavily for default values, so it is likely to be useful to fix bugs in other resources and data sources too.

The idea though is that this new option is completely opt-in, and no change of behavior should occur when not using it. Potentially in the future, it could be useful to have a custom OpenAPI parameter (e.g., `x-stainless-decode-null-to-zero`) that indicates a particular attribute should be marked as `decode_null_to_zero` during translation, but I agree that this should never be added automatically.

The point of this change though is to provide the possibility for certain attributes to be decoded to the zero value when they are omitted from responses. Without this change, I am not aware of another (simple) way to make the Terraform provider work with APIs that have this behavior.

If it helps to review, I opened a draft PR with one example of where this is useful for the `cloudflare_ruleset` resource (but there are many other cases for `cloudflare_ruleset` alone): https://github.com/cloudflare/terraform-provider-cloudflare/pull/5902/commits/22ce5738380bf879b2e860bc4a1043490d5a6cc9.

ms-jpq:
hi @zakcutner thank you so much for the additional context. may I ask how urgently do you want this change up-streamed?

My thinking is that in so far as possible, I would prefer to isolate complexity away from the encoder / decoder stack, since not only do we need to keep the encoders in sync for `apijson` and `apiform`, their correctness is actually fairly tricky to reason through, since there is non-trivial tribal knowledge of the Terraform provider framework involved.

My goal is to iterate a little on alternative approaches to achieving the same outcome, but perhaps at a schema level, instead of additional model annotations. This is not to say that I am not sensitive to the need to have more encoder / decoder capabilities in general. I just want to be able to do a bit more due-diligence first 😄

---

my estimate for the up-streaming turn around is ~2-3 days, given ^^^

zakcutner:
> may I ask how urgently do you want this change up-streamed?

I think this is quite urgent unfortunately, because it blocks several bug fixes to the `cloudflare_ruleset` resource and data source.

> My thinking is that in so far as possible, I would prefer to isolate complexity away from the encoder / decoder stack, since not only do we need to keep the encoders in sync for apijson and apiform, their correctness is actually fairly tricky to reason through, since there is non-trivial tribal knowledge of the Terraform provider framework involved.

You make a fair point and I realize that the `apijson` package is already pretty complex 😅 However, I do think this particular piece of functionality really does belong in `apijson`, since it specifically relates to how decoding should be performed.

I also confirmed that the existing tests in `apijson` (which are quite extensive) continue to pass after this change, and copied every relevant test to use the new `decode_null_to_zero` option to confirm that it behaves as expected. While the test suite is not exhaustive, I was hoping this might give us some more confidence of the change. Also let me know if you have other ideas for testing!

> My goal is to iterate a little on alternative approaches to achieving the same outcome, but perhaps at a schema level, instead of additional model annotations. This is not to say that I am not sensitive to the need to have more encoder / decoder capabilities in general. I just want to be able to do a bit more due-diligence first 😄

I think I would really prefer to avoid doing something at the schema level. I did try some similar ideas before making this change, e.g., creating a custom type where null is equal to the default value. But it did not feel very clean, and I realized I'd probably need to make changes to `apijson` anyway FWIW, since it does not handle custom types so well currently.

More broadly, as I mentioned above I think we do ultimately want to have this functionality in `apijson`. And I'd like to both avoid complexity in the `cloudflare_ruleset` resource during an intermediate state, and the need to refactor our resource later, given the urgency of some of these bug fixes and time constraints on our side.

As an alternative idea, I wonder if in the very short term we could duplicate the `apijson` package with these changes (e.g., into a new `apijsonnext` package) that we would use only for `cloudflare_ruleset` for now? And that would give you more time to review the change before it is used for other resources and data sources.

ms-jpq:
Thanks @zakcutner, I will try to get the fixes to `cloudflare_ruleset` up streamed as soon as bandwidth allows, the goal is before EOD tomorrow.

musa-cf (approved):

zakcutner:
Thank you again for looking at this @ms-jpq! After discussing with @musa-cf, we've moved these changes into a separate `apijsoncustom` package for now, while you review them.

ms-jpq:
Thanks @zakcutner!

ms-jpq:
hi @zakcutner,

We have discussed this internally about what to do when there is more changes to the more sensitive parts of the Terraform provider stack, i.e. serialization / deserializaiton than we can realistically have the bandwidth to ingest.

and we think that the approach to fork `apijson` for now is correct and we will asynchronously pull in the changes into the upstream as you have proposed.

The priority we have is to be minimally blocking for your bandwidth.

cc @musa-cf 

Thanks all!

PR #5903, opened 2025-08-11
fix(list_item): Use proper pagination from client

jeffa5:

## Changes being requested

Updating the go-sdk so that list items resource handling can use the proper pagination client, rather than only using the first page of results.

This also means that we can use the default page size as we expect the item to be near the top but pagination will continue to get it for us anyway.

## Additional context & links

As we are now using the list client to obtain go types for the list item search result we have to implement more of the equality check with the local state model. This isn't ideal but I'm not sure of how else to do it (open to suggestions).


go.mod
```diff
@@ -9,8 +9,8 @@ require (
 	github.com/aws/aws-sdk-go-v2/config v1.27.36
 	github.com/aws/aws-sdk-go-v2/credentials v1.17.34
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.63.0
-	github.com/cloudflare/cloudflare-go v0.104.0
-	github.com/cloudflare/cloudflare-go/v5 v5.0.0
+	github.com/cloudflare/cloudflare-go v0.115.0
+	github.com/cloudflare/cloudflare-go/v5 v5.0.1-0.20250811144937-fd8fec5656c3
```
vaishakdinesh, line 13: I think this is fine since we haven't released the new go sdk yet. and when we do release it this would get overwritten.
@mgirouard @musa-cf @tamas-jozsa thoughts?

vaishakdinesh:
@jeffa5 were you able to run the test locally? Are they passing with this change? Are they comprehensive enough or do you think we need more coverage?

go.mod
jeffa5: Yeah, needed this to drag in the new pagination

jeffa5:
>  were you able to run the test locally? Are they passing with this change? Are they comprehensive enough or do you think we need more coverage?

I did run the tests locally, and still pass with this. I think they enough coverage for now.

go.mod
vaishakdinesh: Yeah, I got that. Unless we do a release, it'd have to be like this or commit something that can't be resolved.

musa-cf (approved):

PR #5910, opened 2025-08-12
feat(migrate): enhance migration tooling

dackerman:
- Add state transformation support for access_application, access_policy, load_balancer, load_balancer_pool, and tiered_cache resources
   - note that these are not fully validated yet (there's unit tests, but we aren't merging the state upgrades and acceptance tests yet into next, so that's in another branch)
- Enhance MigrationTestStep to handle version-specific migrations and state checks
- Add ExpectEmptyPlanExceptFalseyToNull plan checker for v4->v5 migrations with removed defaults
- Add comprehensive test helpers for migration testing
- Add documentation for HCL transformations and testing approach

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,19 @@
+{
+  "tools": {
+    "write": {
+      "blocked_paths": [
+        "cmd/migrate/ast/wrappers.go"
+      ]
+    },
+    "edit": {
+      "blocked_paths": [
+        "cmd/migrate/ast/wrappers.go"
+      ]
+    },
+    "multiedit": {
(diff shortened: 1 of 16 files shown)


vaishakdinesh:
I think we can get this in after the next release. Thoughts @dackerman @musa-cf @mgirouard @tamas-jozsa

tamas-jozsa (approved):

dackerman:
@vaishakdinesh yeah i think it could go in whenever since users aren't actually calling it yet. ideally the sooner the better because we have several migration tests that all depend on it, but i'll defer to you folks!

PR #5958, opened 2025-08-18
fix(ruleset): update data sources and documentation

zakcutner:
* Make data sources consistent with the schema of the resource.

* Deprecate `ruleset_id` in favor of `id` for `cloudflare_ruleset` data source.

* Deprecate `result` in favor of `rulesets` for `cloudflare_rulesets` data source.

* Update documentation to reflect latest changes.

* Improve validation on `account_id` and `zone_id` attributes.

### DIFF
--- a/docs/data-sources/ruleset.md
+++ b/docs/data-sources/ruleset.md
@@ -13,9 +13,8 @@ description: |-
 
 ```terraform
 data "cloudflare_ruleset" "example_ruleset" {
-  ruleset_id = "2f2feab2026849078ba485f918791bdc"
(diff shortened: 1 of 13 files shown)


musa-cf:
Hi @zakcutner! We use tfplugindocs to generate these docs from the `template` directory: https://github.com/cloudflare/terraform-provider-cloudflare/tree/next/templates

We have a few other resources that have custom templating for their docs that you can look at. Once you add your templates, run `scripts/generate-docs` to generate the files in `docs`. The script wraps the `tfplugindocs` tool for convenience.

PR #5970, opened 2025-08-19
fix(cloudflare_list_item) conditionally require exclude_exact_hostname

broswen:

## Changes being requested

`exclude_exact_hostname` is only allowed to be set for hostname list items that are wildcards (start with *.).
If they are omitted when the hostname is a wildcard, then the API return the value set to true, which causes Terraform drift.

We want to conditionally require the user to set `exclude_exact_hostname` if the hostname is a wildcard hostname, and reject if `exclude_exact_hostname` is set when the hostname is not a wildcard hostname.


vaishakdinesh (approved):

vaishakdinesh:
@broswen closing this. Lets talk if this is still required and we can get it done.

PR #5973, opened 2025-08-20
helpers for translating dynamic blocks

ms-jpq:
translate test

sort keys


cmd/migrate/errs/errs.go
```diff
@@ -1,12 +0,0 @@
-package errs
```
ms-jpq, line 1: not used at all, meant to remove eariler

cmd/migrate/list_item_example_test.go
```diff
@@ -0,0 +1,88 @@
...
+  account_id  = local.account_id
+  description = "The set of IPs that internal OpenAI requests may come from. Used for exemption from firewall and bot detection rules. Do not use for general access control!"
+  kind        = "ip"
+  name        = "known_openai_ips"
+}
+`, `
+resource "cloudflare_list_item" "known_openai_ips" {
+  account_id = cloudflare_list_item.known_openai_ips.account_id
+  comment    = each.value.content.comment
+  for_each   = [for ip_entry in [for value in local.ips_data.ips : { key = value, value = value }] : { content = { comment = ip_entry.value.comment, value = { ip = ip_entry.value.ip } } }]
```
ms-jpq, line 40: the double for comprehension translation is ugly, but it handles almost arbitrary cases, since it does not require too much AST rewrite of the dynamic block's bodies

cmd/migrate/list_item_example_test.go
```diff
@@ -0,0 +1,114 @@
...
+  account_id  = local.account_id
+  description = "dogs are great"
+  kind        = "ip"
+  name        = "persona_webhook_ips"
+}
+`, `
+resource "cloudflare_list_item" "persona_webhook_ips" {
+  account_id = cloudflare_list_item.persona_webhook_ips.account_id
+  comment    = each.value.content.comment
+  for_each = [
```
ms-jpq, line 93: I fixed the ugly!

cmd/migrate/ast/utils.go
```diff
@@ -11,20 +11,44 @@ import (
 	"github.com/hashicorp/hcl/v2/hclwrite"
 )
 
-// ExprTransformer mutates an attribute value in place
-// or sets to nil if the attribute should be removed
-type ExprTransformer func(*hclsyntax.Expression, Diagnostics)
+func OrderedAttributes(attri hclsyntax.Attributes) []*hclsyntax.Attribute {
```
ms-jpq, line 14: Note: there is an existing flaky test due to this function not being used where its supposed to. but I dont want to fix it in this PR, since that would result in a noisy diff.

its a simple change, just make sure we use this function and update the expected examples to always use the ordering here

PR #5980, opened 2025-08-20
fix(zone_setting): remove grit patterns for `cloudflare_zone_settings…

mgirouard:
…_override`

Removes the grit patterns that convert block --> attribute syntax. This never produced a valid result.

Zone settings also present a unique challenge for v4 -> v5 migrations: it's a one-to-many type of situation that Grit doesn't seem to handle well. The v4 provider had a `cloudflare_zone_settings_override` resource which handled multiple zone settings under the hood. In v5, zone settings are defined individually with a `cloudflare_zone_setting` resource that handles a specific setting on the zone.

For v4 -> v5 migrations, we'll handle cases like this by converting the HCL + import blocks with `cmd/migrate` instead of Grit.

## Changes being requested
- updates grit patterns to remove all `cloudflare_zone_settings_override`

## Additional context & links
n/a


mgirouard:
Test failures are caused by `next` not pointing to `clouflare-go`'s `next` and are unrelated to this. Going to merge.

PR #5981, opened 2025-08-20
feat(regional_hostname): support migration from v4 to v5, add tests

nickysemenza:
v4 had a timeouts property (https://registry.terraform.io/providers/cloudflare/cloudflare/4.52.1/docs/resources/regional_hostname#timeouts-1) which was never sent to the API, so it can/should be dropped while migrating to v5.

This change also adds some additional tests


internal/services/regional_hostname/migrations.go
```diff
@@ -6,10 +6,83 @@ import (
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework-timetypes/timetypes"
 )
 
 var _ resource.ResourceWithUpgradeState = (*RegionalHostnameResource)(nil)
 
 func (r *RegionalHostnameResource) UpgradeState(ctx context.Context) map[int64]resource.StateUpgrader {
-	return map[int64]resource.StateUpgrader{}
+	return map[int64]resource.StateUpgrader{
```
mgirouard, line 17: Is it possible to have tests for this?
Non-blocking -- can be in another PR if its easier

nickysemenza: yeah, although not sure what coverage is missing from `TestAccCloudflareRegionalHostname_Migration_TimeoutsRemoval`  that should smoke test this right?

nickysemenza: moved the test for this into a migrations_test.go in 7807eda

PR #5995, opened 2025-08-21
feat(cloudflare_list): add nested set list items to resource and data source

broswen:

## Changes being requested

Adds a nested attribute object set of list items. This results in a single bulk operation request that replaces all items in the list.

This avoids the issue that each singular `cloudflare_list_item` creates a bulk operation with a single item for each resource causing users to easily hit API rate limits.

example:
```terraform
resource "cloudflare_list" "example_list" {
  account_id = "023e105f4ecef8ad9ca31a8372d0c353"
  kind = "ip"
  name = "list1"
  description = "This is a note"
  items = [
    {
      ip = "1.1.1.1"
    },
    {
      ip = "1.1.1.2"
    }
    {
      ip = "1.1.1.3"
    }
  ]
}
```

## Additional context & links

Since this replaces all existing items in a list, it is incompatible with the existing singular `cloudflare_list_item` and will cause drift when used together.

Currently `items` is set as `optional` and best effort tries to convert an empty set to a null set to preserve backwards compatibility until users move off of individual `cloudflare_list_item`s.

### DIFF
--- a/.grit/patterns/cloudflare_terraform_v5_block_to_attribute_configuration.grit
+++ b/.grit/patterns/cloudflare_terraform_v5_block_to_attribute_configuration.grit
@@ -158,7 +158,6 @@ pattern cloudflare_terraform_v5_block_to_attribute_configuration() {
     inline_cloudflare_block_to_map(`dns`) as $block where { $block <: within `resource "cloudflare_spectrum_application" $_ { $_ }` },
     inline_cloudflare_block_to_map(`edge_ips`) as $block where { $block <: within `resource "cloudflare_spectrum_application" $_ { $_ }` },
     inline_cloudflare_block_to_map(`origin_dns`) as $block where { $block <: within `resource "cloudflare_spectrum_application" $_ { $_ }` },
-    inline_cloudflare_block_to_list(`origin_port_range`) as $block where { $block <: within `resource "cloudflare_spectrum_application" $_ { $_ }` },
     inline_cloudflare_block_to_list(`tunnels`) as $block where { $block <: within `resource "cloudflare_split_tunnel" $_ { $_ }` },
     inline_cloudflare_block_to_list(`antivirus`) as $block where { $block <: within `resource "cloudflare_teams_account" $_ { $_ }` },
     inline_cloudflare_block_to_list(`notification_settings`) as $block where { $block <: within `resource "cloudflare_teams_account" $_ { $_ }` },
--- a/.stats.yml
+++ b/.stats.yml
@@ -1,4 +1,4 @@
 configured_endpoints: 1809
 openapi_spec_url: https://storage.googleapis.com/stainless-sdk-openapi-specs/cloudflare%2Fcloudflare-77d61495fecd0d26b9adff1af0ab3510b06a3cc2c6781b9a40aabcad2f10588a.yml
 openapi_spec_hash: 95dee3be411dda77306a41dc7d49eb35
-config_hash: 73b82490ed501dcc81bc0ddf7dade975
+config_hash: db0ff227af424d15b1c84f467e04c023
--- a/cmd/migrate/ast/wrappers.go
+++ b/cmd/migrate/ast/wrappers.go
@@ -44,6 +44,11 @@ func (b Block) Attributes(diags Diagnostics) map[string]*hclsyntax.Expression {
 }
 
 func (b Block) SetAttribute(key string, val hclsyntax.Expression, diags Diagnostics) {
+	if val == nil {
+		// If expression is nil, remove the attribute instead of setting it
+		b.RemoveAttribute(key, diags)
+		return
+	}
 	e := Expr2WriteExpr(val, diags)
 	eTokens := e.BuildTokens(nil)
 	b.Body().SetAttributeRaw(key, eTokens)
@@ -58,6 +63,11 @@ func (o Object) Attributes(diags Diagnostics) map[string]*hclsyntax.Expression {
 }
 
 func (o Object) SetAttribute(key string, val hclsyntax.Expression, diags Diagnostics) {
+	if val == nil {
+		// If expression is nil, remove the attribute instead of setting it
+		o.RemoveAttribute(key, diags)
+		return
+	}
 	if o.attrs[key] == nil {
 		// object doesn't have this attribute yet, create it & add to map
 		attr := hclsyntax.ObjectConsItem{
--- a/cmd/migrate/dns_record.go
+++ b/cmd/migrate/dns_record.go
@@ -17,7 +17,7 @@ func isDNSRecordResource(block *hclwrite.Block) bool {
 		(block.Labels()[0] == "cloudflare_dns_record" || block.Labels()[0] == "cloudflare_record")
 }
 
-// ProcessDNSRecordConfig processes cloudflare_dns_record configurations to fix CAA record issues
+// ProcessDNSRecordConfig processes cloudflare_dns_record configurations to fix CAA record issues and ensure TTL is present
 func ProcessDNSRecordConfig(file *hclwrite.File) error {
 	body := file.Body()
 
@@ -34,6 +34,18 @@ func ProcessDNSRecordConfig(file *hclwrite.File) error {
 		resourceType := labels[0]
 		if resourceType != "cloudflare_dns_record" && resourceType != "cloudflare_record" {
 			continue
+		}
+
+		// Ensure TTL is present for v5 (required field)
+		ttlAttr := block.Body().GetAttribute("ttl")
+		if ttlAttr == nil {
+			// TTL is missing - add it with default value of 1 (automatic)
+			// Create the TTL token with value 1
+			ttlToken := &hclwrite.Token{
+				Type:  hclsyntax.TokenNumberLit,
+				Bytes: []byte("1"),
+			}
+			block.Body().SetAttributeRaw("ttl", hclwrite.Tokens{ttlToken})
 		}
 
 		// Check if this is a CAA record
@@ -170,6 +182,13 @@ func transformDNSRecordStateJSON(result string, path string, instance gjson.Resu
 	} else if value.Exists() && content.Exists() {
 		// If both exist, remove value since content takes precedence in v5
 		result, _ = sjson.Delete(result, path+".attributes.value")
+	}
+
+	// Ensure TTL is present (required in v5)
+	ttl := instance.Get("attributes.ttl")
+	if !ttl.Exists() {
+		// TTL is missing - add default value of 1 (automatic)
+		result, _ = sjson.Set(result, path+".attributes.ttl", 1.0)
 	}
 
 	// Rename metadata -> meta and ensure it's a JSON string
--- a/cmd/migrate/dns_record_test.go
+++ b/cmd/migrate/dns_record_test.go
@@ -140,7 +140,46 @@ resource "cloudflare_record" "caa_legacy" {
 }`},
 		},
 		{
-			Name: "Multiple CAA records in same file - content renamed to value",
+			Name: "DNS record without TTL - should add TTL with default value",
+			Config: `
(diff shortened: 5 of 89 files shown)


docs/data-sources/list.md
mgirouard, line 1: Wasn't sure if you generated these or not, but just a heads up that the docs will get regenerated when we do the release later this week. Any custom outside the schemas will be lost.

broswen: yea I generated these with `./scripts/generate-docs` :+1:

internal/services/list/model.go
```diff
@@ -30,3 +32,30 @@ func (m ListModel) MarshalJSON() (data []byte, err error) {
 func (m ListModel) MarshalJSONForUpdate(state ListModel) (data []byte, err error) {
 	return apijson.MarshalForUpdate(m, state)
 }
+
+type ListItemModel struct {
```
vaishakdinesh, line 36: would this need `ID`?

internal/services/list/data_source_schema.go
```diff
@@ -64,6 +65,74 @@ func DataSourceSchema(ctx context.Context) schema.Schema {
 				Description: "The number of [filters](/api/resources/filters/) referencing the list.",
 				Computed:    true,
 			},
+			"search": schema.StringAttribute{
+				Description: " A search query to filter returned items. Its meaning depends on the list type: IP addresses must start with the provided string, hostnames and bulk redirects must contain the string, and ASNs must match the string exactly.",
+				Optional:    true,
+			},
+			"items": schema.SetNestedAttribute{
+				Description: "The items in the list. If set, this overwrites all items in the list. Do not use with `cloudflare_list_item`.",
```
vaishakdinesh, line 73: This feels like a potential footgun...If we really want to enforce this then we might be better off returning an  error for the list item resource and providing users a way to merge their list items into this resource.
@musa-cf @mgirouard thoughts?

internal/services/list/schema.go
```diff
@@ -66,6 +73,102 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				Description: "The number of [filters](/api/resources/filters/) referencing the list.",
 				Computed:    true,
 			},
+			"items": schema.SetNestedAttribute{
```
vaishakdinesh, line 76: what happens if this is an empty array?

internal/services/list/testdata/listwithipitems_datasource.tf
```diff
@@ -0,0 +1,20 @@
+resource "cloudflare_list" "%[1]s" {
+  account_id = "%[4]s"
+  name = "%[2]s"
+  description = "%[3]s"
+  kind = "ip"
+  items = [
+    {
+      ip = "1.1.1.1"
```
vaishakdinesh, line 8: how does the resource behave when you have another entry with the same ip? If recall, the API merges them?

internal/services/list_item/resource.go
```diff
@@ -105,7 +106,7 @@ func (r *ListItemResource) Create(ctx context.Context, req resource.CreateReques
 		return
 	}
 
-	err = pollBulkOperation(ctx, data.AccountID.ValueString(), createEnv.Result.OperationID.ValueString(), r.client)
+	err = list.PollListBulkOperation(ctx, data.AccountID.ValueString(), createEnv.Result.OperationID.ValueString(), r.client)
```
vaishakdinesh, line 109: Would it be possible here in list item to collect all the list items and make the bulk update rather than adding items as a part of list?

vaishakdinesh:
How would import work? Would we not need to update that? We'd need to import the whole into state right?

internal/services/list_item/resource.go
broswen: I'm not sure I understand, can we see the state of other `cloudflare_list_item`s outside of the single resource that is being creating?

We would still need to submit one bulk update per list item.
If you have to manage 1000 list items, then it's still 1000 bulk updates.

internal/services/list/schema.go
broswen: An empty set deletes all list items. A null set doesn't do anything.
I did this to preserve compatibility while users could still be using `cloudflare_list_item`

internal/services/list/model.go
broswen: It doesn't _need_ it, I didn't see the benefit as we won't be managed each item by ID anymore.
I can add it quick if we think it's beneficial.

internal/services/list/data_source_schema.go
broswen: I tried to keep this backwards compatible so users could still use `cloudflare_list_item` while this update goes out. For example, not setting `items` (null) means it will not clear out the list items and delete all individually managed ones.

But if we're ok with a breaking change then I can make `items` required and simplify things a bit.

internal/services/list/testdata/listwithipitems_datasource.tf
broswen: Yes the API merges duplicate entries. When it does that depends on the list item type.

broswen:
> How would import work? Would we not need to update that? We'd need to import the whole into state right?

I did update the `ImportState` method. It paginates through all list items.
https://github.com/cloudflare/terraform-provider-cloudflare/pull/5995/files#diff-5c6ab0396667735a7a7f506c6b0a41849868ccaa4fc85ec31976a6e662cba003R312

internal/services/list/testdata/listwithipitems_datasource.tf
broswen: One thing I just checked, since this is a nested set duplicate identical objects don't show up as different 
```
resource "cloudflare_list" "example" {
  account_id  = local.account_id
  name        = "example"
  description = "example"
  kind        = "ip"
  items = [
    {
      ip = "1.1.1.1"
    }
  ]
}
```
And if you add an identical item, it shows no changes required to remote state.
```
resource "cloudflare_list" "example" {
  account_id  = local.account_id
  name        = "example"
  description = "example"
  kind        = "ip"
  items = [
    {
      ip = "1.1.1.1"
    },
    {
      ip = "1.1.1.1"
    }
  ]
}
```
If you submit a bulk job with duplicate items that have the same "key" that differ in their other properties, the API returns an error.

```
{
│       "code": 10028,
│       "message": "invalid value for redirect at position 1: filters.api.duplicate_item_value",
│       "source": {
│         "parameter_value_index": 1
│       }
│     }
```

internal/services/list/model.go
vaishakdinesh: No, i think adding it would be more problematic than not. As long as the API is happy :)

internal/services/list_item/resource.go
vaishakdinesh: ah never mind, I was thinking of something that probably wasn't possible

vaishakdinesh:
> > How would import work? Would we not need to update that? We'd need to import the whole into state right?
> 
> I did update the `ImportState` method. It paginates through all list items. https://github.com/cloudflare/terraform-provider-cloudflare/pull/5995/files#diff-5c6ab0396667735a7a7f506c6b0a41849868ccaa4fc85ec31976a6e662cba003R312

Gotcha. And it brings everything related into the state yeah? Do you have an example of how the state would look?

broswen:
> > > How would import work? Would we not need to update that? We'd need to import the whole into state right?
> > 
> > 
> > I did update the `ImportState` method. It paginates through all list items. https://github.com/cloudflare/terraform-provider-cloudflare/pull/5995/files#diff-5c6ab0396667735a7a7f506c6b0a41849868ccaa4fc85ec31976a6e662cba003R312
> 
> Gotcha. And it brings everything related into the state yeah? Do you have an example of how the state would look?

Here's the `resources` from a project that has an example hostname list with 1 item.
```
"resources": [
    {
      "mode": "managed",
      "type": "cloudflare_list",
      "name": "example",
      "provider": "provider[\"registry.terraform.io/cloudflare/cloudflare\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "account_id": "<account_id>",
            "created_on": "2025-08-27T13:58:39Z",
            "description": "example",
            "id": "<list_id>",
            "items": [
              {
                "asn": null,
                "comment": null,
                "hostname": {
                  "exclude_exact_hostname": null,
                  "url_hostname": "example.com"
                },
                "ip": null,
                "redirect": null
              }
            ],
            "kind": "hostname",
            "modified_on": "2025-08-27T13:58:39Z",
            "name": "example",
            "num_items": 1,
            "num_referencing_filters": 0
          },
          "sensitive_attributes": []
        }
      ]
    }
  ]
  ```

File: internal/services/list/resource.go [hunk 6]
```diff
@@ -92,40 +94,52 @@ func (r *ListResource) Create(ctx context.Context, req resource.CreateRequest, r
...
+}
+
+func (r *ListResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
 	var state *ListModel
+	var data *ListModel
+	var plan *ListModel
 
 	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
```
ms-jpq, line 128: fyi there is a double `Plan.Get` 👍

broswen: I'm doing that on purpose because `data` is overwritten by the response from the API request and I need to check what the original plan was compared to the existing state.
And I use it to update the list items if they have changed.

https://github.com/cloudflare/terraform-provider-cloudflare/pull/5995/files/4acea093801819374d9edcab1a7a94bc624d1469#diff-5c6ab0396667735a7a7f506c6b0a41849868ccaa4fc85ec31976a6e662cba003R156-R159

ms-jpq: ohh, nice

docs/data-sources/list.md
mgirouard, line 1: Just a heads up that these are auto-generated from the TF schema at release time. Not sure if that's what you did, but just in case you're making some custom updates, they'll get lost on the next release.

broswen: These are autogenerated from the schema.

internal/services/list/data_source_schema.go
broswen: Per our conversation, updated the docs with a warning message like we have in the V4 provider.

vaishakdinesh:
lgtm

vaishakdinesh (approved):

internal/services/list/resource.go
```diff
@@ -141,13 +155,41 @@ func (r *ListResource) Update(ctx context.Context, req resource.UpdateRequest, r
 	}
 	data = &env.Result
 
+	if !plan.Items.Equal(state.Items) {
```
dackerman, line 158: what's the desired behavior if you unset `plan.Items` (e.g. set to `null` when previously state had a value)?  just wondering if this is doing what you want - it seems like you'd essentially unset the list but i'm not sure whether the bulk update call will work as intended if the value is `nil` rather than `[]`. would be good to have an acceptance for that

internal/services/list/resource.go
```diff
@@ -181,6 +223,22 @@ func (r *ListResource) Read(ctx context.Context, req resource.ReadRequest, resp
 	}
 	data = &env.Result
 
+	if !prev.Items.IsNull() {
```
dackerman: nit: it looks like maybe you only need to store `prevItemsIsNull := data.Items.IsNull()` and then you don't need to have a whole separate list in memory

internal/services/list/resource.go
broswen: Yes passing null is equivalent to [] for the API endpoint and clears out all items. I added some steps to an existing acceptance test to confirm.

internal/services/list/resource.go
broswen: Thanks, updated.

Issue #5736, opened 2025-07-10
cloudflare_zero_trust_access_application Computed customlist attributes are not being stored after create


GreenStage:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

5.6

### Affected resource(s)

I can replicate with zero_trust_access_application , it doesnt break anything but its a weird behaviour and IDK what other side effect it may have on other resources.

### Terraform configuration files

```hcl
resource "cloudflare_zero_trust_access_application" "myapp_3" {
  account_id = "c91137350c00a28806157ec3918faff2"
  type = "self_hosted"
  name = "tfapp_3"
  domain = "tfapp6.bacalhauaescadote.org"
}
```

### Link to debug output

https://gist.github.com/GreenStage/4b92730808c10c68b982f6884e2851e8

### Expected output

The output is correct, what is weird is the state.
When not passed, the API populates the response's `self_hosted_domains` based on the 'domain' field.

I expect `self_hosted_domains` to match
```
"self_hosted_domains": [
      "tfapp6.bacalhauaescadote.org"
    ],
```
but instead, its an empty array.

### Actual output

In the state:

"self_hosted_domains": [],

### Steps to reproduce

1. Create a zero_trust_access_application like the one in the example
2. Check the state

### Additional factoids

This isn't blocking 5.7, and it is as far as I cant tell symptomless right now, at least for zero_trust_access_application resources.

Furthermore, the self_hosted_domains array is populated on subsequent Updates. Its just the **Create that does not respect the computed list from the API response, if the list in the config is not set.**

I think this has something to do with the json parser in IfUnset mode + the customfield.List and customfield.NestedObjectList types returned on POST responses.

nlsandler:
Thanks for reporting this! It looks like this attribute is deprecated in favor of `destinations` - could you tell me if the `destinations` attribute is set in the state on create?

Also, when you say it's populated on subsequent updates - did you set `self_hosted_domains` in the config before those updates or was it still absent? If you set it, I think the behavior here is correct (we should ignore the attribute unless it's set explicitly by the user), but if you didn't it's clearly wrong.

GreenStage:
Hi @nlsandler .
[...]

PR #5997, opened 2025-08-22
Add regression check for access application computed optional properties

dackerman:
This adds some assertions to validate that destinations and self_hosted_domains are populated from the API as intended.

It currently fails due to https://github.com/cloudflare/terraform-provider-cloudflare/issues/5736 

We have a codegen fix for this, so once that makes it into next, this can be merged as a regression test.

(Also, as a drive-by improvement, I'm sorting the keys in `acctest.DumpState` because that makes it easier to read)

### DIFF
--- a/internal/acctest/acctest.go
+++ b/internal/acctest/acctest.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"reflect"
+	"sort"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


dackerman:
hello @tamas-jozsa - i just confirmed that this passes against latest `next` now (the fix commit on next is here: https://github.com/cloudflare/terraform-provider-cloudflare/commit/d6c64dc8775540dc5179cf1d655b374fba0c22bc)

therefore, i think we're good to merge this

tamas-jozsa:
Amazing and looks good to me. Merging now.

PR #6003, opened 2025-08-22
Migrate `workers_route`, `workers_script`

mgirouard:

## Changes being requested

- drops all grit support for most workers related things
- adds support for migrating `workers_route` and `workers_script` using `cmd/migrate`
- transforms config and state
- handles attribute renames
- handles renames in references
- handles binding transforms
- adds tests for all the above


internal/services/workers_route/migrations_test.go
```diff
@@ -0,0 +1,149 @@
...
+				// Step 1: Create with v4 provider
+				ExternalProviders: map[string]resource.ExternalProvider{
+					"cloudflare": {
+						Source:            "cloudflare/cloudflare",
+						VersionConstraint: "4.52.1",
+					},
+				},
+				Config: v4Config,
+			},
+			// Step 2: Run migration and verify state (note: resource name should be updated by Grit)
```
mgirouard: This comment is no longer true.

mgirouard:
cc @1000hz for review. May be easiest to go commit-by-commit (see messages for details). Looking for feedback on some of the quirks with workers: resource and attribute renames, bindings, etc. Migration tests should be fairly comprehensive. LMK if I missed anything important.

I needed to drag the migrations for bindable resources into this PR. It's already a big one, but it felt right since they're all so closely coupled in TF. Also looking for feedback on [removals from v4 going to v5](https://github.com/cloudflare/terraform-provider-cloudflare/pull/6003/files#diff-1a6f973f4177e1b9e76d5b3e29477fabc2a15c2ce22303209e71cd2c0f0c3d36R60-R70) (`dispatch_namespace`, `hyperdrive_config_binding`, `module`).

musa-cf:
Not specific to this PR, but these tests are going to take forever to run once we have all of them done. 😅

musa-cf (approved):

vaishakdinesh (approved):

cmd/migrate/renames.go
```diff
@@ -42,12 +42,17 @@ func applyRenames(block *hclwrite.Block) {
...
 		"cloudflare_access_application": "cloudflare_zero_trust_access_application",
-		"cloudflare_access_group": "cloudflare_zero_trust_access_group",
-		// Add more renames as needed
+		"cloudflare_access_group":       "cloudflare_zero_trust_access_group",
+		// Workers resource renames
+		"cloudflare_worker_route":        "cloudflare_workers_route",
+		"cloudflare_worker_script":       "cloudflare_workers_script",
+		"cloudflare_worker_cron_trigger": "cloudflare_workers_cron_trigger",
+		"cloudflare_worker_domain":       "cloudflare_workers_custom_domain",
+		"cloudflare_worker_secret":       "cloudflare_workers_secret",
```
1000hz: These two aren't actually the same resource. In v5, `cloudflare_workers_secret` hits the **Workers for Platforms** APIs (and as a result never worked as users expected, so we dropped it from v5.3.0). Existing `cloudflare_worker_secret` resources should be migrated to a binding of `secret_text` type on the corresponding `cloudflare_workers_script` resource.

cmd/migrate/workers_script.go
```diff
@@ -0,0 +1,233 @@
...
+	// Transform v4 binding attributes to v5 bindings list
+	result = transformBindingsInState(result, path)
+
+	// Remove unsupported attributes that exist in v4 but not in v5
+	dispatchNamespacePath := path + ".attributes.dispatch_namespace"
+	result, _ = sjson.Delete(result, dispatchNamespacePath)
+
+	// Remove hyperdrive_config_binding which exists in v4 but not expected in v5 schema
+	hyperdriveConfigBindingPath := path + ".attributes.hyperdrive_config_binding"
+	result, _ = sjson.Delete(result, hyperdriveConfigBindingPath)
```
1000hz: this should be migrated to a binding of type `hyperdrive` in v5

cmd/migrate/workers_script.go
```diff
@@ -0,0 +1,233 @@
...
+		// Delete the old attribute
+		result, _ = sjson.Delete(result, namePath)
+	}
+
+	// Transform v4 binding attributes to v5 bindings list
+	result = transformBindingsInState(result, path)
+
+	// Remove unsupported attributes that exist in v4 but not in v5
+	dispatchNamespacePath := path + ".attributes.dispatch_namespace"
+	result, _ = sjson.Delete(result, dispatchNamespacePath)
```
1000hz: this should be migrated to `namespace_id`

cmd/migrate/workers_script.go
```diff
@@ -0,0 +1,233 @@
...
+
+	// Remove hyperdrive_config_binding which exists in v4 but not expected in v5 schema
+	hyperdriveConfigBindingPath := path + ".attributes.hyperdrive_config_binding"
+	result, _ = sjson.Delete(result, hyperdriveConfigBindingPath)
+
+	// Remove module attribute which exists in v4 but causes issues in v5
+	modulePath := path + ".attributes.module"
+	result, _ = sjson.Delete(result, modulePath)
```
1000hz: in v4, `module = true` signifies that the Worker uses ES Module syntax and needs to have `main_module = "<some-arbitrary-filename>" # (e.g. "worker.js")` in v5.

`module = false` signifies that the Worker uses ServiceWorker syntax and needs to have `body_part = "<some-arbitrary-filename>" # (e.g. "worker.js")` in v5.

cmd/migrate/workers_script.go
```diff
@@ -0,0 +1,233 @@
...
+	v4BindingAttributes := []string{
+		"plain_text_binding",
+		"kv_namespace_binding",
+		"secret_text_binding",
+		"r2_bucket_binding",
+		"queue_binding",
+		"d1_database_binding",
+		"analytics_engine_binding",
+		"service_binding",
+		"webassembly_binding",
```
1000hz: `webassembly_binding` has no support in v5. any Wasm modules need to be bundled into script content.

cmd/migrate/workers_script.go
```diff
@@ -0,0 +1,233 @@
...
+	var blocksToRemove []*hclwrite.Block
+
+	// Map of v4 block types to v5 binding types
+	blockTypeToBindingType := map[string]string{
+		"plain_text_binding":       "plain_text",
+		"kv_namespace_binding":     "kv_namespace",
+		"secret_text_binding":      "secret_text",
+		"r2_bucket_binding":        "r2_bucket",
+		"queue_binding":            "queue",
+		"d1_database_binding":      "d1_database",
```
1000hz: the correct binding type in v5 is `"d1"`

cmd/migrate/workers_script.go
```diff
@@ -0,0 +1,233 @@
...
+	for _, bindingAttr := range v4BindingAttributes {
+		bindingPath := path + ".attributes." + bindingAttr
+		bindingValue := gjson.Get(jsonStr, bindingPath)
+
+		if bindingValue.Exists() {
+			// Convert binding attribute name to binding type
+			bindingType := bindingAttr
+			if bindingType[len(bindingType)-8:] == "_binding" {
+				bindingType = bindingType[:len(bindingType)-8]
+			}
```
1000hz: slicing off the `_binding` suffix isn't sufficient in all cases, e.g. `d1_database_binding` attributes are now bindings of type `"d1"` in v5, and `hyperdrive_config_binding` => `"hyperdrive"`. those should be the only exceptions though.

cmd/migrate/workers_script.go
```diff
@@ -0,0 +1,233 @@
...
+	})
+
+	// Copy all attributes from the binding block to the binding object
+	// Sort attribute names for deterministic output
+	var attrNames []string
+	attributes := block.Body().Attributes()
+	for attrName := range attributes {
+		attrNames = append(attrNames, attrName)
+	}
+	slices.Sort(attrNames)
```
1000hz, line 333: Some binding attributes have been renamed from v4 => v5.
e.g.
```diff
# v4
d1_database_binding {
  name = "myDb"
- database_id = "0123456789abdef0123456789abdef"
}

# v5
bindings = [{
  name = "myDb"
+ id = "0123456789abdef0123456789abdef"
  type = "d1"
}]
```

cmd/migrate/workers_secret.go
1000hz, line 1: as mentioned above, these need to be transformed into `secret_text` bindings on a `cloudflare_workers_script`

1000hz (changes requested):

cmd/migrate/renames.go
mgirouard: We'll do something like this, yeah?
```hcl
# V4 (source)
resource "cloudflare_workers_secret" "example" {
  account_id  = "..."
  script_name = "my-worker"        # ← Links to worker script
  name        = "MY_SECRET_NAME"   # ← Becomes binding name  
  secret_text = "secret_value"    # ← Becomes binding text
}

# V5 (target)
resource "cloudflare_workers_script" "example" {
  bindings = [{
    type = "secret_text"
    name = "MY_SECRET_NAME"        # From workers_secret.name
    text = "secret_value"          # From workers_secret.secret_text
  }]
}
```

cmd/migrate/workers_script.go
mgirouard: Something like this?

```hcl
# V4 (source)
resource "cloudflare_workers_script" "example" {
  dispatch_namespace = "my-dispatch-namespace"  # ← String attribute
  # ... other attributes
}
```

V5 (target)
```hcl
resource "cloudflare_workers_script" "example" {
  bindings = [{
    type         = "dispatch_namespace"
    namespace_id = "my-dispatch-namespace"  # ← Becomes namespace_id in binding
  }]
  # ... other attributes
}
```

cmd/migrate/workers_script.go
mgirouard: Something like this?
```hcl
# V4 (source) - ES Module syntax
resource "cloudflare_workers_script" "example" {
  module = true  # ← Boolean indicating ES Module syntax
  # ... other attributes
}
```

```hcl
# V5 (target)
resource "cloudflare_workers_script" "example" {
  main_module = "worker.js"  # ← String filename for ES Module
  # ... other attributes
}
```

```hcl
# V4 (source) - Service Worker syntax
resource "cloudflare_workers_script" "example" {
  module = false  # ← Boolean indicating Service Worker syntax
  # ... other attributes
}
```

```hcl
# V5 (target)
resource "cloudflare_workers_script" "example" {
  body_part = "worker.js"  # ← String filename for Service Worker
  # ... other attributes
}
```

cmd/migrate/workers_script.go
mgirouard: Correct V4 → V5 Transformation should be like this?

```hcl
# V4 (actual schema)
hyperdrive_config_binding {
  binding = "HYPERDRIVE"      # ← V4 attribute name
  id = "hyperdrive123"        # ← V4 attribute name
}

# V5 (should be)
bindings = [{
  type = "hyperdrive"
  name = "HYPERDRIVE"         # ← binding → name
  binding_id = "hyperdrive123"  # ← id → binding_id
}]
```

cmd/migrate/workers_script.go
1000hz: yup, exactly

cmd/migrate/workers_script.go
1000hz: ah, no I meant a top-level `namespace_id` property, but I just realized it's undocumented and not included in  the Terraform resource schema. you're right, there's currently no migration path for supporting this attribute in v5.

cmd/migrate/workers_script.go
1000hz: no, in hyperdrive's case it's just `binding → name`. (`id` remains the same)
you can refer to the `workers_binding_kind_*` schema objects from our OpenAPI description for the canonical attributes for each binding type, or here's a quick rundown:

#### Analytics Engine bindings
```diff
# v4
analytics_engine_binding {
	name = string
	dataset = string
}

# v5
bindings = [{
	name = string
	dataset = string
	type = "analytics_engine" 
}]
```

#### D1 Database bindings
```diff
# v4
d1_database_binding {
	name = string
-	database_id = string
}

# v5
bindings = [{
	name = string
+	id = string
	type = "d1" 
}]
```

#### Hyperdrive Config bindings
```diff
# v4
hyperdrive_config_binding {
-	binding = string
	id = string
}

# v5
bindings = [{
+	name = string
	id = string
	type = "hyperdrive" 
}]
```

#### KV Namespace bindings
```diff
# v4
kv_namespace_binding {
	name = string
	namespace_id = string
}

# v5
bindings = [{
	name = string
	namespace_id = string
	type = "kv_namespace" 
}]
```

#### Plain Text bindings
```diff
# v4
plain_text_binding {
	name = string
	text = string
}

# v5
bindings = [{
	name = string
	text = string
	type = "plain_text" 
}]
```

#### Queue bindings
```diff
# v4
queue_binding {
-	binding = string
-	queue = string
}

# v5
bindings = [{
+	name = string
+	queue_name = string
	type = "queue" 
}]
```

#### R2 Bucket bindings
```diff
# v4
r2_bucket_binding {
	name = string
	bucket_name = string
}

# v5
bindings = [{
	name = string
	bucket_name = string
	type = "r2_bucket" 
}]
```

#### Secret Text bindings
```diff
# v4
secret_text_binding {
	name = string
	text = string
}

# v5
bindings = [{
	name = string
	text = string
	type = "secret_text" 
}]
```

#### Service bindings
```diff
# v4
service_binding {
	name = string
	service = string
}

# v5
bindings = [{
	name = string
	service = string
	type = "service" 
}]
```

cmd/migrate/workers_script.go
mgirouard: ACK :+1: Is there a future where something like this would exist? I could keep the existing code for that if it's something you'd find valuable.

cmd/migrate/renames.go
1000hz: yup, exactly. just be sure to target only the `cloudflare_workers_script` with the same `account_id` and `script_name` (which I'm sure is probably obvious, but figured I'd call it out just to be extra safe.)

cmd/migrate/workers_script.go
1000hz: yeah it's probably coming Soon™

mgirouard:
Heads up @1000hz -- I just saw this on the TF docs page 
image

It sounds like we shouldn't be migrating folks onto the `workers_script` resource since `worker_script` is to be its successor?

1000hz:
the existing `cloudflare_workers_script` resource is the most analogous resource for someone migrating from v4 => v5, but yeah we hope to deprecate it eventually in favor of a combination of the new `cloudflare_worker` and  `cloudflare_worker_version` resources we just introduced, and `cloudflare_workers_deployment`.

you could perhaps migrate straight to these new resources, but it would be a slightly more complex migration since you’d go from 1 resource block to 3. seems doable though.

mgirouard:
ACK :+1: I'll continue with the migration to the `cloudflare_workers_script` resource for now.

vaishakdinesh:
lgtm

PR #6013, opened 2025-08-25
feat: migrate zero_trust_access_group migrations

tamas-jozsa:


cmd/migrate/test_dir/test.tf
mgirouard, line 1: `git add .` can be dangerous :stuck_out_tongue: 

This may still be useful if it demonstrates a reproduction of the issue. If that's the case, maybe you can add this to a `testdata/` dir instead.

internal/acctest/acctest.go
```diff
@@ -572,6 +572,9 @@ func isFalseyValue(v interface{}) bool {
 
 var ExpectEmptyPlanExceptFalseyToNull = expectEmptyPlanExceptFalseyToNull{}
 
+// ExpectEmptyPlanExceptZeroTrustAccessGroupOrdering is a specialized plan check for zero trust access group migrations
```
mgirouard: ~Uncommitted patch?~ Ah it's been moved to `zero_trust_access_group.go`

tamas-jozsa: Missed to clean this up. It's removed now.

cmd/migrate/test_dir/test.tf
tamas-jozsa: nice catch! ^^

PR #6014, opened 2025-08-25
feat: list and list items migrations for config and state

vaishakdinesh:

## Additional context & links
Changes:
- remove grit patterns for list and list item
- add logic in migrate pkg to use hcl to load the existing config and transform it for v5
- parse the state in its entirety to move it to v5 rather than instance by instance
- tests


cmd/migrate/list_test.go
```diff
@@ -0,0 +1,1306 @@
...
+				Attributes: attrs,
+			}
+		}
+	}
+
+	return resources
+}
+
+// tokensToString converts HCL tokens to a normalized string
+func tokensToString(tokens hclwrite.Tokens) string {
```
ms-jpq, line 88: some useful thing that I found:

---

you can actually inline this function by just calling on `thing`

`string(thing.BuildTokens(nil).Bytes())`

cmd/migrate/list_dynamic_blocks_test.go
```diff
@@ -0,0 +1,279 @@
...
+			},
+		},
+		{
+			Name: "dynamic block already with toset should not double wrap",
+			Config: `resource "cloudflare_list" "my_ips" {
+  account_id = "abc123"
+  name       = "my_ips"
+  kind       = "ip"
+
+  dynamic "item" {
```
ms-jpq, line 52: ah nice! this was the case that i pretty worry about

cmd/migrate/list_dynamic_blocks_test.go
```diff
@@ -0,0 +1,279 @@
...
+  }
+}`,
+			Expected: []string{
+				`resource "cloudflare_list" "asn_list" {
+  kind       = "asn"
+  account_id = "abc123"
+  name       = "asn_list"
+}`,
+				`resource "cloudflare_list_item" "asn_list_items" {
+  for_each   = toset(var.asn_map)
```
ms-jpq, line 99: humm, I think this is wrong. if you do `toset` the keys and values become identical. 

ie. iterating over `toset(["one", "two", "three"]) would yield

`each.key` -> `"one", "two", "three"`
and
`each.value` -> `"one", "two", "three"`

we want something like a map here where iterating over `for_each = {a="one", b="two"}`

would yield

`each.key` -> `"a", "b"`
and
`each.value` -> `"one", "two"`

cmd/migrate/list.go
```diff
@@ -0,0 +1,926 @@
...
+		}
+	}
+
+	// Return original tokens if not a quoted string
+	return tokens
+}
+
+// wrapForEachForObjects creates a for expression for lists of objects
+// The key field is determined by the list kind
+func wrapForEachForObjects(expr *hclwrite.Expression, listKind string) hclwrite.Tokens {
```
ms-jpq, line 646: imo, it might be easier in the future to use something like 


```go
// just an example I had, i think this is slightly easier to read and modify
syn := &hclsyntax.Block{
				Type:   "resource",
				Labels: labels,
				Body: &hclsyntax.Body{
					Attributes: hclsyntax.Attributes{
						"for_each": &hclsyntax.Attribute{
							Name: "for_each",
							Expr: item.ValueExpr,
						},
						"account_id": &hclsyntax.Attribute{
							Name: "account_id",
							Expr: ast.NewScopeTraversal(append(slices.Clone(labels), "account_id")...),
						},
						"list_id": &hclsyntax.Attribute{
							Name: "list_id",
							Expr: ast.NewScopeTraversal(append(slices.Clone(labels), "id")...),
						},
						"comment": &hclsyntax.Attribute{Name: "comment",
							Expr: ast.NewScopeTraversal("each", "value", "content", "comment"),
						},
						"ip": &hclsyntax.Attribute{Name: "ip",
							Expr: ast.NewScopeTraversal("each", "value", "content", "ip", "value"),
						},
					},
				},
			}

blockString := ast.Blocks2S(newBlocks, diags)
// and then just reparse the blockString into hclwrite.Tokens, which we currently do not have a helper for, but it should be easy to write one.
```

I personally find it kind of scary & hard to read when we are doing token concat rather than AST composition.

ms-jpq (commented):
some comments and suggestions :D

PR #6024, opened 2025-08-26
test: Add acceptance-tests to zero_trust_tunnel_cloudflared_route

jcsf:
Extra tests for zero_trust_tunnel_cloudflared_route resource

### DIFF
--- a/internal/services/zero_trust_tunnel_cloudflared_route/resource_test.go
+++ b/internal/services/zero_trust_tunnel_cloudflared_route/resource_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/cloudflare/cloudflare-go"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
 
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/acctest"
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/consts"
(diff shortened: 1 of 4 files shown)


mgirouard (approved):

internal/services/zero_trust_tunnel_cloudflared_route/resource_test.go
```diff
@@ -71,15 +72,28 @@ func TestAccCloudflareTunnelRoute_Exists(t *testing.T) {
...
 					resource.TestCheckResourceAttrSet(name, "tunnel_id"),
-					resource.TestCheckResourceAttr(name, "network", "10.0.0.20/32"),
+					resource.TestCheckResourceAttr(name, "network", "10.10.10.10/32"),
 					resource.TestCheckResourceAttr(name, "comment", rnd),
 				),
 			},
+			// Update
+			{
+				Config: testAccCloudflareTunnelRouteSimple(rnd, rnd, accountID, "20.20.20.20/32"),
+				Check:  resource.TestCheckResourceAttr(name, "network", "20.20.20.20/32"),
```
vaishakdinesh, line 88: Can we add a config plan check to ensure nothing else changes. And maybe one more on an update which doesn't update anything but rather just verifies its a noop?

vaishakdinesh: Don;t wanna block this PR, we can add it in a later one too if you think thats fine!

mgirouard:
Agree with @vaishakdinesh's comment. Merging now to keep the fixes flowing.

internal/services/zero_trust_tunnel_cloudflared_route/resource_test.go
jcsf: Added the proposed changes in this PR: https://github.com/cloudflare/terraform-provider-cloudflare/pull/6034

PR #6026, opened 2025-08-26
feat: add case-insensitive location handling for R2 bucket resources

pvail-cf:

### DIFF
--- a/internal/services/r2_bucket/plan_modifiers.go
+++ b/internal/services/r2_bucket/plan_modifiers.go
@@ -1,0 +1,30 @@
+package r2_bucket
+
+import (
+	"context"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+func modifyPlan(ctx context.Context, req resource.ModifyPlanRequest, res *resource.ModifyPlanResponse) {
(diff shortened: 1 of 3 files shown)


internal/services/r2_bucket/plan_modifiers.go
```diff
@@ -0,0 +1,27 @@
...
+	if stateApp != nil && !planApp.Location.IsNull() && !stateApp.Location.IsNull() &&
+		!planApp.Location.IsUnknown() && !stateApp.Location.IsUnknown() {
+		planAppLocation := planApp.Location.ValueString()
+		stateAppLocation := stateApp.Location.ValueString()
+		if strings.EqualFold(planAppLocation, stateAppLocation) {
+			planApp.Location = stateApp.Location
+		}
+	}
+
+	res.Diagnostics.Append(res.Plan.Set(ctx, &planApp)...)
```
vaishakdinesh: probably don't need reset the whole plan reduce chance of changing other things unintentionally? Just the value if it changes?
Like resp.Plan.SetAttribute(ctx, path.Root("location"), stateApp.Location)?

Also wanna make sure that's the case the Read sees too. Just confirming

pvail-cf: Fixed, want to give it another look?

vaishakdinesh (approved):

PR #6031, opened 2025-08-27
feat: migrate dns records

vaishakdinesh:

## Additional context & links
- use grit to update the config
- migrate pkg updates to handle special cases for config and state upgrades
- update schema and add plan modifiers to normalize diffs

### DIFF
--- a/cmd/migrate/test_helpers.go
+++ b/cmd/migrate/test_helpers.go
@@ -160,4 +160,4 @@ func normalizeWhitespace(s string) string {
 	s = strings.ReplaceAll(s, "] ", "]")
 
 	return s
-}c
+}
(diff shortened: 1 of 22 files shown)


cmd/migrate/test_helpers.go
```diff
@@ -160,4 +160,4 @@ func normalizeWhitespace(s string) string {
 	s = strings.ReplaceAll(s, "] ", "]")
 
 	return s
-}
+}c
```
mgirouard: Looks like an accidental add?

internal/services/dns_record/schema.go
```diff
@@ -27,6 +30,7 @@ var _ resource.ResourceWithConfigValidators = (*DNSRecordResource)(nil)
 
 func ResourceSchema(ctx context.Context) schema.Schema {
 	return schema.Schema{
+		Version: 1,
```
vaishakdinesh: todo: remove

vaishakdinesh: done!

PR #6038, opened 2025-08-27
making token tests runnable in staging

steve-thousand:

## Changes being requested

Rewriting token tests to not be environment specific. Generally better for development.

## Additional context & links

Hardcoded permission group IDs are not portable across environment. I rewrote the test HCL to use permission group datasource lists to find the desired permission group by name.

I also updated the naming convention for the pre-existing API token test files. I swear I even did a `git mv ...` to ensure they would be renamed instead of delete/add but it looks like that didn't work


internal/services/account_token/testdata/account_token-permissiongroup-order.tf
```diff
@@ -1,22 +1,28 @@
-resource "cloudflare_account_token" "%[1]s" {
```
steve-thousand, line 1: An aside - these `"%[1]s"` resource names are for some reason removed when I save the file in my vscode. Not sure what is causing that but giving it a definite resource name stops that. I see no reason why these can't have definite resource names.

scripts/run-ci-tests
```diff
@@ -4,11 +4,10 @@
 # Supports both acceptance tests and migration tests
 # Dynamically detects which services have sweeper implementations by grepping for AddTestSweepers
 
-# ./internal/services/api_token
-# 'api_token' requires token-based auth, and our current CI assumes key-based auth.
-# These tests otherwise pass.
-# TODO: support multiple auth schemes in CI for acceptance tests
```
steve-thousand, line 10: I was the one who recommended this and I was incorrect in that I assumed the CI was using an account-owned token. It is really using a global API key which is capable of creating user API tokens (I have been using one myself to run these tests locally).

However, long term these tests should move away from the global API key because that is a feature that we would one day like to eliminate. The future is in tokens

scripts/run-ci-tests
```diff
@@ -4,11 +4,10 @@
...
-
+# ./internal/services/account
+# account creation can only be done by a tenant admin and the current CI does 
+# have that support.
+# TODO: support tenant admin auth for acceptance tests
```
steve-thousand, line 10: This part has been mentioned already but the account tests are currently set to skip so they don't need to be removed from this list

internal/services/account_token/testdata/account_token-permissiongroup-order.tf
mgirouard: I've seen that happen if I run `terraform fmt` on these files. I'll be you might have a format on save type of thing setup in your editor.

Issue #6062, opened 2025-08-29
Restore v3 and v4 schemas for migrations


andrewmwhite:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

The current migration path from v3/v4 to v5 seems to require doing state surgery via grit. This isn't feasible when many workspaces are involved. 

I understand there's ongoing work to improve the migration in the `next` branch, which is great! But from the documentation it seems to still have the same issue.

We need some declarative way to migrate our configs from v3/v4 to v5. If the old schemas were backported into v5, we could use `removed`/`import` blocks to do so.

I have code locally that does exactly this for a number of resources and will happily submit a PR if the maintainers are willing to backport these schemas.

### Affected resource(s)

Many, but especially:

- cloudflare_argo_tunnel
- most cloudflare zero trust resources

### Terraform configuration files

```hcl
n/a
```

### Link to debug output

n/a

### Expected output

n/a

### Actual output

n/a

### Steps to reproduce

n/a

KaydeeDee:
Hi there - I completely understand where you're coming from and am extremely sympathetic to the situation folks are in around v5 and migration. Let me bring this to the team to discuss how we can help.

chris-pinola-rf:
> we could use `removed`/`import` blocks to do so.

I recently learned that Terraform now supports using `moved` blocks with mixed types...given that the provider has implemented state migration logic. See: https://github.com/hashicorp/terraform/issues/27310#issuecomment-2316319684

chris-pinola-rf:
Still relevant IMO

KaydeeDee:
Hi All - We have discussed with and came to the decision that we will not be restoring the schemas mentioned in this ticket. We are actively working on v5 stabilization (every member of our team!), which includes migration issues (see list here: https://github.com/cloudflare/terraform-provider-cloudflare/issues/6237). We expect to be in a very good place by the end of Q4 and an even better place by Q1. We know it's been difficult and we appreciate how patient you have been - we care a lot about your experience with our provider and want to make sure we stay focused on v5 fixes to get it as stable as possible as fast as possible.

PR #6064, opened 2025-08-29
Restore v3 and v4 schemas to enable removed/import

andrewmwhite:

## Changes being requested
This PR restores schemas from v3/v4 in order to enable declarative migration from v3/v4 to v5 using `removed`/`import` blocks. It is intended more as a conversation starter than an fully fleshed-out contribution.

The currently suggested migration path involves state surgery, which isn't feasible in our case (with dozens of workspaces). We could instead use pairs of `removed` / `import` blocks to remove old resources and import under the new resource types. However, v4 and later v5 completely removed the schemas for the old resource types, which means we cannot remove and import at the same time. This PR restores the old schemas under the old names. This allows `removed` blocks to target the old resource types and `import` blocks to target the new resource types, enabling a smooth migration via TF config.

Here is a partial implementation of https://github.com/cloudflare/terraform-provider-cloudflare/issues/6062 but only covers certain resources.

## Additional context & links


tamas-jozsa:
This pr is incorrectly targeting main. We are merging against next. Please also add description to your MR.

vaishakdinesh:
Hi @andrewmwhite I understand this intended more as a conversation starter but i believe we currently dont have plans for those paths. 
@KaydeeDee @rupalims

KaydeeDee:
Hi there - we aren't going to be moving forward with v3 to v4 migrations. We're almost done with our v4 to v5 migration readiness, and that's where our focus is currently.

PR #6067, opened 2025-08-29
Fix cmd/migrate tests, add some helpers, and fix a couple small bugs

dackerman:
## Summary

There were some failing tests in cmd/migrate and some flakiness, so I fixed that and also tried to improve the helpers a little.

## Problem

The migration tool tests were experiencing intermittent failures due to non-deterministic map iteration in Go, which caused attributes to appear in random order across test runs. Additionally, the test helpers had accumulated technical debt with multiple layers of fallback comparison strategies that made the code difficult to maintain.

### 1. Fixed Test Flakiness
- Used AttributesOrdered() helper function to ensure deterministic attribute ordering when iterating over HCL body attributes
- Applied deterministic ordering to tiered_cache.go and zone_settings.go transformations
- Tests now pass consistently with -count 100 (verified with no failures across 100 consecutive runs)

### 2. Refactored Test Helpers
- Consolidated test_helpers_hcl.go into test_helpers.go to eliminate code duplication
- Simplified comparison logic from 6+ fallback strategies to 2 clear approaches (one for HCL, one for JSON)
- Removed fragile manual token extraction that was prone to edge cases
- Eliminated resource-specific coupling (hardcoded checks for include/exclude/require arrays)
- Reduced code from 367+ lines to ~250 lines while improving functionality

### 3. Improved Test Infrastructure
- HCL comparisons: Aggressive normalization handles formatting differences (e.g., single-line vs multi-line arrays)
- JSON comparisons: Proper semantic comparison using reflect.DeepEqual on normalized values
- Better error messages: Consistent use of cmp.Diff for clear, actionable test failure output
- Shared utilities: Moved createMovedBlock() to helpers.go for reuse across multiple transformations

### 4. Bug Fixes
- Fixed zone settings security header test to expect correct strict_transport_security wrapper structure

## Testing

- All existing tests pass
- Verified with go test -count 100 to ensure no flakiness remains
- Test output is now deterministic and consistent across runs

## Impact

This change improves the reliability of the migration tool test suite and makes it easier to maintain and debug test failures. The deterministic output also makes it easier to review changes in generated HCL configurations.


dackerman:
@tamas-jozsa @musa-cf @vaishakdinesh @mgirouard this PR should be good to go - mostly just fixing the tests and attempting to improve the helpers & assertions, but there was one edge case that access group wasn't handling for the migration.

are we running this in CI yet? we probably should add it - shouldn't add much time since they are unit tests

cmd/migrate/test_helpers.go
```diff
@@ -27,24 +26,195 @@ func RunTransformationTests(t *testing.T, tests []TestCase, transformFunc func([
...
+	if strings.Contains(actualNorm, expectedNorm) {
+		return true
+	}
+
+	// For more complex cases, try semantic matching
+	return hclSemanticMatch(actual, expected)
+}
+
+// normalizeHCLContent aggressively normalizes HCL for comparison
+func normalizeHCLContent(s string) string {
```
dackerman, line 82: there's probably a better way to do this kind of check, but this works for now - can try to simplify later

vaishakdinesh:
> @tamas-jozsa @musa-cf @vaishakdinesh @mgirouard this PR should be good to go - mostly just fixing the tests and attempting to improve the helpers & assertions, but there was one edge case that access group wasn't handling for the migration.
> 
> are we running this in CI yet? we probably should add it - shouldn't add much time since they are unit tests

Oh the `cmd/migrate` unit tests!? I don't think we do. Good call out. We totally should.

Issue #898, opened 2020-12-20
Error: invalid value for api_key


bcruik:
### Terraform version
Terraform v0.14.3
+ provider registry.terraform.io/cloudflare/cloudflare v2.14.0
+ provider registry.terraform.io/hashicorp/aws v3.22.0

### Affected resource(s)
+ provider registry.terraform.io/cloudflare/cloudflare v2.14.0

### Terraform configuration files
```
provider "cloudflare" {
  email   = "var.cloudflare_email"
  api_key = "var.cloudflare_api_key"
  account_id ="var.cloudflare_account_id"
}
```
cloudflare.tfvars 
cloudflare_api_key starts with number and lowercase letters
```
// Cloudflare API KEY
variable "cloudflare_api_key" {
  default = "<SNIP>"
}
```
### Debug output
Error: invalid value for api_key (API key must only contain characters 0-9 and a-f (all lowercased))
```
2020-12-20T02:38:12.244Z [DEBUG] plugin: plugin process exited: path=.terraform/providers/registry.terraform.io/cloudflare/cloudflare/2.14.0/linux_amd64/terraform-provider-cloudflare_v2.14.0 pid=55950
2020-12-20T02:38:12.244Z [DEBUG] plugin: plugin exited
2020-12-20T02:38:12.244Z [WARN]  plugin.stdio: received EOF, stopping recv loop: err="rpc error: code = Unavailable desc = transport is closing"
2020-12-20T02:38:12.245Z [WARN]  plugin.stdio: received EOF, stopping recv loop: err="rpc error: code = Unavailable desc = transport is closing"
2020-12-20T02:38:12.248Z [DEBUG] plugin: plugin process exited: path=.terraform/providers/registry.terraform.io/hashicorp/aws/3.22.0/linux_amd64/terraform-provider-aws_v3.22.0_x5 pid=55943
2020-12-20T02:38:12.248Z [DEBUG] plugin: plugin exited
```
### Panic output
n/a

### Expected behavior
Cloudflare should accept API key generated in portal (I'm using the global API key from the portal).
This also fails when using api_token in like issue 842

### Actual behavior
Fails with error `Error: invalid value for api_key (API key must only contain characters 0-9 and a-f (all lowercased))`

cloudflare_api_key starts with number and lowercase letters

### Steps to reproduce
1. `terraform apply`

### Important factoids
Using Terraform on Oracle Cloud VM (always free), running Ubuntu 20.04. 
[...]

PR #6075, opened 2025-08-30
fix: comment_modified_on drift in DNS records

dackerman:
## Summary

This PR fixes the remaining DNS record drift issue where `comment_modified_on` was showing as "known after apply" on every plan, even when no changes were made to the comment field.

## Issue Fixed

**Problem**: The `comment_modified_on` field was exhibiting the same drift behavior that was previously fixed for `modified_on` and `tags_modified_on` fields, showing "(known after apply)" on every terraform plan even when the comment hadn't changed.

**Solution**: Extended the existing `ModifyPlan()` implementation to handle `comment_modified_on` with the same logic used for other computed timestamp fields:
- When comment is empty/null: Set `comment_modified_on` to null to prevent drift
- When comment exists but hasn't changed: Preserve `comment_modified_on` from state
- When comment has actually changed: Allow it to remain unknown (will be updated by API)

## Technical Details

This builds on the existing workaround for terraform-plugin-framework issue [#898](https://github.com/hashicorp/terraform-plugin-framework/issues/898) where computed fields adjacent to optional+computed collections show perpetual drift.

## Testing

- Verified that DNS records with comments no longer show spurious `comment_modified_on` drift
- Confirmed that `comment_modified_on` still updates correctly when comments are actually modified
- All existing drift fixes remain functional

This completes the DNS record drift fixes, providing users with clean terraform plans.


dackerman:
@vaishakdinesh i think this gets rid of the last bit of drift for DNS record. can go into the next release.

vaishakdinesh (approved):

PR #6109, opened 2025-09-04
Add graceful rotation support for client secrets

ajholland:


internal/sdkv2provider/resource_cloudflare_access_service_tokens_test.go
```diff
@@ -289,6 +298,129 @@ func TestAccCloudflareAccessServiceToken_WithDuration(t *testing.T) {
...
+		},
+	})
+}
+
+func TestAccCloudflareAccessServiceToken_DefaultClientSecretVersion(t *testing.T) {
+	// Temporarily unset CLOUDFLARE_API_TOKEN if it is set as the Access
+	// Service Tokens endpoint does not yet support the API tokens and it
+	// results in misleading state error messages.
+	if os.Getenv("CLOUDFLARE_API_TOKEN") != "" {
+		t.Setenv("CLOUDFLARE_API_TOKEN", "")
```
mgirouard, line 349: I forget how this works with go tests -- does the env get shared between tests or does it automatically get reset? If not, should we consider a `defer` to set this back?

(same comment elsewhere)

mgirouard (approved):

PR #6149, opened 2025-09-12
fix: snippet and load balancer migration tests

vaishakdinesh:

### DIFF
--- a/cmd/migrate/snippet_rules_test.go
+++ b/cmd/migrate/snippet_rules_test.go
@@ -338,11 +338,6 @@ func TestTransformSnippetRulesStateJSON(t *testing.T) {
 			instance := instances[0].(map[string]interface{})
 			attributes := instance["attributes"].(map[string]interface{})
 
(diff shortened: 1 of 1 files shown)


internal/services/snippet/model.go
```diff
@@ -36,7 +36,7 @@ var SnippetFileType = snippetFileType{
 type SnippetModel struct {
 	SnippetName types.String          `tfsdk:"snippet_name" path:"snippet_name,required"`
 	ZoneID      types.String          `tfsdk:"zone_id" path:"zone_id,required"`
-	Files       *[]SnippetFile        `tfsdk:"files" json:"files,metadata,required"`
+	Files       *[]SnippetFile        `tfsdk:"files" json:"files,required"`
```
vaishakdinesh: this looks right yeah? @musa-cf @mgirouard @tamas-jozsa

vaishakdinesh: The test pass with or without it. But this seems right to me, but put it back?

PR #6160, opened 2025-09-12
feat: support for handling Workers Assets uploads

1000hz:

## Changes being requested
This PR introduces a new `assets.directory` attribute to the `cloudflare_workers_script` and `cloudflare_worker_version` resources. If this attribute is set, the referenced assets directory will be uploaded as part of the resource creation or update operation.

```tf
resource "cloudflare_worker_version" "static_site" {
  account_id = var.account_id
  worker_id  = cloudflare_worker.static_site.id
  assets = {
    directory = "../path/to/assets"
  }
}
```
```tf
resource "cloudflare_workers_script" "static_site" {
  account_id  = var.account_id
  script_name = "static-site"
  assets = {
    directory = "../path/to/assets"
  }
}
```

### DIFF
--- a/internal/services/worker_version/schema.go
+++ b/internal/services/worker_version/schema.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework-jsontypes/jsontypes"
 	"github.com/hashicorp/terraform-plugin-framework-timetypes/timetypes"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listplanmodifier"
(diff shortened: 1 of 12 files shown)


1000hz:
An earlier version of this PR set `assets.jwt` as a `WriteOnly` attribute. After some internal discussion, we've decided to avoid using `WriteOnly` attributes in the provider for the time being since they require Terraform v1.11+ and many of our users are still on older versions.

This PR has been updated to no longer use `WriteOnly` attributes.

mgirouard (approved):

1000hz:
@mgirouard @musa-cf @vaishakdinesh @tamas-jozsa I rebased this PR on top of the latest Workers resource changes. Mind taking another look?

internal/services/worker_version/schema.go
```diff
@@ -254,9 +255,7 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 			},
 			"assets": schema.SingleNestedAttribute{
 				Description: "Configuration for assets within a Worker.\n\n[`_headers`](https://developers.cloudflare.com/workers/static-assets/headers/#custom-headers) and\n[`_redirects`](https://developers.cloudflare.com/workers/static-assets/redirects/) files should be\nincluded as modules named `_headers` and `_redirects` with content type `text/plain`.",
-				Computed:    true,
 				Optional:    true,
-				CustomType:  customfield.NewNestedObjectType[WorkerVersionAssetsModel](ctx),
```
1000hz, line 259: `assets` recently became codegen'd as `Computed: true` (and switched to using a custom type to handle this) because we made the nested `assets.config.*` attributes have defaults. But the top-level `assets` property isn't always returned by the API (i.e. if your Version doesn't include any assets) so bubbling up the `Computed: true` to the top-level `assets` attribute was wrong.

PR #6163, opened 2025-09-14
fix: ruleset migration issues

musa-cf:

### DIFF
--- a/cmd/migrate/cloudflare_ruleset_test.go
+++ b/cmd/migrate/cloudflare_ruleset_test.go
@@ -8,6 +8,72 @@ import (
 	"github.com/tidwall/gjson"
 )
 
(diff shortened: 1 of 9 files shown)


cmd/migrate/cloudflare_ruleset_test.go
```diff
@@ -8,6 +8,72 @@ import (
 	"github.com/tidwall/gjson"
 )
 
+func TestDebugQueryStringTransform(t *testing.T) {
```
vaishakdinesh, line 11: Dont think we need this? Looks like its just debugging?

musa-cf: True, let me remove it

vaishakdinesh (approved):

PR #6196, opened 2025-09-16
fix: preserve heredoc expressions in cloudflare_ruleset migration

tamas-jozsa:
- Add special post-processing to restore escaped heredoc expressions in cloudflare_ruleset resources
- Scope double-dollar-sign fixes to only cloudflare_ruleset resources to avoid breaking variable references in other resources
- Fix 'Unterminated template string' errors during v4 to v5 migration
- Add comprehensive tests for complex ruleset configurations with heredoc expressions and nested ratelimit blocks

### DIFF
--- a/cmd/migrate/transformations/config_block_to_attribute.go
+++ b/cmd/migrate/transformations/config_block_to_attribute.go
@@ -56,6 +56,13 @@ func transformBlocksToList(config *TransformationConfig, body *hclwrite.Body, bl
 		return transformOriginsBlocksToList(body, blocks)
 	}
 
+	// Special handling for rules blocks in cloudflare_ruleset which may contain heredoc expressions
+	// Use standard transformation with enhanced post-processing for now
+	// The token-based approach has indentation issues that need more work
+	// if resourceType == "cloudflare_ruleset" && blockName == "rules" {
+	//     return transformRulesBlocksToList(body, blocks)
+	// }
+
 	// Build list of objects from blocks
 	var objects []cty.Value
 	for _, block := range blocks {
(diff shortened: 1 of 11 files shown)


cmd/migrate/transformations/config_block_to_attribute.go
```diff
@@ -56,6 +56,13 @@ func transformBlocksToList(config *TransformationConfig, body *hclwrite.Body, bl
 		return transformOriginsBlocksToList(body, blocks)
 	}
 
+	// Special handling for rules blocks in cloudflare_ruleset which may contain heredoc expressions
```
mgirouard, line 59: Can we remove the commented out code?

cmd/migrate/transformations/transformer.go
```diff
@@ -77,6 +86,13 @@ func (ht *HCLTransformer) TransformFile(inputPath, outputPath string) error {
 
 	// Write the transformed content with formatting
 	output := hclwrite.Format(ht.file.Bytes())
+	
+	// Post-process to fix double dollar signs and heredoc issues in cloudflare_ruleset expressions
+	// Only apply if we're dealing with heredoc expressions that need it
+	if strings.Contains(string(output), "<<-EOF") || strings.Contains(string(output), "<<-EOT") {
+		output = fixCloudflareRulesetDoubleDollarSigns(output)
```
mgirouard, line 93: Nit: I suspect that it's not _specifically_ the double dollar signs that we're trying to fix. I assume that was just the symptom because they're getting somehow escaped/double encoded/etc. Maybe we can call this something better?

cmd/migrate/transformations/transformer.go
```diff
@@ -85,6 +101,9 @@ func (ht *HCLTransformer) TransformFile(inputPath, outputPath string) error {
 		return fmt.Errorf("failed to write output file: %w", err)
 	}
 
+	// Skip terraform fmt during transformation to avoid working directory issues
```
mgirouard, line 104: These comment lines seem to be orphaned. Can it be removed?

cmd/migrate/transformations/transformer.go
```diff
@@ -141,4 +160,92 @@ func (ht *HCLTransformer) HasAttributeRename(resourceType, attrName string) (str
 // ShouldRemoveAttribute is a helper that delegates to the attribute_renames module
 func (ht *HCLTransformer) ShouldRemoveAttribute(resourceType, attrName string) bool {
 	return ShouldRemoveAttribute(ht.config, resourceType, attrName)
-}
\ No newline at end of file
+}
+// fixCloudflareRulesetDoubleDollarSigns fixes double dollar sign escaping and heredoc issues ONLY within cloudflare_ruleset resources
```
mgirouard, line 164: > heredoc issues ONLY within cloudflare_ruleset resources

is it true that this is specific to heredocs in this one specific resource? I would have imagined that this would be a more general heredoc handling issue.

cmd/migrate/transformations/transformer.go
mgirouard, line 93: It's kind of a bummer we have to operate on raw bytes instead of being able to use the HCL tooling we have access to just a few lines up. Are there no settings or options we can define to that `hclwrite` understands to control the escaping in heredocs?

PR #6200, opened 2025-09-16
chore: fix variable interpolation in a post process step

vaishakdinesh:

### DIFF
--- a/cmd/migrate/transformations/config_block_to_attribute_interpolation_test.go
+++ b/cmd/migrate/transformations/config_block_to_attribute_interpolation_test.go
@@ -3,6 +3,7 @@ package transformations
 import (
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmd/migrate/transformations/post_process.go
```diff
@@ -0,0 +1,478 @@
...
+	lines := strings.Split(content, "\n")
+	inHeredoc := false
+	isIndentedHeredoc := false
+
+	for i := 0; i < len(lines); i++ {
+		line := lines[i]
+		trimmed := strings.TrimSpace(line)
+
+		// Check if we're entering a heredoc
+		if strings.Contains(line, "<<-EOF") {
```
mgirouard, line 57: This works, but it probably should be a flag we pass in since not everyone is guaranteed to use `EOF`.

mgirouard (approved):

PR #6219, opened 2025-09-18
chore: make CAA test more accurate/useful

svenr-cf:

## Changes being requested

Make CAA test more correct and useful.

## Additional context & links

The "flags" field should be defined as an integer, so that's the behavior we need to ensure is working correctly.


vaishakdinesh:
No longer needed.

PR #6263, opened 2025-09-25
fix(workers_custom_domain): require resource replacement on update

1000hz:

## Changes being requested
`cloudflare_workers_custom_domain` doesn't actually support update operations, so we should require resource replacement on update.

## Additional context & links
Refs #6082


mgirouard (approved):

1000hz:
handling this via codegen instead

PR #6268, opened 2025-09-26
Fix failing GW schema parity check test

kian2attari:


internal/services/zero_trust_gateway_certificate/model.go
```diff
@@ -16,7 +16,7 @@ type ZeroTrustGatewayCertificateModel struct {
 	ID                 types.String      `tfsdk:"id" json:"id,computed"`
 	AccountID          types.String      `tfsdk:"account_id" path:"account_id,required"`
 	ValidityPeriodDays types.Int64       `tfsdk:"validity_period_days" json:"validity_period_days,optional,no_refresh"`
-	Activate           types.Bool        `tfsdk:"activate" json:"-"`
+	Activate           types.Bool        `tfsdk:"activate" json:"optional"`
```
alyssamw: I think that means we expect a json field named "optional" to contain this field, which is not the desired behavior... I think what you want is to add the line `errs.Ignore(".@ZeroTrustGatewayCertificateModel.activate")` to the specific test that's failing

https://github.com/cloudflare/terraform-provider-cloudflare/actions/runs/18016161094/job/51261493711?pr=6233#step:5:7137

tamas-jozsa:
Can you explain why is this the solution? I'm hesitant approving a solution that only ignores the error.

alyssamw:
Hi Tamas, there is no activate field in our API and the field is terraform only

vaishakdinesh:
Dont think this is required anymore.

PR #6282, opened 2025-10-01
APIOPS-11384-02: Add acceptance tests for Hostname Routes

GoncaloGarcia:

### DIFF
--- a/internal/services/zero_trust_network_hostname_route/data_source_test.go
+++ b/internal/services/zero_trust_network_hostname_route/data_source_test.go
@@ -1,0 +1,64 @@
+package zero_trust_network_hostname_route_test
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/acctest"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/utils"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+)
+
+// generateRandomTunnelSecret generates a random base64-encoded tunnel secret
+func generateRandomTunnelSecret(t *testing.T) string {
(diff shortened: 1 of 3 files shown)


internal/services/zero_trust_network_hostname_route/testdata/basic.tf
```diff
@@ -0,0 +1,17 @@
+resource "cloudflare_zero_trust_tunnel_cloudflared" "%[1]s" {
+  account_id    = "%[2]s"
+  name          = "%[1]s"
+  tunnel_secret = "AQIDBAUGBwgBAgMEBQYHCAECAwQFBgcIAQIDBAUGBwg="
```
chungthuang: Is this a secret for an actual tunnel?

GoncaloGarcia: Yes but it's a tunnel created in the test account

mgirouard: Please roll or invalidate this secret immediately.

For any kind of secrets, regardless if they're in a test account or not, they'll need to be defined as secrets in GitHub so we can reference them in CI via environment variables.

jcsf: You can generate the secret in the test, randomly since no one will connect to this tunnel.

GoncaloGarcia: @mgirouard removed the secret

internal/services/zero_trust_network_hostname_route/data_source_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+)
+
+// generateRandomTunnelSecret generates a random base64-encoded tunnel secret
+func generateRandomTunnelSecret() string {
+	bytes := make([]byte, 32) // 32 bytes for tunnel secret
+	if _, err := rand.Read(bytes); err != nil {
+		panic(err) // This should never happen in tests
```
vaishakdinesh: can you pass in `t` here and call fatal instead of panic?

GoncaloGarcia: Updated

PR #6283, opened 2025-10-01
DRAFT: Rex/improve error reporting at gateway policies

rexscaria:
Going to clean this up so we can provide accurate picture what's drifting in gateway policies. Our customers has been having lot of issues here, this should help


vaishakdinesh:
@rexscaria is this still needed?

vaishakdinesh:
Closing since it has seen an activity in over a month @rexscaria

PR #6307, opened 2025-10-04
Fix Account Member policies acceptance test by selecting the correct account-scoped resource group

AdamBouhmad:

## Changes being requested
The account_member TestAccCloudflareAccountMember_Policies test fails locally with the following error: 

```
  | POST
  | "https://api.cloudflare.com/client/v4/accounts/a67e14daa5f8dceeb91fe5449ba496eb/members":
  | 400 Bad Request
  | {"result":[],"success":false,"errors":[{"code":1006,"message":"Unknown error
  | when processing member, please try again. Error: invalid permission group and
  | resource group combination"}],"messages":[]}
```
After digging into the issue, the tests were failing because the test was grabbing data.cloudflare_resource_groups.example_resource_groups.result[0].id without constraining the query, so it could return a non–account-scoped (or empty) resource group, which would result in an invalid request. 

The fix was to filter the data source by name to always select an account resource group, and update the template to use positional specifiers (%[n]s) where account_id is referenced multiple times. After adding this, the test passes locally 

## Additional context & links
N/A

### DIFF
--- a/internal/services/account_member/testdata/cloudflareaccountmemberpoliciesconfig.tf
+++ b/internal/services/account_member/testdata/cloudflareaccountmemberpoliciesconfig.tf
@@ -1,18 +1,20 @@
 data "cloudflare_resource_groups" "example_resource_groups" {
-  account_id = "%s"
+  account_id = "%[1]s"
+  name       = "com.cloudflare.api.account.%[1]s"
 }
 
-resource "cloudflare_account_member" "%s" {
-  account_id = "%s"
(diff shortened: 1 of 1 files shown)


vaishakdinesh:
@AdamBouhmad can you make sure you're branched off next and target next?

AdamBouhmad:
closing in favor of https://github.com/cloudflare/terraform-provider-cloudflare/pull/6352

PR #6345, opened 2025-10-15
fix: cloudflare_pages_project continually reports changes to apply

christhorwarth:

## Changes being requested

Fixes to the pages_project resource so that only valid changes are detected.


internal/services/pages_project/resource.go
```diff
@@ -176,7 +176,7 @@ func (r *PagesProjectResource) Read(ctx context.Context, req resource.ReadReques
 		return
 	}
 	bytes, _ := io.ReadAll(res.Body)
-	err = apijson.Unmarshal(bytes, &env)
+	err = apijson.UnmarshalComputed(bytes, &env)
```
1000hz, line 179: Hmm, I don't think we should use `UnmarshalComputed()` in `Read()` since it'll leave most of the resource unrefreshed. It's meant to be used inside of `Create()` and `Update()`, when we know the API result will reflect whatever input we gave it and just want to fill in the attributes that the API has authority over.

We need to find which attributes are resulting in false drift and understand why (e.g. the attribute isn't properly marked as Optional/Computed, or the API response doesn't quite match the resource schema, etc.) and most likely fix it with schema updates or custom handling in `Create()`/`Update()`.

PR #6346, opened 2025-10-16
[DRAFT] fix: allow r2_bucket_event_notification to be applied twice without f…

Carolx715:
…ailing


internal/services/r2_bucket_event_notification/schema.go
Carolx715, line 1: I'm not super sure if I make these changes in the schema or not. I know for sure I need to change optional to required on R2 API's side, but what about the validators?

PR #6356, opened 2025-10-21
fix: allow r2_bucket_event_notification to be applied twice without failing

Carolx715:


tamas-jozsa:
Could you add integration tests to cover this case? And please link the result of your test run in the description.

Carolx715:
@tamas-jozsa integration tests are currently blocked on another ticket but i got the go ahead to put this one out first

musa-cf:
@tamas-jozsa 

Carol and I ran through this resource together and this is Step 1 of 2 for getting it completely working. There is a recurring plan diff happening due to an inconsistency between the PUT and GET responses, which will be the follow-up for getting the acceptance tests passing.

tamas-jozsa:
sounds good

PR #6357, opened 2025-10-21
fix: cloudflare_worker and cloudflare_worker_version can be cleanly imported

maxwellpeterson:

## Changes being requested

- Update `cloudflare_worker` resource to prevent changes on import
  - Add plan modifiers for created_on and updated_on to prevent these properties from incorrectly appearing in the diff
  - Fill in all default values to prevent user-configurable properties from being marked as unknown
  - These were causing an unnecessary update to be performed on import
- Update `cloudflare_worker_version` resource to prevent replacement on import
  - Allow in-place updates to write provider-only attributes (module content_file) to state
  - This attribute will always need to be updated on import (it only exists in the provider and won't we returned by the API)

## Additional context & links

Closes ticket raised through internal support.

## Acceptance test run output
```
CLOUDFLARE_ACCOUNT_ID=$CLOUDFLARE_ACCOUNT_ID CLOUDFLARE_API_TOKEN=$CLOUDFLARE_API_TOKEN TF_ACC=1 go test ./internal/services/worker -run "^TestAccCloudflareWorker_Basic" -v -count 1
=== RUN   TestAccCloudflareWorker_Basic
--- PASS: TestAccCloudflareWorker_Basic (8.11s)
PASS
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/services/worker    8.135s
```

```
$ CLOUDFLARE_ACCOUNT_ID=$CLOUDFLARE_ACCOUNT_ID CLOUDFLARE_API_TOKEN=$CLOUDFLARE_API_TOKEN TF_ACC=1 go test ./internal/services/worker_version -run "^TestAccCloudflareWorkerVersion_Basic" -v -count 1
=== RUN   TestAccCloudflareWorkerVersion_Basic
--- PASS: TestAccCloudflareWorkerVersion_Basic (14.38s)
PASS
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/services/worker_version    14.396s
```

### DIFF
--- a/internal/acctest/acctest.go
+++ b/internal/acctest/acctest.go
@@ -236,11 +236,6 @@ func SharedV1Client() (*cfv1.API, error) {
 // SharedClient returns a common Cloudflare V2 client setup needed for the
 // sweeper functions.
 func SharedClient() *cloudflare.Client {
-	if apiToken, ok := os.LookupEnv("CLOUDFLARE_API_TOKEN"); ok {
-		return cloudflare.NewClient(
-			option.WithAPIToken(apiToken),
-		)
-	}
 	return cloudflare.NewClient(
 		option.WithAPIKey(os.Getenv("CLOUDFLARE_API_KEY")),
 		option.WithAPIEmail(os.Getenv("CLOUDFLARE_EMAIL")),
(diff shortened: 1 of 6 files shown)


internal/acctest/acctest.go
```diff
@@ -236,6 +236,11 @@ func SharedV1Client() (*cfv1.API, error) {
 // SharedClient returns a common Cloudflare V2 client setup needed for the
 // sweeper functions.
 func SharedClient() *cloudflare.Client {
+	if apiToken, ok := os.LookupEnv("CLOUDFLARE_API_TOKEN"); ok {
```
tamas-jozsa: this doesn't make any sense here, because the next line is going to re-initialize the client with API Key and email.

tamas-jozsa:
can you attach the run result of your acceptance tests to the description?

internal/acctest/acctest.go
maxwellpeterson: What if I want to use an API token instead of API key and email? Shouldn't we support both?

maxwellpeterson:
@tamas-jozsa I dropped the changes to `internal/acctest/acctest.go` and no longer need any follow-up from your team

PR #6364, opened 2025-10-22
WDAPI-3261 fix drift in zero_trust_device_custom_profile

TylerStanish:

## Changes being requested
When configuring a zero_trust_device_custom_profile, various attributes drift even if they're unspecified, in particular: exclude, include, default, fallback_domains, gateway_unique_id, service_mode_v2, target_tests.

To fix this, this change:
- Adds UseStateForUnknown plan modifier for some computed attributes

### DIFF
--- a/internal/services/zero_trust_device_custom_profile/schema.go
+++ b/internal/services/zero_trust_device_custom_profile/schema.go
@@ -12,6 +12,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/float64default"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listplanmodifier"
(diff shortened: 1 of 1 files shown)


internal/services/zero_trust_device_custom_profile/schema.go
```diff
@@ -212,13 +217,21 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 			"default": schema.BoolAttribute{
 				Description: "Whether the policy is the default policy for an account.",
 				Computed:    true,
+				// Required and optional are both explicitly set to false to make clear this is a read-only attribute.
+				// Custom profiles will always be non-default profiles.
+				// Default profiles are managed through zero_trust_device_default_profile.
+				Required: false,
+				Optional: false,
+				Default:  booldefault.StaticBool(false),
```
TylerStanish: Would it be better for this to be configured in api-schemas rather than this PR?

tamas-jozsa: yes, please configure this in the schema instead

TylerStanish: Ok, removed

PR #6368, opened 2025-10-24
TUN-9900: Remove warp_routing from cloudflared_config

jcsf:
## Summary
The warp_routing configuration is deprecated and is supposed to be ignored on our API's, therefore, it shouldn't be stored in terraform. The only reason our API's still return it is for really old versions of cloudflared to not break.

## Testing
Run the acceptance tests and they passed.


jcsf:
Tests run
image

musa-cf:
This change was integrated with https://github.com/cloudflare/terraform-provider-cloudflare/pull/6471, so closing this out.

PR #6385, opened 2025-10-29
ci: fix acceptance tests

mgirouard:

### DIFF
--- a/internal/services/zero_trust_tunnel_cloudflared_virtual_network/resource_test.go
+++ b/internal/services/zero_trust_tunnel_cloudflared_virtual_network/resource_test.go
@@ -47,6 +47,7 @@ func testSweepCloudflareTunnelVirtualNetwork(r string) error {
 		context.Background(), zero_trust.NetworkVirtualNetworkListParams{})
 	if err != nil {
 		tflog.Error(ctx, fmt.Sprintf("Failed to fetch Cloudflare Tunnel Virtual Networks: %s", err))
+		return err
 	}
 
 	if len(tunnelVirtualNetworks.Result) == 0 {


internal/services/list/resource_test.go
```diff
@@ -620,14 +620,12 @@ func TestAccCloudflareList_ImportWithSeparateListItem(t *testing.T) {
 					return fmt.Sprintf("%s/%s", accountID, s.RootModule().Resources[resourceNameList].Primary.ID), nil
 				},
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"modified_on", "num_items"},
+				ImportStateVerifyIgnore: []string{"modified_on", "num_items", "items"},
```
vaishakdinesh: This is all you need i think. 
And with the above, this line will ignore items on import anyway

mgirouard: Unfortunately this doesn't seem to be enough:
```bash
mgirouard@sixteen:~/src/terraform-provider-cloudflare$ git diff
diff --git a/internal/services/list/resource_test.go b/internal/services/list/resource_test.go
index d6e9a5e4f0..9fb842d187 100644
--- a/internal/services/list/resource_test.go
+++ b/internal/services/list/resource_test.go
@@ -620,7 +620,7 @@ func TestAccCloudflareList_ImportWithSeparateListItem(t *testing.T) {
                                        return fmt.Sprintf("%s/%s", accountID, s.RootModule().Resources[resourceNameList].Primary.ID), nil
                                },
                                ImportStateVerify:       true,
-                               ImportStateVerifyIgnore: []string{"modified_on", "num_items"},
+                               ImportStateVerifyIgnore: []string{"modified_on", "num_items", "items"},
                                        resource.TestCheckResourceAttr(resourceNameList, "account_id", accountID),
                                        resource.TestCheckResourceAttr(resourceNameList, "name", listName),
```
```bash
mgirouard@sixteen:~/src/terraform-provider-cloudflare$ TF_ACC=1 go test ./internal/services/list/... -run ^TestAccCloudflareListWithItems_IP -v -count=1
=== RUN   TestAccCloudflareListWithItems_IP_Datasource
--- PASS: TestAccCloudflareListWithItems_IP_Datasource (9.93s)
=== RUN   TestAccCloudflareListWithItems_IP
    resource_test.go:357: Step 3/6 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.

          map[string]string{
        -       "items.#":    "2",
        -       "items.0.%":  "5",
        -       "items.0.ip": "1.1.1.1",
        -       "items.1.%":  "5",
        -       "items.1.ip": "1.1.1.2",
          }
--- FAIL: TestAccCloudflareListWithItems_IP (7.11s)
FAIL
FAIL    github.com/cloudflare/terraform-provider-cloudflare/internal/services/list      17.051s
FAIL
```

vaishakdinesh: hmmm. @broswen what do you think?

mgirouard: I'm fine with pulling this from my patch, but we'll need to get the acceptance tests running else we can't release. Tests run fine when I check out the previous release tag.

broswen: > This is all you need i think. And with the above, this line will ignore items on import anyway

It looks like you added it to a different test than what was failing?

`TestAccCloudflareList_ImportWithSeparateListItem` vs `TestAccCloudflareListWithItems_IP`


I wonder if there is a way to validate what ImportState is doing for `cloudflare_list` with the test framework.

We want to validate that imports do not load any list items into the nested `items` property. But on the next plan/read it will load them dynamically if `items` is set.

We'll probably have to use `ImportStateCheck` or `ImportPlanChecks`

mgirouard: Aye, fair. There were a few tests that were failing in the release PR and I just zoomed in on this one.
https://github.com/cloudflare/terraform-provider-cloudflare/actions/runs/18904266697/job/53958806404?pr=6292

broswen: I think the commits related to this change should be reverted. 
https://github.com/cloudflare/terraform-provider-cloudflare/pull/6262

Import is actually not working on all configurations for this resource.

mgirouard: ACK :+1: Will pull it out.

broswen: I don't think it's actually possible what I was trying to do there because `Read` doesn't give you access to the config state.

mgirouard: Removed my patch in ef07f5ea29...98723b3f87

mgirouard: Revert PR is here: https://github.com/cloudflare/terraform-provider-cloudflare/pull/6389

PR #6407, opened 2025-11-03
chore: adding boolean support for run_worker_first

christhorwarth:

## Changes being requested

This PR adds support for a worker version's run_worker_first property to be a boolean in addition to a list of strings.


vaishakdinesh:
@christhorwarth have you tested an upgrade scenario for this change? How would the state look once this is released?

christhorwarth:
> have you tested an upgrade scenario for this change? How would the state look once this is released?

I just added the TestAccCloudflareWorkerVersion_RunWorkerFirstUpgrade test to verify that transitions between all value types work correctly. Users with existing run_worker_first = [...] configurations will see no changes on first refresh after upgrading. They can then optionally switch to boolean syntax if desired.

vaishakdinesh (approved):

PR #6440, opened 2025-11-10
token policy order and resources

steve-thousand:

## Changes being requested

1. Fixing policy order by removing unnecessary computed fields and switching to a set
2. Changing resources to a string so that we can use jsonencode to represent more complex resource forms

- [X] I have run acceptance tests for my changes and included the results below 
```
TF_ACC=1 go test internal/services/account_token/resource_test.go
ok      command-line-arguments  180.053s
TF_ACC=1 go test internal/services/api_token/resource_test.go    
ok      command-line-arguments  278.908s
```

## Additional context & links

The changes to resources will result in a backwards-compatibility breaking change, but we were unable to find a better way to support polymorphic resources. Users who already have tokens will need to wrap their resources definition in a jsonencode function call.

Example previous config:

```
resources = {
  "com.cloudflare.api.account.ABC" = "*"
}
```

New conifg with added jsonencode

```
resources = jsonencode({
  "com.cloudflare.api.account.ABC" = "*"
})
```

We did also verify that removing the computed fields shouldn't cause any issues for users with existing state files.


internal/services/api_token/schema.go
```diff
@@ -57,29 +75,15 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 										Description: "Identifier of the permission group.",
 										Required:    true,
 									},
-									"meta": schema.SingleNestedAttribute{
```
vaishakdinesh, line 60: @steve-thousand do we need to remove this?

internal/services/account_token/schema.go
```diff
@@ -63,31 +80,15 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 										Description: "Identifier of the permission group.",
 										Required:    true,
 									},
-									"meta": schema.SingleNestedAttribute{
```
vaishakdinesh, line 66: @steve-thousand Do we need to remove this ?

internal/services/api_token/schema.go
steve-thousand: The reason this and name need to be removed is because they are computed fields that are interfering with the policy set's identify function which results in unexpected drift.

Furthermore, in the case of "meta", I don't think the meta field actually exists. It may exist in the case of account member and user group permission groups but not tokens.

internal/services/account_token/schema.go
steve-thousand: The reasoning here is the same as for API tokens https://github.com/cloudflare/terraform-provider-cloudflare/pull/6440#discussion_r2525222462

PR #6441, opened 2025-11-10
Migrate zero trust device posture rule

SirCortly:
A couple notes: 
* `name` attribute went from optional in V4 to required in V5. We will check the state during config transforms and use that value if it contains one, but otherwise we will not migrate the name attribute and the user will have to manually add the required name to their HCL. 
* The `input` attribute block is computed in V4 and defaults to empty strings. In V5 it defaults to null for attributes that are undefined in the HCL. For now we are going to migrate any empty values to null. If this causes issues it should be discovered in E2E testing. 

```
=== RUN   TestMigrateDevicePostureRuleBasic
--- PASS: TestMigrateDevicePostureRuleBasic (14.41s)
=== RUN   TestMigrateDevicePostureRuleWithoutName
    migrations_test.go:99: Skipping: v4 configs without 'name' field will need manual intervention during migration. Users must add the name from their state to the config for v5 compatibility. This is documented as a known migration limitation since TransformConfig cannot access state.
--- SKIP: TestMigrateDevicePostureRuleWithoutName (0.00s)
=== RUN   TestMigrateDevicePostureRuleFirewall
--- PASS: TestMigrateDevicePostureRuleFirewall (16.12s)
=== RUN   TestMigrateDevicePostureRuleDiskEncryption
--- PASS: TestMigrateDevicePostureRuleDiskEncryption (16.51s)
=== RUN   TestMigrateDevicePostureRuleMultiplePlatforms
--- PASS: TestMigrateDevicePostureRuleMultiplePlatforms (14.81s)
PASS
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/services/zero_trust_device_posture_rule    (cached)
```

### DIFF
--- a/internal/services/zero_trust_device_posture_rule/migrations_test.go
+++ b/internal/services/zero_trust_device_posture_rule/migrations_test.go
@@ -1,0 +1,365 @@
+package zero_trust_device_posture_rule_test
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/acctest"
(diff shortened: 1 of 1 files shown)


musa-cf:
tf-migrate PR: https://github.com/cloudflare/tf-migrate/pull/21

PR #6520, opened 2025-12-05
cloudflare_account_member refactor

steve-thousand:

## Changes being requested

This PR has a number of changes that are meant to address bugs with account_member Imports as well as drift issues around roles and policies.

### Resource Validation: ExactlyOneOf Roles or Policies

It is possible to pass both roles and policies to the API, but the API prefers policies and in that case will override the roles. This can cause unexpected errors if the API hands back roles different from what the user has configured. So, to reduce confusion we enforce one of either roles or policies, and error on both or neither.

### Changing resource Create function to perform POST + GET

There is a known bug in the POST endpoint for members whereby the response does not return roles. This causes part of our difficulties supporting roles, and some provider-side logic was attempting to fix this by simply preserving the state roles, which isn't great practice.

Roles may be removed in the future, and changing the endpoint is difficult. Instead, we can perform a second GET after the POST to retrieve the full correct state.

### Changing roles to a Set

Like other collection fields on policies, list order isn't guaranteed. To avoid unexpected drift, we can make roles into a set.

### Changing `ModifyPlan` to conditionally set Roles and Policies to "Unknown"

Roles vs Policies has been a problematic bit of the Members API. If the API caller provides roles, the server will hand back roles _and_ inferred policies. And If the API caller provides policies, the server will hand back policies _and_ roles. This is for backwards compatibility. But, the unknown response values can cause unintended drift or crashes in terraform.

Earlier fixes for this proved to be problematic. The new fix is to use the user's config to decide if they intend to change roles or policies, and if we can detect that change then we set the corresponding other collection to unknown (eg. if a user changes roles then the planned policies are set to unknown).

### DIFF
--- a/internal/services/account_member/custom.go
+++ b/internal/services/account_member/custom.go
@@ -3,23 +3,46 @@ package account_member
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/apijson"
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/customfield"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
+type ConfiguredPermissionType int
+
+const (
+	Unknown ConfiguredPermissionType = iota
+	Policies
+	Roles
+)
+
+// Given a config value, determine which permission type is configured. This
+// should only be called with the config value, never the plan or state value.
+func checkConfiguredPermissionType(configuredModel *AccountMemberModel) ConfiguredPermissionType {
+	permissionType := Unknown
+
+	if !configuredModel.Roles.IsNull() && len(configuredModel.Roles.Elements()) > 0 {
+		permissionType = Roles
+	}
+	if !configuredModel.Policies.IsNull() && len(configuredModel.Policies.Elements()) > 0 {
+		if permissionType == Roles {
+			// if both are found, return Unknown
+			return Unknown
+		}
+		return Policies
+	}
+	return permissionType
+}
+
 func (m AccountMemberModel) marshalCustom() (data []byte, err error) {
 	return apijson.MarshalRoot(m)
 }
 
-func (m AccountMemberModel) marshalCustomForUpdate(state AccountMemberModel) (data []byte, err error) {
+func (m AccountMemberModel) marshalCustomForUpdate(state AccountMemberModel, configuredPermissionType ConfiguredPermissionType) (data []byte, err error) {
 	data, err = apijson.MarshalForUpdate(m, state)
 	if err != nil {
-		return
-	}
-
-	if m.Roles == nil || len(*m.Roles) == 0 {
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


internal/services/account_member/model.go
```diff
@@ -17,7 +17,7 @@ type AccountMemberModel struct {
 	AccountID types.String                                            `tfsdk:"account_id" path:"account_id,required"`
 	Email     types.String                                            `tfsdk:"email" json:"email,required"`
 	Status    types.String                                            `tfsdk:"status" json:"status,computed_optional"`
-	Roles     *[]types.String                                         `tfsdk:"roles" json:"roles,optional,no_refresh"`
+	Roles     types.Set                                               `tfsdk:"roles" json:"roles,computed_optional,no_refresh"`
```
steve-thousand: This needed to change so that we can set its value to Unknown during the planning phase (can't be done with a native go slice)

internal/services/account_member/resource.go
```diff
@@ -262,27 +305,36 @@ func (r *AccountMemberResource) ModifyPlan(ctx context.Context, req resource.Mod
...
-
-	// Set the modified plan
-	resp.Diagnostics.Append(resp.Plan.Set(ctx, &plan)...)
+	if config != nil {
+		configurePermissionType := checkConfiguredPermissionType(config)
+		switch configurePermissionType {
+		case Roles:
+			if state == nil || !reflect.DeepEqual(plan.Roles, state.Roles) {
+				// if roles are changing, set policies to unknown
+				resp.Diagnostics.Append(resp.Plan.SetAttribute(ctx, path.Root("policies"), plan.Policies.UnknownValue(ctx))...)
```
steve-thousand: TBD: technically these "Unknown" cases should always be set. Every time we write the member back to the server, the server tries to rebuild roles/policies. Currently they won't ever change, but if we ever did change the server-side code that does that, then the values that come back will change which will cause terraform to crash (actual values != expected)

internal/services/account_member/custom.go
```diff
@@ -3,43 +3,80 @@ package account_member
...
+	}
+	return permissionType
+}
+
 func (m AccountMemberModel) marshalCustom() (data []byte, err error) {
 	return apijson.MarshalRoot(m)
 }
 
-func (m AccountMemberModel) marshalCustomForUpdate(state AccountMemberModel) (data []byte, err error) {
+func (m AccountMemberModel) marshalCustomForUpdate(state AccountMemberModel, configuredPermissionType ConfiguredPermissionType) (data []byte, err error) {
```
steve-thousand, line 43: Updated this function to check `configuredPermissionType` and correctly set/delete roles/policies based on that value. Also needed to update the serialization for roles now that it is a Set type.

internal/services/account_member/custom.go
```diff
@@ -51,75 +88,76 @@ func (m AccountMemberModel) marshalCustomForUpdate(state AccountMemberModel) (da
...
 		return nil, err
 	}
+	result := &env.Result
+
+	result.AccountID = configuredModel.AccountID
+
+	return parsePoliciesAndRoles(ctx, data, result)
+}
 
+func unmarshalComputedCustom(data []byte, configuredModel *AccountMemberModel) (*AccountMemberModel, error) {
```
steve-thousand, line 103: There are now two custom unmarshal functions, one for only computed values.

internal/services/account_member/resource_test.go
```diff
@@ -238,82 +253,96 @@ func TestAccCloudflareAccountMember_RolesUpdate(t *testing.T) {
...
+					statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("policies").AtSliceIndex(0).AtMapKey("permission_groups").AtSliceIndex(0).AtMapKey("id"), knownvalue.StringExact(adminPermissionGroupID)),
+					statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("policies").AtSliceIndex(0).AtMapKey("resource_groups"), knownvalue.ListSizeExact(1)),
 				},
 			},
-		},
-	})
-}
-
-// Test that using roles ignores auto-generated policies to avoid diffs
-func TestAccCloudflareAccountMember_RolesVsPolicies(t *testing.T) {
```
steve-thousand: I deleted this test because it didn't seem to be distinct at all from the test above

internal/services/account_member/resource_test.go
```diff
@@ -374,29 +464,40 @@ func TestAccCloudflareAccountMember_Policies(t *testing.T) {
...
+				ImportState:         true,
+				ImportStateVerify:   true,
+				ImportStateIdPrefix: fmt.Sprintf("%s/", accountID),
+			},
 		},
 	})
 }
 
 func TestAccCloudflareAccountMember_PoliciesAddResourceGroup(t *testing.T) {
+	// once this is enabled, the test may also be flaky due to caching issues
```
steve-thousand, line 502: I did comment out the skip here and run this test to ensure my changes work and I was unable to encounter any caching bugs

Issue #6526, opened 2025-12-07
Inconsistent results after apply for `cloudflare_pages_project` updates


FidelusAleksander:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

OpenTofu Version v1.10.7

Cloudflare provider version  ~> 5.14

### Affected resource(s)

cloudflare_pages_project updates produce inconsistent results after apply

> 
> cloudflare_pages_project.this: Modifying... 
> ╷
> │ Error: Provider produced inconsistent result after apply
> │ 
> │ When applying changes to cloudflare_pages_project.this, provider
> │ "provider[\"registry.opentofu.org/cloudflare/cloudflare\"]" produced an
> │ unexpected new value: .deployment_configs: inconsistent values for
> │ sensitive attribute.
> │ 
> │ This is a bug in the provider, which should be reported in the provider's
> │ own issue tracker.
> ╵

### Terraform configuration files

```hcl
terraform {
  required_version = ">= 1.10"

  required_providers {
    cloudflare = {
      source  = "cloudflare/cloudflare"
      version = "~> 5.14"
    }
  }
}

provider "cloudflare" {
}

resource "cloudflare_pages_project" "this" {
  account_id        = var.cloudflare_account_id
  name              = "xxxxx"
  production_branch = "master"
  build_config = {
    build_command       = "hugo"
    destination_dir     = "public"
    root_dir            = ""
    web_analytics_tag   = null
    web_analytics_token = null
  }
  deployment_configs = {
    preview = {
      always_use_latest_compatibility_date = false
      build_image_major_version            = 2
      compatibility_date                   = "2024-01-17"
      compatibility_flags                  = []
      env_vars = {
        HUGO_VERSION = {
          type  = "plain_text"
          value = "0.145.0"
        }
      }
      fail_open = true
    }
    production = {
      always_use_latest_compatibility_date = false
      build_image_major_version            = 2
      compatibility_date                   = "2024-01-17"
      compatibility_flags                  = []
      env_vars = {
        HUGO_VERSION = {
          type  = "plain_text"
          value = "0.145.0"
        }
      }
      fail_open = true
    }
  }
  source = {
    config = {
      owner                          = "XXXX"
[...]

Issue #5928, opened 2025-08-14
`cloudflare_pages_project` optional fields `build_config` & `deployment_configs` generate drift


domenikk:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.12.2
on darwin_arm64

Cloudflare provider version: 5.8.2

### Affected resource(s)

cloudflare_pages_project

### Terraform configuration files

```hcl
resource "cloudflare_pages_project" "this" {
  account_id = var.cf_account_id
  name       = var.cf_project_name

  production_branch = "main"
}
```

### Link to debug output

https://gist.github.com/domenikk/fbc6b4d55826855317b8b7fdd56ea99b

### Expected output

Both `build_config` and `deployment_configs` are listed as optional, so they can be ommited or set to `null`, according to this provider's documentation:

<img width="509" height="315" alt="Image" src="https://github.com/user-attachments/assets/f50dc625-7ae3-4067-8e4c-e6b7f5efd291" />

If I don't make any changes to the resource after applying, I expect it to not show in the next `terraform plan`.

### Actual output

Without any changes to the resource, the next `terraform plan` shows these actions to be performed:
```terraform
- build_config         = {} -> null
- deployment_configs   = {
  - preview    = {
      - compatibility_date  = "2025-08-05" -> null
      - compatibility_flags = *** -> null
    } -> null
  - production = {
      - compatibility_date  = "2025-08-05" -> null
      - compatibility_flags = *** -> null
    } -> null
} -> null
```

### Steps to reproduce

1. Write the resource block as mentioned above
2. `terraform apply`
3. `terraform plan`

### Additional factoids

I think the principle that a provider should not show any differences when there are no changes to the resource, is important and an expectation from most Terraform users.
We have automated workflows for drift detection and PR checks for the affected resources, that we can't rely on anymore because the V5 of this provider seems to have a lot of issues like this.

**Workaround**
I have found, through some trial and error, that using the values below, stop the provider from generating drift:
```terraform
build_config = {}
deployment_configs = {
  preview = {
    compatibility_date  = var.cf_compatibility_date
    compatibility_flags = []
  }
  production = {
    compatibility_date  = var.cf_compatibility_date
[...]

PR #6545, opened 2025-12-16
fix: addressing drift issues in cloudflare_pages_project

christhorwarth:

## Changes being requested

Fix "inconsistent result after apply" error for secret env vars (#6526) and prevent drift when omitting build_config/deployment_configs (#5928) in `cloudflare_pages_project`.

### DIFF
--- a/internal/services/pages_project/custom.go
+++ b/internal/services/pages_project/custom.go
@@ -5,12 +5,12 @@ import (
 
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/customfield"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
-// PreserveSecretEnvVars returns a copy of the destination deployment configs with
-// secret_text environment variable values carried over from the source deployment configs,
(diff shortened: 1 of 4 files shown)


mgirouard:
I think this is totally mergable but when I run the full test suite for this resource locally I get an error:

```
=== RUN   TestMigratePagesProject_V4ToV5_FullResource
    migration_test.go:480: Step 2/2 error: Pre-apply plan check(s) failed:
        expected empty plan except for falsey-to-null changes, but cloudflare_pages_project.cftftestwmmhtcvvob.build_config has change from map[build_caching:false build_command:npm run build destination_dir:dist root_dir: web_analytics_tag: web_analytics_token:] to map[build_command:npm run build destination_dir:dist]
--- FAIL: TestMigratePagesProject_V4ToV5_FullResource (17.01s)
=
```

I don't have a lot of context on this particular API's semantics but, I'm pretty sure the v4 -> v5 migration that was done previously is outdated -- so I don't think there's anything wrong with this PR. Will get a second opinion and will try to get this in the 5.15 release.

---

Edit: Going to merge this since it addresses reported issues. Migrations are different and we'll handle that separately.

PR #6449, opened 2025-11-11
chore(zone_settings): document configurations and examples (non-exhaustive)

vaishakdinesh:
- [x] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested
Add documentation and examples for zone settings. Note: This is not an exhaustive set.

## Acceptance test run results

NA
- [x] I have run acceptance tests for my changes and included the results below

PR #6605, opened 2025-12-31
update workers cron trigger example with schedules

rzhade3:

## Changes being requested

The docs for the `cloudflare_workers_cron_trigger ` resource were incorrectly displaying an example using the `body` attribute. The correct attribute that is used here is `schedules`. 

N/A, just a docs change

### Test output

N/A, just a docs change

### DIFF
--- a/docs/resources/workers_cron_trigger.md
+++ b/docs/resources/workers_cron_trigger.md
@@ -15,7 +15,7 @@ description: |-
 resource "cloudflare_workers_cron_trigger" "example_workers_cron_trigger" {
   account_id = "023e105f4ecef8ad9ca31a8372d0c353"
   script_name = "this-is_my_script-01"
-  body = [{
+  schedules = [{
     cron = "*/30 * * * *"
   }]
 }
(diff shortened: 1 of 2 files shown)


docs/resources/workers_cron_trigger.md
```diff
@@ -15,7 +15,7 @@ description: |-
 resource "cloudflare_workers_cron_trigger" "example_workers_cron_trigger" {
   account_id = "023e105f4ecef8ad9ca31a8372d0c353"
   script_name = "this-is_my_script-01"
-  body = [{
+  schedules = [{
```
mgirouard, line 18: This *should* be automatic if your OpenAPI definition includes this definition.

Can you confirm this is the case? Or is this particular resource all custom code?

mgirouard:
@rzhade3 See https://github.com/cloudflare/terraform-provider-cloudflare/pull/6449/files for an example for how to do this in a way such that it won't be overwritten when we regenerate the docs.

docs/resources/workers_cron_trigger.md
rzhade3: It's not entirely clear to me how the documentation is generated, because the autogenerated Model seems to use the [`schedules` parameter](https://github.com/rzhade3/terraform-provider-cloudflare/blob/374f1c12c308a40bd8cd3a742faa9aaf4fd4e36a/internal/services/workers_cron_trigger/model.go#L18) in code, but not in the documentation here.

mgirouard:
cc @1000hz

vaishakdinesh:
closing due to inactivity.

PR #6609, opened 2026-01-02
[HOLD]add checks for duplicate resource group

mya-jin:
**Issue**
The prod_api_call.txt  shows that the Cloudflare API correctly returns only 3 unique resource groups. However, the terraform_plan.txt shows that the data.cloudflare_resource_groups data source is returning a list where those same groups are repeated multiple times.

Cause of issue
In the account_member.tf, we have a hardcoded mapping:
```
# Nebula Prd
"4e5a271b3668903c7010afb65b7e958d" = {
  nebula_admin = "b67c066acc8348a2a6163608cec6c914"
  release_r2   = "null"
}
```
When Terraform runs the plan for cloudflare_account_member resource, it follows the steps:

It looks at the this mapping and sees the aim of assigning ID "b67..."
The Provider code uses acc = append(acc, env.Result.Elements()...). This command is "blind"—it doesn't check for existing IDs; it just keeps stacking items on top of each other.
For each matches found, it keeps adding to this acc list, which results in the duplication

**Output after running terraform plan:**
```
Changes to Outputs:
  + all_resource_group_names = [
      + "Nebula",
      + "Nebula",
      + "Nebula",
      + "Nebula",
      + "Release",
      + "Account-Group",
    ]
  + data_source_result_count = 6
```

**Solution:**
A "Clean Sweep" deduplication block was added to the pagination loop:

Logic: Uses a map[string]bool (seenIDs) to track unique Resource Group IDs.
Technical Execution: Iterates through the raw acc slice, type-asserts the attr.Value elements to types.Object, and uses basetypes.ObjectAsOptions to safely extract the model for ID validation.
Output: Reconstructs the final list using only unique elements before committing to the Terraform state.

**Test Output:**
```
Changes to Outputs:
  + all_resource_group_names = [
      + "Nebula",
      + "Release",
      + "Account-Group",
    ]
  + data_source_result_count = 3
```

### DIFF
--- a/internal/services/resource_group/list_data_source.go
+++ b/internal/services/resource_group/list_data_source.go
@@ -11,6 +11,8 @@ import (
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/customfield"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 )
 
 type ResourceGroupsDataSource struct {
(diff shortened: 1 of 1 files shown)


internal/services/resource_group/list_data_source.go
```diff
@@ -91,6 +93,34 @@ func (d *ResourceGroupsDataSource) Read(ctx context.Context, req datasource.Read
 		}
 	}
 
+    uniqueAcc := []attr.Value{}
+    seenIDs := make(map[string]bool)
```
mgirouard, line 97: What sort of upper-bound (size wise) do you think this map can grow to? This could be micro-optimized to a `map[string]struct{}`.

I'd probably do this only if there are potentially thousands of keys -- else it's probably not worth it.

mgirouard (approved):

musa-cf (approved):

PR #6624, opened 2026-01-07
draft: map managed by to unit in unmarshal for legacy compatibility

mya-jin:
## Issue: 
1. The cloudflare_account API does not consistently return the unit object in its response, instead defaulting to `managed_by.parent_org_id` structure. This caused a critical failure in the Terraform Plugin Framework during resource creation and updates, as Terraform expects every value in the "Planned" state to be saved exactly as-is in the "Response" state. Because the API returned `"unit": null` while the plan had a specific `unit.id`, the provider threw an error - _"Error: Provider produced inconsistent result after apply"_ and failed the apply.

2. Import Drift: Standard unmarshaling was stripping the account type during import steps because the `configuredModel `was empty, leading to `ImportStateVerify` failures.

## Solution:
Added logic to unmarshalCustom that detects if a unit.id is expected but missing from the API response. If so, it extracts the parent_org_id from the managed_by object and populates unit.id manually.

## Test:
Successfully ran acceptance tests, confirming that the ImportStateVerify attributes are now equivalent and the "inconsistent result" error is resolved.

### Test output
1. TestAccCloudflareAccount_Basic
```
mya@G9JRYWDQ25 terraform-provider-cloudflare % go test ./internal/services/account -run "^TestAccCloudflareAccount_Basic" -v -count 1 timeout=30m -parallel=1
=== RUN   TestAccCloudflareAccount_Basic
--- PASS: TestAccCloudflareAccount_Basic (6.00s)
PASS
ok  github.com/cloudflare/terraform-provider-cloudflare/internal/services/account 7.039s
```
2. TestAccCloudflareAccount_2FAEnforced
```
mya@G9JRYWDQ25 terraform-provider-cloudflare % go test ./internal/services/account -run "^TestAccCloudflareAccount_2FAEnforced" -v -count 1 timeout=30m -parallel=1
=== RUN   TestAccCloudflareAccount_2FAEnforced
--- PASS: TestAccCloudflareAccount_2FAEnforced (5.41s)
PASS
ok  github.com/cloudflare/terraform-provider-cloudflare/internal/services/account 6.477s 
```
3. TestAccCloudflareAccount_WithMulti
```
mya@G9JRYWDQ25 terraform-provider-cloudflare % go test ./internal/services/account -run "^TestAccCloudflareAccount_WithMulti" -v -count 1 timeout=30m -parallel=1  
=== RUN   TestAccCloudflareAccount_WithMulti
--- PASS: TestAccCloudflareAccount_WithMulti (7.05s)
PASS
ok  github.com/cloudflare/terraform-provider-cloudflare/internal/services/account 8.071s 
```

4. TestAccCloudflareAccount_WithUnit
```
mya@G9JRYWDQ25 terraform-provider-cloudflare % go test ./internal/services/account -run "^TestAccCloudflareAccount_WithUnit" -v -count 1 timeout=30m -parallel=1  
=== RUN   TestAccCloudflareAccount_WithUnit
--- PASS: TestAccCloudflareAccount_WithUnit (2.50s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/services/account	4.009s
```


mgirouard:
> The cloudflare_account API does not consistently return the unit object in its response, instead defaulting to managed_by.parent_org_id structure.

Is it possible to fix the API instead of adding custom code to the auto-generated code?

internal/services/account/custom.go
```diff
@@ -33,5 +36,20 @@ func unmarshalCustom(bytes []byte, configuredModel *AccountModel) (*AccountModel
 	}
 	// Setting type to whatever the configured type is to avoid state/drift issues
 	env.Result.Type = configuredModel.Type
+
+	if configuredModel.Unit != nil && !configuredModel.Unit.ID.IsNull() {
```
steve-thousand: It seems we are only reading the response's unit ID if the resource is configured that way, but I think we also need to do it if the API has it, regardless of whether or not the configured model expects it.

I am thinking of this workflow:

1. User creates an account through terraform
2. User or some other user adds account to a tenant through the API or dashboard
3. User tries to plan or apply through terraform: they need to see the unit ID as drift

aadithpm: +1, there are already multiple ways this association (self-serve / by CF) can be created post account provisioning so makes sense to me to support this.

mya-jin:
> > The cloudflare_account API does not consistently return the unit object in its response, instead defaulting to managed_by.parent_org_id structure.
> 
> Is it possible to fix the API instead of adding custom code to the auto-generated code?

We have discussed this in our team before. https://gitlab.cfdata.org/cloudflare/www/cf-www/-/merge_requests/16497

So instead of making a duplicate field which might cause confusion to the user, the custom code would be a better option.

internal/services/account/custom.go
mya-jin: Thanks for bringing this up, new checks has been posted.

internal/services/account/resource_test.go
```diff
@@ -113,18 +114,15 @@ func TestAccCloudflareAccount_Basic(t *testing.T) {
 			},
 			// Import step
 			{
-				ResourceName:      resourceName,
-				ImportState:       true,
-				ImportStateVerify: true,
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"type"},
```
vaishakdinesh, line 120: Would the user need to know the type if they do an import? Why is the `configuredModel` empty?

internal/services/account/custom.go
```diff
@@ -33,5 +36,24 @@ func unmarshalCustom(bytes []byte, configuredModel *AccountModel) (*AccountModel
...
+	if !env.Result.ManagedBy.IsNull() && !env.Result.ManagedBy.IsUnknown() {
+		var managedBy AccountManagedByModel
+		diag := env.Result.ManagedBy.As(context.Background(), &managedBy, basetypes.ObjectAsOptions{})
+
+		if !diag.HasError() {
+			// Backfill unit from managed_by if user expects it OR if we are in discovery/import mode
+			// This allows drift detection while preventing 'Inconsistent Result' panics during Apply
+			if env.Result.Unit == nil || env.Result.Unit.ID.IsNull() || env.Result.Unit.ID.IsUnknown() {
+				unitInConfig := configuredModel.Unit != nil && !configuredModel.Unit.ID.IsNull()
+				if (unitInConfig) || configuredModel.ID.IsNull() {
```
vaishakdinesh: `nit`:  the parentheses around unitInConfig isn't really required and bool vars are better named `isUnitInConfig`. Not a big deal and not a blocker if there are no other changes required.

internal/services/account/resource_test.go
mya-jin: I think during a standard terraform import, the user or the test provides an ID but no actual configuration logic for the attributes yet. Because the HCL block for the resource is empty, the framework passes a nil or empty configuredModel to the unmarshalCustom function.

mya-jin: And looking at this code block - https://github.com/cloudflare/terraform-provider-cloudflare/blob/main/internal/services/account/resource.go#L205-L209, we are only asking the user for one thing: the account_id

internal/services/account/custom.go
```diff
@@ -33,5 +36,24 @@ func unmarshalCustom(bytes []byte, configuredModel *AccountModel) (*AccountModel
...
 	env.Result.Type = configuredModel.Type
+
+	if !env.Result.ManagedBy.IsNull() && !env.Result.ManagedBy.IsUnknown() {
+		var managedBy AccountManagedByModel
+		diag := env.Result.ManagedBy.As(context.Background(), &managedBy, basetypes.ObjectAsOptions{})
+
+		if !diag.HasError() {
+			// Backfill unit from managed_by if user expects it OR if we are in discovery/import mode
+			// This allows drift detection while preventing 'Inconsistent Result' panics during Apply
+			if env.Result.Unit == nil || env.Result.Unit.ID.IsNull() || env.Result.Unit.ID.IsUnknown() {
```
steve-thousand: I don't think the unit ID would ever be unknown at this point

internal/services/account/custom.go
```diff
@@ -33,5 +36,24 @@ func unmarshalCustom(bytes []byte, configuredModel *AccountModel) (*AccountModel
...
+				unitInConfig := configuredModel.Unit != nil && !configuredModel.Unit.ID.IsNull()
+				if (unitInConfig) || configuredModel.ID.IsNull() {
+					env.Result.Unit = &AccountUnitModel{
+						ID: managedBy.ParentOrgID,
```
steve-thousand: Don't we want to set this value regardless of whether or not it's in the config? I thought a possible workflow was this:

1. User configures a new account (or imports one) without a unit id
2. Someone adds account to a tenant/organization on the backend
3. User runs a plan

Don't we want the user to see that the remote state has changed and has a unit ID now? I would expect you wouldn't need the unitInConfig checks at all

mya-jin: Actually we couldn't set this value regardless of whether or not it's in the config? As if we map the value when we create an account without setting unit.id, it will give "Inconsistent Result" as seeing null in config, but a value it returned - likely the same reason why we need to map unit.id with managed_by when we create an account with unit.id.

I will update the logic to only allow mapping of the unit during Read or Import operations. This allows the drift detection workflow to work for existing accounts while keeping the initial creation stable.

steve-thousand: If you create an account without a unit ID, does the server assign one?

aadithpm: > If you create an account without a unit ID, does the server assign one?

An account can be created only by an organization admin - and without an unit ID only in the instance where the user creating the account administers a single root organization. So, yes it can, in a specific cirscumstance, otherwise, a unit ID would be required.

mya-jin: > > If you create an account without a unit ID, does the server assign one?
> 
> An account can be created only by an organization admin - and without an unit ID only in the instance where the user creating the account administers a single root organization. So, yes it can, in a specific cirscumstance, otherwise, a unit ID would be required.

Agreed. This is exactly what’s happening in our testing environment. When we create the account without a unit block using a tenant account, the API returns a parent_org_id immediately.

internal/services/account/resource_test.go
vaishakdinesh: Ok. So the model looks like
```
type AccountModel struct {
	ID        types.String                                    `tfsdk:"id" json:"id,computed"`
	Unit      *AccountUnitModel                               `tfsdk:"unit" json:"unit,optional,no_refresh"`
	Name      types.String                                    `tfsdk:"name" json:"name,required"`
	Type      types.String                                    `tfsdk:"type" json:"type,optional"`
	ManagedBy customfield.NestedObject[AccountManagedByModel] `tfsdk:"managed_by" json:"managed_by,computed_optional"`
	Settings  customfield.NestedObject[AccountSettingsModel]  `tfsdk:"settings" json:"settings,computed_optional"`
	CreatedOn timetypes.RFC3339                               `tfsdk:"created_on" json:"created_on,computed" format:"date-time"`
}
```

Unit has `no_refresh` tag. So this will be ignored in the import.

mya-jin: Well, I think normally, that would prevent the Unit from being populated during a refresh or import.

However, because we are using `unMarshalCustom`, we are manually injecting the `parent_org_id` into the `Unit` model during the unmarshalling phase. This allows us to support:
- our manual mapping during the import's Read phase ensures the Unit ID is captured.
- ensure that we only "refresh" this value by adding the unit block from the user's HCL.

PR #6636, opened 2026-01-13
docs(api_shield): add jwt session identifier documentation

mattrighetti:


ghost:
lgtm

internal/services/api_shield/schema.go
```diff
@@ -33,7 +34,7 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				NestedObject: schema.NestedAttributeObject{
 					Attributes: map[string]schema.Attribute{
 						"name": schema.StringAttribute{
-							Description: "The name of the characteristic field, i.e., the header or cookie name.",
+							Description: "The name of the characteristic field, i.e., the header or cookie name. When using type \"jwt\", this must be a claim location expressed as `$(token_config_id):$(json_path)`, where `token_config_id` is the ID of the token configuration used in validating the JWT, and `json_path` is a RFC 9535 JSONPath expression.",
```
musa-cf, line 37: The attribute description and resource description in the `schema.go` file can be upstreamed into OpenAPI instead. That means this useful context will also be present in all of the SDKs and the API docs too.

mattrighetti: Will do, should I remove it from this PR in the meantime?

mgirouard: It's probably fine to leave it. When codegen runs again it *might*  conflict -- but the code would likely be the same.

musa-cf (approved):

musa-cf:
You'll need to branch off `next` and target `next`, but the changes look good. We'll merge it once the branch target is fixed.

mattrighetti:
Should look good now

PR #6652, opened 2026-01-15
provides PlanModifiers for web_analytics_site resource schema

MildMax:
provides PlanModifiers for web_analytics_site resource schema to import state without overwriting existing fields

### DIFF
--- a/internal/services/web_analytics_site/resource_test.go
+++ b/internal/services/web_analytics_site/resource_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/utils"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 )
 
@@ -71,7 +72,7 @@ func testSweepCloudflareWebAnalyticsSites(r string) error {
 	return nil
 }
 
-func TestAccCloudflareWebAnalyticsSite_Create(t *testing.T) {
+func TestAccCloudflareWebAnalyticsSite_Create_ImportState(t *testing.T) {
 	t.Parallel()
 	accountID := os.Getenv("CLOUDFLARE_ACCOUNT_ID")
 	domain := os.Getenv("CLOUDFLARE_DOMAIN")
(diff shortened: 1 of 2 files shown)


vaishakdinesh:
@MildMax can you target next instead of main

internal/services/web_analytics_site/resource_test.go
```diff
@@ -91,9 +92,19 @@ func TestAccCloudflareWebAnalyticsSite_Create(t *testing.T) {
 					resource.TestCheckResourceAttr(name, "host", domain),
 					resource.TestCheckResourceAttr(name, "auto_install", "false"),
 					resource.TestCheckResourceAttrSet(name, "site_token"),
-					resource.TestCheckResourceAttrSet(name, "snippet"),
```
vaishakdinesh, line 94: Was this intentional to skip this assertion?

internal/services/web_analytics_site/schema.go
```diff
@@ -35,93 +39,113 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
-				Optional:    true,
+				Description:   "If enabled, the JavaScript snippet is automatically injected for orange-clouded sites.",
+				Optional:      true,
+				PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()},
 			},
 			"enabled": schema.BoolAttribute{
-				Description: "Enables or disables RUM. This option can be used only when auto_install is set to true.",
-				Optional:    true,
+				Description:   "Enables or disables RUM. This option can be used only when auto_install is set to true.",
+				Optional:      true,
```
vaishakdinesh, line 48: I would add test cases, if it doesn't exists already, whether the plan modifier is working as expected and actual updates to the attribute are taking place and not resulting in stale values.

For optional fields. Computed and computed optionals _should_ be fine

internal/services/web_analytics_site/resource_test.go
MildMax: Yes, the RUM API does not provide this value from the POST /accounts/{account_id}/rum/site_info endpoint -- I'm not sure when this was changed, but even in main, this test fails on this assertion.

internal/services/web_analytics_site/schema.go
MildMax: So this change is strictly for imports -- without these plan modifiers, import statements will attempt to overwrite and change the values that are imported from the API, but if a zone is importing their web_analytics_site configuration, we don't want these overwritten. Is there a better way to accomplish this without these plan modifiers?? 

edit:: I discovered that when any optional fields are provided in an import, our provider will attempt to re-create all optional + read-only values. Is there a better way to prevent our provider from overwriting optional/read-only fields during import if the API already provides them?

PR #6686, opened 2026-01-23
feat: state upgrader

vaishakdinesh:
related tf-migrate PR: https://github.com/cloudflare/tf-migrate/pull/125 

**What changed:**
Added provider-side state migration infrastructure for 3 resources transitioning from SDKv2 (v4) to Plugin Framework (v5).

Resources migrated:
- 🔄 dns_record - renamed resource + state transformations
- 🔄 page_rule - complex JSON transformations
- 🔄 load_balancer_pool - nested object transformations

New infrastructure (per resource):  migration/v500/
├── handler.go     - StateUpgrader functions
├── model.go       - Source (v4) and target (v5) models
├── schema.go      - v4 schema definition for parsing old state
├── transform.go   - State transformation logic
├── migrations_test.go - Migration tests (v4→v5)
└── testdata/      - Test configs (v4 + v5)

**Key features:**
- ✅ Controlled rollout: TF_MIG_TEST=1 enables migrations during development
- ✅ Automatic migration: StateUpgraders run during terraform apply (no manual steps)
- ✅ Comprehensive tests: Dual test cases validate both v4→v5 and v5→v5 upgrades
- ✅ Moved old tests: Deleted 1,448 lines of old migration tests (consolidated into  new structure)

**Notable transformations:**
- load_balancer_pool: SDK v2 TypeList MaxItems:1 (arrays) → Plugin Framework SingleNestedAttribute (objects)
- dns_record: Resource rename + MoveState implementation
- page_rule: Complex nested JSON structure changes

Result: +3,687 lines of migration infrastructure that enables seamless v4→v5 upgrades for users.


internal/acctest/acctest.go
```diff
@@ -1166,8 +1160,14 @@ func MigrationV2TestStep(t *testing.T, v4Config string, tmpDir string, exactVers
 	return resource.TestStep{
 		PreConfig: func() {
 			WriteOutConfig(t, v4Config, tmpDir)
-			debugLogf(t, "Running migration command for version: %s (%s -> %s)", exactVersion, sourceVersion, targetVersion)
-			RunMigrationV2Command(t, v4Config, tmpDir, sourceVersion, targetVersion)
+			// Only run migration command if source version is v4
+			// v5→v5 "migrations" only test StateUpgrader idempotency, no config changes needed
+			if sourceVersion == "v4" {
```
vaishakdinesh: I think we don't need this? `tf-migrate` _should_ just passover or i think we should build that into `tf-migrate`, thoughts?

internal/services/dns_record/migrations.go
```diff
@@ -1,15 +1,56 @@
...
+			StateMover:   _00.MoveState,
+		},
+	}
+}
+
+// UpgradeState handles schema version upgrades for cloudflare_dns_record.
+// This is triggered when users manually run `terraform state mv` (Terraform < 1.8).
 func (r *DNSRecordResource) UpgradeState(ctx context.Context) map[int64]resource.StateUpgrader {
-	return map[int64]resource.StateUpgrader{}
+	sourceSchema := _00.SourceCloudflareRecordSchema()
```
vaishakdinesh: `_00.`?

vaishakdinesh: Does this compile? Can maybe use a pkg alias ?

internal/services/dns_record/schema.go
```diff
@@ -27,6 +28,7 @@ var _ resource.ResourceWithConfigValidators = (*DNSRecordResource)(nil)
 
 func ResourceSchema(ctx context.Context) schema.Schema {
 	return schema.Schema{
+		Version: acctest.GetSchemaVersion(0, 500),
```
vaishakdinesh: This will be called outside the context of the tests. Lets keep its out of the `acctest` pkg? I know its just temporary but we will have a release where this function is gonna be called. 

This might cause merge conflicts, fml. But thats true for a int64 or even a func that returns int64. But a problem for another day.

internal/acctest/acctest.go
```diff
@@ -247,6 +247,13 @@ func SharedClient() *cloudflare.Client {
 	)
 }
 
+func GetSchemaVersion(preMigration, postMigration int64) int64 {
+	if os.Getenv("TF_ACC") == "" {
+		return preMigration
+	}
+	return postMigration
+}
+
```
vaishakdinesh: `nit`: the names might be a little confusing? Its really not pre or post migration the versions we're returning. Rather the intended use is to return right schema version if its not a test.

Thinking about it. more, i'm not sure how normal acceptance might behave. Should be fine since we have no-op upgrader in the v5 series of releases.

internal/services/page_rule/migrations.go
```diff
@@ -6,10 +6,35 @@ import (
...
+			PriorSchema:   nil,
+			StateUpgrader: v500.UpgradeFromV0,
+		},
+		// Migration: legacy v4.x SDKv2 provider with schema_version=3 (if explicitly set)
+		3: {
+			PriorSchema:   &sourceSchema,
+			StateUpgrader: v500.UpgradeFromLegacyV3,
+		},
+		// No-op: released version bump (v5.0.0-v5.16.0 releases that set Version: 4)
+		4: {
```
vaishakdinesh: Dont think we need this?
V4 last schema version was 3 and v5 is 0.

vaishakdinesh:
Note: switch branch to next. Probably needs a rebase. 
@tamas-jozsa I can handle that

internal/services/dns_record/migration/v500/handler.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+// UpgradeFromV0 handles state upgrades from earlier v500 versions (schema_version=0) to current v500.
+// This is a no-op upgrade since the schema is compatible - just copy state through.
+//
+// Why this exists: Terraform requires explicit upgraders to be defined for version tracking,
```
mgirouard: nit: "Why this exists:" is just fluff. the rest is solid

internal/services/dns_record/migration/v500/migrations_test.go
```diff
@@ -0,0 +1,891 @@
...
+// - currentProviderVersion: Current v5 release (auto-updates with releases)
+//
+// Based on breaking changes analysis:
+// - All breaking changes happened between 4.x and 5.0.0
+// - No breaking changes between v5 releases (testing against latest v5)
+// - Key changes: cloudflare_record → cloudflare_dns_record, data block → attribute
+
+// Embed migration test configuration files
+//
+//go:embed testdata/v4_a_record.tf
```
mgirouard, line 38: oh this is a very welcomed pattern addition

internal/migrations/schema_version.go
mgirouard, line 1: There's always a better name than `utils` -- what about `migrations`?

internal/version.go
```diff
@@ -3,3 +3,7 @@
 package internal
 
 const PackageVersion = "5.16.0" // x-release-please-version
+
+// LastV4Version is the last stable v4 release before v5 breaking changes.
+// This is used in migration tests across all resources to validate the v4 → v5 upgrade path.
+const LastV4Version = "4.52.5"
```
mgirouard: what maintains this? perhaps a ci test that checks last v4.x versions?

scripts/run-ci-tests
mgirouard, line 1: i forget how exactly the split of previous migration tests vs new migration tests will work. will they both coexist? or is it an opt-in type of thing?

scripts/run-ci-tests
```diff
@@ -576,35 +586,49 @@ run_tests() {
...
-    
+
+    log "${WHITE}Running $test_name: $test_path${NC}" > "$log_file"
+    log "Running $test_name: $test_path"
+
+    # Build command with optional TF_MIGRATE_BINARY_PATH for migration tests
+    local env_prefix=""
+    if [ "$TEST_TYPE" = "migration" ]; then
+        # Set TF_MIGRATE_BINARY_PATH for migration tests
+        local tf_migrate_path="${TF_MIGRATE_BINARY_PATH:-/Users/tjozsa/cf-repos/sdks/migration-work/agent-a/tf-migrate/bin/tf-migrate}"
```
mgirouard: we might need a better default here :stuck_out_tongue_closed_eyes:

scripts/run-ci-tests
tamas-jozsa: We don't care about the outcome of previous migration tests, because even though they pass, they are incorrect, because they are not using the correct state upgraders. Since now that we are looking for migration tests under migration/ subdirectory, CI will only find the new migration tests.

scripts/run-ci-tests
tamas-jozsa: uf good call!

internal/version.go
tamas-jozsa: good call, I moved this to an env var

PR #6690, opened 2026-01-26
fix(resources): handle empty/null/unknown IDs in Read methods

aditmeno:
## Summary

This PR addresses issues where Terraform/OpenTofu runs refresh before create for new resources, causing API errors when trying to read non-existent resources.

### Problem

When using `tofu apply` (or `terraform apply`) with the `-refresh=true` flag (which is the default), Terraform attempts to refresh the state of all resources before making changes. For resources that don't yet exist (i.e., resources being created for the first time), the `Read` method is called with an empty/null/unknown ID. Without proper handling, this causes the provider to make API calls with empty identifiers, resulting in errors.

### Solution

- **Add early return checks in `Read` methods for 31 resources** to gracefully handle null/unknown/empty resource identifiers by calling `resp.State.RemoveResource()`. This signals to Terraform that the resource needs to be created.

- **Fix `calls_sfu_app`**: Consolidate `app_id` and `uid` fields (they were duplicates) by mapping the API's `uid` response to `app_id` with proper `Computed+Optional` schema and `UseStateForUnknown` plan modifier.

- **Fix `zero_trust_access_service_token`**: Replace `AlsoRequires` validator with a custom `ConfigValidator` that only requires `previous_client_secret_expires_at` when `client_secret_version > 1` (during rotation). The previous validator was problematic because `client_secret_version` has a default value of `1`, so it's always set, incorrectly triggering the validator on initial creation.

### Affected Resources

- account, api_shield_operation, api_shield_schema, calls_sfu_app, calls_turn_app
- connectivity_directory_service, d1_database, list_item, notification_policy
- queue, queue_consumer, schema_validation_schemas, stream, stream_live_input
- stream_watermark, turnstile_widget, web_analytics_site, worker
- zero_trust_access_application, zero_trust_access_custom_page
- zero_trust_access_service_token, zero_trust_device_custom_profile
- zero_trust_device_managed_networks, zero_trust_dex_test, zero_trust_dlp_dataset
- zero_trust_dns_location, zero_trust_tunnel_cloudflared, zone

## Test Plan

- [ ] Verify `tofu plan` / `terraform plan` works for new resources without errors
- [ ] Verify `tofu apply` / `terraform apply` successfully creates resources
- [ ] Verify existing resources continue to work (refresh/update/delete)
- [ ] Test `calls_sfu_app` resource creation and verify `app_id` is properly populated
- [ ] Test `zero_trust_access_service_token` initial creation without `previous_client_secret_expires_at`
- [ ] Test `zero_trust_access_service_token` rotation with `client_secret_version > 1` requires `previous_client_secret_expires_at`

### DIFF
--- a/internal/services/account/resource.go
+++ b/internal/services/account/resource.go
@@ -146,6 +146,14 @@ func (r *AccountResource) Read(ctx context.Context, req resource.ReadRequest, re
 		return
 	}
 
+	// If ID is empty, null, or unknown, the resource doesn't exist yet.
+	// This occurs when terraform/tofu runs refresh before create for new resources.
+	// Return early to signal that the resource needs to be created.
+	if data.ID.IsNull() || data.ID.IsUnknown() || data.ID.ValueString() == "" {
+		resp.State.RemoveResource(ctx)
+		return
+	}
+
 	res := new(http.Response)
 	_, err := r.client.Accounts.Get(
 		ctx,
--- a/internal/services/api_shield_operation/resource.go
+++ b/internal/services/api_shield_operation/resource.go
@@ -109,6 +109,14 @@ func (r *APIShieldOperationResource) Read(ctx context.Context, req resource.Read
 		return
 	}
 
+	// If OperationID is empty, null, or unknown, the resource doesn't exist yet.
+	// This occurs when terraform/tofu runs refresh before create for new resources.
+	// Return early to signal that the resource needs to be created.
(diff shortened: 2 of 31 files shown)


vaishakdinesh:
@aditmeno can you create a github issue for the problems you're facing and we can get it resolved? I'd like to understand the exact issues you're facing and then see how to best tackle them

vaishakdinesh:
@aditmeno closing this in favor for the github issue. Thank you!

PR #6707, opened 2026-01-30
fix(apijson): use 512-bit precision for Float64 to match HCL parsing

maxwellpeterson:
### The Symptom

When importing an existing `cloudflare_worker` resource with `observability.logs.head_sampling_rate = 0.0001`, Terraform would always report that the resource needed to be updated, even when the configuration exactly matched the imported state. The plan output showed:

```hcl
~ resource "cloudflare_worker" "riot_games_redirector" {
    ...
    ~ references     = { ... } -> (known after apply)
    ~ updated_on     = "2026-01-30T20:19:15Z" -> (known after apply)
  }
```

### The Root Cause

The issue was a precision mismatch in how `Float64Value` objects were created between two code paths:

| Source | Creation Method | Internal `big.Float` Precision |
|--------|-----------------|-------------------------------|
| HCL Config (user's `.tf` file) | `big.ParseFloat(s, 10, 512, big.ToNearestEven)` | 512-bit |
| API Response (via `apijson.Unmarshal`) | `types.Float64Value(float64)` → `big.NewFloat(float64)` | 53-bit |

When Terraform compared the imported state (from API) against the proposed new state (from HCL config), it used `Float64Value.Equal()`, which internally calls `big.Float.Cmp()`. Two `big.Float` values with the same numeric value but different precisions have different mantissa representations, causing `Cmp()` to return non-zero.

### The Chain of Events

1. **Import**: Terraform fetches the worker from the Cloudflare API
2. **Decode**: `apijson.Unmarshal` creates `Float64Value` with 53-bit precision for `head_sampling_rate = 0.0001`
3. **Plan**: Terraform parses the HCL config and creates `Float64Value` with 512-bit precision for the same value
4. **Compare**: During `PlanResourceChange`, the framework compares observability objects using `ObjectValue.Equal()`, which recursively calls `Float64Value.Equal()`
5. **Mismatch**: `big.Float.Cmp()` returns non-zero because the internal representations differ
6. **Trigger**: The framework detects a "value change" for observability, which triggers the resource-level `ModifyPlan`
7. **Cascade**: In `ModifyPlan`, the comparison `plan.Observability.Equal(state.Observability)` returns `false`, so computed attributes like `updated_on` and `references` are NOT copied from state to plan
8. **Result**: These computed attributes show as `(known after apply)`, indicating an update is needed

### Why Certain Values Were Affected

The bug primarily affected values that cannot be represented exactly in binary floating-point, such as:
- `0.0001` (the actual value causing issues)
- `0.1`, `0.333333`, etc.

Values like `1.0` or `0.5` might coincidentally have matching representations at both precisions.

### The Fix

The fix modifies `apijson/decoder.go` to create `Float64Value` objects with 512-bit precision, matching how `terraform-plugin-go` parses HCL values:

```go
// Before (53-bit precision):
value.Set(reflect.ValueOf(types.Float64Value(node.Float())))

// After (512-bit precision):
bigF, _, err := big.ParseFloat(node.Raw, 10, 512, big.ToNearestEven)
tfVal := tftypes.NewValue(tftypes.Number, bigF)
f64Val, _ := basetypes.Float64Type{}.ValueFromTerraform(ctx, tfVal)
value.Set(reflect.ValueOf(f64Val))
```

This ensures that API-sourced float values and HCL config float values have identical internal representations, so `Equal()` returns `true` when the numeric values match.

### Impact

This fix applies to all resources using the `apijson` decoder with `types.Float64` fields, not just the worker resource. Any resource with `float64` attributes that might have non-exact decimal values will benefit from this fix.

### DIFF
--- a/internal/apijson/decoder.go
+++ b/internal/apijson/decoder.go
@@ -16,6 +16,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
 	"github.com/tidwall/gjson"
 
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/customfield"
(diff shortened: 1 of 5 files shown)


tamas-jozsa (changes requested):
I don't understand how this change would solve the "known after apply" issues you mentioned in your description. Please provide correct examples and proof that this change would solve your issue.

Futhermore, changes to decoder would affect every other resources as well, so this change is potentially dangerous and needs sufficient testing with every other resource.

mgirouard:
@tamas-jozsa I think that it might be that the difference in precision is causing TF to see a change on a computed attribute, which then makes it think that all other computed fields _might_ change.

(edit) Also, agreed on the testing.

PR #6711, opened 2026-02-02
fix(authenticated_origin_pulls): handle array response and implement full lifecycle

ang-cloudflare:
## Summary
Fixes the `cloudflare_authenticated_origin_pulls` resource which was completely broken due to API response format mismatch.

### Primary Issue: Array Response Not Handled
The API endpoint `PUT /zones/{zone_id}/origin_tls_client_auth/hostnames` returns an **array** of all hostname associations for the zone, not just the single hostname being created/updated:

```json
{
  "result": [
    {"hostname": "example.com", "cert_id": "abc-123", "enabled": true, ...},
    {"hostname": "api.example.com", "cert_id": "def-456", "enabled": true, ...}
  ],
  "success": true
}
```

The auto-generated provider code expected a single object:
```json
{
  "result": {"hostname": "example.com", "cert_id": "abc-123", ...}
}
```

This caused Create/Update operations to **silently fail** - the API call succeeded but the response couldn't be parsed, leaving the resource state empty or incorrect.

### Secondary Issues
1. **Delete failed with "Certificate ID required"** - The API requires `cert_id` in the delete payload, but it wasn't being sent
2. **State drift on `private_key`** - This is a write-only field (sent to API, never returned), causing Terraform to detect drift on every refresh

### 1. Array Response Handling (`custom.go`)
Created custom types to properly handle the array response:

```go
// AuthenticatedOriginPullsArrayResultEnvelope handles the array response from the API
type AuthenticatedOriginPullsArrayResultEnvelope struct {
    Result []AuthenticatedOriginPullsModel `json:"result"`
}

// FindByHostname searches the array for a specific hostname
func (e *AuthenticatedOriginPullsArrayResultEnvelope) FindByHostname(hostname string) (*AuthenticatedOriginPullsModel, error) {
    for i := range e.Result {
        if e.Result[i].Hostname.ValueString() == hostname {
            return &e.Result[i], nil
        }
    }
    return nil, fmt.Errorf("hostname %q not found in response", hostname)
}
```

### 2. Fixed CRUD Operations (`resource.go`)

**Create/Update:**
```go
// Parse array response
env := AuthenticatedOriginPullsArrayResultEnvelope{}
err = apijson.UnmarshalComputed(bytes, &env)

// Find matching hostname from the array
result, err := env.FindByHostname(targetHostname)

// Copy computed fields to state
data.CERTID = result.CERTID
data.Enabled = result.Enabled
// ... etc
```

**Delete:**
```go
// Include cert_id as required by API
deletePayload := map[string]interface{}{
    "config": []map[string]interface{}{
        {
            "hostname": hostname,
            "cert_id":  data.CERTID.ValueString(), // Required!
            "enabled":  nil, // null voids the association
        },
    },
}
```

**Read:**
```go
// PrivateKey is write-only, set to null to prevent drift
data.PrivateKey = types.StringNull()
```

### 3. Comprehensive Lifecycle Test (`resource_test.go`)
Rewrote tests to cover the full lifecycle:
- **Step 1:** Create hostname association with `enabled=true`
- **Step 2:** Update to `enabled=false` 
- **Step 3:** Import and verify state matches

## API Documentation Verification

Verified implementation against [Cloudflare API docs](https://developers.cloudflare.com/api/resources/origin_tls_client_auth/subresources/hostnames/methods/update/):

| API Behavior | Implementation | Status |
|--------------|----------------|--------|
| PUT returns array of all hostname associations | `AuthenticatedOriginPullsArrayResultEnvelope` with `FindByHostname()` | ✅ |
| GET returns single hostname association | Uses `AuthenticatedOriginPullsResultEnvelope` (single object) | ✅ |
| Delete uses `enabled: null` to void association | Sends `enabled: nil` in payload | ✅ |
| `cert_id` required in request | Included in Create/Update/Delete payloads | ✅ |
| `private_key` is write-only | Set to `types.StringNull()` after API calls | ✅ |

## Files Changed
| File | Purpose |
|------|---------|
| `custom.go` | New file - Array envelope type and `FindByHostname()` helper |
| `resource.go` | Fixed Create/Update/Delete/Read methods |
| `resource_test.go` | Rewrote with comprehensive lifecycle test |
| `testdata/*.tf` | Removed outdated test configs |

## Cherry-picked Dependencies
This PR includes cherry-picked certificate normalization fixes required for the test to pass (the test creates a certificate first, then associates it with a hostname):
- Certificate normalization from PR #6710 (`authenticated_origin_pulls_certificate`)
- Certificate normalization from PR #6703 (`authenticated_origin_pulls_hostname_certificate`)

## Test Results
```
=== RUN   TestAccAuthenticatedOriginPulls_FullLifecycle
--- PASS: TestAccAuthenticatedOriginPulls_FullLifecycle (8.21s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/services/authenticated_origin_pulls	9.633s
```

## JIRA
SECENG-12970

### DIFF
--- a/internal/services/authenticated_origin_pulls/resource.go
+++ b/internal/services/authenticated_origin_pulls/resource.go
@@ -65,14 +65,17 @@ func (r *AuthenticatedOriginPullsResource) Create(ctx context.Context, req resou
 		return
 	}
 
-	// Extract the target hostname from config before making the API call
-	// The config has no_refresh tag so we need to preserve it
-	var targetHostname string
-	if data.Config != nil && len(*data.Config) > 0 {
-		targetHostname = (*data.Config)[0].Hostname.ValueString()
+	// Validate config has exactly one item - this resource manages a single hostname association
+	if data.Config == nil || len(*data.Config) != 1 {
+		resp.Diagnostics.AddError(
+			"invalid config",
+			"config must contain exactly one hostname association. Create separate resources to manage multiple hostnames.",
+		)
+		return
 	}
+	targetHostname := (*data.Config)[0].Hostname.ValueString()
 	if targetHostname == "" {
-		resp.Diagnostics.AddError("missing hostname", "config must contain at least one item with a hostname")
+		resp.Diagnostics.AddError("missing hostname", "config[0].hostname must not be empty")
 		return
 	}
 
@@ -156,13 +159,17 @@ func (r *AuthenticatedOriginPullsResource) Update(ctx context.Context, req resou
 		return
 	}
 
-	// Extract the target hostname from config
-	var targetHostname string
-	if data.Config != nil && len(*data.Config) > 0 {
-		targetHostname = (*data.Config)[0].Hostname.ValueString()
+	// Validate config has exactly one item - this resource manages a single hostname association
+	if data.Config == nil || len(*data.Config) != 1 {
+		resp.Diagnostics.AddError(
+			"invalid config",
+			"config must contain exactly one hostname association. Create separate resources to manage multiple hostnames.",
+		)
+		return
 	}
+	targetHostname := (*data.Config)[0].Hostname.ValueString()
 	if targetHostname == "" {
-		resp.Diagnostics.AddError("missing hostname", "config must contain at least one item with a hostname")
+		resp.Diagnostics.AddError("missing hostname", "config[0].hostname must not be empty")
 		return
 	}
 
--- a/internal/services/authenticated_origin_pulls/resource_test.go
+++ b/internal/services/authenticated_origin_pulls/resource_test.go
@@ -171,7 +171,7 @@ func TestAccAuthenticatedOriginPulls_FullLifecycle(t *testing.T) {
 				ResourceName:            resourceName,
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"config", "created_at", "status"},
+				ImportStateVerifyIgnore: []string{"config", "created_at", "updated_at", "status"},
 				ImportStateIdFunc:       testAccAuthenticatedOriginPullsImportStateIdFunc(resourceName),
 			},
 		},
--- a/internal/services/authenticated_origin_pulls_certificate/resource_test.go
+++ b/internal/services/authenticated_origin_pulls_certificate/resource_test.go
@@ -112,16 +112,7 @@ func testAccAuthenticatedOriginPullsCertificateImportStateIdFunc(resourceName st
 }
 
 func testAccAuthenticatedOriginPullsCertificateConfig(rnd, zoneID, cert, key string) string {
-	return fmt.Sprintf(`
-resource "cloudflare_authenticated_origin_pulls_certificate" "%[1]s" {
-  zone_id     = "%[2]s"
-  certificate = <<EOT
-%[3]s
-EOT
-  private_key = <<EOT
-%[4]s
-EOT
-}`, rnd, zoneID, cert, key)
+	return acctest.LoadTestCase("authenticatedoriginpullscertificatelifecycle.tf", rnd, zoneID, cert, key)
 }
 
 // TestAccAuthenticatedOriginPullsCertificate_FullLifecycle tests the full lifecycle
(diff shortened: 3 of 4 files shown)


internal/services/authenticated_origin_pulls/resource.go
```diff
@@ -113,13 +156,27 @@ func (r *AuthenticatedOriginPullsResource) Update(ctx context.Context, req resou
 		return
 	}
 
+	// Extract the target hostname from config
+	var targetHostname string
+	if data.Config != nil && len(*data.Config) > 0 {
+		targetHostname = (*data.Config)[0].Hostname.ValueString()
```
vaishakdinesh: Just checking if the first item in the array is fine?

internal/services/authenticated_origin_pulls/resource_test.go
```diff
@@ -25,88 +33,147 @@ func init() {
...
-	return acctest.LoadTestCase("authenticatedoriginpullsglobalconfig.tf", zoneID, name)
-}
-
-func testAccCheckCloudflareAuthenticatedOriginPullsPerZoneConfig(zoneID, name, aopType string) string {
-	return acctest.LoadTestCase("authenticatedoriginpullsperzoneconfig.tf", name, zoneID, aopType)
-}
-
-func testAccCheckCloudflareAuthenticatedOriginPullsConfig(zoneID, name, aopType, hostname string) string {
-	return acctest.LoadTestCase("authenticatedoriginpullsconfig.tf", name, zoneID, aopType, name+"."+hostname)
-}
```
vaishakdinesh, line 112: These cases are longer required/invalid?

internal/services/authenticated_origin_pulls_certificate/resource_test.go
```diff
@@ -5,226 +5,173 @@ import (
...
-	})
+		return fmt.Sprintf("%s/%s", zoneID, certID), nil
+	}
 }
 
-func TestAccCloudflareAuthenticatedOriginPullsCertificatePerHostname(t *testing.T) {
-	acctest.TestAccSkipForDefaultZone(t, "Pending investigation into correct test setup for reproducibility.")
+func testAccAuthenticatedOriginPullsCertificateConfig(rnd, zoneID, cert, key string) string {
+	return fmt.Sprintf(`
+resource "cloudflare_authenticated_origin_pulls_certificate" "%[1]s" {
```
vaishakdinesh: can you make this as file in testdata instead?

vaishakdinesh:
> The API endpoint PUT /zones/{zone_id}/origin_tls_client_auth/hostnames returns an array of all hostname associations for the zone, not just the single hostname being created/updated:


Is the openapi definition correct?

ang-cloudflare:
> updated

Yes, the OpenAPI spec in api-docs correctly defines the response as hostname_aop_response_collection with result: type: array

internal/services/authenticated_origin_pulls/resource_test.go
ang-cloudflare: - The old tests (TestAccCloudflareAuthenticatedOriginPullsGlobal, PerZone, PerHostname) were all skipped with TestAccSkipForDefaultZone
- They used outdated patterns and testdata files that didn't match the current resource schema
- We replaced them with TestAccAuthenticatedOriginPulls_FullLifecycle which is comprehensive and actually runs

internal/services/authenticated_origin_pulls_certificate/resource_test.go
ang-cloudflare: Done

internal/services/authenticated_origin_pulls/resource.go
ang-cloudflare: The config array in the Terraform schema is input only - it's what the user specifies in their HCL configuration. With the validation we added, we enforce that config must have exactly one item:
resource "cloudflare_authenticated_origin_pulls" "example" {
  zone_id = "abc123"
  config = [{
    cert_id  = cloudflare_authenticated_origin_pulls_hostname_certificate.example.id
    enabled  = true
    hostname = "api.example.com"  # This specific hostname
  }]
}
So config[0] is always the hostname the user explicitly configured for this resource.

We might want to added this somewhere in documentations

vaishakdinesh: > We might want to added this somewhere in documentations


I think that;s a good idea. The documentation for the registry is generated using `scripts/generate-docs` you can add template, which allows you have the docs generated as well add custom informations. @ang-cloudflare lemme know if you need a pr as a reference

PR #6718, opened 2026-02-03
feat(account): add RequiresReplace modifier to unit.id field

mya-jin:

## Changes being requested
The unit.id update should trigger recreate, so need to add the planModifier `stringplanmodifier.RequiresReplace(),`, and a plan check should be added to acceptance tests to make sure updating unit.id would trigger destroy and create.

- [X] I have added or updated acceptance tests for my changes
- [X] I have run acceptance tests for my changes and included the results below

### Test output
 
```
mya@G9JRYWDQ25 terraform-provider-cloudflare % cd ./internal/services/account && TF_LOG='' go test -v -run "^TestAccCloudflareAccount_WithUnit" 2>&1 | tail -5
=== RUN   TestAccCloudflareAccount_WithUnit
    resource_test.go:247: CLOUDFLARE_UNIT_ID is not set
--- SKIP: TestAccCloudflareAccount_WithUnit (0.00s)
PASS
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/services/account   1.540s
```

### DIFF
--- a/internal/services/account/resource_test.go
+++ b/internal/services/account/resource_test.go
@@ -3,6 +3,7 @@ package account_test
 import (
 	"context"
 	"fmt"
+	"os"
 	"regexp"
 	"testing"
 
@@ -243,7 +244,7 @@ func TestAccCloudflareAccount_WithUnit(t *testing.T) {
 	resourceName := fmt.Sprintf("cloudflare_account.%s", rnd)
 
 	// Get organization IDs to determine unit_id and alternate_unit_id
-	orgIDs, err := getOrganizationIDs()
+	orgIDs, err := getOrganizationIDs(t)
 	if err != nil {
 		t.Fatalf("Failed to get organization IDs: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


mgirouard (approved):

internal/services/account/resource_test.go
```diff
@@ -267,6 +267,16 @@ func TestAccCloudflareAccount_WithUnit(t *testing.T) {
...
 			},
+			// Changing unit.id should force replacement (destroy before create)
+			{
+				Config: testAccCheckCloudflareAccountWithUnit(rnd, rnd, "invalid-unit-id"),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(resourceName, plancheck.ResourceActionDestroyBeforeCreate),
+					},
+				},
+				ExpectError: regexp.MustCompile(`failed to make http request|403|not found|invalid`),
```
vaishakdinesh: Why would we be expecting a 4xx error?

vaishakdinesh: Also the test output says the test was skipped. 
```
CLOUDFLARE_UNIT_ID is not set

```

mya-jin: > Why would we be expecting a 4xx error?
I was using a invalid-unit-id to test updating, that is why an error was expected. 

> Also the test output says the test was skipped.
I updated the test, making the test more robust, will read the unit.id from organizations instead of the tester need to manually find it.

internal/services/account/resource_test.go
```diff
@@ -306,6 +360,25 @@ func testAccCheckCloudflareAccountWithUnit(rnd, name, unitID string) string {
 	return acctest.LoadTestCase("accountwithunit.tf", rnd, name, unitID)
 }
 
+// getOrganizationIDs fetches all organizations and returns their IDs
+func getOrganizationIDs() ([]string, error) {
```
vaishakdinesh: I think this will run before `resource.TestStep` which is where we check for `TF_ACC` to run acceptance tests. So this will accidentally run in the CI tests. Simple fix would be to do that check here.

PR #6743, opened 2026-02-11
fix: upgrade scenario

vaishakdinesh:

## Changes being requested

When upgrading the 

### DIFF
--- a/internal/services/certificate_pack/migrations.go
+++ b/internal/services/certificate_pack/migrations.go
@@ -4,6 +4,7 @@ package certificate_pack
 
 import (
 	"context"
+	"os"
 
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/services/certificate_pack/migration/v500"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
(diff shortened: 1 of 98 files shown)


mgirouard (approved):

internal/services/certificate_pack/migrations.go
```diff
@@ -22,10 +23,26 @@ var _ resource.ResourceWithUpgradeState = (*CertificatePackResource)(nil)
...
 //   - Just bumps the version number, no data transformation
+//
+// In production (no TF_MIG_TEST), only a no-op upgrader is registered at slot 0
+// to safely bump existing v5 users from version 0 to 1 without triggering the
+// v4→v5 transformation (which would fail on v5-format state).
 func (r *CertificatePackResource) UpgradeState(ctx context.Context) map[int64]resource.StateUpgrader {
-	sourceSchema := v500.SourceCloudflareCertificatePackSchema()
 	targetSchema := ResourceSchema(ctx)
 
+	if os.Getenv("TF_MIG_TEST") == "" {
```
tamas-jozsa, line 33: nit: should we put this behind a function? like `migrations.MigrationGateEnabled()`

tamas-jozsa (approved):

internal/services/certificate_pack/migrations.go
vaishakdinesh: We can do that, sounds good!

vaishakdinesh: Hmmm actually. It might a little hard to put the check in side the function. probably have something like bool, func return and then 
```
if isGateEnabled, gater := migrations.MigrationGateEnabled(); ok {
 return gater(targetSchema)
}
```

At which point i think this is fine? @tamas-jozsa

tamas-jozsa: yeah it's fine

PR #6790, opened 2026-02-18
feat(ruleset): add new actions for http_response_cache_settings phase

anthocf:

## Changes being requested

I am adding support for the new http_response_cache_settings rulesets phase and the correspond actions.

0. Follow steps in CONTRIBUTING.md. Edit `.terraformrc` to force cloudflare provider to point to local build
1. Retrieve the API key, zone id, account id, and email from personal Cloudflare account
2. Run `./scripts/build 0.0.0-dev` to build the provider with changes

```
$ CLOUDFLARE_ZONE_ID=<snip> CLOUDFLARE_ACCOUNT_ID=<snip> CLOUDFLARE_EMAIL=<snip> CLOUDFLARE_API_KEY=<snip> TF_ACC=1 go test -v ./internal/services/ruleset/... -run "TestAccCloudflareRuleset_SetCacheControlRules"
...

$ CLOUDFLARE_ZONE_ID=<snip> CLOUDFLARE_ACCOUNT_ID=<snip> CLOUDFLARE_EMAIL=<snip> CLOUDFLARE_API_KEY=<snip> TF_ACC=1 go test -v ./internal/services/ruleset/... -run "TestAccCloudflareRuleset_SetCacheTagsRules"
...

$ CLOUDFLARE_ZONE_ID=<snip> CLOUDFLARE_ACCOUNT_ID=<snip> CLOUDFLARE_EMAIL=<snip> CLOUDFLARE_API_KEY=<snip> TF_ACC=1 go test -v ./internal/services/ruleset/... -run "TestAccCloudflareRuleset_SetCacheSettingsRules"
```

### Test output
 

```
=== RUN   TestAccCloudflareRuleset_SetCacheControlRules
--- PASS: TestAccCloudflareRuleset_SetCacheControlRules (14.99s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/services/ruleset	16.658s
```

```
=== RUN   TestAccCloudflareRuleset_SetCacheTagsRules
--- PASS: TestAccCloudflareRuleset_SetCacheTagsRules (22.41s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/services/ruleset	24.149s
```

```
=== RUN   TestAccCloudflareRuleset_SetCacheSettingsRules
--- PASS: TestAccCloudflareRuleset_SetCacheSettingsRules (45.83s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal/services/ruleset	47.525s
```

### DIFF
--- a/internal/services/queue_consumer/data_source_schema_test.go
+++ b/internal/services/queue_consumer/data_source_schema_test.go
@@ -15,5 +15,6 @@ func TestQueueConsumerDataSourceModelSchemaParity(t *testing.T) {
 	model := (*queue_consumer.QueueConsumerDataSourceModel)(nil)
 	schema := queue_consumer.DataSourceSchema(context.TODO())
 	errs := test_helpers.ValidateDataSourceModelSchemaIntegrity(model, schema)
+	errs.Ignore(t, ".@QueueConsumerDataSourceModel.created_on")
 	errs.Report(t)
 }
--- a/internal/services/snippet_rules/resource_test.go
+++ b/internal/services/snippet_rules/resource_test.go
@@ -108,8 +108,10 @@ func testAccCheckCloudflareSnippetRulesDestroy(s *terraform.State) error {
 		}
 
 		// Check if any rules still exist
-		if rules != nil && len(rules.Result) > 0 {
-			return fmt.Errorf("snippet rules still exist")
+		if rules != nil {
+			if rulesList, ok := (*rules).([]interface{}); ok && len(rulesList) > 0 {
+				return fmt.Errorf("snippet rules still exist")
+			}
 		}
 	}
 


internal/services/ruleset/schema.go
```diff
@@ -1645,6 +1649,435 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 										},
 									},
 								},
+								"strip_etags": schema.BoolAttribute{
```
zakcutner: Could you put these with the other `set_cache_settings` action parameters please? The parameters are ordered lexicographically by action and then alphabetically. You'll also need to update `data_source_model.go`, `data_source_schema.go`, and `model.go` to be consistent.

anthocf: Will do! Thanks for checking this out

internal/services/ruleset/schema.go
```diff
@@ -1645,6 +1649,435 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+											Description: "The operation to perform.",
+											Required:    true,
+											Validators: []validator.String{
+												stringvalidator.OneOfCaseInsensitive("set", "remove"),
+											},
+										},
+										"value": schema.Int64Attribute{
+											Description: "The value for the directive in seconds.",
+											Optional:    true,
+										},
```
zakcutner, line 1710: I think we should add a validator to indicate that it can only be set for the `set` operation (same for the other cache control directives).
```go
customvalidator.RequiresOtherStringAttributeToBe(
    path.MatchRelative().AtParent().AtName("operation"),
    "set"
),
```

internal/services/ruleset/schema.go
```diff
@@ -1645,6 +1649,435 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+									Validators: []validator.Object{
+										customvalidator.RequiresOtherStringAttributeToBe(
+											path.MatchRelative().AtParent().AtParent().AtName("action"),
+											"set_cache_control",
+										),
+									},
+									CustomType: customfield.NewNestedObjectType[RulesetRulesActionParametersSetCacheControlValueModel](ctx),
+									Attributes: map[string]schema.Attribute{
+										"operation": schema.StringAttribute{
+											Description: "The operation to perform.",
```
zakcutner: Could you add `Available values` to the `Description` (you can copy how it's done for other enums in this file)? You'll want to do the same thing with the other enums you've added.

internal/services/ruleset/schema.go
```diff
@@ -1645,6 +1649,435 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+								},
+								"values": schema.ListAttribute{
+									Description: "The cache tag values for set_cache_tags action.",
+									Optional:    true,
+									Validators: []validator.List{
+										customvalidator.RequiresOtherStringAttributeToBe(
+											path.MatchRelative().AtParent().AtParent().AtName("action"),
+											"set_cache_tags",
+										),
+									},
```
zakcutner, line 2105: Could you add a validator both here and to `expression` to prevent both from being set?
```go
objectvalidator.ConflictsWith(path.MatchRelative().AtParent().AtName("expression")),
```

internal/services/ruleset/model.go
```diff
@@ -106,6 +106,25 @@ type RulesetRulesActionParametersModel struct {
 	RequestFields            customfield.NestedObjectList[RulesetRulesActionParametersRequestFieldsModel]            `tfsdk:"request_fields" json:"request_fields,optional"`
 	ResponseFields           customfield.NestedObjectList[RulesetRulesActionParametersResponseFieldsModel]           `tfsdk:"response_fields" json:"response_fields,optional"`
 	TransformedRequestFields customfield.NestedObjectList[RulesetRulesActionParametersTransformedRequestFieldsModel] `tfsdk:"transformed_request_fields" json:"transformed_request_fields,optional"`
+	StripEtags               types.Bool                                                                              `tfsdk:"strip_etags" json:"strip_etags,optional"`
```
zakcutner: nit

```suggestion
	StripETags               types.Bool                                                                              `tfsdk:"strip_etags" json:"strip_etags,optional"`
```

internal/services/ruleset/model.go
```diff
@@ -321,3 +340,21 @@ type RulesetRulesRatelimitModel struct {
 	ScorePerPeriod          types.Int64                    `tfsdk:"score_per_period" json:"score_per_period,optional"`
 	ScoreResponseHeaderName types.String                   `tfsdk:"score_response_header_name" json:"score_response_header_name,optional"`
 }
+
+// Cache control models for set_cache_control action
```
zakcutner: nit: I think the comment is probably not needed for consistency with other types (the type name you chose is already very self-explanatory IMO)

internal/services/ruleset/model.go
```diff
@@ -321,3 +340,21 @@ type RulesetRulesRatelimitModel struct {
...
+type RulesetRulesActionParametersSetCacheControlValueModel struct {
+	Operation      types.String `tfsdk:"operation" json:"operation,required"`
+	Value          types.Int64  `tfsdk:"value" json:"value,optional"`
+	CloudflareOnly types.Bool   `tfsdk:"cloudflare_only" json:"cloudflare_only,optional"`
```
zakcutner: This should be marked as `computed_optional` (rather than `optional`) to match the schema, and as `decode_null_to_zero` to tell the provider to interpret an omitted value as `false` (otherwise Terraform will complain that a computed attribute cannot be omitted). Also, the same applies to all the other `CloudflareOnly` attributes.
```suggestion
	CloudflareOnly types.Bool   `tfsdk:"cloudflare_only" json:"cloudflare_only,computed_optional,decode_null_to_zero"`
```

internal/services/ruleset/ruleset_test.go
```diff
@@ -7315,3 +7395,595 @@ func TestAccCloudflareRuleset_SkipRules(t *testing.T) {
 		},
 	})
 }
+
+func TestAccCloudflareRuleset_SetCacheControlRules(t *testing.T) {
```
zakcutner, line 7431: I think to be more comprehensive, it would be great to have three test configs, where each attribute is in all three possible states in one of the configs:
1. Omitted
2. Operation set to `set`
3. Operation set to `remove`

internal/services/ruleset/testdata/TestAccCloudflareRuleset_SetCacheControlRules/1.tf
```diff
@@ -0,0 +1,68 @@
+variable "zone_id" {}
+
+resource "cloudflare_ruleset" "my_ruleset" {
+  zone_id = var.zone_id
+  name    = "My ruleset"
+  phase   = "http_response_cache_settings"
+  kind    = "zone"
+  rules = [
+    {
+      expression = "any(http.request.headers.values[*] == \"application/json\")"
```
zakcutner, line 10: nit: Could you use `ip.src eq 1.1.1.1` as the expression in all the tests, to be consistent with other tests

internal/services/ruleset/testdata/TestAccCloudflareRuleset_SetCacheSettingsRules/9.tf
zakcutner, line 1: nit: Could you add the new fields to `2.tf` and `3.tf` instead (like `respect_strong_etags`), to keep the tests simpler?

internal/services/ruleset/data_source_schema.go
zakcutner, line 1: I think you'll also want to add your new phase here?

zakcutner, line 1: Could you update this to match `schema.go` (aside from all the attributes being `Computed`-only obviously)? For example, the validators should be the same.

internal/services/ruleset/data_source_model.go
```diff
@@ -332,3 +351,20 @@ type RulesetRulesRatelimitDataSourceModel struct {
 	ScorePerPeriod          types.Int64                    `tfsdk:"score_per_period" json:"score_per_period,computed"`
 	ScoreResponseHeaderName types.String                   `tfsdk:"score_response_header_name" json:"score_response_header_name,computed"`
 }
+
+type RulesetRulesActionParametersSetCacheControlValueDataSourceModel struct {
+	Operation      types.String `tfsdk:"operation" json:"operation,computed"`
+	Value          types.Int64  `tfsdk:"value" json:"value,computed"`
+	CloudflareOnly types.Bool   `tfsdk:"cloudflare_only" json:"cloudflare_only,computed"`
```
zakcutner: You'll need to add `decode_null_to_zero` to the `CloudflareOnly` attributes, same as in `model.go`

```suggestion
	CloudflareOnly types.Bool   `tfsdk:"cloudflare_only" json:"cloudflare_only,computed,decode_null_to_zero"`
```

vaishakdinesh:
@anthocf could you target `next` instead of main? :)

internal/services/ruleset/testdata/TestAccCloudflareRuleset_SetCacheControlRules/1.tf
anthocf: The ip attributes aren't available for response cache rules. Would you prefer using another expression?

internal/services/ruleset/testdata/TestAccCloudflareRuleset_SetCacheSettingsRules/9.tf
anthocf: Customers shouldn't be able to use the new fields along side the others in the http_response_cache_settings phase. The action is the same but the fields can't be used in both phases.

internal/services/ruleset/data_source_model.go
anthocf: Thank you!

internal/services/ruleset/model.go
anthocf: thanks!

internal/services/ruleset/schema.go
zakcutner: Should we do this for the directives that take `qualifiers` too?

internal/services/ruleset/schema.go
```diff
@@ -1646,6 +1679,431 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+										customvalidator.RequiresOtherStringAttributeToBe(
+											path.MatchRelative().AtParent().AtParent().AtName("action"),
+											"set_cache_control",
+										),
+									},
+									CustomType: customfield.NewNestedObjectType[RulesetRulesActionParametersSetCacheControlValueModel](ctx),
+									Attributes: map[string]schema.Attribute{
+										"operation": schema.StringAttribute{
+											Description: "The operation to perform.\nAvailable values: \"set\", \"remove\".",
+											Optional:    true,
```
zakcutner: Is this meant to be required?

internal/services/ruleset/testdata/TestAccCloudflareRuleset_SetCacheControlRules/1.tf
zakcutner: That's fair, I think what you've done is fine in that case!

internal/services/ruleset/testdata/TestAccCloudflareRuleset_SetCacheSettingsRules/9.tf
zakcutner: Ah, fair point! Would you mind splitting into two different test steps though, to keep it consistent with each test step having a single rule (and so we test the change from false to true)?

internal/services/ruleset/data_source_schema.go
```diff
@@ -1630,6 +1664,381 @@ func DataSourceSchema(ctx context.Context) schema.Schema {
...
+											Description: "The operation to perform.\nAvailable values: \"set\", \"remove\".",
+											Computed:    true,
+											Validators: []validator.String{
+												stringvalidator.OneOfCaseInsensitive("set", "remove"),
+											},
+										},
+										"value": schema.Int64Attribute{
+											Description: "The value for the directive in seconds.",
+											Computed:    true,
+										},
```
zakcutner, line 1694: This is missing the same validators as in `schema.go` (same for other `value` properties)

internal/services/ruleset/data_source_model.go
```diff
@@ -111,11 +111,30 @@ type RulesetRulesActionParametersDataSourceModel struct {
...
+	NoTransform              customfield.NestedObject[RulesetRulesActionParametersSetCacheControlSimpleDataSourceModel]        `tfsdk:"no_transform" json:"no-transform,computed"`
+	Immutable                customfield.NestedObject[RulesetRulesActionParametersSetCacheControlSimpleDataSourceModel]        `tfsdk:"immutable" json:"immutable,computed"`
+	NoStore                  customfield.NestedObject[RulesetRulesActionParametersSetCacheControlSimpleDataSourceModel]        `tfsdk:"no_store" json:"no-store,computed"`
+	Public                   customfield.NestedObject[RulesetRulesActionParametersSetCacheControlSimpleDataSourceModel]        `tfsdk:"public" json:"public,computed"`
+	Operation                types.String                                                                                      `tfsdk:"operation" json:"operation,computed"`
+	Values                   customfield.List[types.String]                                                                    `tfsdk:"values" json:"values,computed"`
+	CacheTagsExpression      types.String                                                                                      `tfsdk:"expression" json:"expression,computed"`
```
zakcutner: Could you rename this to `Expression` to match `model.go` please?

zakcutner (approved):

PR #6892, opened 2026-03-06
test(r2_data_catalog): add acceptance tests

garvit-gupta:

## Changes being requested

Add acceptance tests for the `cloudflare_r2_data_catalog` resource.

`TF_ACC=1 go test -count 1 ./internal/services/r2_data_catalog -run "TestAcc" -v`

### Test output
```
❯ TF_ACC=1 go test -count 1 ./internal/services/r2_data_catalog -run "TestAcc" -v
=== RUN   TestAccCloudflareR2DataCatalog_Basic
--- PASS: TestAccCloudflareR2DataCatalog_Basic (5.81s)
PASS
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/services/r2_data_catalog   7.570s
```

### DIFF
--- a/internal/services/r2_data_catalog/resource_test.go
+++ b/internal/services/r2_data_catalog/resource_test.go
@@ -1,0 +1,104 @@
+package r2_data_catalog_test
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/cloudflare/cloudflare-go/v6"
+	"github.com/cloudflare/cloudflare-go/v6/r2_data_catalog"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/acctest"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/consts"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/utils"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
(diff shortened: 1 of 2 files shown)


internal/services/r2_data_catalog/resource_test.go
```diff
@@ -0,0 +1,103 @@
...
+					statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("id"), knownvalue.NotNull()),
+					statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("name"), knownvalue.NotNull()),
+				},
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateId:     fmt.Sprintf("%s/%s", accountID, rnd),
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
```
vaishakdinesh, line 62: @garvit-gupta is this intentional? What happens to these fields when you import them?

garvit-gupta: Yes, this is intentional. I've trimmed the list from 5 to 4 though (removed name since Enable does return it) and added a comment to explain.

The Enable (POST) and Get (GET) APIs return different response shapes:
- Enable returns only {id, name}
- Get returns {id, name, bucket, status, maintenance_config, credential_status}

After Create, these 4 fields are null in state because Enable doesn't include them. On Import, the provider calls Get which does return them. `ImportStateVerify` seems to be a strict state-equality check, so the 4 fields mismatch if I don't use `ImportStateVerifyIgnore`.

This should be safe because all 4 fields are `Computed: true` only. Users can't set them in config, so there's no drift possible. The `PostApplyPostRefresh: ExpectEmptyPlan()` check in Step 1 confirms the resource is stable after creation.

PR #6920, opened 2026-03-13
fix(apijson): return empty object from MarshalForPatch when no fields are serialisable

tamas-jozsa:
When every serialisable field in a patch update is either unchanged (omitted by the diff check) or computed_optional unknown (omitted by handleNullAndUndefined), MarshalForPatch previously returned nil bytes. Passing nil bytes to option.WithRequestBody produced an empty HTTP body, which the Cloudflare API rejected with "400: invalid json request".

At the root encoder level, return []byte("{}") instead of nil so the caller always has a structurally valid JSON body to send. Nested sub-object encoders still return nil (causing them to be omitted from the parent object) so existing patch diffing behaviour is unaffected.

The isRoot flag is captured at the start of newTypeEncoder, before e.root is reset to false, and threaded into newStructTypeEncoder as a parameter so the closure reads the correct value at call time.

Fixes the 400 error observed on cloudflare_zero_trust_device_default_profile updates triggered by a moved block, where the only pending diff was include/exclude being "known after apply".

### DIFF
--- a/internal/apijson/encoder.go
+++ b/internal/apijson/encoder.go
@@ -148,6 +148,10 @@ func (e *encoder) typeEncoder(t reflect.Type) encoderFunc {
 }
 
 func (e *encoder) newTypeEncoder(t reflect.Type) encoderFunc {
+	// Capture root before any mutation below (e.root is set to false at line 179
+	// before the switch dispatches to newStructTypeEncoder).
+	isRoot := e.root
+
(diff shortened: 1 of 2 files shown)


ms-jpq (commented):
makes total sense!

tamas-jozsa:
closing the PR, bc it will be included in codegen

PR #6940, opened 2026-03-20
feat(secrets_store): add cloudflare_secrets_store and cloudflare_secrets_store_secret resources

trafgals:
## Summary

Implements `cloudflare_secrets_store` and `cloudflare_secrets_store_secret` resource types to allow managing Secrets Store resources via Terraform.

## What this PR does

- Adds `cloudflare_secrets_store` resource for creating/managing Secrets Store containers
- Adds `cloudflare_secrets_store_secret` resource for managing secrets within a store
- Registers both resources in the provider

## Example Usage

```hcl
resource "cloudflare_secrets_store" "worker" {
  account_id = var.account_id
  name       = "my-app-secrets-\${terraform.workspace}"
}

resource "cloudflare_secrets_store_secret" "api_key" {
  account_id  = var.account_id
  store_id    = cloudflare_secrets_store.worker.id
  secret_name = "API_KEY"
  secret_text = var.api_key
}
```

## References

- Resolves issue #6939
- Cloudflare Secrets Store API: https://developers.cloudflare.com/secrets-store/

PR #6938, opened 2026-03-20
feat(ai_gateway): add AI Gateway resource for issue #6720

trafgals:
## Summary

Implements the `cloudflare_ai_gateway` resource to address issue #6720 - the missing AI Gateway support in the Terraform provider.

## What this PR does

Adds a new Terraform resource `cloudflare_ai_gateway` that allows users to manage AI Gateway configurations. This resource enables users to create, read, update, and delete AI Gateways without needing to use the generic `http` provider.

## Resources Added

- `cloudflare_ai_gateway` - Main AI Gateway resource

## Supported Configuration Options

- **Basic Settings**: `id`, `account_id`
- **Cache Settings**: `cache_invalidate_on_update`, `cache_ttl`
- **Logging**: `collect_logs`, `log_management`, `log_management_strategy`, `logpush`, `logpush_public_key`
- **Rate Limiting**: `rate_limiting_interval`, `rate_limiting_limit`, `rate_limiting_technique`
- **Authentication**: `authentication`
- **DLP**: `dlp` block with `action`, `enabled`, `profiles`
- **OpenTelemetry**: `otel` block with `authorization`, `headers`, `url`, `content_type`
- **Stripe**: `stripe` block with `authorization`, `usage_events`
- **Workers AI**: `workers_ai_billing_mode`
- **Other**: `is_default`, `store_id`, `zdr`

## Example Usage

\`\`\`hcl
resource "cloudflare_ai_gateway" "example" {
  account_id = var.account_id
  id         = "my-ai-gateway"
  
  collect_logs            = true
  rate_limiting_limit     = 100
  rate_limiting_interval  = 60
  rate_limiting_technique = "fixed"
  
  workers_ai_billing_mode = "postpaid"
}
\`\`\`

## References

- Issue: #6720
- API Documentation: https://developers.cloudflare.com/api/resources/ai_gateway/

## Note

This implementation assumes the `github.com/cloudflare/cloudflare-go/v6/ai_gateway` SDK package exists. If the SDK package structure differs, adjustments may be needed.

PR #6945, opened 2026-03-20
fix: align secrets_store and ai_gateway resources with cloudflare-go SDK

trafgals:
## Summary

Fixes the secrets_store and ai_gateway resource implementations to match the actual cloudflare-go SDK v6.8.1 API.

### secrets_store resource
- Fixed StoreNewParams to use Body.Name structure instead of direct Name field
- Fixed service path (already correct: client.SecretsStore.Stores.*)

### secrets_store_secret resource  
- Fixed StoreSecretNewParams to use Body structure with Name, Value, Scopes, Comment
- Fixed service path to client.SecretsStore.Stores.Secrets.*
- Changed Update to Edit (SDK uses Edit, not Update)
- Fixed scopes handling for []types.String model field

### ai_gateway resource
- Fixed service path from r.client.AIGateway.Gateways.* to r.client.AIGateway.*
- Fixed response handling (direct response, not wrapped in Result)

### schema fixes
- Removed invalid stringvalidator.ValueStringsAre validator
- Changed SetAttribute to ListAttribute for scopes field

## Motivation

The PRs #6940 and #6938 were written against an API design that doesn't match the actual cloudflare-go SDK. The implementations call methods and use params structures that don't exist in the SDK.

### DIFF
--- a/internal/services/ai_gateway/resource.go
+++ b/internal/services/ai_gateway/resource.go
@@ -143,32 +143,69 @@ func (r *AIGatewayResource) Update(ctx context.Context, req resource.UpdateReque
 		return
 	}
 
-	result, err := r.client.AIGateway.Update(
-		ctx,
-		state.ID.ValueString(),
-		ai_gateway.AIGatewayUpdateParams{
-			AccountID:               cloudflare.F(data.AccountID.ValueString()),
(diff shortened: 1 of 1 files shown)


trafgals:
Closing in favor of consolidated PRs #6946 and #6947

Issue #6952, opened 2026-03-23
Update cloudflare_workers_cron_trigger example with schedules


BradleyYeo:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform 1.14.7
Cloudflare provider: 5.19.0-beta.2

### Affected resource(s)

cloudflare_workers_cron_trigger

### Terraform configuration files

```hcl
resource "cloudflare_workers_cron_trigger" "example_workers_cron_trigger" {
  account_id = "023e105f4ecef8ad9ca31a8372d0c353"
  script_name = "this-is_my_script-01"
  body = [{
    cron = "*/30 * * * *"
  }]
}
```

### Link to debug output

-

### Panic output

Error: Missing required argument
│ 
│   on main.tf line 23, in resource "cloudflare_workers_cron_trigger" "example_workers_cron_trigger":
│   23: resource "cloudflare_workers_cron_trigger" "example_workers_cron_trigger" {
│ 
│ The argument "schedules" is required, but no definition was found.
╵
╷
│ Error: Unsupported argument
│ 
│   on main.tf line 26, in resource "cloudflare_workers_cron_trigger" "example_workers_cron_trigger":
│   26:   body = [{
│ 
│ An argument named "body" is not expected here.

### Expected output

  # cloudflare_workers_cron_trigger.example_workers_cron_trigger will be created
  + resource "cloudflare_workers_cron_trigger" "example_workers_cron_trigger" {
      + account_id  = "xyz"
      + id          = (known after apply)
      + schedules   = [
          + {
              + created_on  = (known after apply)
              + cron        = "*/30 * * * *"
              + modified_on = (known after apply)
            },
        ]
      + script_name = "script-01"
    }

### Actual output

Error: Missing required argument
│ 
│   on main.tf line 23, in resource "cloudflare_workers_cron_trigger" "example_workers_cron_trigger":
│   23: resource "cloudflare_workers_cron_trigger" "example_workers_cron_trigger" {
│ 
│ The argument "schedules" is required, but no definition was found.
╵
╷
│ Error: Unsupported argument
│ 
[...]

PR #6953, opened 2026-03-23
Fix cloudflare_workers_cron_trigger documentation example to use schedules instead of body

BradleyYeo:

## Changes being requested
The docs for the cloudflare_workers_cron_trigger  resource were incorrectly displaying an example using the body attribute. The correct attribute that is used here is schedules.

N/A, just a docs change

### Test output
 
N/A, just a docs change

## Additional context & links
Issue below:
https://github.com/cloudflare/terraform-provider-cloudflare/issues/6952

### DIFF
--- a/docs/resources/workers_cron_trigger.md
+++ b/docs/resources/workers_cron_trigger.md
@@ -2,7 +2,7 @@
 page_title: "cloudflare_workers_cron_trigger Resource - Cloudflare"
 subcategory: ""
 description: |-
-  
+
 ---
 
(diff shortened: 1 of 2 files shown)


vaishakdinesh:
@BradleyYeo unfortunately this wont work the way you expect. We can get this fixed correctly. Thank you for the PR though!

BradleyYeo:
Thank you for reviewing! Could you share with me how it would be fixed as I would like to learn how to do it.🙏

PR #6970, opened 2026-03-26
fix: multiple provider state upgrade and drift fixes discovered during v4→v5 migration testing

tamas-jozsa:
### `zero_trust_gateway_policy` — `filters` causes persistent drift

The `filters` attribute was `Optional`-only. Because the Cloudflare API always returns a value for `filters` (e.g. `["dns"]`) even when not explicitly set in config, the provider planned to remove it on every plan and re-read it on every apply — making the resource permanently unstable.

- `schema.go`: added `Computed: true` and `CustomType: customfield.NewListType[types.String](ctx)` so the framework can represent the field as unknown during planning
- `model.go`: changed `Filters` from `*[]types.String` to `customfield.List[types.String]`
- `resource.go`: removed the stale `priorFilters` preservation workaround that was a band-aid for the missing `Computed` flag

---

### `leaked_credential_check_rule` — empty ID from v4 provider bug causes plan failure

The v4 provider had a bug where it created `cloudflare_leaked_credential_check_rule` resources successfully but failed to store the returned `detection_id` in state (`id = ""`). When the v5 provider's `Read()` called `Get("")`, the SDK rejected it immediately with `"missing required detection_id parameter"`.

`Read()` now detects the empty ID, removes the resource from state with a warning, and lets Terraform plan to recreate it — which correctly stores the real `detection_id` on the next apply.

Also: removed `t.Skip` from migration tests (product now enabled in CI account), and added unit tests for the empty ID migration path.

---

### `zero_trust_access_policy` — `email_domain` migration test

Added `TestMigrateZeroTrustAccessPolicyEmailDomainTransformation` verifying that `email_domain = ["domain"]` (v4 list syntax) is correctly migrated to `email_domain = { domain = "domain" }` (v5 object syntax).

---

### State upgrade failures (`zone`, `notification_policy`, `ruleset`, `zero_trust_access_policy`)

Five resources failed during `terraform plan` after migration with errors like:

```
Error: Unable to Read Previously Saved State for UpgradeResourceState
  AttributeName("filters"): invalid JSON, expected "{", got "["
  AttributeName("plan"): invalid JSON, expected "{", got "enterprise"
  AttributeName("output_options"): invalid JSON, expected "{", got "["
  AttributeName("action_parameters"): invalid JSON, expected "{", got "["
  AttributeName("connection_rules"): invalid JSON, expected "{", got "["
```

**Root cause:** The Terraform Plugin Framework's `PriorSchema`-based state deserialization fails when v4 SDKv2 stored `TypeList MaxItems:1` fields as JSON arrays `[]` — the Plugin Framework's internal state encoding for `ListNestedAttribute` is different. The framework validates the raw state against `PriorSchema` before calling the upgrader and rejects the mismatch.

**Fix:** Set `PriorSchema: nil` in each affected `UpgradeState` registration, and update each handler to unmarshal raw state bytes directly via `req.RawState.Unmarshal(sourceType)` + `tfsdk.State.Get()`. This bypasses the framework's schema validation and is the same pattern already working in the `logpush_job` migration.

`zero_trust_access_policy` additionally needed a v4/v5 format detector in `UpgradeFromSchemaV0` since both v4 and early v5 state have `schema_version=0` — it checks for `application_id`, `precedence`, or `connection_rules=[]` to distinguish v4 state from v5 state and transform accordingly.

---

### `zone_setting` — migration test harness

Updated `RunMigrationV2Command` in the acctest harness to pass `--skip-phase-check` (previously `--yes`) so tf-migrate runs the full migration directly instead of entering the phased migration flow.

---

### DIFF
--- a/internal/services/zero_trust_access_policy/migration/v500/migrations_test.go
+++ b/internal/services/zero_trust_access_policy/migration/v500/migrations_test.go
@@ -709,7 +709,7 @@ func TestMigrateZeroTrustAccessPolicyEmailDomainTransformation(t *testing.T) {
 	resourceName := "cloudflare_zero_trust_access_policy." + rnd
 	domain := os.Getenv("CLOUDFLARE_DOMAIN")
 	if domain == "" {
-		domain = consts.TestCloudflareAlternativeDomain
+		t.Skip("CLOUDFLARE_DOMAIN must be set for this test")
 	}
 	tmpDir := t.TempDir()
 


vaishakdinesh:
```
========================================
✓ E2E Test Complete!
========================================

Summary:

  Step 1: v4 terraform apply
    Status: ✓ SUCCESS

  Step 2: Migration (v4 → v5)
    Status: ✓ SUCCESS

  Step 3: v5 plan (before apply)
    Status: ✓ SUCCESS - Drift matched exemptions
    Result: 69 material changes (69 matched exemptions, 0 unmatched)
    Terraform: Plan: 16 to import, 8 to add, 63 to change, 0 to destroy.

  Step 4: v5 terraform apply
    Status: ✓ SUCCESS

  Step 5: v5 plan (after apply)
    Status: ✓ SUCCESS - Stable state achieved
    Result: 11 material changes matched exemptions (0 unmatched)
```

Issue #2907, opened 2023-11-02
Rule resource (for ruleset)


yevhenvolchenko:
### Current Terraform and Cloudflare provider version

Terraform: 1.6.3
Cloudflare provider: 4.18.0

### Description

A rule resource, that can be added to an existing ruleset

### Use cases

We frequently add new domains and WAF rules specific to those domains. At the moment there is no way to package an application in a single, independent TF module, because all the rules must be in a single TF resource

This problem was not present in pre-RulesetAPI era, BTW

### Potential Terraform configuration

```hcl
...
```

jacobbednarz:
duplicate of #2423 and #2688

petero-dk:
I added a workaround using multiple repositories here: https://github.com/cloudflare/terraform-provider-cloudflare/issues/3027#issuecomment-2138260815

Issue #2688, opened 2023-08-16
Allows to have rules split in different repos


peruzzof:
### Current Terraform and Cloudflare provider version

Terraform v1.5.5
+ provider registry.terraform.io/cloudflare/cloudflare v4.12.0

### Description

Create a flag that allows the usage of cloudflare_rulesets that are not complete. This flag will allow to merge new and existing rulesets per kind.

### Use cases

We have  a single Cloudflare zone (domain.io) that has multiple hosts (api-dev, api-test, api-uat...) that are being created together with our product, so we have one for dev, another test and so on.
As one extra resource that is being create is the WAF filter to allow some problematic traffic (graphql queries).

With the introduction of the resource "cloudflare_ruleset" we are not allowed to manage it in the same way and we will need to decouple the DNS/WAF from our product pipeline.

While I understand the issues as explained at #2423 I would prefer to assume the risk to create multiple resources, with the configuration per hots in multiple terraform states.

### Potential Terraform configuration

```hcl
resource "cloudflare_ruleset" "zone_custom_firewall" {
  merge_existing_rulesets = enabled

  zone_id = var.cloudflare_zone_id

  name        = "Rule for ${var.project_id}"
  description = "This rule is managed by Terraform in the project XYZ"
  kind        = "zone"
  phase       = "http_request_firewall_custom"

..
```

### References

This is similar to #2423

jacobbednarz:
closing this off as a duplicate. 

we aren't currently tracking this externally due to some unknowns. if/when this comes to fruition, the team will add support to the provider.

petero-dk:
I added a workaround using multiple repositories here: https://github.com/cloudflare/terraform-provider-cloudflare/issues/3027#issuecomment-2138260815

acdha:
I just wanted to voice support for this: like many large organizations, we have multiple parties creating rules from fully-managed DevOps projects using Terraform to security response in the console and, especially in the latter case, this spreads across applications in ways which are not trivial to model in source code structures.

[...]

Issue #2423, opened 2023-05-09
cloudflare_ruleset_rule resource


eidam:
### Current Terraform and Cloudflare provider version

Terraform v1.4.6
on darwin_arm64

### Description

Currently, all rules (in our case http_ratelimit) needs to be managed within a single **cloudflare_ruleset** resource per zone. However, we are managing rate limit policies from our Terraform modules, and not always from the same codebase. 

Introducing `cloudflare_ruleset_rule` would help to decouple rules. Once a single ruleset is created, it would be referenced by its ID from all `cloudflare_ruleset_rule` resources.

### Use cases

decouple the config/state, our stack does not allow to manage rate limit rules from a single place per zone

### Potential Terraform configuration

```hcl
resource "cloudflare_ruleset" "ruleset" {
  zone_id     = var.zone_id
  name        = "rate limit ruleset"
  kind        = "zone"
  phase       = "http_ratelimit"

  # rules are empty, managed by individual cloudflare_ruleset_rule resources
  # rules {}
}

resource "cloudflare_ruleset_rule" "example" {
  zone_id     = var.zone_id
  ruleset_id = cloudflare_ruleset.ruleset.id
  name        = "one of the rate limit rules"

  action = "block"
  ratelimit {
    requests_to_origin = true # dont count cache HITs

    characteristics = [
      "cf.colo.id",
      "ip.src"
    ]
    period              = 60
    requests_per_period = 180
    mitigation_timeout  = 600
  }

  expression  = "http.host eq \"${var.examplecom}\""
  description = "rate limit for ${var.examplecom}"
  enabled     = true
}
```

jacobbednarz:
👋 @eidam this is explicitly not supported as there isn't a way within the service to denote priority/ordering outside of providing the ruleset rules in the one payload. we've spoke about it in the past but never progressed due to the addition complexity in management that it introduces to an already complex resource. 

i'm going to close this off as won't do for now as there isn't a way of supporting it within the service. if there is ever supported added there, we can revisit the provider supporting it too.

eidam:
Oh that's unfortunate as its currently blocking us to migrate to the new ruleset engine. I would expect `cloudflare_ruleset_rule` would just follow the "[new placed at the top](https://developers.cloudflare.com/api/operations/createZoneRulesetRule )" ordering/priority, and this could be a warning on the resource documentation.

[...]

PR #7003, opened 2026-04-04
feat: Allow for ruleset composition using new rule data source

fholzer:

In the Cloudflare API, rulesets are structured as a single entity per phase, containing an array of rules. While this matches the API's design, it creates a significant bottleneck in enterprise-scale Terraform deployments.

### The Problem: Lack of Encapsulation

In large organizations, a single Zone is often managed by multiple teams (e.g., Security, Platform, and Apps). Currently, because all rules for a phase must reside within a single cloudflare_ruleset resource, it is impossible to fully encapsulate logic within independent Terraform modules.

Users are forced to centralize rule definitions, which breaks the modularity benefits of Terraform. Previous attempts to address this (e.g., #2907, #2423, #2688) show a clear demand for a more flexible approach.

### The Limitation of Current Workarounds

The common workaround is to have modules output rule objects/JSON and then flatten() or concat() them in a root module. However, this is fraught with issues:

* Strict Type Checking: Terraform requires all objects in a list to have identical schemas. If one module defines a rule with a logging block and another doesn't, Terraform throws a type mismatch error.
* Boilerplate: Users must define every possible attribute (setting unused ones to null) to satisfy the type system, which is unmaintainable as the Cloudflare provider evolves.

### Proposed Solution: A "Construction" Data Source

I propose implementing a data source similar to the AWS provider’s aws_iam_policy_document. This data source would:

* Not perform API calls: It serves as an in-memory utility to "stage" rule configurations.
* Normalize Types: It would handle the internal complexity of mapping diverse rule attributes into a consistent object type that the cloudflare_ruleset resource accepts.
* Enable Composition: Modules can each define their own "rule fragments" via this data source, which are then passed to a central ruleset resource without type-mismatch headaches.

## Example Usage

In my fork I named the utility data source "cloudflare_ruleset_rules".

```hcl
# From modules/blog/main.tf
data "cloudflare_ruleset_rules" "config" {
    phase = "http_config_settings"
    rules = [
        {
            action = "set_config"
            action_parameters = {
                security_level = "under_attack"
            }
            expression = "(http.host == \"blog.example.com\" and http.request.uri.path contains \"/admin\")"
        }
    ]
}

output "config_rules" {
  value = data.cloudflare_ruleset_rules.config.rules
}

# From modules/hr/main.tf
data "cloudflare_ruleset_rules" "config" {
    phase = "http_config_settings"
    rules = [
        {
            action = "set_config"
            action_parameters = {
                automatic_https_rewrites = true
            }
            expression = "(http.host == \"hr.example.com\" and starts_with(http.request.uri.path, \"/timetracking\"))"
        }
    ]
}

output "config_rules" {
  value = data.cloudflare_ruleset_rules.config.rules
}

# From main.tf

module "blog" {
    source = "./modules/blog"
}

module "hr" {
    source = "./modules/hr"
}

resource "cloudflare_ruleset" "test" {
    // zone id for example.com zone
    zone_id = "12345678901234567890123456789012"
    name = "default"
    kind = "zone"
    phase = "http_config_settings"
    rules = flatten([
        module.blog.config_rules,
        module.hr.config_rules
    ])
}
```

I have been working on a proof-of-concept for this and would like to gauge interest from the community and the maintainers. Opening this as MR because it seems you're only allowed to open Issues for bugs.


vaishakdinesh:
@fholzer Thank you for this! 
I think you can open an issue as a general issue and label it a feature request? I havent had a chance to look over this yet but will make some time.

musa-cf:
Closing this in favor of the issue I've opened using your PR as the contents. Thank you @fholzer! We'll forward the feature request to the Rulesets team.

PR #7026, opened 2026-04-15
test(waiting_room): Add test case for waiting room minimal idempotency

jeffa5:

## Changes being requested

Adding an idempotency test for waiting room with a minimal config.

```
go test ./internal/services/waiting_room/resource_test.go -run=TestAccCloudflareWaitingRoom_Idempotency -v
```

### Test output

```
=== RUN   TestAccCloudflareWaitingRoom_Idempotency
=== PAUSE TestAccCloudflareWaitingRoom_Idempotency
=== CONT  TestAccCloudflareWaitingRoom_Idempotency
--- PASS: TestAccCloudflareWaitingRoom_Idempotency (4.89s)
PASS
ok  	command-line-arguments	4.900s
```

### DIFF
--- a/internal/services/waiting_room/resource_test.go
+++ b/internal/services/waiting_room/resource_test.go
@@ -9,16 +9,15 @@ import (
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/acctest"
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/consts"
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/utils"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/plancheck"
-	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 )
 
 func TestMain(m *testing.M) {
 	resource.TestMain(m)
 }
-
 
 func init() {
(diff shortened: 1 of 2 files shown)


internal/services/waiting_room/schema.go
```diff
@@ -60,7 +60,9 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 			},
 			"cookie_suffix": schema.StringAttribute{
 				Description: "Appends a '_' + a custom suffix to the end of Cloudflare Waiting Room's cookie name(__cf_waitingroom). If `cookie_suffix` is \"abcd\", the cookie name will be `__cf_waitingroom_abcd`. This field is required if using `additional_routes`.",
+				Computed:    true,
 				Optional:    true,
+				Default:     stringdefault.StaticString(""),
```
musa-cf, line 65: These should be upstreamed into OpenAPI using `default: ""`. That will also force `computed_optional` configurability for the Terraform attribute, and it'll update the models to match the schema for you too.

musa-cf (changes requested):

internal/services/waiting_room/schema.go
jeffa5: Yeah, we've updated the openapi spec - happy for this PR to wait for that to be out and it can just add the test.

jeffa5: The update is now in the next branch.

rotem-cloud (changes requested):
can you change the commit message to use a conventional commit format? (ie feat/fix/chore etc)

rotem-cloud (approved):

PR #7045, opened 2026-04-21
fix(zero_trust_gateway_policy): add UseStateForUnknown to computed fields to prevent permadiff

rexscaria:
fix(zero_trust_gateway_policy): add UseStateForUnknown to computed fields to prevent permadiff
---
Description:
## Problem
`cloudflare_zero_trust_gateway_policy` produces a permadiff — fields show
`(known after apply)` on every `terraform plan` even when nothing changed.
Affects any account managing gateway policies as code: a minor unrelated
change (e.g. editing `description`) causes Terraform to flag 10+ fields as
unknown, making plans noisy and untrustworthy. Customers have reported this
affects 26 out of 105 policies in a single account.
Affected `rule_settings` sub-fields:
`allow_child_bypass`, `block_page_enabled`, `block_reason`,
`ignore_cname_category_matches`, `insecure_disable_dnssec_validation`,
`ip_categories`, `ip_indicator_feeds`, `override_host`, `override_ips`,
`resolve_dns_through_cloudflare`
Affected top-level fields:
`created_at`, `updated_at`, `deleted_at`, `read_only`, `sharable`,
`source_account`, `version`, `warning_status`
**Root cause:** These fields are `Computed: true` but missing
`UseStateForUnknown()` plan modifiers. Without them, Terraform cannot predict
their post-apply value at plan time and marks them `(known after apply)` on
every plan that touches the resource. Workarounds attempted by customers
(`lifecycle { ignore_changes = [rule_settings] }`, provider upgrades,
`generate-config-out`) do not resolve it.
## Fix
Adds `custom_schema.go` following the established pattern used by
`magic_wan_ipsec_tunnel`, `magic_wan_gre_tunnel`, and `magic_wan_static_route`.
It wraps the generated `ResourceSchema()` and appends `UseStateForUnknown()`
to only the affected fields — all other attribute definitions are untouched.
One line changed in the generated `schema.go` to call
`customResourceSchema(ctx)` instead of `ResourceSchema(ctx)`, consistent with
how Stainless auto-wires custom schemas when a `custom_schema.go` is present.
## Testing
Reproduce with any `cloudflare_zero_trust_gateway_policy` resource — make a
minor change (e.g. edit `description`) and run `terraform plan`. Before this
fix: 10+ fields show `(known after apply)`. After: only the changed field
shows a diff.
## References
- Fixes #6580
- Same pattern: #6364 (`zero_trust_device_custom_profile`),
[...]

PR #6759, opened 2026-02-12
fix(zero_trust_device_default_profile): add UseStateForUnknown

jlu-cloudflare:
…ero_trust_device_default_profile_local_domain_fallback to nested set

- [x] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested

We're simply adding `UseStateUnknown` to fields in zero_trust_device_default_profile as well as changing zero_trust_device_default_profile_local_domain_fallback to a nested set instead of list to prevent drift issues.

## Acceptance test run results

- [x] I have added or updated acceptance tests for my changes
- [x] I have run acceptance tests for my changes and included the results below

### Test output
jlu@M6YD7V50CG terraform-provider-cloudflare % TF_ACC=1 go test -v ./internal/services/zero_trust_device_default_profile/... -run 'TestAcc'
=== RUN   TestAccCloudflareZeroTrustDeviceDefaultProfile_Basic
--- PASS: TestAccCloudflareZeroTrustDeviceDefaultProfile_Basic (5.47s)
=== RUN   TestAccCloudflareZeroTrustDeviceDefaultProfile_Update
--- PASS: TestAccCloudflareZeroTrustDeviceDefaultProfile_Update (5.24s)
=== RUN   TestAccCloudflareZeroTrustDeviceDefaultProfile_Complete
--- PASS: TestAccCloudflareZeroTrustDeviceDefaultProfile_Complete (3.17s)
=== RUN   TestAccCloudflareZeroTrustDeviceDefaultProfile_WithExclude
--- PASS: TestAccCloudflareZeroTrustDeviceDefaultProfile_WithExclude (2.90s)
=== RUN   TestAccCloudflareZeroTrustDeviceDefaultProfile_WithInclude
--- PASS: TestAccCloudflareZeroTrustDeviceDefaultProfile_WithInclude (3.88s)
=== RUN   TestAccUpgradeZeroTrustDeviceDefaultProfile_FromPublishedV5
--- PASS: TestAccUpgradeZeroTrustDeviceDefaultProfile_FromPublishedV5 (17.81s)
=== RUN   TestAccCloudflareZeroTrustDeviceDefaultProfile_ServiceModeProxy
--- PASS: TestAccCloudflareZeroTrustDeviceDefaultProfile_ServiceModeProxy (3.52s)
=== RUN   TestAccCloudflareZeroTrustDeviceDefaultProfile_UnsetOptionals
[...]

PR #7112, opened 2026-05-13
fix(zero_trust_gateway_policy): remove Computed from user-configurabl…

murtas:
## Changes being requested
Removes the `Computed` flag from user-configurable `rule_settings` sub-fields that cause `(known after apply)` permadiffs on every `terraform plan` when any unrelated field changes.

Changing any field (e.g. `description`) on a `cloudflare_zero_trust_gateway_policy` resource causes 10 `rule_settings` sub-fields to show `(known after apply)`, making plans noisy and blocking effective infrastructure-as-code workflows. Customers report this affects all gateway policies managed via Terraform.

Reproduction:

```hcl
resource "cloudflare_zero_trust_gateway_policy" "example" {
  account_id  = var.account_id
  action      = "allow"
  name        = "terraform policy"
  description = "This policy is managed by Terraform"
  enabled     = false
  filters     = ["http"]
  precedence  = 5000
  rule_settings = {
    add_headers = {
      X-Custom-Header-Name = ["value"]
    }
  }
  traffic = "http.request.uri matches \".*a/partial/uri.*\""
}
```

Make a minor change (e.g. edit `description`) and run `terraform plan`:

```
  ~ resource "cloudflare_zero_trust_gateway_policy" "example" {
      ~ description    = "..." -> "..."
      ~ rule_settings  = {
          + allow_child_bypass                 = (known after apply)
          + block_page_enabled                 = (known after apply)
          + block_reason                       = (known after apply)
          + ignore_cname_category_matches      = (known after apply)
          + insecure_disable_dnssec_validation = (known after apply)
          + ip_categories                      = (known after apply)
          + ip_indicator_feeds                 = (known after apply)
          + override_host                      = (known after apply)
          + override_ips                       = (known after apply)
          + resolve_dns_through_cloudflare     = (known after apply)
        }
    }
```

Root cause:

The generated schema marks these fields as `Computed+Optional` because they appear in both the API request and response schemas. However, they are not server-computed, the API returns what the user set, or null when unset. The `Computed` flag causes Terraform to mark them unknown on every update, producing the permadiff.

Adds a `custom_schema.go` wrapper (same pattern as `magic_wan_ipsec_tunnel`, `magic_wan_gre_tunnel`, `magic_wan_static_route`) that removes `Computed` from the 10 user-configurable `rule_settings` sub-fields. This makes Terraform treat them as `Optional`-only: the plan uses the config value (or null) directly, with no unknown marking.

[UseStateForUnknown](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification) doesn't solve this because it copies the prior state value into the plan. When that value is null (field never set), the permadiff remains.

After fix:

```
  ~ resource "cloudflare_zero_trust_gateway_policy" "example" {
      ~ description    = "..." -> "..."
        # (N unchanged attributes hidden)
    }
```

The `rule_settings` block is completely absent from the diff.

Note: some top-level server-computed fields (`deleted_at`, `version`, etc.) still show `(known after apply)`. These are genuinely server-owned.

Fields changed:

All within `rule_settings`:

| Field | Before | After |
|---|---|---|
| `allow_child_bypass` | `Computed+Optional` | `Optional` | 
| `block_page_enabled` | `Computed+Optional` | `Optional` | 
| `block_reason` | `Computed+Optional` | `Optional` | 
| `ignore_cname_category_matches` | `Computed+Optional` | `Optional` | 
| `insecure_disable_dnssec_validation` | `Computed+Optional` | `Optional` | 
| `ip_categories` | `Computed+Optional` | `Optional` | 
| `ip_indicator_feeds` | `Computed+Optional` | `Optional` | 
| `override_host` | `Computed+Optional` | `Optional` | 
| `override_ips` | `Computed+Optional` | `Optional` | 
| `resolve_dns_through_cloudflare` | `Computed+Optional` | `Optional` |

Tested locally with `dev_overrides` against a live Cloudflare account:
1. Created a gateway policy with `rule_settings`
2. Changed `description` and ran `terraform plan`
3. Before fix: 10 `rule_settings` fields show `(known after apply)`
4. After fix: `rule_settings` absent from diff entirely

- Same `custom_schema.go` pattern: #6364 (`zero_trust_device_custom_profile`), #6759 (`zero_trust_device_default_profile`)
- Related (closed, not merged): #7045

### DIFF
--- a/internal/services/zero_trust_gateway_policy/custom_schema.go
+++ b/internal/services/zero_trust_gateway_policy/custom_schema.go
@@ -1,0 +1,58 @@
+package zero_trust_gateway_policy
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+)
+
+// customResourceSchema wraps the generated ResourceSchema and removes the
+// Computed flag from user-configurable rule_settings sub-fields.
+//
+// The generated schema marks these fields as Computed+Optional because they
+// appear in both the API request and response schemas. However, they are not
+// server-computed — the API returns what the user set, or null when unset.
+// The Computed flag causes Terraform to mark them unknown on every update,
+// producing noisy plans with (known after apply) on unrelated changes.
+//
+// Removing Computed makes Terraform treat them as Optional-only: the plan
+// uses the config value (or null) directly, with no unknown marking.
+func customResourceSchema(ctx context.Context) schema.Schema {
+	s := ResourceSchema(ctx)
+
+	rs := s.Attributes["rule_settings"].(schema.SingleNestedAttribute)
+	removeBoolComputed(rs.Attributes, "allow_child_bypass")
+	removeBoolComputed(rs.Attributes, "block_page_enabled")
+	removeStringComputed(rs.Attributes, "block_reason")
+	removeBoolComputed(rs.Attributes, "ignore_cname_category_matches")
+	removeBoolComputed(rs.Attributes, "insecure_disable_dnssec_validation")
+	removeBoolComputed(rs.Attributes, "ip_categories")
+	removeBoolComputed(rs.Attributes, "ip_indicator_feeds")
+	removeStringComputed(rs.Attributes, "override_host")
+	removeListComputed(rs.Attributes, "override_ips")
+	removeBoolComputed(rs.Attributes, "resolve_dns_through_cloudflare")
+	s.Attributes["rule_settings"] = rs
+
+	return s
+}
+
+func removeBoolComputed(attrs map[string]schema.Attribute, name string) {
+	if a, ok := attrs[name].(schema.BoolAttribute); ok {
+		a.Computed = false
+		attrs[name] = a
+	}
+}
+
+func removeStringComputed(attrs map[string]schema.Attribute, name string) {
+	if a, ok := attrs[name].(schema.StringAttribute); ok {
+		a.Computed = false
+		attrs[name] = a
+	}
+}
+
+func removeListComputed(attrs map[string]schema.Attribute, name string) {
+	if a, ok := attrs[name].(schema.ListAttribute); ok {
+		a.Computed = false
+		attrs[name] = a
+	}
+}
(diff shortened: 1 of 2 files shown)


internal/services/zero_trust_gateway_policy/custom_schema.go
```diff
@@ -0,0 +1,58 @@
...
+// The Computed flag causes Terraform to mark them unknown on every update,
+// producing noisy plans with (known after apply) on unrelated changes.
+//
+// Removing Computed makes Terraform treat them as Optional-only: the plan
+// uses the config value (or null) directly, with no unknown marking.
+func customResourceSchema(ctx context.Context) schema.Schema {
+	s := ResourceSchema(ctx)
+
+	rs := s.Attributes["rule_settings"].(schema.SingleNestedAttribute)
+	removeBoolComputed(rs.Attributes, "allow_child_bypass")
```
vaishakdinesh, line 24: Are they actually optional? Can a user set values for them? If so the openapi should also reflect that.

vaishakdinesh:
The provider uses the `terraform-plugin-framework` which is much more verbose in its output. Every computed attribute will be marked as `known after apply` if something is changing. 

If these attributes are indeed computed, as in the user can not provider values for them and the server sets a default, then they should remain computed. There might be an actual changes that's happening from the context of terraform causing the plan to be that.

If the change is indeed a user induced change, because of the framework used and the version we're at, the plan is accurate. Because from terraform pov, the server side values may have changed and terraform wont know until after it is applied.

internal/services/zero_trust_gateway_policy/custom_schema.go
murtas: These fields are user-settable. The Create endpoint docs (https://developers.cloudflare.com/api/resources/zero_trust/subresources/gateway/subresources/rules/methods/create/) list all 10 under the request body as optional parameters (e.g., allow_child_bypass: optional boolean, block_page_enabled: optional boolean). Aligned with bypass_parent_rule has the identical MSP restriction ("Only MSP child accounts can set this"), the same API behavior, and is already annotated as optional in the spec — it never had Computed in the generated schema.

murtas:
The Open API spec was already fixed as well to mark those fields as Optional (not computed). Once that propagates through the Stainless pipeline, the generated schema will be correct.

Issue #7049, opened 2026-04-22
cloudflare_snippet_rules Read is a no-op (no drift detection) and neither snippet resource supports terraform import


devchampps:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

  ---
  Terraform & provider version (terraform -v):
  Terraform v1.x.x
  + provider registry.terraform.io/cloudflare/cloudflare v5.x.x

  ---

### Affected resource(s)

  Affected resources:
  - cloudflare_snippet_rules
  - cloudflare_snippet

### Terraform configuration files

```hcl
resource "cloudflare_snippet_rules" "example" {
    zone_id = "<zone_id>"
    rules = [
      {
        snippet_name = "my-snippet"
        expression   = "http.request.uri.path contains \"/test\""
        enabled      = true
        description  = "Test rule"
      }
    ]
  }

  resource "cloudflare_snippet" "example" {
    zone_id      = "<zone_id>"
    snippet_name = "my-snippet"
    files = [
      {
        name    = "main.js"
        content = "export default { async fetch(req) { return fetch(req); } }"
      }
    ]
    metadata = {
      main_module = "main.js"
    }
  }
```

### Link to debug output

 (not applicable — the bugs are in the provider source, not configuration)

### Expected output

  1. After creating `cloudflare_snippet_rules `and then manually modifying a rule via the Cloudflare dashboard, running terraform plan should detect the drift and show a diff.
  2. Running terraform import `cloudflare_snippet_rules.example <zone_id>` and terraform import `cloudflare_snippet.example <zone_id>/<snippet_name>` should succeed and populate state
   with the existing configuration.

### Actual output

  ****Bug 1 — cloudflare_snippet_rules Read is a no-op:****

  In Terraform, the Read function is called during terraform plan and terraform refresh to query the provider API and sync real-world state back into Terraform's state file. For  `cloudflare_snippet_rules`, this function exists in the source but has no body — it makes no API call and returns nothing:

```
  // internal/services/snippet_rules/resource.go
  func (r *SnippetRulesResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {

  }
```

[...]

PR #7123, opened 2026-05-20
fix(snippet,snippet_rules): implement Read for snippet_rules and ImportState for both resources

maurizioabba:
## Summary

Fixes https://github.com/cloudflare/terraform-provider-cloudflare/issues/7049

This PR addresses two bugs in the `cloudflare_snippet` and `cloudflare_snippet_rules` resources:

### Bug 1: `cloudflare_snippet_rules` Read is a no-op

The `Read` function had an empty body — no API call, no state population. This meant:
- `terraform plan` never detected drift for snippet rules modified outside of Terraform
- The resource was effectively write-only after initial creation

**Fix:** Implement `Read` by calling `Snippets.Rules.List` with the zone ID, unmarshalling the response via `apijson.Unmarshal`, and setting state. Handles 404 by removing the resource from state.

### Bug 2: Neither resource supports `terraform import`

Neither `cloudflare_snippet` nor `cloudflare_snippet_rules` implemented the `ResourceWithImportState` interface, so `terraform import` failed with "Resource does not support import".

**Fix:** Add `ImportState` for both resources following established provider patterns:
- `cloudflare_snippet`: import ID format `<zone_id>/<snippet_name>`
- `cloudflare_snippet_rules`: import ID format `<zone_id>` (rules are zone-scoped singletons)

The `cloudflare_snippet` import makes two API calls (same as Read): `Snippets.Get` for metadata and `Snippets.Content.Get` for the multipart file content.

### Testing

Both changed files pass `go vet`. The changes follow the same patterns used by 193+ other resources in this provider (e.g., `waiting_room_rules`, `zone_lockdown`, `ruleset`).

### DIFF
--- a/internal/services/snippet/resource.go
+++ b/internal/services/snippet/resource.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cloudflare/cloudflare-go/v6/option"
 	"github.com/cloudflare/cloudflare-go/v6/snippets"
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/apijson"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/importpath"
 	"github.com/cloudflare/terraform-provider-cloudflare/internal/logging"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/types"
@@ -20,6 +21,7 @@ import (
 // Ensure provider defined types fully satisfy framework interfaces.
 var _ resource.ResourceWithConfigure = (*SnippetResource)(nil)
 var _ resource.ResourceWithModifyPlan = (*SnippetResource)(nil)
+var _ resource.ResourceWithImportState = (*SnippetResource)(nil)
 
 func NewResource() resource.Resource {
 	return &SnippetResource{}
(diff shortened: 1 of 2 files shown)


internal/services/snippet_rules/resource.go
```diff
@@ -168,6 +206,47 @@ func (r *SnippetRulesResource) Delete(ctx context.Context, req resource.DeleteRe
 	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
 }
 
+func (r *SnippetRulesResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
```
vaishakdinesh, line 209: This shouldnt be fully done via custom code

internal/services/snippet/resource.go
```diff
@@ -235,6 +237,75 @@ func (r *SnippetResource) Delete(ctx context.Context, req resource.DeleteRequest
 	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
 }
 
+func (r *SnippetResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
```
vaishakdinesh, line 240: Shouldnt be done via custom code

maurizioabba:
Closing in favor of a config-driven approach via cloudflare-config. The import functionality and snippet_rules Read fix should be generated from the Stainless config rather than hand-written.

PR #7124, opened 2026-05-21
fix(zero_trust_gateway_certificate): accept activate arg in create call

alyssamw:

This ensures that the activate argument for this resource is processed during create, not only update. This behavior was previously added by https://github.com/cloudflare/terraform-provider-cloudflare/pull/6248

```
export TF_ACC=1
export CLOUDFLARE_ACCOUNT_ID=<account_id>
export CLOUDFLARE_API_TOKEN=<api_token>

go test ./internal/services/zero_trust_gateway_certificate/... \
        ./internal/services/zero_trust_gateway_certificate/... \
        -v -run TestAcc -count 1 -timeout 30m

```
### Test output
% go test ./internal/services/zero_trust_gateway_certificate/... ./internal/services/zero_trust_gateway_certificate/... -v -run TestAccCloudflareTeamsCertificate_CreateWith -count 1 -timeout 1m
=== RUN   TestAccCloudflareTeamsCertificate_CreateWithActivate
--- PASS: TestAccCloudflareTeamsCertificate_CreateWithActivate (6.14s)
=== RUN   TestAccCloudflareTeamsCertificate_CreateWithDeactivate
--- PASS: TestAccCloudflareTeamsCertificate_CreateWithDeactivate (2.17s)


internal/services/zero_trust_gateway_certificate/resource.go
```diff
@@ -92,6 +95,68 @@ func (r *ZeroTrustGatewayCertificateResource) Create(ctx context.Context, req re
...
+			_, err := r.client.ZeroTrust.Gateway.Certificates.Activate(
+				ctx,
+				data.ID.ValueString(),
+				zero_trust.GatewayCertificateActivateParams{
+					AccountID: cloudflare.F(data.AccountID.ValueString()),
+					Body:      struct{}{}, // Empty body as required by the API
+				},
+				option.WithMiddleware(logging.Middleware(ctx)),
+			)
+			if err != nil {
```
vaishakdinesh: If there's an error here and we return, then the cert gets created on the account but not tracked in state. A subsequent terraform apply will try to create it again and conflict

internal/services/zero_trust_gateway_certificate/resource.go
```diff
@@ -92,6 +95,68 @@ func (r *ZeroTrustGatewayCertificateResource) Create(ctx context.Context, req re
...
+		_, err = r.client.ZeroTrust.Gateway.Certificates.Get(
+			ctx,
+			data.ID.ValueString(),
+			zero_trust.GatewayCertificateGetParams{
+				AccountID: cloudflare.F(data.AccountID.ValueString()),
+			},
+			option.WithResponseBodyInto(&res),
+			option.WithMiddleware(logging.Middleware(ctx)),
+		)
+		if err != nil {
```
vaishakdinesh, line 127: I think the same happens if the GET fails.

internal/services/zero_trust_gateway_certificate/resource.go
```diff
@@ -92,6 +95,68 @@ func (r *ZeroTrustGatewayCertificateResource) Create(ctx context.Context, req re
...
+					AccountID: cloudflare.F(data.AccountID.ValueString()),
+					Body:      struct{}{}, // Empty body as required by the API
+				},
+				option.WithMiddleware(logging.Middleware(ctx)),
+			)
+			if err != nil {
+				resp.Diagnostics.AddError("failed to activate certificate", err.Error())
+				return
+			}
+		} else {
```
vaishakdinesh: Would this be executed on every fresh create? Are certs inactive by default? Can a user specify active on the hcl?

alyssamw: Certs are inactive by default -- removed this block that attempts to deactivate

internal/services/zero_trust_gateway_certificate/resource.go
alyssamw: A user can create a new certificate with the same exact parameters without a conflict. Is there a way to save the cert in the state but still return an error? The state would be out of sync anyways, because the activate field would not be accurate. So in my opinion it makes sense to either import the failed cert and respecify activate, or create a new cert.

vaishakdinesh: >Is there a way to save the cert in the state but still return an error?

if its created, yes. I was calling out the activation part failing. In which case the cert is already present on cloudflare.

>So in my opinion it makes sense to either import the failed cert and respecify activate, or create a new cert.

well the apply fails, it stops the entire apply. This is fine if the user is only configuring this one resource.

alyssamw: Hi Vaishak, asking this question here since it got lost in chat -- is there a way to return the error without failing the entire create/apply? I don't see anything similar in the repo

internal/services/zero_trust_gateway_certificate/resource_schema_test.go
```diff
@@ -5,15 +5,16 @@ package zero_trust_gateway_certificate_test
 import (
 	"context"
 	"fmt"
-	"github.com/cloudflare/cloudflare-go"
+	"os"
+	"testing"
+
+	cfv6 "github.com/cloudflare/cloudflare-go/v6"
```
musa-cf: Can you switch to using the normal v7 Go SDK? Was this a workaround for failing builds earlier?

It was updated recently and shouldn't cause build issues anymore!

alyssamw: Done! I was moreso updating the tests as they were on the older SharedV1Client... but I guess my autofill picked v6 instead of v7 :'(

Issue #7160, opened 2026-06-07
cloudflare_account: Update fails with "missing required account_id parameter"


synmux:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Cloudflare provider: v5.19.1 (latest)

### Affected resource(s)

`cloudflare_account`

### Terraform configuration files

```hcl
resource "cloudflare_account" "example" {
  name = "example"
}
```

### Link to debug output

n/a

### Panic output

```plaintext
│ Error: failed to make http request
│
│   with cloudflare_account.example,
│   on main.tf line 1, in resource "cloudflare_account" "example":
│    1: resource "cloudflare_account" "example" {
│
│ [ERROR] missing required account_id parameter: failed to make http request
```

### Expected output

The account should update successfully (or be a no-op if nothing changed).

### Actual output

Any terraform apply that triggers an update on a cloudflare_account resource fails before an HTTP request is even made. The cloudflare-go client rejects the call because AccountID is not set in AccountUpdateParams.

Root cause

In internal/services/account/resource.go, the Update() function passes an empty params struct:

```go
_, err = r.client.Accounts.Update(
    ctx,
    accounts.AccountUpdateParams{},  // AccountID never set
    option.WithRequestBody("application/json", dataBytes),
    ...
)
```

However, AccountUpdateParams.AccountID is declared api:"required" in cloudflare-go, so the client rejects it client-side:

```go
// cloudflare-go accounts/account.go
type AccountUpdateParams struct {
    AccountID param.Field[string] `path:"account_id" api:"required"`
    Account   AccountParam        `json:"account" api:"required"`
}
```

Partial fix reference

Commit 0f55796 ("fix: clean up global account and zone id, account resource path param", included in v5.19.1) fixed the same missing-param pattern for Read and Delete in the same file, but Update was not included in that change.

Suggested fix

Apply the same pattern used for the Read fix:

```go
_, err = r.client.Accounts.Update(
    ctx,
    accounts.AccountUpdateParams{
        AccountID: cloudflare.F(state.ID.ValueString()),
    },
    option.WithRequestBody("application/json", dataBytes),
    ...
)
```

### Steps to reproduce

[...]

PR #7162, opened 2026-06-09
fix: make sure required AccountID is passed for Update

TomTucka:

## Changes being requested

As described in #7160, Updates to the account recourse fail because the required value of AccountID is never passed.

## Additional context & links

https://github.com/cloudflare/terraform-provider-cloudflare/issues/7160


vaishakdinesh:
fixed in https://github.com/cloudflare/terraform-provider-cloudflare/blob/release-tf-v5.22.0/internal/services/account/resource.go#L119

Issue #7152, opened 2026-06-02
cloudflare_worker_route - RawState had no JSON or flatmap data set error when migrating v4 to v5


scott-pearson:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

1.3.2

### Affected resource(s)

cloudflare_worker_route

### Terraform configuration files

```hcl
{
            "instances": [
                {
                    "attributes_flat": {
                        "enabled": "true",
                        "id": "REDACTED",
                        "multi_script": "false",
                        "pattern": "REDACTED",
                        "zone": "REDACTED",
                        "zone_id": "REDACTED"
                    },
                    "schema_version": 0,
                    "sensitive_attributes": []
                }
            ],
            "mode": "managed",
            "name": "REDACTED",
            "provider": "provider[\"registry.terraform.io/cloudflare/cloudflare\"]",
            "type": "cloudflare_worker_route"
        },
```

### Link to debug output

https://gist.github.com/scott-pearson/8bab40121e192fa1ab16e861ee776928

### Panic output

tack trace from the terraform-provider-cloudflare_v5.19.0 plugin:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=[REDACTED_ADDRESS]]
goroutine 8 [running]:
github.com/cloudflare/terraform-provider-cloudflare/internal/services/workers_route/migration/v500.MoveState({[REDACTED_ADDRESS], [REDACTED_ADDRESS]}, {[REDACTED_ADDRESS], {[REDACTED_ADDRESS], 0x2b}, [REDACTED_ADDRESS], 0x0, 0x0, {[REDACTED_ADDRESS], 0x17}, ...}, ...)
github.com/cloudflare/terraform-provider-cloudflare/internal/services/workers_route/migration/v500/move_state.go:29 +0x255
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).MoveResourceState([REDACTED_ADDRESS]?, {[REDACTED_ADDRESS], [REDACTED_ADDRESS]}, [REDACTED_ADDRESS], [REDACTED_ADDRESS])
[...]

PR #7163, opened 2026-06-09
fix: guard against nil SourceState in MoveState migrations

ssicard:

## Changes being requested
The v5 provider's MoveState implementations panic with a nil pointer dereference when the source Terraform state uses the legacy attributes_flat (flatmap) format from pre-Terraform 0.12. The framework intentionally leaves req.SourceState nil when it cannot unmarshal the source state, but our code calls req.SourceState.Get() without checking.

This adds a nil guard to all 32 MoveState implementations. Instead of crashing, the provider now returns a clear diagnostic telling the user to run terraform apply -refresh-only with the v4 provider to normalize their state before retrying the migration.

Also adds a prerequisite callout to the v5 upgrade guide.
Fixes https://github.com/cloudflare/terraform-provider-cloudflare/issues/7152

## Acceptance Test Results

`TF_ACC=1 ./scripts/test`

### DIFF
--- a/internal/services/argo_smart_routing/migration/v500/move.go
+++ b/internal/services/argo_smart_routing/migration/v500/move.go
@@ -3,6 +3,7 @@ package v500
 import (
 	"context"
 
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/migrations"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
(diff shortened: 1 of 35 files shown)


mgirouard (approved):

internal/services/argo_smart_routing/migration/v500/move.go
```diff
@@ -26,6 +26,16 @@ func MoveArgoToSmartRouting(ctx context.Context, req resource.MoveStateRequest,
 
 	// Parse the source state (legacy v4 cloudflare_argo format)
 	var sourceState SourceCloudflareArgoModel
+	if req.SourceState == nil {
+		resp.Diagnostics.AddError(
+			"Unable to Read Source State",
+			"The source state for "+req.SourceTypeName+" could not be decoded. "+
```
mgirouard: Nit: This (and others) could probably be tightened up with a strings.Sprintf + a constant since it's the same across all resources.

PR #7137, opened 2026-05-27
fix(load_balancer_pool): add UseStateForUnknown for load_shedding attribute

emmanuelflores-cf:
Resolves issue with `load_balancer_pool` `load_shedding` attribute(s) producing continual Terraform plan drift across applies by adding `UseStateForUnknown` plan modifier. This plan modifier should persists the prior Terraform state for field(s) that are not explicitly configured.

- [X] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested

- Add `UseStateForUnknown` plan modifier to `load_balancer_pool` `load_shedding` attribute

Reference commit: https://github.com/cloudflare/terraform-provider-cloudflare/commit/248c746ec

## Acceptance test run results

Noting that the original reference commit does not add/modify acceptance tests. It may not be strictly necessary for this changes since it is relatively minor, but I can proceed with additional testing if needed.

- [x] I have added or updated acceptance tests for my changes
- [ ] I have run acceptance tests for my changes and included the results below

## Additional context & links

For additional context, please see Cloudflare internal escalation: https://jira.cfdata.org/browse/ESCALATION-2542

Issue #7028, opened 2026-04-16
Constant drift on turnstile & LB_pool resources when not alphabetically ordered


mveroone:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```# tf -v
Terraform v1.14.7
on darwin_arm64
+ provider registry.terraform.io/cloudflare/cloudflare v5.18.0
```

### Affected resource(s)

* `cloudflare_load_balancer_pool`
* `cloudflare_turnstile_widget`
(And maybe more

### Terraform configuration files

```hcl
resource "cloudflare_turnstile_widget" "example" {
  account_id = "abc123def456"
  name       = "my-widget"
  mode       = "invisible"

  # Declared in numeric-friendly order
  # During apply, 10 will be reordered between 1 & 2
  domains = [
    "example.io",
    "staging-1.example.com",
    "staging-2.example.com",
    ...
    "staging-10.example.com",
    "staging-11.example.com"
  ]
}
resource "cloudflare_load_balancer_pool" "checkout" {
  account_id = "xyz789abc123"
  name       = "web-checkout"

  # Declared in logical order: primary pool, then secondary, then backup
  # During apply, provider attempts to reorder alphabetically
  origin {
    name    = "primary-pool"
    address = "checkout.prod.internal"
    weight  = 0.5
  }

  origin {
    name    = "legacy-pool"
    address = "legacy-checkout.prod.internal"
    weight  = 0.5
  }

  origin {
    name    = "backup-pool"
    address = "backup.prod.internal"
    weight  = 0
  }

  origin {
    name    = "canary-pool"
    address = "canary.prod.internal"
    weight  = 0
  }
}
```

### Link to debug output

Sorry I tried to get a DEBUG log, but even the terraform resource names reveal too many corporate information.
I can see however that the payload is passed in the right order and the API response contains the re-ordered array along with confirmation.
### Panic output

N/A

### Expected output

Reordering silently computed by the provider and no changes applied if the list is the same as order doesn't matter.

### Actual output

Constant resource update to attempt reordering the domains in turnstile or origins in LB Pool

### Steps to reproduce

1. Create a turnstile widget where domains aren't alphabetically ordered in the list
[...]

Issue #6140, opened 2025-09-11
cloudflare_load_balancer_pool keeps being changed


darkweaver87:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

Terraform v1.13.0
Cloudflare v5.9.0

### Affected resource(s)

cloudflare_load_balancer_pool

### Terraform configuration files

```hcl
resource "cloudflare_load_balancer_pool" "xxx" {
  account_id = data.cloudflare_accounts.xxx.result[0].id

  name = "cluster"

  monitor = cloudflare_load_balancer_monitor.tcp_443.id

  origins = [
    {
      name    = "clusterB"
      address = "x.x.x.y"
      weight  = 0
      enabled = true
    },
    {
      name    = "clusterA"
      address = "x.x.x.x"
      weight  = 1
      enabled = true
  }]
}
```

### Link to debug output

you don't need it :-)

### Expected output

Runs should be consistent.

### Actual output

`origins` list keeps being updated.

### Steps to reproduce

1. create a `cloudflare_load_balancer_pool` with 2 origins
2. order origins items descending
3. apply
4. run a `terraform plan`

### Additional factoids

This bugs only happens when origins items are not ordered by name ascending.
The diff should be done on name ordered lists.

TyranVdMerwePaytec:
It keeps changing even when there is a single origin too.

TyranVdMerwePaytec:
The issue seems to be caused by the API omitting properties from the response.

**Examples:**

- disabled_at
- load_shedding, load_shedding.default_percent, load_shedding.session_percent
- notification_filter.origin.disable
- notification_filter.pool.disable
-  origins[].disabled_at (when enabled), origins[].port (when set to 0),

rrabelloo:
Same here!
I've tested many resources combinations including/excluding default values.

Tested with:
- Terraform 1.13.0 / 1.13.1
- Terragrut 0.87.0 / 0.86.2
- Cloudflare provider 5.10.1 / 5.9.0 / 5.8.4

<details>
<summary>Plan output</summary>

```
# cloudflare_load_balancer_pool.pools["lb-pool-test"] will be updated in-place
[...]

PR #7202, opened 2026-06-19
fix: stop perpetual plan diff on load_balancer_pool and turnstile_widget

andrei-marinache:

## Changes being requested

`cloudflare_load_balancer_pool` and `cloudflare_turnstile_widget` plan an in-place
update on every apply even when nothing has changed. Two causes:

- The API returns a list in a different order than the config: `origins` for the
  pool, `domains` (sorted) for the widget. Both are ordered lists, so any other
  order shows as a permanent diff (#7028).
- Several computed attributes are not returned on every read, so they replan as
  `(known after apply)`: for the pool `disabled_at`, `networks`,
  `notification_filter`, `origin_steering`, `load_shedding`, and per-origin
  `disabled_at`/`port`; for the widget `created_on`, `secret`, `clearance_level`,
  `ephemeral_id`, `offlabel` (#6140 tracks this for the pool).

Both resources already declare `ResourceWithModifyPlan` with empty methods. This
fills them in: the list is rebuilt in the prior (state) order and the computed
values are taken from state when the plan leaves them unknown. The result is then
compared against prior state. If it matches, the run was pure churn (and
`modified_on` is pinned too, so the plan is empty); if not, there is a real
change, so the original order and `modified_on` are restored and the change plans
normally.

On the overlap with #7137: that change added `UseStateForUnknown` on the pool's
`load_shedding`, but the stock modifier returns early when the prior state is
null, which is the case for pools that never set these fields. Handling it in
`ModifyPlan` covers that case and the other fields without touching the generated
`schema.go`.

Addresses #7028. Also addresses the `load_balancer_pool` part of #6140.

Each test creates the resource, re-applies the same config (no-op), re-applies
with the list in a different order (no-op), then makes a real change (update):

```
TF_ACC=1 CLOUDFLARE_ACCOUNT_ID=… CLOUDFLARE_API_TOKEN=… \
  go test ./internal/services/load_balancer_pool/ ./internal/services/turnstile_widget/ \
    -run 'NoReorderOrComputedDrift|NoDomainReorderDrift' -v
```

### Test output

```
--- PASS: TestAccCloudflareLoadBalancerPool_NoReorderOrComputedDrift (10.35s)
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/services/load_balancer_pool 12.893s
--- PASS: TestAccCloudflareTurnstileWidget_NoDomainReorderDrift (14.18s)
ok      github.com/cloudflare/terraform-provider-cloudflare/internal/services/turnstile_widget   18.618s
```

## Additional context & links

- #7028
- #6140
- #7137


ssicard:
This change is being made internally and will be available in a future terraform release.

PR #7226, opened 2026-06-26
fix(zero_trust_access_ai_controls_mcp_portal): migrate servers state to set

Michael9127:
## Changes being requested

Fixes state compatibility for `cloudflare_zero_trust_access_ai_controls_mcp_portal` after v5.22.0 changed `servers` from a List to a Set without incrementing the schema version.

- Models `servers` as an order-independent Set.
- Bumps the resource schema from version 500 to 501.
- Migrates version 0 and 500 List-shaped state to Set state.
- Preserves null, unknown, empty, and nested unknown values.
- Reads `server_id` during refresh and import by removing the stale `no_refresh` tag.
- Adds migration unit tests and an acceptance test covering reorder stability and import.
- Updates the generated resource documentation.

The API/spec fixes providing `server_id` and Set semantics are already deployed. Both List- and Set-era version-500 states serialize as JSON arrays, allowing the migration to support v5.21.1 and v5.22.0 state.

- [x] I have added or updated acceptance tests for my changes.
- [x] I have run acceptance tests for my changes.

```text
--- PASS: TestAccZeroTrustAccessAIControlsMcpPortal_serversSetOrderIndependent (15.26s)
PASS
Additional verification:
go test ./internal/services/zero_trust_access_ai_controls_mcp_portal -count=1
./scripts/check-schema-versions
git diff --check
Ticket: MCP-203

### DIFF
--- a/docs/resources/zero_trust_access_ai_controls_mcp_portal.md
+++ b/docs/resources/zero_trust_access_ai_controls_mcp_portal.md
@@ -117,4 +117,3 @@ Import is supported using the following syntax:
 $ terraform import cloudflare_zero_trust_access_ai_controls_mcp_portal.example '<account_id>/<id>'
 ```
 
-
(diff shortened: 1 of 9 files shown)


ssicard:
This was fixed in v5.24.0

PR #7236, opened 2026-07-01
fix(migrations): handle v4→v5 state upgrader schema drift (ruleset asset_name, lb_pool flatten_cname)

vaishakdinesh:
## Summary

Two silent regressions in the v4→v5 state upgraders that surface when a full
v4-apply → v5-plan pipeline actually runs. They've been dormant since May
because the tf-migrate e2e workflow was running against pre-existing state (no-op
refresh), so the state upgraders were never exercised. They fire the moment CI
recreates resources from an empty state — which is what we hit today.

Both are one-file logic fixes; existing migration test suites keep passing.

## 1. Ruleset — v4 source schema missing `asset_name`

The `asset_name` attribute on `action_parameters` was added to the v4 provider
in [3d97253 (2026-05-28)][1] and shipped in v4.52.8. Any v4 provider from that
release onward writes state containing `asset_name` (typically null unless the
user configures it).

The v4→v5 state upgrader parses the incoming raw JSON using
`SourceV4RulesetSchema()`. That schema was last touched in March and doesn't
declare `asset_name`, so raw-state unmarshal fails with:

```
Could not parse raw state as v4 format:
AttributeName("rules").ElementKeyInt(0).AttributeName("action_parameters").ElementKeyInt(0).AttributeName("asset_name"):
unsupported attribute "asset_name"
```

Fix:
- Add `asset_name` to `SourceV4RulesetSchema` (`schema.go`).
- Add `AssetName types.String` to `SourceV4ActionParametersModel` (`model.go`).
- In `transformActionParameters` (`transform.go`), forward `src.AssetName`
  through to the v5 model instead of nulling it out (the code was assuming
  asset_name didn't exist in v4).

[1]: https://github.com/cloudflare/terraform-provider-cloudflare/commit/3d972534

## 2. Load balancer pool — `SourceOriginsModel` has a phantom `flatten_cname`

The v4 provider's `cloudflare_load_balancer_pool` origins block has no
`flatten_cname` attribute (verified against v4.52.7 and v4.52.8: 0 hits).
So v4 state never contains it.

But `SourceOriginsModel` (the v4-side model used to parse that state)
declares `FlattenCNAME types.Bool`. When the framework tries to convert
the object into the struct, it fails:

```
An unexpected error was encountered trying to convert tftypes.Value into
v500.SourceOriginsModel. This is always an error in the provider.
mismatch between struct and object: Struct defines fields not found in object: flatten_cname.
```

Also confirmed by inspecting the v500 migration schema
(`internal/services/load_balancer_pool/migration/v500/schema.go`), which
correctly does **not** declare `flatten_cname` under `origins`.

Fix:
- Remove `FlattenCNAME` from `SourceOriginsModel` in
  `internal/services/load_balancer_pool/migration/v500/model.go`.
- Leave it on `TargetLoadBalancerPoolOriginsModel` (v5) since v5 introduced
  the attribute.

## Verification

- `go build ./...` — clean.
- `go test ./internal/services/ruleset/migration/...` — ok.
- `go test ./internal/services/load_balancer_pool/migration/...` — ok.

## Impact

Unblocks the tf-migrate e2e workflow, which is currently failing at v5-plan
with 27 ruleset unmarshal errors + 24 load_balancer_pool value-conversion
errors — both caused by these two schema drift bugs.

### DIFF
--- a/internal/services/load_balancer_pool/migration/v500/model.go
+++ b/internal/services/load_balancer_pool/migration/v500/model.go
@@ -40,13 +40,20 @@ type SourceCloudflareLoadBalancerPoolModel struct {
 
 // SourceOriginsModel represents a single origin within the pool from v4.x provider.
 // Stored in TypeSet, so in state it's an array of these objects.
+//
+// NOTE: flatten_cname is intentionally NOT included here. The v4 provider
+// (SDKv2 cloudflare_load_balancer_pool) has no flatten_cname attribute on
+// origins, so v4 state never contains it. Declaring it on this Source model
+// causes tftypes → struct conversion to fail with:
+//   "Struct defines fields not found in object: flatten_cname"
+// It appears on TargetLoadBalancerPoolOriginsModel (v5) below because v5
+// introduced the field.
 type SourceOriginsModel struct {
-	Name             types.String       `tfsdk:"name"`
-	Address          types.String       `tfsdk:"address"`
-	VirtualNetworkID types.String       `tfsdk:"virtual_network_id"`
-	Weight           types.Float64      `tfsdk:"weight"`
-	Enabled          types.Bool         `tfsdk:"enabled"`
-	FlattenCNAME     types.Bool                                     `tfsdk:"flatten_cname"`
+	Name             types.String        `tfsdk:"name"`
+	Address          types.String        `tfsdk:"address"`
+	VirtualNetworkID types.String        `tfsdk:"virtual_network_id"`
+	Weight           types.Float64       `tfsdk:"weight"`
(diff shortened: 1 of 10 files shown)


vaishakdinesh:
The migration fix commit was cherry-picked into the v5.22.0 release PR #7253 (branch `release-tf-v5.22.0`):

- [`baa1b9267b`](https://github.com/cloudflare/terraform-provider-cloudflare/commit/baa1b9267b) — `fix(migrations): handle v4→v5 state upgrader schema drift`

Closing this PR since the fix will ship with v5.22.0. The other commits on this branch (snippet import fix, agent skills) belong to `next` and can go through their own release path.

PR #7285, opened 2026-07-31
docs: correct pre-1.8 rename migration guidance

### LINKED ISSUE #7283: Misleading migration guide documentation for terraform <1.8
Migration [documentation](https://github.com/cloudflare/terraform-provider-cloudflare/blob/main/docs/guides/version-5-migration.md) for v4-v5 states that in case you use terraform version <1.8 it's possible to use "terraform state mv" command in order to migrate renamed resources.

> Using terraform state mv (Terraform < 1.8)
> If you are on a Terraform version older than 1.8, you cannot use moved blocks for cross-resource-type renames. Instead, use terraform state mv to rename resources directly in the state. The provider's state upgraders will automatically transform the state on the next terraform plan or terraform apply.

It's not true and according to the official terraform [documentation](https://developer.hashicorp.com/terraform/cli/v1.5.x/commands/state/mv) the command handles the address change only in case it's the same resource type.

> Both the source and destination addresses must use resource address syntax, and they must both refer to the same kind of object: you can only move a resource instance to another resource instance, a whole module instance to another whole module instance, etc. Furthermore, if you are moving a resource or a resource instance then you can only move it to a new address with the same resource type.

terraform – 1.5.7
provider – migrating from 4.52.5 to 5.19.0

### Affected resource(s)

all resources with changed resource type

### Expected output

resource moved

### Actual output

```
│ Error: Invalid state move request
│ 
│ Cannot move cloudflare_argo.main to cloudflare_argo_smart_routing.main: resource types don't match.

```

lntutor:

## Changes being requested

Correct the v5 migration guide for Terraform versions older than 1.8. `terraform state mv` cannot move state between different resource types, so the guide now recommends Terraform 1.8+ moved blocks or a state removal and import fallback.

The editable guide template and generated documentation are updated together.

Fixes #7283

Not applicable for this documentation-only change. Verified with `git diff --check`.

### Test output

```
git diff --check
```

## Additional context & links

Terraform documents that `state mv` requires source and destination resources to have the same type.

### DIFF
--- a/docs/guides/version-5-migration.md
+++ b/docs/guides/version-5-migration.md
@@ -45,8 +45,8 @@ Before starting the migration:
 - **Terraform 1.8+** is recommended if you use any
   [renamed resources](#resource-rename-reference). It enables `moved` blocks,
   which provide the smoothest migration experience. On older Terraform
-  versions, you can use `terraform state mv` instead -- see
-  [Using `terraform state mv` (Terraform < 1.8)](#using-terraform-state-mv-terraform--18).
+  versions, cross-resource-type moves are not supported; use
+  `terraform state rm` followed by `terraform import` instead.
 - **tf-migrate** -- Install the [tf-migrate] CLI tool for automatic HCL
   configuration changes. Download the latest release for your platform from
   the [tf-migrate releases page](https://github.com/cloudflare/tf-migrate/releases):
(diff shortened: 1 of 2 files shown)


lntutor:
Current head `d24503b6afba76bcbafc6a4e2f0b5752e5b24a92` is mergeable, and there are no unresolved review threads on the current PR state. This looks ready for maintainer review when convenient.

PR #7291, opened 2026-08-04
feat(ruleset): add dry-run validation to plan

jahnavimalhotra1:

## Changes being requested
Configuration errors in Cloudflare Rulesets are currently reported during `terraform apply`, because the provider does not call the API until it writes. This means a plan can look completely healthy and then fail partway through an apply. This change adds dry-run validation for `terraform plan`, allowing customers to validate their setups beforehand and catch errors early.

We validate the planned configuration against the API during terraform plan by issuing the same request the apply would issue, with `?dry_run=true`. The Rulesets API performs full validation and returns the same errors it would for a real write, but does not persist anything.

```
export TF_ACC=1
export CLOUDFLARE_API_TOKEN=<token>
export CLOUDFLARE_ACCOUNT_ID=<account id>
export CLOUDFLARE_ZONE_ID=<zone id>
go test ./internal/services/ruleset/ -run '^TestAccCloudflareRuleset_DryRun' -v -timeout 15m
```

### Test output
 
```
=== RUN   TestAccCloudflareRuleset_DryRunInvalidOnCreate
--- PASS: TestAccCloudflareRuleset_DryRunInvalidOnCreate (2.67s)
=== RUN   TestAccCloudflareRuleset_DryRunInvalidOnUpdate
--- PASS: TestAccCloudflareRuleset_DryRunInvalidOnUpdate (5.28s)
=== RUN   TestAccCloudflareRuleset_DryRunSkippedWhenDependencyIsUnknown
--- PASS: TestAccCloudflareRuleset_DryRunSkippedWhenDependencyIsUnknown (1.23s)
=== RUN   TestAccCloudflareRuleset_DryRunDependencyOnCreateThenInvalidUpdate
--- PASS: TestAccCloudflareRuleset_DryRunDependencyOnCreateThenInvalidUpdate (3.78s)
=== RUN   TestAccCloudflareRuleset_DryRunSkippedWhenNothingChanges
--- PASS: TestAccCloudflareRuleset_DryRunSkippedWhenNothingChanges (4.65s)
=== RUN   TestAccCloudflareRuleset_DryRunSkippedOnDelete
--- PASS: TestAccCloudflareRuleset_DryRunSkippedOnDelete (5.11s)
=== RUN   TestAccCloudflareRuleset_DryRunSkippedOnReplacement
--- PASS: TestAccCloudflareRuleset_DryRunSkippedOnReplacement (5.25s)
=== RUN   TestAccCloudflareRuleset_DryRunSkippedOnReferencedDelete
--- PASS: TestAccCloudflareRuleset_DryRunSkippedOnReferencedDelete (7.92s)
=== RUN   TestAccCloudflareRuleset_DryRunSkippedOnReferencedReplacement
--- PASS: TestAccCloudflareRuleset_DryRunSkippedOnReferencedReplacement (8.05s)
=== RUN   TestAccCloudflareRuleset_DryRunSkippedOnEntryPointRename
--- PASS: TestAccCloudflareRuleset_DryRunSkippedOnEntryPointRename (6.22s)
=== RUN   TestAccCloudflareRuleset_DryRunSkippedOnInvalidEntryPointReplacement
--- PASS: TestAccCloudflareRuleset_DryRunSkippedOnInvalidEntryPointReplacement (6.99s)
=== RUN   TestAccCloudflareRuleset_DryRunInvalidUpdateWithNewDependency
--- PASS: TestAccCloudflareRuleset_DryRunInvalidUpdateWithNewDependency (5.79s)
PASS
```


internal/services/ruleset/testdata/TestAccCloudflareRuleset_DryRunInvalidUpdateWithNewDependencyZone/2.tf
zakcutner, line 1: Let's add two more cases:
1. Dependency during creation, and then update with validation error.
2. No dependency during creation, and then update with dependency and validation error.

internal/services/ruleset/ruleset_test.go
zakcutner, line 1: Let's run each test case for an account-level and zone-level ruleset

internal/services/ruleset/resource.go
```diff
@@ -430,3 +446,158 @@ func transformQueryStringJSON(jsonBytes []byte) []byte {
...
+	diagnostics.AddWarning(
+		"Could not validate against the API",
+		"The planned configuration could not be validated, so errors in it may "+
+			"only be reported during apply.\n\n"+err.Error(),
+	)
+}
+
+// planIsKnownForDryRun reports whether enough of the plan is known to build a
+// request that the API can validate
+func planIsKnownForDryRun(plan *RulesetModel) bool {
```
zakcutner: I think we want to check if anything in the plan (recursively) is unknown.

internal/services/ruleset/ruleset_test.go
```diff
@@ -340,6 +341,64 @@ func TestAccCloudflareRuleset_Name(t *testing.T) {
 	})
 }
 
+var missingRulesetErrorPattern = regexp.MustCompile(`20226`)
```
zakcutner: Let's do something a little more robust (checking the shape of the error, etc.)

internal/services/ruleset/testdata/TestAccCloudflareRuleset_DryRunDependencyOnCreateThenInvalidUpdateZone/1.tf
zakcutner, line 1: I think it might be simpler to test account and zone rulesets in the same test (I believe you can have multiple resources per test). Will also make it easier to ensure that the account and zone tests stay in sync

internal/services/ruleset/testdata/TestAccCloudflareRuleset_DryRunDependencyOnCreateThenInvalidUpdateAccount/2.tf
zakcutner, line 1: You're removing the description in this step, is that intended?

internal/services/ruleset/resource.go
```diff
@@ -309,6 +310,25 @@ func (r *RulesetResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 		return
 	}
 
+	// Validate the plan once the planned rules below have been populated
+	defer func() {
```
zakcutner, line 323: I'm not sure if we want this in a `defer`?

internal/services/ruleset/resource.go
```diff
@@ -309,6 +310,25 @@ func (r *RulesetResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
...
+			return
+		}
+
+		// A ruleset that is not changing is not written during apply, so skip dry-run
+		if !req.State.Raw.IsNull() && !resp.Plan.Raw.IsNull() && req.State.Raw.Equal(resp.Plan.Raw) {
+			return
+		}
+
+		// Check if enough of the plan is known to build a request for the dry-run
+		if !req.Config.Raw.IsFullyKnown() {
```
zakcutner, line 334: You're checking `req.Config` here, but you later use `plan` (derived from `req.Plan`) instead

internal/services/ruleset/resource.go
```diff
@@ -430,3 +450,89 @@ func transformQueryStringJSON(jsonBytes []byte) []byte {
...
+	res := new(http.Response)
+	requestOptions := []option.RequestOption{
+		option.WithQuery("dry_run", "true"),
+		option.WithResponseBodyInto(&res),
+		option.WithMiddleware(logging.Middleware(ctx)),
+	}
+
+	var err error
+	switch {
+	// The plan is absent only when the resource is being deleted
```
zakcutner: I think it would be nicer to check explicitly that it's a delete event, instead of that the plan is empty. Terraform actually has a lot of different events beyond simple CRUD (e.g., imports), so I think there could be another event (now or in the future) where the plan is also empty.

internal/services/ruleset/resource.go
```diff
@@ -430,3 +450,89 @@ func transformQueryStringJSON(jsonBytes []byte) []byte {
...
+			params.ZoneID = cloudflare.F(state.ZoneID.ValueString())
+		}
+
+		err = r.client.Rulesets.Delete(ctx, state.ID.ValueString(), params, requestOptions...)
+
+		if res != nil && res.StatusCode == http.StatusNotFound {
+			return
+		}
+
+	// A plan that leaves the computed ID unknown is creating the ruleset
```
zakcutner: Same here, would be nice to check for create events explicitly, rather than inferring from the ID being unknown

internal/services/ruleset/resource.go
```diff
@@ -430,3 +460,114 @@ func transformQueryStringJSON(jsonBytes []byte) []byte {
 	}
 	return jsonBytes
 }
+
+func planIsKnown(plan, config tftypes.Value) bool {
```
zakcutner: Why do we need the config to tell if the plan is known?

internal/services/ruleset/resource.go
```diff
@@ -430,3 +460,114 @@ func transformQueryStringJSON(jsonBytes []byte) []byte {
...
+	var err error
+	switch {
+	// Terraform plans a create as a null state
+	case req.State.Raw.IsNull():
+		// skip the dry run when creating as part of a replacement
+		if replacesExistingRuleset(ctx, req) {
+			return
+		}
+
+		dataBytes, marshalErr := plan.MarshalJSON()
```
zakcutner, line 488: Is it possible/desired to use the `config` instead of the `plan`?

internal/services/ruleset/ruleset_test.go
```diff
@@ -340,6 +346,260 @@ func TestAccCloudflareRuleset_Name(t *testing.T) {
...
+			{
+				ConfigFile:      config.TestNameFile("1.tf"),
+				ConfigVariables: configVariables,
+				PlanOnly:        true,
+			},
+		},
+	})
+}
+
+func TestAccCloudflareRuleset_DryRunSkippedOnDelete(t *testing.T) {
```
zakcutner: I don't think this test is needed

internal/services/ruleset/ruleset_test.go
```diff
@@ -340,6 +346,260 @@ func TestAccCloudflareRuleset_Name(t *testing.T) {
...
+			{
+				ConfigFile:      config.TestNameFile("1.tf"),
+				ConfigVariables: configVariables,
+			},
+			{
+				ConfigFile:         config.TestNameFile("2.tf"),
+				ConfigVariables:    configVariables,
+				PlanOnly:           true,
+				ExpectNonEmptyPlan: true,
+			},
```
zakcutner, line 511: We should check this gives an error on apply

internal/services/ruleset/testdata/TestAccCloudflareRuleset_DryRunSkippedOnEntryPointRename/2.tf
zakcutner, line 1: I don't think this case is needed?

internal/services/ruleset/ruleset_test.go
```diff
@@ -340,6 +346,260 @@ func TestAccCloudflareRuleset_Name(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				ConfigFile:      config.TestNameFile("1.tf"),
+				ConfigVariables: configVariables,
+			},
+			{
+				ConfigFile:         config.TestNameFile("2.tf"),
+				ConfigVariables:    configVariables,
+				PlanOnly:           true,
+				ExpectNonEmptyPlan: true,
```
zakcutner, line 486: We should check there's an error on apply

internal/services/ruleset/ruleset_test.go
```diff
@@ -340,6 +346,260 @@ func TestAccCloudflareRuleset_Name(t *testing.T) {
...
+			{
+				ConfigFile:      config.TestNameFile("1.tf"),
+				ConfigVariables: configVariables,
+			},
+			{
+				ConfigFile:         config.TestNameFile("1.tf"),
+				ConfigVariables:    configVariables,
+				Destroy:            true,
+				PlanOnly:           true,
+				ExpectNonEmptyPlan: true,
```
zakcutner: We should check this gives an error on apply

internal/services/ruleset/testdata/TestAccCloudflareRuleset_DryRunSkippedOnReferencedDelete/1.tf
zakcutner, line 1: You need a second step where the resources are deleted

internal/services/ruleset/ruleset_test.go
```diff
@@ -340,6 +346,260 @@ func TestAccCloudflareRuleset_Name(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				ConfigFile:      config.TestNameFile("1.tf"),
+				ConfigVariables: configVariables,
+			},
+			{
+				ConfigFile:         config.TestNameFile("2.tf"),
+				ConfigVariables:    configVariables,
+				PlanOnly:           true,
+				ExpectNonEmptyPlan: true,
```
zakcutner, line 462: Let's check for an error on apply

internal/services/ruleset/testdata/TestAccCloudflareRuleset_DryRunSkippedOnReplacement/1.tf
zakcutner, line 1: I think this is already covered in other tests?

internal/services/ruleset/ruleset_test.go
```diff
@@ -340,6 +346,260 @@ func TestAccCloudflareRuleset_Name(t *testing.T) {
...
+func TestAccCloudflareRuleset_DryRunSkippedWhenDependencyIsUnknown(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acctest.TestAccPreCheck(t) },
+		ProtoV6ProviderFactories: acctest.TestAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				ConfigFile:         config.TestNameFile("1.tf"),
+				ConfigVariables:    configVariables,
+				PlanOnly:           true,
+				ExpectNonEmptyPlan: true,
```
zakcutner, line 399: Let's check apply fails

internal/services/ruleset/testdata/TestAccCloudflareRuleset_DryRunSkippedWhenNothingChanges/1.tf
zakcutner, line 1: I don't think this one is needed either

internal/services/ruleset/ruleset_test.go
```diff
@@ -340,6 +346,260 @@ func TestAccCloudflareRuleset_Name(t *testing.T) {
...
+func TestRulesetRequiresReplaceAttributes(t *testing.T) {
+	t.Parallel()
+
+	// Kept in step with replacesRuleset in resource.go
+	expected := []string{"account_id", "kind", "name", "phase", "zone_id"}
+
+	requiresReplace := reflect.TypeOf(stringplanmodifier.RequiresReplace())
+
+	var found []string
+	for name, attribute := range ruleset.ResourceSchema(context.TODO()).Attributes {
```
zakcutner: nit: Let's use `context.Background()`

internal/services/ruleset/ruleset_test.go
zakcutner, line 1: Let's add tests for computed + optional attributes being unknown/null

internal/services/ruleset/ruleset_test.go
zakcutner: Also, could we just do this in the code to avoid the hardcoded list completely?

internal/services/ruleset/resource.go
jahnavimalhotra1: I think it might be better to use plan here. Apply serializes the plan, so it makes sense to use the exact same thing for the dry run as well. Also config wouldn't have the applied default values like plan does.

internal/services/ruleset/ruleset_test.go
jahnavimalhotra1: InvalidOnCreate covers the null case with ref/logging (both computed+optional), SkippedWhenDependencyIsUnknown covers the unknown case when the computed+optional attribute also has a default (description), SkippedWhenRefIsUnknown covers the unknown case when there is no default (ref)

internal/services/ruleset/testdata/TestAccCloudflareRuleset_DryRunSkippedOnReferencedDelete/1.tf
jahnavimalhotra1: This test had a valid referenced delete so we wouldn't have been able to check for an error when doing apply, for it to have an error we would've had to do a replacement which is already covered by SkippedOnReferencedReplacement so this test was redundant anyways.

zakcutner (approved):

Issue #7296, opened 2026-08-06
zero_trust_access_ai_controls_mcp_portal: refresh fails with "duplicate values" because server_id is tagged no_refresh


mafigit:
### Confirmation

- [x] This is a bug with an existing resource and is not a feature request or enhancement. Feature requests should be submitted with Cloudflare Support or your account team.
- [x] I have searched the issue tracker and my issue isn't already found.
- [x] I have replicated my issue using the latest version of the provider and it is still present.

### Terraform and Cloudflare provider version

```
$ tofu version
OpenTofu v1.12.5
on darwin_arm64
+ provider registry.opentofu.org/cloudflare/cloudflare v5.22.0
```

v5.23.0 has the same bug: its CHANGELOG credits [`9766b3a`](https://github.com/cloudflare/terraform-provider-cloudflare/commit/9766b3a5e438e17fe07715cced2a6acdee0a040a) for this resource, but only half of that commit shipped. See Additional factoids.

### Affected resource(s)

cloudflare_zero_trust_access_ai_controls_mcp_portal

### Terraform configuration files

```hcl
# Two MCP servers with no per-tool overrides, so updated_tools is null on both
# and their portal members come out equal.

resource "cloudflare_zero_trust_access_ai_controls_mcp_server" "alpha" {
  account_id = var.account_id
  id         = "server-alpha"
  name       = "server alpha"
  hostname   = "https://alpha.example.com/mcp"
  auth_type  = "bearer"
}

resource "cloudflare_zero_trust_access_ai_controls_mcp_server" "beta" {
  account_id = var.account_id
  id         = "server-beta"
  name       = "server beta"
  hostname   = "https://beta.example.com/mcp"
  auth_type  = "bearer"
}

import {
  to = cloudflare_zero_trust_access_ai_controls_mcp_portal.this
  id = "<account_id>/<portal_id>"
}

resource "cloudflare_zero_trust_access_ai_controls_mcp_portal" "this" {
  account_id = var.account_id
  id         = "my-portal"
  name       = "My Portal"
  hostname   = "portal.example.com"

  servers = [
    { server_id = cloudflare_zero_trust_access_ai_controls_mcp_server.alpha.id },
    { server_id = cloudflare_zero_trust_access_ai_controls_mcp_server.beta.id },
  ]
}
```

### Link to debug output

Not published. The failure follows from the source path below; happy to attach a log on request.

### Panic output

From a v5.22.0 run against a five-member portal:

[...]

Issue #7271, opened 2026-07-24
mcp_portal servers list causes perpetual diff (read positionally instead of by server_id)


JVenberg:
Running into a perpetual plan with `cloudflare_zero_trust_access_ai_controls_mcp_portal` and I think it comes down to how the `servers` list is read back.

`servers` is an ordered list, and on refresh the provider seems to line it up positionally rather than matching each entry on `server_id`. The API returns the members in its own order (looks like it sorts by the server's `created_at`), which doesn't match the order I declare in config, so every plan shows the whole list being reordered even though nothing actually changed:

```
~ servers = [
    - { server_id = "statsig" -> null }
    - { server_id = "atlassian" -> null }
    + { server_id = "atlassian" }
    + { server_id = "statsig" }
  ]
```

It gets worse once a member has a nested `updated_tools` block. Because the match is positional, that block gets attributed to the wrong `server_id` in the refreshed state, so the diff never converges no matter how many times I apply. I can work around it by pre-sorting my config to match the API's `created_at` order, but servers created within the same second have no stable tiebreak I can see from the API, so a couple of them still flap.

Expected: the order of `servers` shouldn't matter. Matching members by `server_id` on read (treating it as a set keyed by id) would make config order irrelevant and fix both the reorder noise and the `updated_tools` misattribution.

Provider version: 5.22.0 (also saw it on 5.21.x).

Minimal repro:

```hcl
resource "cloudflare_zero_trust_access_ai_controls_mcp_portal" "example" {
  account_id = var.account_id
  id         = "example"
  name       = "Example"
  hostname   = "example.mcp.example.com"

  servers = [
    { server_id = "server-a", on_behalf = true },
    { server_id = "server-b", on_behalf = true, updated_tools = [{ name = "some_read_tool", enabled = true }] },
  ]
}
```

If the API happens to return `server-b` before `server-a`, `plan` reorders them on every run and the `updated_tools` block ends up associated with `server-a` in state. Happy to provide a fuller repro or state/debug logs if useful. Thanks!

jessicayingg:
Thank you for submitting an issue! This issue has been flagged for an internal Cloudflare team to take a look. If this is urgently blocking your development or deployment, please open a support case so we can serve you faster.

ssicard:
This has been fixed in v5.24.0

PR #7328, opened 2026-08-20
release: v5.24.0

rotem-cloud:
# Changelog

### BREAKING CHANGES

The following upstream API schema changes required corresponding provider schema updates. Existing state files load without user action (the plugin framework silently drops attributes that no longer exist in the schema), but Terraform **configurations** that still reference removed attributes must be updated before `terraform plan` will succeed.

* **hostname_tls_setting:** the `cloudflare_hostname_tls_setting` data source now requires `hostname` as an input parameter instead of returning it as a computed attribute. The computed `id` attribute has been removed; use `setting_id` instead. ([ece27fe](https://github.com/cloudflare/terraform-provider-cloudflare/commit/ece27fee47f3fb7229d16fc73354ea755b96b285))
* **image_variant:** the computed `variant` nested attribute has been removed. The API no longer returns this wrapper object; the `id`, `options`, and `never_require_signed_urls` fields remain available at the top level. Remove any references to `cloudflare_image_variant.<name>.variant` from your configurations. ([738bd28](https://github.com/cloudflare/terraform-provider-cloudflare/commit/738bd28e14408e9012fd6487e6e4c8e46bf397ea))
* **organization:** the computed `meta.flags` nested attribute has been renamed to `meta.tenant_flags`. Two new computed sub-attributes (`enterprise_capability`, `member_management`) have been added. Update any references from `.meta.flags` to `.meta.tenant_flags`. ([ece27fe](https://github.com/cloudflare/terraform-provider-cloudflare/commit/ece27fee47f3fb7229d16fc73354ea755b96b285))
* **zero_trust_access_ai_controls_mcp_portal:** the `allow_code_mode` attribute is now deprecated and no longer computed with a default of `true`. Use the new `code_mode` attribute (`"off"`, `"opt_in"`, `"default_on"`, `"enforced"`) instead. Configurations that relied on the computed default must explicitly set `code_mode` or `allow_code_mode`. ([ece27fe](https://github.com/cloudflare/terraform-provider-cloudflare/commit/ece27fee47f3fb7229d16fc73354ea755b96b285))
[...]

PR #7330, opened 2026-08-21
fix(zero_trust_access_ai_controls_mcp_portal): remove no_refresh from server_id

rotem-cloud:
## Summary

Removes the `no_refresh` tag from `server_id` in `ZeroTrustAccessAIControlsMcpPortalServersModel`.

## Problem

`server_id` was tagged `no_refresh`, causing the decoder to skip it during refresh/import. With `server_id` blanked, portal members were compared only by their remaining four attributes (`default_disabled`, `on_behalf`, `updated_tools`, `updated_prompts`). Two servers with identical tool mirrors (e.g. two freshly registered servers with no per-tool overrides) produced duplicate set elements, resulting in:

```
unexpected error creating NestedObjectSet: This attribute contains duplicate values
```

## Fix

Remove `no_refresh` so `server_id` is read back from the API on every refresh, keeping members distinct.

## References

- Jira: MCP-319
- Related PR: #7226
- Related issues: #7271, #7296


rotem-cloud:
Closing in favour of #7328.

PR #7335, opened 2026-08-25
fix(provider): send user_agent_operator_suffix unquoted and honor its env var

### LINKED ISSUE #7323: user_agent_operator_suffix is sent double-quoted in the User-Agent header
### Summary

`user_agent_operator_suffix` is emitted into the `User-Agent` header wrapped in literal double quotes, because `Configure` formats it with `String()` instead of `ValueString()`. The same branch also doesn't guard against unknown values, so an unknown suffix is sent as the literal text `<unknown>`.

### Details

https://github.com/cloudflare/terraform-provider-cloudflare/blob/main/internal/provider.go#L435-L440

```go
if !data.UserAgentOperatorSuffix.IsNull() {
	operatorSuffix := data.UserAgentOperatorSuffix.String()
	userAgentParams.OperatorSuffix = &operatorSuffix
} else {
	userAgentParams.TerraformVersion = &req.TerraformVersion
}
```

`basetypes.StringValue.String()` is the human-readable debug representation, not the value:

```go
func (s StringValue) String() string {
	if s.IsUnknown() { return attr.UnknownValueString }  // "<unknown>"
	if s.IsNull()    { return attr.NullValueString }     // "<null>"
	return fmt.Sprintf("%q", s.value)                    // quoted
}
```

So there are two problems:

1. **Quoting.** A known value goes out as `"mycorp/1.0"` (with the quotes) rather than `mycorp/1.0`.
2. **Unknown values.** `IsNull()` is false for an unknown value, so the first branch is taken and the header gets the literal string `<unknown>`.

### Reproduction

```hcl
provider "cloudflare" {
  user_agent_operator_suffix = "mycorp/1.0"
}
```

Point `base_url` at a local HTTP server and inspect the request.

Actual:

```
User-Agent: terraform-provider-cloudflare/5.23.0 terraform-plugin-framework/1.19.0 "mycorp/1.0"
```

Expected:

```
User-Agent: terraform-provider-cloudflare/5.23.0 terraform-plugin-framework/1.19.0 mycorp/1.0
```

### Impact

This defeats the purpose of the feature added in #2831 — the suffix exists so that non-Terraform operators can be identified in traffic analysis, and the stray quotes mean the token doesn't match what the operator configured.

### Suggested fix

```diff
-	if !data.UserAgentOperatorSuffix.IsNull() {
-		operatorSuffix := data.UserAgentOperatorSuffix.String()
+	if !data.UserAgentOperatorSuffix.IsNull() && !data.UserAgentOperatorSuffix.IsUnknown() {
+		operatorSuffix := data.UserAgentOperatorSuffix.ValueString()
 		userAgentParams.OperatorSuffix = &operatorSuffix
 	} else {
```

### Unrelated but adjacent

The schema description for `user_agent_operator_suffix` says it "can be configured using the `CLOUDFLARE_USER_AGENT_OPERATOR_SUFFIX` environment variable", but `Configure` never reads that variable — every other provider setting has an `os.LookupEnv` fallback and this one doesn't. Happy to split that into a separate issue if you'd prefer.

### Versions

Reproduced against `v5.23.0` and current `main`.

---

Happy to open a PR for the fix if that's useful.

jkodroff:
Fixes #7323.

Filed as one PR with two distinct commits per @mgirouard's direction on the issue. Happy to move any of this into codegen upstream instead — the intent here is to make the expected output concrete.

## Changes being requested

**1. `fix(provider): send user_agent_operator_suffix unquoted`**

`Configure` formatted the suffix with `basetypes.StringValue.String()`, which is the human-readable debug representation, not the value. A configured suffix of `mycorp/1.0` reached the wire as `"mycorp/1.0"`, quotes included:

```
User-Agent: terraform-provider-cloudflare/5.23.0 terraform-plugin-framework/1.19.0 "mycorp/1.0"
```

The token therefore no longer matched what the operator configured, defeating the point of the feature added in #2831.

The same branch only guarded on `IsNull()`, which is false for an unknown value, so an unknown suffix was sent as the literal text `<unknown>`. Both are fixed by guarding on `IsUnknown()` as well and using `ValueString()` — the shape every other option in `Configure` already uses.

**2. `fix(provider): read user_agent_operator_suffix from the environment`**

The schema description promises the setting "can be configured using the `CLOUDFLARE_USER_AGENT_OPERATOR_SUFFIX` environment variable", but `Configure` never read it. Every other provider setting has an `os.LookupEnv` fallback; this adds the missing one in the same shape, so explicit configuration still wins over the environment.

No acceptance test — this is provider `Configure` behaviour, not a resource, so it is covered by a plain unit test instead. `internal/provider_test.go` is new (the repo has no provider-level test today): it builds a `tfsdk.Config` from the provider schema, points `base_url` at an `httptest` server, and asserts the exact `User-Agent` that reaches the wire. No credentials, no `TF_ACC`, no network — it runs in the normal sharded unit test job.

```sh
go test ./internal/ -run TestProviderUserAgentOperatorSuffix -v
```

### Test output

```
=== RUN   TestProviderUserAgentOperatorSuffix
=== RUN   TestProviderUserAgentOperatorSuffix/configured_suffix_is_sent_verbatim
=== RUN   TestProviderUserAgentOperatorSuffix/unknown_suffix_falls_back_to_the_Terraform_version
=== RUN   TestProviderUserAgentOperatorSuffix/suffix_can_come_from_the_environment
=== RUN   TestProviderUserAgentOperatorSuffix/configuration_takes_precedence_over_the_environment
=== RUN   TestProviderUserAgentOperatorSuffix/no_suffix_falls_back_to_the_Terraform_version
--- PASS: TestProviderUserAgentOperatorSuffix (0.00s)
    --- PASS: TestProviderUserAgentOperatorSuffix/configured_suffix_is_sent_verbatim (0.00s)
    --- PASS: TestProviderUserAgentOperatorSuffix/unknown_suffix_falls_back_to_the_Terraform_version (0.00s)
    --- PASS: TestProviderUserAgentOperatorSuffix/suffix_can_come_from_the_environment (0.00s)
    --- PASS: TestProviderUserAgentOperatorSuffix/configuration_takes_precedence_over_the_environment (0.00s)
    --- PASS: TestProviderUserAgentOperatorSuffix/no_suffix_falls_back_to_the_Terraform_version (0.00s)
PASS
ok  	github.com/cloudflare/terraform-provider-cloudflare/internal	0.936s
```

The first three cases pass on commit 1 alone; the two environment cases arrive with commit 2. Each commit is green on its own.

## Additional context & links

- Issue: #7323
- Feature this restores: #2831

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0184GUxExLR6iaL6wqdwdsQB

### DIFF
--- a/internal/provider.go
+++ b/internal/provider.go
@@ -435,6 +435,8 @@ func (p *CloudflareProvider) Configure(ctx context.Context, req provider.Configu
 	if !data.UserAgentOperatorSuffix.IsNull() && !data.UserAgentOperatorSuffix.IsUnknown() {
 		operatorSuffix := data.UserAgentOperatorSuffix.ValueString()
 		userAgentParams.OperatorSuffix = &operatorSuffix
+	} else if o, ok := os.LookupEnv(consts.UserAgentOperatorSuffixEnvVarKey); ok {
+		userAgentParams.OperatorSuffix = &o
 	} else {
 		userAgentParams.TerraformVersion = &req.TerraformVersion
 	}
--- a/internal/provider.go
+++ b/internal/provider.go
@@ -432,8 +432,8 @@ func (p *CloudflareProvider) Configure(ctx context.Context, req provider.Configu
 		PluginVersion:   pluginVersion,
 	}
 
-	if !data.UserAgentOperatorSuffix.IsNull() {
-		operatorSuffix := data.UserAgentOperatorSuffix.String()
+	if !data.UserAgentOperatorSuffix.IsNull() && !data.UserAgentOperatorSuffix.IsUnknown() {
+		operatorSuffix := data.UserAgentOperatorSuffix.ValueString()
 		userAgentParams.OperatorSuffix = &operatorSuffix
 	} else {
 		userAgentParams.TerraformVersion = &req.TerraformVersion
--- a/internal/provider_test.go
+++ b/internal/provider_test.go
@@ -1,0 +1,138 @@
+package internal_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/cloudflare/cloudflare-go/v7"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal"
+	"github.com/cloudflare/terraform-provider-cloudflare/internal/consts"
+	"github.com/hashicorp/terraform-plugin-framework/provider"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+const testProviderVersion = "1.0"
+const testTerraformVersion = "1.9.0"
+
+// providerConfig builds a `tfsdk.Config` from the provider schema, with every
+// attribute null unless overridden in `attrs`.
+func providerConfig(t *testing.T, ctx context.Context, attrs map[string]tftypes.Value) tfsdk.Config {
+	t.Helper()
+
+	schema := internal.ProviderSchema(ctx)
+	objType, ok := schema.Type().TerraformType(ctx).(tftypes.Object)
+	if !ok {
+		t.Fatalf("provider schema type is %T, expected tftypes.Object", schema.Type().TerraformType(ctx))
+	}
+
+	values := map[string]tftypes.Value{}
+	for name, attrType := range objType.AttributeTypes {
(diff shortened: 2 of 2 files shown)


jkodroff:
Closing this PR since the actual fix is upstream anyway, but still waiting on the functionality to be implemented upstream.

